PR #65, opened 2015-04-02
Feature/add logging

moonso:
- Fixed update case
- Added some logging


moonso:
I've updated the changes from #64

scout/log/log.py
```diff
@@ -0,0 +1,63 @@
...
+    
+    Arguments:
+    
+        filename (str): Path to a file. Or None if logging is to
+                         be disabled.
+        loglevel (str): Determines the level of the log output.
+    """
+    
+    formatter = logging.Formatter(
+                    '[%(asctime)s] %(levelname)s: %(name)s: %(message)s'
```
robinandeer: I just learned something new about formatting strings from this link: https://docs.python.org/3/howto/logging-cookbook.html#logging-to-multiple-destinations

Basically you can align each section of the log record so it's easier to read:

```python
formatter = logging.Formatter( 
    '[%(asctime)s] %(levelname)-8s: %(name)-25s: %(message)s')
```

scout/log/log.py
```diff
@@ -0,0 +1,63 @@
+import os
+import sys
+import logging
+
+def init_log(logger, filename=None, loglevel=None):
```
robinandeer: I'm not super sure but why not initialize the root logger in this function? Then all the rest of the loggers will inherit it's formatting and handler, no?

```python
root_logger = logging.getLogger()
```

scout/log/log.py
moonso: Yes that is nice!

scout/log/log.py
moonso: Haha we just discussed it yeserday, remember?

moonso: We concluded that it should allways be up to the user how the loggers should be formated.
So if you import the package you set the loggers yourself.

robinandeer: Yeah :sweat_smile:, I'm still a little confused about this whole implicit inheritance works and what really is a root logger to some other type of logger, hehe. But what you say make sense - keep up the good work!

moonso: Merge it when you feel ready.

robinandeer: Sure thing - I think we can update the formatted template line first though :smiley:

PR #72, opened 2015-04-09
Feature/smaller fixes

robinandeer:
Some fixes that @vwirta suggested.

- Remove IGV link until it works
- Add ability to unmark a causative variant
- You are no longer allowed to mark *an additional* causative variant (overwriting the first)
- Log more events to variant activity log
- Allow resetting manual variant rank
- use a sorted list field for cases in institute model (this will go away I guess with the new ODM updates...)


robinandeer:
Let's move the code review to the release branch instead - how about that?

PR #74, opened 2015-04-10
Release/v1.0.0 rc1

robinandeer:
I will open this mostly to discuss. We can do the actual merge on the command line.


scout/ext/backend/load_mongo.py
```diff
@@ -22,166 +22,231 @@
...
 from vcf_parser import VCFParser
 
-from pprint import pprint as pp
-
 import scout
 
 
-def load_mongo_db(scout_configs, config_file=None, family_type='cmms',
+
+def load_mongo_db(scout_configs, vcf_configs=None, family_type='cmms',
```
robinandeer: I was wondering about the ``vcf_configs`` - is it supposed to be a list or a path to a file?

moonso: Yes i agree, this is unclear.
I think that they should by dictionaries with configurations.
So one could send in a configobj object or a dictionary. 
If you think that sound reasonable I can update that and fix a proper docstring.

robinandeer:
Check this: https://github.com/Clinical-Genomics/scout/blob/da4d0f2755d314c98dde24c5dd34822f25274126/README.md

scout/ext/backend/load_mongo.py
robinandeer: Yeah, sure - do that

PR #92, opened 2015-05-06
Updated requirements. Deletes all variants of a certain case and variant...

moonso:
... type before new variants are added

### DIFF
--- a/scout/models/variant.py
+++ b/scout/models/variant.py
@@ -300,12 +300,6 @@ class Variant(Document):
   acmg_evaluation = StringField(choices=ACMG_TERMS)
 
   @property
-  def local_requency(self):
-    """Returns a float with the local freauency for this position."""
-    return (Variant.objects(variant_id=self.variant_id).count /
-              Case.objects.count())
-
-  @property
   def omim_annotations(self):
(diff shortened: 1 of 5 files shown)


scout/models/variant.py
```diff
@@ -442,6 +436,7 @@ def end_position(self):
 
     return self.position + (bases - 1)
 
+  # This is exactly the same as variant_id...
```
robinandeer: you can remove this!

robinandeer: Change this across the server/templates to use "variant_id"

robinandeer:
I'm not sure which field you wanted me to use instead of "id_string" but it seems "display_name" is almost the same so I use that! Correct?

moonso:
No it should be ```variant_id```

robinandeer:
Hmmm.. variant id is an MD5 sum though, no?

I'm using ``id_string`` in the sanger email to five the variant an id
https://github.com/Clinical-Genomics/scout/pull/92/files#diff-6415cc6a9574f468510a1ada00d9e49cR762

moonso:
Ok you are probebly right:)

robinandeer:
:+1:

PR #109, opened 2015-05-19
Feature/comments in own collection

moonso:
@robinandeer  please help and have a look at this. I'm not at all certain of how we should handle this new structure...


moonso:
Begin by looking at the new format of Event, then we need to discuss how to handle the rest.

robinandeer:
I'll wait until you've pushed the recent updates

moonso:
Ok things start to look ready please have a look.
But we should discuss a bit more before merging.

scout/ext/backend/mongo.py
```diff
@@ -19,11 +19,12 @@
 import io
 import json
 import click
+import logging
 from mongoengine import connect, DoesNotExist, Q
 
 from . import BaseAdapter
 from .config_parser import ConfigParser
-from ...models import (Variant, Compound, Case)
+from scoutmodels import (Variant, Compound, Case, Event)
```
robinandeer: I think this is missing a ``.``

scout/ext/backend/mongo.py
```diff
@@ -269,6 +271,384 @@ def previous_variant(self, document_id):
...
+      
+    """
+    self.logger.info("Creating event for updating the synopsis for case {0}".format(
+      case.display_name
+    ))
+    
+    self.create_event(
+      institute=institute,
+      case=case,
+      user=user,
```
robinandeer: You might need to trick MongoEngine here by adding an id directly... ``.to_dbref()``

scout/ext/backend/mongo.py
```diff
@@ -269,6 +271,384 @@ def previous_variant(self, document_id):
...
+        case=case,
+        user=user,
+        link=link,
+        category='case',
+        verb='comment',
+        subject=case.display_name,
+        content=content
+      )
+    
+    return
```
robinandeer: Maybe return the event if nothing else? Or ``self`` for chain-ability

scout/ext/backend/utils/get_case.py
```diff
@@ -112,26 +114,32 @@ def get_case(scout_configs, family_type):
...
+        logger.debug("Madeline file was read succesfully")
 
     # Add the coverage report
     coverage_report = scout_configs.get('coverage_report', None)
     if coverage_report:
-      mongo_case['coverage_report_path'] = coverage_report
-      logger.debug("Setting coverage report to: {0}".format(coverage_report))
+      logger.debug("Found a coverage report")
+      with coverage_report_path.open('rb') as handle:
+        mongo_case['coverage_report_path'] = handle.read()
```
robinandeer: Shouldn't we rename this field to something more relevant now that it isn't a path?

scout/models/case/case.py
```diff
@@ -113,7 +51,8 @@ class Case(Document):
   madeline_info = StringField()
   vcf_file = StringField()
 
-  coverage_report_path = StringField()
+  # The coverage report will be read as a binary blob
+  coverage_report_path = BinaryField()
```
robinandeer: same here, I would consider a new name for the field!

scout/models/event.py
```diff
@@ -26,34 +29,78 @@
...
+      "comment" : "commented on",
+      "synopsis" : "updated synopsis for",
+      "pin" : "pinned variant",
+      "unpin" : "removed pinned variant",
+      "sanger" : "ordered sanger sequencing for",
+      "archive" : "archived",
+      "open_research" : "opened research mode for",
+    }
+    
+    return display_info.get(self.verb, "")
```
robinandeer: Any reason for returning an empty string by default? I was thinking it's more robust to test for ``None``

scout/models/variant/variant.py
```diff
@@ -280,6 +78,17 @@ class Variant(Document):
   # Predicted deleteriousness:
   cadd_score = FloatField()
   clnsig = IntField()
+  
+  @property
+  def has_comments(self):
```
robinandeer: I will need to remember to update my code to use this!

Should I push to this branch or handle it separately?

robinandeer:
I noticed that some of the new "event" methods handle the actual action like updating the status of a case whereas other simply create a new event.

Is this intended behaviour or will you update so that for example ``pin_variant`` actually updates the vase with a new pinned variant?

robinandeer:
Hej,

I was thinking on implementing the global comment etc. I noticed a lot of unresolved issues with imports and missing "OmimPhenotype" e.g. Can you push the most recent changes of this branch? Or should I fix the problems I find locally?

ping @moonso

moonso:
I will do it as soon as I have time, the branch is not ready yet...

PR #114, opened 2015-05-21
Feature/superficial updates

robinandeer:
Only visual updates according to issues for v1.1


robinandeer:
Should compare with develop :stuck_out_tongue:

PR #118, opened 2015-05-22
Feature/case status groups

robinandeer:
Orders cases base on status
Skips cases marked archived
Enables archiving of cases


robinandeer:
:+1:

PR #121, opened 2015-05-28
Feature/global comments

robinandeer:
(I managed to do something good from waking up at 4 AM at least)

Please review, merge, and open a release branch!

I tried to go over all the parts of the site to make sure everything is updated,
hopefully I didn't miss too many things

Coverage reports seems to work, global comments, and the indicator for if a variant has comments. I decided to skip this indicator if there is a manual rank set, OK?

Maybe we should do a sanity check that 5000 variants is enough to make compounds useful? Don't seem like many compound pairs are loaded


robinandeer:
:+1:

PR #155, opened 2015-10-01
Demo setup

robinandeer:
I'll just merge this to get the rest of the fixes in and have traceback

It doesn't touch the server code besides adding support for guest users when the config is set up that way


robinandeer:
Testing one more time, then merging :)

robinandeer:
We should totally move to conda for all this but let's get this in master for now :P

robinandeer:
I'll include the fix for sorting cases by rerun date
Will need to upload recompiled JS templates first

PR #186, opened 2015-11-02
Multiple causative variants

robinandeer:
Should just work with the modifications I made

@moonso: I did update the adapter so check that out :)

### DIFF
--- a/scout/ext/backend/mongo.py
+++ b/scout/ext/backend/mongo.py
@@ -44,7 +44,6 @@ class MongoAdapter(BaseAdapter):
     # self.variant_collection = self.db.variant
 
   def __init__(self, app=None):
-
     self.logger = logging.getLogger(__name__)
 
     if app:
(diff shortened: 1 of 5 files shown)


scout/ext/backend/mongo.py
```diff
@@ -823,7 +822,7 @@ def mark_causative(self, institute, case, user, link, variant):
 
     """
     # mark the variant as causative in the case model
-    case.causative = variant
+    case.causatives.append(variant)
```
moonso: Sure this should work!

scout/models/case/case.py
```diff
@@ -13,75 +13,77 @@
...
+    owner = StringField(required=True)
+    # These are the names of all the collaborators that are allowed to view the
+    # case, including the owner
+    collaborators = ListField(StringField())
+    assignee = ReferenceField('User')
+    individuals = ListField(EmbeddedDocumentField(Individual))
+    created_at = DateTimeField(default=datetime.now)
+    updated_at = DateTimeField(default=datetime.now)
+    suspects = ListField(ReferenceField('Variant'))
+    causatives = ListField(ReferenceField('Variant'))
```
moonso: Yes!

PR #190, opened 2015-11-04
add new rank model version attr to case model, fix #150

### LINKED ISSUE #150: Dispay rank model version in case view. FIXED

robinandeer:
add version to interface

Left to do:
- [x] load from scout config
- [x] populate case model


moonso:
Ok so I expect a new line in the config that looks like:
```ini
rank_model_version = 'cmms_v10.3'
```
The name of the key is what is important

moonso:
Guess you only have to fix the config files with pymip now :hamster:

Issue #151, opened 2015-09-25
Display analysisType in case view (i.e. WES or WGS). FIXED


henrikstranneheim:

moonso:
- [ ] Add a new field to case model ```analysis_type```
- [ ] A badge to case and cases

moonso:
@henrikstranneheim is the analysistype specified in qc_sample_info?
If not, can you add it.
Alternatives should be ```wgs, wes, mixed or unknown```

henrikstranneheim:
It is as 'analysisType'. However you will have to recode it as entries are: genomes and exomes.
Theses are not easy to change and would require a rerun of all samples. 
It can never be unknown and mixed are always processed as genomes. 

Skickat från min iPad

PR #192, opened 2015-11-04
add analysis type to model and interface, adress #151

robinandeer:
Left to complete:

- [x] read analysis type from scout config


moonso:
Now you need to update the scout config from pymip again @robinandeer

PR #197, opened 2015-11-06
Fix clinsig

moonso:
Show clnsig if 5 is found


moonso:
I'll merge this little one :small_blue_diamond:

PR #199, opened 2015-11-06
Fix clnsig accs

moonso:
Adds CLNSIG accession code


moonso:
@robinandeer can you update the interface so that when there is a CLNSIG value add a link to CLNSIG page. The accession code is in ```mongo_variant['clnsigacc']```.

Also they want to change the name of CLNSIG on variant page. We can annotate like:
```
Clinvar: Pathogenic(link to clinvar)
```

PR #204, opened 2015-11-10
include display name + version in default coverage report, fix #203

### LINKED ISSUE #203: Add version of gene panels as default in report FIXED

robinandeer:
add property to combine name and version of gene panel

### DIFF
--- a/scout/core/views.py
+++ b/scout/core/views.py
@@ -65,7 +65,7 @@ def case(institute_id, case_id):
     case_events = store.events(institute, case=case_model)
 
     # default coverage report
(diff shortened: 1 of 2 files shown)


scout/core/views.py
```diff
@@ -65,7 +65,7 @@ def case(institute_id, case_id):
     case_events = store.events(institute, case=case_model)
 
     # default coverage report
-    default_panel_names = [panel.panel_name for panel
+    default_panel_names = [panel.name_and_version for panel
```
moonso: Aha now I understand, it's for the coverage report!

PR #211, opened 2015-11-17
Add pileup

robinandeer:
After some tweaking I think we can evaluate if this is ready for production!

It got some positive surprised reactions when I showed the genome viewer at the meeting with the geneticists today, hehe

We need to test this on the actual BAM alignments on rasta before we merge tho :)


dnil:
👍🏻

moonso:
Super :palm_tree:

robinandeer:
Didn't have time to do any testing but...

Issue #84, opened 2015-05-06
Allow search by position


moonso:
Is it possible to allow searches by position?
Questions like the following should work:

```
chr1:4657
chr1:10000-20000
chr5
```

robinandeer:
Just one position/range or multiple?

robinandeer:
Interface part done :)

moonso:
:+1:

PR #260, opened 2016-02-22
Search position

robinandeer:
In response to #84 

Interface will put a named tuple to the query dict with attributes: "chrom", "start", "end"

The input is on the standard format for genomic positions


moonso:
Ok so I will do the backend stuff for this?

robinandeer:
Yes!

Issue #102, opened 2015-05-08
Reintroduce gene inheritance model


robinandeer:
- [x] Add gene level inheritance model (CMMS gene) @ingkebil 
- [ ] OMIM phenotype (rename) @robinandeer
- [x] Pick this new annotation up and populate database @moonso 

@vwirta: can we change this in Scout and MIP (very minor changes) and call it bugs that don't require a re-validation?

ingkebil:
Fixed the GeneLists.
Just holler when I can update them in _references_.

moonso:
Could anyone, @robinandeer  or @ingkebil , be a little more specific here?
What needs to be done?

robinandeer:
I'm not sure if @henrikstranneheim is picking this up and annotating it in the VCF?

henrikstranneheim:
Sure is. However, it is recorded as ""Genetic_disease_model" to separate it from the variant inheritance model.

henrikstranneheim:
I'm not sure what you want me to do Robin. All the info is in the vcf taken from the clinical list populated by Kenny. This info should go below variant in the table "Inheritance model" using the display name "Disorder"

moonso:
Is there any file on rasta with an example of `Genetic_disease_model`?
I've tried to look at some but could not find it...
@henrikstranneheim @emmser

ingkebil:
Check out cust003-CMMS.txt, and clinical master list. All CMMS related lists have a chance of having `Genetic_disease_model` filled in.

moonso:
Yes I understand that they are annotated in the gene list files. But it should be annotated in the VCF according to @henrikstranneheim comment above.

PR #261, opened 2016-02-22
show expected inheritance if annotated for a variant

robinandeer:
In response to #102

@moonso: can you check if you populate this field from the VCF?

Also: only show expected column in "clinical" mode

### DIFF
--- a/scout/core/templates/variant.html
+++ b/scout/core/templates/variant.html
@@ -421,7 +421,19 @@
             </div>
           </div>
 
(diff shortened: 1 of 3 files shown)


moonso:
Is expected inheritance from the gene panels? 
Yes I'll have a look at this

ingkebil:
Yes, this comes from the gene panel, from column _Genetic_disease_model_

moonso:
Ok @robinandeer I've added the extraction so feel free to merge when you feel like it

Issue #107, opened 2015-05-19
Add detailed variant frequencies


moonso:
We should include frequencies for different populations on variant level when it exists.
Also there has been a discussion to use the highest frequency among all populations when scoring since if the varaint is common in any population it should not be disease causing.

Input from clinics?

dnil:
Agree. 
Likely not disease causing for a high-penetrant rare monogenic. Although different genetic backgrounds could add complication, this is common practice, and the preferred mode in the ACMG guidelines: e.g. BS1 - strong benign support - is tagged when >5% is observed any subpopulation.

moonso:
Ok so this is a MIP issue for now.
@henrikstranneheim can you annotate the allele frequencies from different populations when there are such information?

moonso:
ping @henrikstranneheim

henrikstranneheim:
Yeah, I will have to see if I have time for that one.

Skickat från min iPad

henrikstranneheim:
I think we should support two frequencies from EXAC and 1000G: a MAX_AF (MIP will add this to the vcf references for both 1000G and EXAC) and the usual AF. 
If you want to look into more detail one can follow the EXAC link in Scout for subpopulations and the 1000G browser for instance: http://browser.1000genomes.org/Homo_sapiens/Variation/Population?db=core;r=1:128213-129213;source=dbSNP;v=rs545245702;vdb=variation;vf=78547415

However, that requires an RS number. I think the easiest would be to use the same RS as recorded in the 1000G file. 

@moonso : Can you modify GENMOD to include the rs number and MAX_AF in addition to the AF your extracting now from 1000G?

The frequency table in Scout should expand to display both the MAX_AF and AF as well as a link to the 1000G browser.

robinandeer:
I will guess what the model will look like and update the interface

PR #262, opened 2016-02-22
add link to 1000G page with detailed freq information (#107)

robinandeer:
I've added the link and will display the max freq when the variant model is updated, relates to #107

@moonso: I named the fields as "max_" followed by the normal AF frequency name


moonso:
Oki

moonso:
Think this is ready for merge @robinandeer

PR #293, opened 2016-06-08
Add is_canonical to transcript model

moonso:
Also checks the vep annotation if transcript is canonical


moonso:
@robinandeer interface left

PR #302, opened 2016-06-22
Adds disease associated transcripts on gene level

moonso:
Parses the Disease_associated_transcript info field of the vcf and adds a list
of manually annotated disease associated transcripts to the gene model


moonso:
Check out the `gene` model @robinandeer

PR #303, opened 2016-06-22
Adds cadd score to filters

moonso:


moonso:
Check if it is fine @robinandeer

PR #331, opened 2016-10-03
Changes load functionality

### LINKED ISSUE #329: Adapt loading to position sorted VCFs

moonso:
Only use rank score treshold when loading variants.
Loop through whole vcf and load all variants that meets the rank score treshold.
Current default is rank score 0.
Fix #329


moonso:
@robinandeer please check if you need to modify uploading. I removed one option in the upload command (variant number treshold).

PR #332, opened 2016-10-03
add interface part of clinsig search

robinandeer:
I send in an int with the query or None


robinandeer:
I added the if block to perform the search in the adapter - now it just matches the exact CLINSIG level

If we should handle it differently you can update before you merge :)

PR #386, opened 2016-12-12
Adds some tests and things that where missed when building the varian…

moonso:
…t model


moonso:
Oops sorry, here are some additions

PR #419, opened 2017-01-30
Update hpo

moonso:
Adds functionality to reload HPO


moonso:
This is ready for merge, please have a look @robinandeer

robinandeer:
Looks good as far as I can tell!

What's the plan here? Do we just need to reload the terms or do we need to do any further migration and updating of the interface?

moonso:
I don't think it will affect the interface, reloading should be fine

moonso:
Ok @robinandeer I merged this, please try to rerun `scout load hpo`and check that it works

PR #435, opened 2017-02-10
Suggests new way of building transcripts

moonso:
THIS IS NOT AT ALL READY TO MERGE

@robinandeer have a look at the build hgnc gene file, there are some suggestions there of how this migration could be handled.
Parse just parses but build makes sure that the database is consistent.


moonso:
Hi @robinandeer maybe test this out now?

robinandeer:
skit i den då!

PR #436, opened 2017-02-14
Update case

moonso:
@robinandeer can you add event and variant stuff?


robinandeer:
Now I'm done @moonso! 🎻

PR #444, opened 2017-02-20
Use omim info

moonso:
Use omim to get inheritance info for genes


moonso:
This is rebased on skip-mongoengine-genes

PR #459, opened 2017-03-29
Adds  and adds the necessary files

moonso:


moonso:
Not ready yet! Will try to fix so that scout serve uses demo database

moonso:
It works perfekt with `scout --demo serve` 🏭 🍇 🇭🇺

Issue #443, opened 2017-02-20
sv variant view


dnil:
It would be good to be able to see

- [x] Size. Only in variants list view now.
- [ ] ISCN variant description (canonical clinical nomenclature - compare HGVS for SNV/MNV)
- [x] OR/AND, if too much for now, coordinates of variant close to type, size, cytogenetic coordinate (chromosomal band), zygosity.
- [x] Link to Decipher - the de facto clinical repository of SVs. Can be deep linked via coordinates at least: https://decipher.sanger.ac.uk/browser#q/15:78276969-79060601%20/location/15:78151969-79185601

moonso:
Hi @dnil , are there any examples of ISCN descriptions? Can we collect these from the VCF? How should the filter look?

dnil:
Emailed about ISCN; basically @J35P312 and I have an idea for that. Let's see how it pans out - and just shout if you would like to fix it first!

moonso:
Maybe you could just add a function?
Like:
```python
def get_iscn(variant_obj):
    """Returns ISCN coordinates

    Args:
        variant_obj(dict)

    Returns:
        issn_coordinates(str)
    """
    pass
```

SV variant obj should include most information. Might be something for @J35P312 to contribute with 😸

moonso:
Since most of these are in place I will open a new issue for the ISCN coordinates

PR #466, opened 2017-04-06
Added SV variant filter.

dnil:
Initiated SV variant filter!

### DIFF
--- a/scout/server/blueprints/variants/forms.py
+++ b/scout/server/blueprints/variants/forms.py
@@ -60,4 +60,3 @@ class SvFiltersForm(FlaskForm):
     svtype = SelectMultipleField('SVType', choices=SV_TYPE_CHOICES)
 
     thousand_genomes_frequency = DecimalField('1000 Genomes', places=2)
-    exac_frequency = DecimalField('ExAC', places=2)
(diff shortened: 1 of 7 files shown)


moonso:
Nice! I will have a look 😄

dnil:
Regarding editing commits, I checked "allow edits from maintainers" - looks like this might help you edit the PR anyhow?

dnil:
Adresses #443 3/4 items. ISCN will have to be discussed more. We could perhaps try it on a variant-by-variant basis on the pipeline or loading side first.

scout/server/blueprints/variants/templates/variants/sv-variants.html
```diff
@@ -59,8 +95,101 @@
   <div class="container-fluid">
     {{ pagination() }}
   </div>
+</div>
 {% endblock %}
 
+{% macro filters_form() %}
```
robinandeer: Perhaps this macro can be shared between the templates?

scout/server/blueprints/variants/forms.py
```diff
@@ -41,3 +41,22 @@ class FiltersForm(FlaskForm):
 
     thousand_genomes_frequency = DecimalField('1000 Genomes', places=2)
     exac_frequency = DecimalField('ExAC', places=2)
+
+class SvFiltersForm(FlaskForm):
```
robinandeer: Maybe this class can inherit the common fields from another class?

We will look into it after merging this PR 👍

PR #478, opened 2017-04-18
Fix files

moonso:


moonso:
Ready to merge @robinandeer , fixes the problem with updating dynamic gene list

moonso:
Think we need to merge and deploy this one. Also check if `du -sh` looks better 😄

robinandeer:
I'll publish a new version on pypi!

moonso:
👍

PR #490, opened 2017-04-21
Refseq ids will only be added once per enst

### LINKED ISSUE #489: Duplicated refseq ids
For some transcripts there are duplicated NM ids, for example:
screen shot 2017-04-21 at 2 27 35 pm

moonso:
That is, there can still be multiple refseq transcripts per enst.
Before there could be douplicates.
Fix #489


moonso:
I'm merging this. We need to run `scout load genes --update` to see the difference

Issue #423, opened 2017-01-31
Cohort tag


henrikstranneheim:
- [x] Family level
- [x] Multiple tags per case
- [ ] Searchable

moonso:
Just provide a list here with cohort tags

moonso:
- other
- endo
- mito
- ketogenic diet

PR #496, opened 2017-04-26
Adds functionality to handle cohort tags

moonso:
Addresses #423 
Following needs to be added:

- [x] Update frontend so user can modify from case page
- [ ] Allow to add cohort tags when loading(in config)
- [x] Add tests


moonso:
Could you add a dropdown on case page @robinandeer ?
The cohorts are presented in #423 , these are the CMMS cohorts but we might use them globally for now.

PR #509, opened 2017-05-04
Use cyvcf2

moonso:


moonso:
Guess it works but need to fix tests

moonso:
Ok I think this is finally ready for merging @robinandeer ! Please have a look 👾

.travis.yml
```diff
@@ -28,6 +28,7 @@ addons:
             - mongodb-org-server
 
 install:
+    - pip install cython
```
robinandeer: is this a new dependency? why not add it to `requirements.txt`?

PR #537, opened 2017-05-31
Acmg ui

robinandeer:
you can now set an ACMG class on a variant
and you can set critera and calculate a class

missing:

- [ ] fill-in a classification with critera when `submit_classification` is called on the adapter
- [ ] fetch classifications for a variant for all cases and display for a variant


moonso:
Great! I will continue to work on this

PR #545, opened 2017-06-07
Adds functionality to check if a variant is in the PAR regions

moonso:
- [x] Add function to check if coordinates is in PAR
- [x] Add test for the above function
- [x] Add `is_par` to variant in adapter
- [x] Add label on variant page to display if variant is in PAR


moonso:
Ok @robinandeer , when you add the label this is done

moonso:
🕋

PR #546, opened 2017-06-07
add form to update default gene panels on a case

robinandeer:


moonso:
Ok @robinandeer , redat to merge

PR #580, opened 2017-08-29
Clinsig clinical filter

dnil:
Fix CLINSIG filter - and enable command line debug loglevel without running app in debug mode.


dnil:
Fixes #581. Deploy as hotfix?

scout/server/app.py
```diff
@@ -36,7 +36,8 @@ def create_app(config_file=None, config=None):
     if config_file:
         app.config.from_pyfile(config_file)
```
moonso: Great!

PR #593, opened 2017-09-05
Fix #304 again.

### LINKED ISSUE #304: CADD score filter monday-issue!
Thank you kindly for the quick inclusion of CADD score filtering! Will make a couple of our doctors very happy.

One major caveat though: the current version seems to filter out unknown CADD scores as well (similar to the unknown frequency bug)! Not intended usage..

dnil:


scout/adapter/mongo/query.py
```diff
@@ -114,15 +114,19 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
             logger.debug("Adding cadd_score: %s to query" % cadd)
 
-            if query.get('cadd_inclusive') == 'yes':
-                cadd_query = {
-                    '$or': [
-                        cadd_query,
-                        {'cadd_score': {'$exists': False}}
-                    ]}
-                logger.debug("Adding cadd inclusive to query")
+            if query.get('cadd_inclusive') is not None:
```
moonso: I don't really follow. First you check that `cadd_inclusive != None`. 
Then you check if `cadd_inclusive == True`... Why not check that from the beginning? And also cadd_inclusive is set to 'yes' in `query` so this would not work if I understand everything right

dnil: After parsing the query, it will be True rather than 'y' as it appears on the html query. I haven't tested the undefined part, but at least the "cadd_inclusive" part is not on the http request line if the box is not checked..

dnil: It seems the test is not correct?

tests/adapter/test_query.py
```diff
@@ -57,7 +57,7 @@ def test_build_cadd_exclusive(adapter):
 def test_build_cadd_inclusive(adapter):
     case_id = 'cust000'
     cadd = 10.0
-    cadd_inclusive = 'yes'
+    cadd_inclusive = True
```
moonso: 👍

scout/adapter/mongo/query.py
moonso: Couldn't you skip some lines by only checking `if cadd_inclusive is True:` at the top?

PR #599, opened 2017-09-08
Update panel handling

moonso:


moonso:
Hi @robinandeer , there is a `##TODO` in `scout/server/blueprints/panels/views.py` line 24. Please have a look at that before merging

PR #602, opened 2017-09-08
Fix #577: toggle for passthrough of confident clinsig, enabled in cli…

### LINKED ISSUE #577: Always allow ClinVar Pathogenic, Likely Pathogenic through Clinical filter
It is counterintuitive to find known clinical pathogenic mutation filtered out as a result of a "clinical filter". Also apply to local "marked causatives" - but there we already have a sufficient mechanism?

dnil:
…nical filter.


dnil:
Here's a working (i think) solution - grateful for input on aesthetics and ease of extension.

moonso:
Hi @dnil , nice contribution! 
I think that this is a rather big change to how build_query behaves, I have a hard time to follow the code. Could you add some comments about `mongo_minor` and the intention of the new code blocks? Also some tests that illustrates how the new query looks/works would also be nice.

dnil:
yep, that’s why I want a thorough review! :)

I’ll add a few sentences and devise a test for the new query flags; the others should hopefully do the same thing they did before. Perhaps some kind of interaction test as well. Likely on Wednesday.

moonso:
👍 👍 👍

dnil:
Ok, comments in; tests added. Should work now. Are the failing tests from something that was added/removed from master (general conftest.py - missing family_id)?

moonso:
Well done! I have the same problems with the tests in my PR #603 , no idea what is happening 😱

moonso:
Hi @dnil , we had some problems with our git history. Could you fetch the latest version of master and then rebase this branch?

dnil:
Yiekes, that was kind of messy! I'll see what I can do!

dnil:
Ok, I totally messed up my own repo trying to rebase, but simply editing in the master versions of all files I didn't touch here seems to do the trick?

moonso:
All tests works 😸

dnil:
@monso ok, how's this? 😸

dnil:
Hm, no - looks like there are still issues with the history here. I'm seeing some 3.3 comments etc disappearing on the deltas. Perhaps easiest for you to say which ones are acutally supposed to be removed?

moonso:
I'll look into it

dnil:
See #615 instead - that should be easier.

PR #615, opened 2017-09-19
Fix #577 rebase

### LINKED ISSUE #577: Always allow ClinVar Pathogenic, Likely Pathogenic through Clinical filter

dnil:
Ok, lets try it this way instead. I've cherry-picked the relevant commits into current master.

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -273,12 +273,12 @@ class QueryHandler(object):
 
             else:
                 logger.debug("add CLINSIG filter for rank: %s" %
-                             ', '.join(query['clinsig']))
+                             ', '.join(str(query['clinsig'])))
                 if mongo_query_minor:
                     mongo_query_minor.append({'clnsig.value': {'$in': rank}})
                 else:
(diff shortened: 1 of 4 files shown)


scout/adapter/mongo/query.py
```diff
@@ -13,11 +13,13 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
                 'thousand_genomes_frequency': float,
                 'exac_frequency': float,
                 'cadd_score': float,
+                'cadd_inclusive": boolean,
```
robinandeer: the type in Python is `bool`, isn't it?

tests/adapter/test_query.py
```diff
@@ -91,6 +91,101 @@ def test_build_thousand_g_and_cadd(adapter):
         }
     ]
 
+def test_build_clinsig(adapter):
```
robinandeer: very nice with tests!

PR #620, opened 2017-09-20
Allows multiple alleles in VEP's MAF-fields, separated by ampersands

bjhall:
Our VEP-annotated VCFs sometimes contain allele frequencies for multiple alleles like this "A:0.00483&T:0.0001725", which causes "scout load  case" to crash. This should fix that.

### DIFF
--- a/scout/parse/variant/transcript.py
+++ b/scout/parse/variant/transcript.py
@@ -122,16 +122,21 @@ def parse_transcripts(raw_transcripts, allele=None):
         thousandg_frequencies = []
         for key in entry:
             if key.endswith('MAF'):
-                # If the frequency starts with 'ExAC' we know it is a exac freq
-                splitted_entry = entry[key].split(':')
-                if splitted_entry[0] == allele:
-                    value = float(splitted_entry[1])
-                    if value > 0:
-                        if key.startswith('ExAC'):
-                            exac_frequencies.append(value)
-                # Otherwise we know it is a 1000G frequency
-                        else:
-                            thousandg_frequencies.append(value)
+
+                maf_allele_entries = entry[key].split('&')
+
+                for maf_allele in maf_allele_entries:
+                    
+                    # If the frequency starts with 'ExAC' we know it is a exac freq
+                    splitted_entry = maf_allele.split(':')
+                    if splitted_entry[0] == allele:
+                        value = float(splitted_entry[1])
+                        if value > 0:
+                            if key.startswith('ExAC'):
+                                exac_frequencies.append(value)
+                                # Otherwise we know it is a 1000G frequency
+                            else:
+                                thousandg_frequencies.append(value)
 
         if exac_frequencies:
             transcript['exac_maf'] = sum(exac_frequencies)/len(exac_frequencies)


henrikstranneheim:
You should consider using vt or bcftools to normalize and decompose your vcf before annotation. That will reduce the complexity of your vcf annotation process and the information in the INFO fields. Then you do not have to add parsing of multiple alleles for every annotation.

moonso:
Hi @bjhall , yes like @henrikstranneheim said we made a decision some time ago to only operate on normalised and splited vcfs. The problems will grow and grow when trying to handle multi allelic calls...
There is a great guide [here](https://gemini.readthedocs.io/en/latest/) in the gemini documentation that explains how to do it. They also refer to [this](http://www.cureffi.org/2014/04/24/converting-genetic-variants-to-their-minimal-representation/) blog post where it is discussed in detail.

bjhall:
Sorry for being unclear, but these sites are NOT multiallelic in our VCFs (which we are splitting and normalizing btw). However they are multi-allelic in ExAC, so VEP adds allele frequencies for all minor alleles in ExAC, whether they are in the VCF or not.

As an example, here's the variant in the VCF before any annotation/splitting/normalization (straight out of GATK):
```
1       906769  .       C       A       1766.13 .       AC=1;AF=0.167;AN=6;BaseQRankSum=-1.766e+00;ClippingRankSum=0.00;DP=362;ExcessHet=3.0103;FS=0.000;MLEAC=1;MLEAF=0.167;MQ=60.00;MQRankSum=0.00;QD=6.35;ReadPosRankSum=0.815;SOR=0.676     GT:AD:DP:GQ:PL  0/
0:36,0:36:99:0,102,874  0/0:48,0:48:99:0,116,1162       0/1:170,108:278:99:1797,0,3266
```

These links show that the position has multiple minor alleles (A and T) in ExAC:
http://exac.broadinstitute.org/variant/1-906769-C-T
http://exac.broadinstitute.org/variant/1-906769-C-A

And here is part of the CSQ INFO-field after running VEP (the full VCF string is too long...)
```
CSQ=A|missense_variant|MODERATE|PLEKHN1|84069|Transcript|NM_001160184.1|protein_coding|8/15||NM_001160184.1:c.810C>A|NP_001153656.1:p.Asp270Glu|845|810|270|D/E|gaC/gaA|rs78850589||1||SNV|||||||NP_001153656.1||||||tolerated(0.15)|benign(0.077)|||A:0.0014|A:0|
A:0|A:0|A:0.007|A:0|A:0.0002285|A:0.001632|A:1.714e-03&T:5.018e-05|A:0.003665&T:0.0001136|A:0.0002693&T:0|A:0.008489&T:0|A:0&T:0|A:0.014&T:0|A:0.00483&T:0.0001725|A:0.01527&T:0|A:0.001484&T:0.0001142|||||||||10.03|0.891488|0.926
```

As you can see, VEP has added EXaC MAFs for both the A and T alleles (ex: A:0.001484&T:0.0001142), even though the VCF only contains the A allele.

It is very curious that you've never encountered this? Maybe there is some setting in VEP that controls this? Regardless, I think this fix is worth adding since it generalizes the case loader. Our biggest problem with Scout so far has been that is makes quite a few assumptions about the VCF.

moonso:
Aha i understand, guess we don't use VEP to annotate ExAC.. But I totally agree that scout expects the files to look in a certain way. Of course we need to be more general if other labs are going to use Scout.

moonso:
Thanks @bjhall !

PR #621, opened 2017-09-20
Support defining a unique family id in the config

robinandeer:
I've added a method that can migrate a case using the `CUSTOMER-FAMILY` syntax over to using a unique family id.

I'm not updating variant documents since I think we should only run this function when we re-upload a case and then all the variants will be deleted either way!


robinandeer:
I've tested this on a real case with all the included events simulated and it transferred fine

It will do this transfer automatically when it detects a conflicting "old" case it without the need to load the new case using an "update" flag.

PR #624, opened 2017-09-21
Fixes so that gene panels is converted to include correct hgnc id

moonso:


moonso:
Can you check and merge @robinandeer , it is rebased

PR #629, opened 2017-10-02
Fix small issues

robinandeer:
and enable peddy files to be given in the config!


scout/parse/case.py
```diff
@@ -62,11 +62,11 @@ def parse_case_data(config=None, ped=None, owner=None, vcf_snv=None,
                                               config_data['default_gene_panels']]
 
     if peddy_ped:
```
moonso: 👍

robinandeer:
I forget... Was the idea to merge this in to master now? Or do we create the release branch already?

PR #634, opened 2017-10-05
use sample ids when making link to coverage report

robinandeer:
this better reflects how samples can be part of multiple famlies


robinandeer:
chanjo report v4.2.0

PR #636, opened 2017-10-10
enable adding new gene panel from server

robinandeer:


robinandeer:
I've tested with the example panel file you can download from the documentation

PR #657, opened 2017-11-10
Uses regex to parse chr-prefix. Parse GNOMAD field in vcf

moonso:


moonso:
Hi @robinandeer could you add interface part?
You should look for `'gnomad_frequency'` and `'max_gnomad_frequency'`

PR #663, opened 2017-11-16
Export causatives

moonso:


scout/commands/base.py
```diff
@@ -48,7 +49,8 @@
 @click.pass_context
 def cli(context, mongodb, username, password, authdb, host, port, loglevel, config, demo):
     """scout: manage interactions with a scout instance."""
-    coloredlogs.install(level=loglevel)
+    log_format = "%(message)s" if sys.stdout.isatty() else None
+    coloredlogs.install(level=loglevel, fmt=log_format)
```
robinandeer: Nice - would be better with colored text for "info" logs to make it stand out but it's work in progress 🙂

scout/export/variant.py
```diff
@@ -1,26 +1,26 @@
+from pprint import pprint as pp
 
 
 def export_causatives(adapter, collaborator):
```
robinandeer: Should we perhaps add the option to pass a list of collaborators to export causative variants from multiple institutes? Could do in separate PR

scout/export/variant.py
```diff
@@ -1,26 +1,26 @@
...
+        collaborator(str)
+    """
+    #put variants in a dict to get unique ones
     variant_string = ("{0}\t{1}\t{2}\t{3}\t{4}\t{5}\t{6}\t{7}")
     
-    for variant_id in variants:
-        variant = variants[variant_id]
+    for document_id in adapter.get_causatives(institute_id=collaborator):
+        variant_obj = adapter.variant(document_id)
         yield variant_string.format(
```
robinandeer: are all these fields garuanteed to exist?

scout/export/variant.py
moonso: Yes I also thought about that. One should also be able to export all variants from all institutes.

scout/export/variant.py
moonso: Jepp

moonso:
Just realized that the variants are not exported in a sorted way...

moonso:
Guess I have to fix that

robinandeer:
Does it matter that they are sorted? Sorted on RankScore or chromosome:pos?

robinandeer:
It's fine to merge!

moonso:
They should allways be sorted on chromosome:pos

PR #711, opened 2018-01-15
SPIDEX filter. Fix #693

### LINKED ISSUE #693: Filter on SPIDEX prediction

dnil:
I ended up commiting the human readable version of SPIDEX to the mongodb on load. Not the most elegant or backwards compatible solution. Other suggestions? 

My issue was figuring out how to conduct queries on abs(SPIDEX) - or for that matter queries on the spidex_human transform of that used in display?

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -146,11 +146,9 @@ class QueryHandler(object):
                 ]
             })
 
-        if query.get('spidex_human') is not None:
+        if query.get('spidex_human'):
             spidex_human = query['spidex_human']
-            mongo_query_minor.append({
-                        {'spidex_human': {'$in': spidex_human}}
-                   })
+            mongo_query_minor.append({'spidex_human': {'$in': spidex_human}})
 
         if query.get('clingen_ngi') is not None:
             mongo_query_minor.append({
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -146,6 +146,12 @@ class QueryHandler(object):
                 ]
             })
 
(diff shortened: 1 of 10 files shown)


moonso:
I think it would be a bettet solution to display the human readable version in the filter and then convert that to float. The filter can use(pseudo code)  `{'spidex': {'gt': x, '$lt': y}}`

dnil:
Hm, that was my first thought too, but as abs() didn't seem to work in the queries, the get.. well.. complex. But, perhaps that's a small price for not having to update the db. I'll give you an edited PR so you can appreciate what the queries look like - perhaps on the trainride home... :)

moonso:
say that the rules are
```
medium =  0.2 < abs(spidex) < 0.6
high = abs(spidex) > 0.6
```
The filer would then look like

```python
if spidex == 'medium':
    spidex_filter = {'spidex': {'gt': 0.2, '$lt': -0.2}}
elif spidex == 'high':
    spidex_filter = {'spidex': {'gt': 0.6, '$lt': -0.6}}
```
or am I missing something?

dnil:
Yep, in principle, but it grows a little more complicated when you wish to consider missing values, multiple choice and three categories. Not impossible, just a bit messy. Also in general a little dubious to have the values hard coded like that in several places.

Anyway, I have an almost complete version of that stashed; will get back with it when I can!

moonso:
Ok I get it, but still better right. We could solve the problem with hardcoded values on different place by adding some kind of data structure in `scout/constants/__init__.py`

Great!

dnil:
Ok, so I got that other version out and made a makeshift constant data type instead of changing the db schema. See #770 instead. Closing this PR.

PR #716, opened 2018-01-24
added parse of rank_model_version to fix #712

### LINKED ISSUE #712: Add missing rank model version to snv/Indel view

patrikgrenfeldt:
Added parse of rank_model_version
Added test of rank_model_version in parsing and loading

### DIFF
--- a/scout/parse/case.py
+++ b/scout/parse/case.py
@@ -72,7 +72,7 @@ def parse_case_data(config=None, ped=None, owner=None, vcf_snv=None,
     config_data['vcf_sv'] = vcf_sv if vcf_sv else config_data.get('vcf_sv')
     config_data['vcf_cancer'] = vcf_cancer if vcf_cancer else config_data.get('vcf_cancer')
 
-    # Set default rank score treshold to 0
+    config_data['rank_model_version'] = config_data.get('rank_model_version')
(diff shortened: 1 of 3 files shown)


scout/parse/case.py
```diff
@@ -72,7 +72,7 @@ def parse_case_data(config=None, ped=None, owner=None, vcf_snv=None,
     config_data['vcf_sv'] = vcf_sv if vcf_sv else config_data.get('vcf_sv')
     config_data['vcf_cancer'] = vcf_cancer if vcf_cancer else config_data.get('vcf_cancer')
 
-    # Set default rank score treshold to 0
+    config_data['rank_model_version'] = config_data.get('rank_model_version')
```
moonso: Aha so it was never picked out from load file?

tests/load/test_load_case.py
```diff
@@ -19,3 +19,21 @@ def test_load_case(case_obj, panel_database):
 
     for panel in loaded_case['panels']:
         assert panel['display_name']
+
+
+def test_load_case_rank_model_version(case_obj, panel_database):
```
moonso: Could add an assertment in the beginning that `assert case_obj['rank_model_version'] is not None`or something?

Issue #741, opened 2018-02-17
Error when updating gene panel


moonso:
```python
2018-02-16 12:18:09,659 - scout.server.app - ERROR: Exception on /panels/5a782ffc124f63138ce1ff86 [POST] [in python3.5/site-packages/flask/app.py:1587]
Traceback (most recent call last):
  File "python3.5/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "python3.5/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "python3.5/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "python3.5/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "python3.5/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "scout/scout/server/utils.py", line 19, in decorated_function
    context = f(*args, **kwargs)
  File "scout/scout/server/blueprints/panels/views.py", line 68, in panel
    hgnc_id = int(raw_hgnc_id)
ValueError: invalid literal for int() with base 10: 'GFM1'
```

ingkebil:
```python
2018-02-22 09:17:22,046 - scout.server.app - ERROR: Exception on /panels/5a659baf124f63029e8acca1 [POST] [in /home/hiseq.clinical/miniconda/envs/prod/lib/python3.5/site-packages/flask/app.py:1587]
Traceback (most recent call last):
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.5/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.5/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.5/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
[...]

PR #744, opened 2018-02-23
fix issue with HGNC id when adding a new gene to a panel

northwestwitch:
This should fix the bug #741 , hopefully!


scout/server/blueprints/panels/views.py
```diff
@@ -72,12 +72,17 @@ def panel(panel_id):
         raw_hgnc_id = request.form['hgnc_id']
         if '|' in raw_hgnc_id:
             raw_hgnc_id = raw_hgnc_id.split(' | ', 1)[0]
-        hgnc_id = int(raw_hgnc_id)
+        hgnc_id = 0
```
moonso: Nice this should do it!

PR #752, opened 2018-02-27
fix #698 , PDF coverage empty

### LINKED ISSUE #698: PDF coverage report empty
Coverage report PDF-download generates empty reports

northwestwitch:
I think this bug might be caused by the fact that the sample_ids are passed to the function as arguments and not as form inputs. I have no way to test it on my local computer because I don't have chanjo-report running. What do you think @moonso ?

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/actionbar.html
+++ b/scout/server/blueprints/cases/templates/cases/actionbar.html
@@ -202,10 +202,12 @@
     </form>
     <p>Based on: {{ case.panels|selectattr('is_default')|map(attribute='panel_name')|join(', ') }}</p>
     <p>
-      <form method="POST" action="{{ url_for('report.pdf', sample_id=case.individual_ids, level=institute.coverage_cutoff, dl='yes') }}">
+      <form method="POST" action="{{ url_for('report.pdf', level=institute.coverage_cutoff, dl='yes') }}">
         <input type="hidden" name="gene_ids" value="{{ case.default_genes|join(',') }}" />
         <input type="hidden" name="panel_name" value="{{ case.panel_names|join(', ') }}" />
-
+        {% for sample_id in case.individual_ids %}
+				    <input type="text" name="sample_id" value="{{ sample_id }}" hidden>
+			  {% endfor %}
         <div class="form-group">
           <button type="submit" class="btn btn-default form-control">Download PDF</button>
         </div>


moonso:
Hmm I know to little about chanjo report to answer that. I guess we need some way to check if this really work. Either you could try to setup chanjo and chanjo-report locally, which shouldn't be to hard. Or you you could try to connect to the production database and run your branch on that.

northwestwitch:
The function in chanjo-report that is called from the form in scout is this:

@report_bp.route('/report/pdf', methods=['GET', 'POST'])
def pdf():
    data_dict = request.form if request.method == 'POST' else request.args
    # make a PDF from another view
    response = render_pdf(url_for('report.report', **data_dict))

    # check if the request is to download the file right away
    if 'dl' in request.args:
        fname = 'coverage-report.pdf'
        header = "attachment; filename={}".format(fname)
        response.headers['Content-Disposition'] = header

    return response


I think the key might be this line:  "data_dict = request.form if request.method == 'POST' else request.args"

Since the method is POST (in the scout actionbar form ) then the data dictionary is created based on the request.form, but as the code was written the sample ids were provided as arguments (request.args). OMG installing chanjo and chanjo-report on my local computer is going to be painful and I'd rather avoid if there's another way, also because I'll be missing the database part. Let's discuss about it tomorrow! :)

moonso:
Sounds reasonable. Why do you think chanjo etc is so hard to install? It would be great if you could give it a try and see where the problems arise!

northwestwitch:
I don't think the difficulty is in the installation, but the interaction with the rest, don't I need to have access to some actual data (apart from the demo stuff of scout) to create the coverage reports?

moonso:
Sure but then we need to update documentation and add more test data. It would be nice if you could put a little time on it. Also it is good for clinical if you spend some time with chanjo and the interaction with scout since that competence is lost.

northwestwitch:
Ok, I managed to install chanjo-report and connect it to a demo mysql database containing transcripts and coverage for those 3 samples that are in the scout demo. Tested that this fix works and creates both the html report page and the pdf one (both not empty). So yes, @moonso you can merge! I will now update scout readme file with instructions on how to set up the compatibility with chanjo-report, as well as how to fix the demo mysql db chanjo and chanjo-report connect to.

moonso:
Amazing 💠! Great work @northwestwitch

henrikstranneheim:
Fantastic!

northwestwitch:
Thanks guys, but @henrikstranneheim you should be doing the stuff you are supposed to do when pappaledig! ;)

henrikstranneheim:
Let's call it being a modern but multitasking dad, 😄

PR #760, opened 2018-03-07
updated scout documentation with chanjo-report howto

northwestwitch:
I've included also a short description of chanjo and chanjo-report database since we don't have chanjo-report in our git repository. I'm including also two .sql schemas for this db: one is the empty schema that can be used to set up the database from scratch and one is the database containing the coverage data for the 3 demo samples.


henrikstranneheim:
Great! 
@robinandeer Would you consider migrating Chanjo-Report to Clinical genomics?

moonso:
@henrikstranneheim we can just fork it to Clinical and continue development from there, just like we did with chanjo

PR #772, opened 2018-03-15
enable hpo filtering for SVs, fix #710

### LINKED ISSUE #710: Enable HPO list filter for SV list
I have not tested this explicitly yet. Anyone else?

northwestwitch:
This should fix the issue. The problem was that the HPO genes were not passed to the form by the view.

### DIFF
--- a/scout/server/blueprints/variants/templates/variants/variants.html
+++ b/scout/server/blueprints/variants/templates/variants/variants.html
@@ -138,11 +138,11 @@
   {% endif %}
   {% if comment_count > 0 %}
     <a href="#"
-       class="badge pull-right" 
-       data-toggle="popover" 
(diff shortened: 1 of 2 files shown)


northwestwitch:
If you refer to the whitespaces of the file that otherwise I left untouched, I have to be honest with you, I didn't realize I changed that file as well! :)

moonso:
Good stuff @northwestwitch 👍

dnil:
Hotfix on that one perhaps? Easy enough, and adds quite a bit of functionality back to the SV-view.

moonso:
@dnil There is a problem with hot fixes at the moment. We need to make a big update of scout and then we will move over to github flow and make releases at least once per week.

PR #773, opened 2018-03-15
Dismiss variants. Fix #605

### LINKED ISSUE #605: Requested tag from Clinical Genetcs //
Regarding "Clinical report" Clinical Genetics would like to have the following options regarding variants (new)

Too common in public databases
Too common in local database
Not relevant phenotype
Inconsistent inheritance pattern
No compound variant
Not in RefSeq transcript
Does not affect splicing

WE would like to have the opportunity to check several options (krysslista motsvarande ex paneler)
bild

If any chosen we would like to see this as a grey dot (maybe) on the variants page.

dnil:


moonso:
Wow fast work @dnil ! I will need some time to look this through...

scout/adapter/mongo/event.py
```diff
@@ -871,6 +871,57 @@ def update_manual_rank(self, institute, case, user, link, variant,
         logger.debug("Variant updated")
         return updated_variant
 
+    def update_dismiss_variant(self, institute, case, user, link, variant,
+                           dismiss_variant):
+        """Create an event for updating the manual rank of a variant
```
moonso: Update the docstring?

dnil: Ahem, yes. 😊

moonso: 😸

moonso:
Also @dnil , could you add a test or two? The event tests are pretty straightforward, it would be nice to have for this one as well.

dnil:
Added event test. Also got some feedback on the color scheme. Since the info-color is used for e.g. causative transcripts, it is not ideal for marking dismissed variants. Added a new gray to fit the role.

northwestwitch:
Nice!

moonso:
Hi @dnil, looks nice! One question: It looks like the same event will be created regardless if a variant is dismissed or if it is "reset dismiss". It could be a bit confusing when looking at the history of events for a variant or case. Could you update the verbs perhaps?

dnil:
Right; sending an empty list will unset. I thought it simple and robust, but didn't really consider logging.  It has the same granularity as set/unset manual_rank. Has that been a problem?

moonso:
Maybe not a problem. Lets go for this! We can change that one later

Issue #702, opened 2018-01-09
HPO term problem


4WGH:
we got the following answer when trying to add the HPO term: "HP:0012335"

Bad Request

unable to add phenotype: HP:0012335

what does it mean?

this seams to be a general problem, as we tried with several other numbers.

moonso:
Will check

northwestwitch:
@moonso if you have not starting working on this one I can take a look at it!

moonso:
@northwestwitch yes please 🌾 🥋

moonso:
The reason is that the term has been created after we updated the HPO terms in scout.
So the term is not in scout.
I've created a new issue #774

PR #779, opened 2018-04-04
Update hpo terms

moonso:


northwestwitch:
Wow that was a big update, good job @moonso !

northwestwitch:
So this solves the issue #702 (Adding HPO phenotypes not associated to any gene)?

moonso:
Yes

northwestwitch:
Nice, so we can close that issue e insert the add "HP:0012335" to the scout checklist.

northwestwitch:
Ok, it's closed already, but I definitely add it to the checklist, right?

moonso:
Yes sounds good. I've added `scout update hpo` to the checklist

northwestwitch:
Hello again. I was doing some testing with the new master, that contains also this PR, and after the "scout update hpo" command it looks like it is still not possible to add HPO terms not associated to genes (like HPO:0012335). The difference from before is that now the program doesn't crash anymore, but the term is not added to the "Added phenotypes" panel. Am I missing something?

moonso:
skarmavbild 2018-05-28 kl 18 39 29
skarmavbild 2018-05-28 kl 18 40 01
skarmavbild 2018-05-28 kl 18 40 16

northwestwitch:
I believed you even without pics! :)

moonso:
👍 if you want to keep use that you can run `scout --demo update hpo` and then `scout --demo serve`

PR #790, opened 2018-04-17
Clinvar

northwestwitch:
Clinvar submission functionality. Monster pull request, good luck with reviewing it! :)


scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
...
+    """Parses the form from blueprints/variants/clinvar.html
+    and creates the lines for the .Variant.csv clinvar submission file.
+    """
+
+    clinvar_header  = list(CLIVAR_HEADER)
+    optional_fields = dict(CLINVAR_OPTIONAL)
+    clinvars_dict = {}
+
+    if 'all_vars' in form_fields:
+        clinvars = form_fields['local'] #this is a list
```
moonso: 👍 nice with those small comments

scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
...
+
+    variants = [] #one line for each variant
+    for clinvar in clinvars:
+        variant_line = []
+        clinvar_dict = clinvars_dict[clinvar]
+        for field in clinvar_header:
+            variant_line.append('"'+clinvar_dict[field]+'"')
+        variants.append(variant_line)
+
+    #returning a list and a list of lists
```
moonso: Try to return one thing from each function. If you have to return multiple things, pack them into a dictionary like 
```python
variant_lines = {
    'headers': clinvar_headers,
    'variants': variants
}
return variant_lines
```

scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
+import urllib.request
+import sys
+from scout.constants import CLIVAR_HEADER, CLINVAR_OPTIONAL, CASEDATA_HEADER, CASEDATA_OPTIONAL
+
+def get_variant_lines(form_fields):
+    """Parses the form from blueprints/variants/clinvar.html
+    and creates the lines for the .Variant.csv clinvar submission file.
```
moonso: Please elaborate here. And follow the convention with args and returns

scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
...
+    clinvar_header  = list(CLIVAR_HEADER)
+    optional_fields = dict(CLINVAR_OPTIONAL)
+    clinvars_dict = {}
+
+    if 'all_vars' in form_fields:
+        clinvars = form_fields['local'] #this is a list
+    else:
+        clinvars = [form_fields['main_var'][0]] #also a list, but has one element
+
+    for clinvar in clinvars:
```
moonso: If there is any possibility to simplify this function it would be nice. Hard to follow at the moment. Perhaps break it into smaller parts

scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
...
+        variants.append(variant_line)
+
+    #returning a list and a list of lists
+    return clinvar_header, variants
+
+
+def get_casedata_lines(form_fields):
+    """Parses the form from blueprints/variants/clinvar.html
+    and creates the lines for the .CaseData.csv clinvar submission file.
+    """
```
moonso: Elaborate and Args and Returns

scout/parse/clinvar.py
```diff
@@ -0,0 +1,216 @@
...
+                for cd_field in casedata_header:
+                    temp_cd_fields.append('"'+case[cd_field.replace(' ','_')]+'"')
+                casedata_lines.append(','.join(temp_cd_fields))
+
+    return '"'+'","'.join(variants_header)+'"', '"'+'","'.join(casedata_header)+'"', clinvar_lines, casedata_lines
+
+
+def create_clinvar_submission_dict(variant_header, variant_lines, casedata_header, casedata_lines):
+    """
+    Creates a list of variants used for a clinvar submission. The returned list has the following format:
```
moonso: This is great!

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.clinvars_from_clinvarid(submission_id=submission_obj[0]['clinvar_submission']):
+            return "clinvar submission id "+submission_obj[0]['clinvar_submission']+" already exists in database!","danger"
+        elif self.clinvars_from_variantids(ids):
+            return "One of more variants your are trying to save is already present in a clinvar submission!","danger"
+        else:
+            logger.info("Adding clinvar submission id: {0}".format(submission_obj[0]['clinvar_submission']))
+            result = self.clinvar_collection.insert_many(submission_obj)
+            return "variants with ids {0} were saved into clinvar submission collection".format(result.inserted_ids), "success"
+
+
```
moonso: These three functions could be collapsed to one with multiple arguments like

```python
def clinvars(self, variant_ids=None, case_id=None, submission_id=None)
    """Something here"""
    # Check that one of the parameters is used
```

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,127 @@
...
+        ids = []
+        for subm_variant in submission_obj:
+            subm_variant['user'] = user
+            subm_variant['institute'] = institute
+            subm_variant['case'] = case
+            subm_variant['created_at'] = datetime.now()
+            ids.append(subm_variant['_id'])
+
+        if self.clinvars_from_clinvarid(submission_id=submission_obj[0]['clinvar_submission']):
+            return "clinvar submission id "+submission_obj[0]['clinvar_submission']+" already exists in database!","danger"
```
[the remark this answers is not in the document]
northwestwitch: But I need it this way, I'm using the two strings returned directly in the view, where the function is invoked, to flash the message on the HTML page.

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,127 @@
...
+        elif self.clinvars_from_variantids(ids):
+            return "One of more variants your are trying to save is already present in a clinvar submission!","danger"
```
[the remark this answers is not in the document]
northwestwitch: Same here

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,127 @@
...
+        else:
+            return submission_objs
+
+    def add_clinvar_accession(self, variant_id, clinvar_accession):
+        """Introduces the field 'clinvar accession' to a variant submitted to clinvar
+
+            Args:
+                variant_ids(str)
+
+            Returns: the number of updated documents
```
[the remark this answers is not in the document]
northwestwitch: Thanks! That happens when you are afraid to forget the comments and you create them before knowing what the subroutine is supposed to do! :)

scout/constants/clinvar.py
```diff
@@ -0,0 +1,103 @@
+# clinvar Variant sheet
+CLIVAR_HEADER = [
```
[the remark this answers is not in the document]
northwestwitch: Yes, OMG I wrote this everywhere!!!

scout/constants/clinvar.py
```diff
@@ -0,0 +1,103 @@
...
+    '##Local ID',
```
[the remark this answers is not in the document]
northwestwitch: Yes, this has to be written like this. These constants (CLINVAR_HEADER, CLINVAR_OPTIONAL, CASEDATA_HEADER, CASEDATA_OPTIONAL) are the fields of the the submission excel sheet. They named the first field of the variant sheet that way (see full submission template here:https://www.ncbi.nlm.nih.gov/clinvar/docs/submit/#spreadsheet) so I'm saving it with the same nomenclature.

scout/parse/clinvar.py
northwestwitch: But I don't need a dictionary, I'm using them as a list and a list of lists immediately after when generating the CSV files and in several other places in the code. Otherwise I would need to extract them from the object all the time..

scout/parse/clinvar.py
northwestwitch: I know, it looks quite complicated, but trust me, I re-wrote it for this reason and this is the simplest I could get. The fact is that I'm parsing this huge clinvar submission form and to do it in an elegant way I gave the form fields the same names as the clinvar excel sheet columns. These strings are defined in the constants file.
And I need all these loops to understand which columns I have to include in the generated files (both variants file and case data file). For instance if I have to include 4 variants, some with optional data, some no. Then I have to check if any of the variants has a field filled in and in that case I have to include the field in the document header and an empty cell for the variants where the field is blank instead. On the other hand, if the field is not used I have to remove it from the header so I don't print the column at all.

I think this explanation might be more complicated than the code itself! :)

Anyway, I've tried to document these 2 subroutines a bit better. We can discuss about it!

scout/adapter/mongo/clinvar.py
northwestwitch: Yes, you are right, much better this way!!

moonso:
Hi @northwestwitch ! Could you add some tests?

northwestwitch:
Sure, but I don't know how to do that. Could you point me to a good example in scout where there are tests, so I get an idea?

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -423,6 +427,43 @@ <h4 class="modal-title">Edit synopsis</h4>
   </div>
 {% endmacro %}
 
+{% macro clinvar_panel() %}
+  <div class="panel panel-default">
+    <div class="panel-heading">Variants submitted to clinvar</div>
```
dnil: "Clinvar submissions" perhaps - could be listing unsubmitted ones as well?

northwestwitch: Yes, of course. The only way I have to know if it was submitted is that the variant has been published and the clinvar ID is saved in scout database. But we want to keep it broad here. Then I'll change it also in the cases page where under "Clinvar" the label says "submitted". I'll write instead "Clinvar submissions" and the label "created"

dnil: Right, thanks! Otw agree with Måns we should have a few tests - perhaps at least for the mongo adapter part?

moonso:
@northwestwitch you can check out the exing tests under `tests/adapter` for example

northwestwitch:
Thanks, I'll do that!

dnil:
Also the Platform-related fields and testing lab should probably be like the Method fields - prefilled but editable: in case more different techniques have been used, we might want to reflect that?

northwestwitch:
Sure, it's easy to modify. Are you satisfied with the pre-filled values?

scout/server/blueprints/variants/templates/variants/clinvar_update.html
```diff
@@ -0,0 +1,146 @@
...
+      {{ institute.display_name }}
+    </a>
+  </li>
+  <li>
+    <a href="{{ url_for('cases.case', institute_id=institute._id, case_name=case.display_name) }}">
+      {{ case.display_name }}
+    </a>
+  </li>
+  <li>
+    {% if cancer %}
```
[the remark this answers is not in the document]
northwestwitch: Not my doing, it's in many pages already!

northwestwitch:
Test still in progress but I've the rest of the changes, plus a couple of improvements : 
1) if you click on the submission link from the case page you arrive directly to the submission page.
2) you can now save the clinvar variation id for one variant at the time

northwestwitch:
Hi guys, I've introduces the most important tests: 
1) That it creates a valid csv Variants file from the input form.
2) That it creates a valid csv Casedata file from the input form.
3) That these fields are converted in valid submission objects to be saved in scout database.

Today I've fixed the code to include the submission of structural variants, so one can either submit a list of snvs, a list of SVs or SVs and snvs together.

Let me know!

northwestwitch:
YESSSSS! After 6 hours of painful attempts and a big headache I finally managed to understand the whole concept behind pytest and how to use mongomock, fixtures, etc. (I think!) This morning all tests failed because I didn't create a clinvar_database in the conftest.py, but look at the Travis now!!! :D 
I also rewrote almost everything in the parser, as requested by @moonso . Same for the subroutines in the adapter, that don't return flash messages any more, but objects and numbers instead.

henrikstranneheim:
🎈 🎉

scout/constants/clinvar.py
```diff
@@ -0,0 +1,125 @@
...
+    'Method',
+    'Method purpose',
+    'Method citations',
+    'Software name and version',
+    'Testing laboratory',
+    'Date variant was reported to submitter'
+]
+
+# clinvar CaseData sheet, optional fields for clinvar but some are pre-determined for clinvar scout form.
+CASEDATA_OPTIONAL = {
```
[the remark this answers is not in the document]
northwestwitch: It's optional fields and they are set to false by default. Their value changes if the user fills in the respective field in the html form. In the end, when I loop over the fields to use in the downloadable files, I use only those that are not False.

northwestwitch:
Oh come on, I've just added a line of code (a comment!) and now I have the same tests failing on this pull request as well!!

northwestwitch:
There is definitely stuff I need to learn about testing. To make it pass the Travis-CI all I did is removing some cache folder, et voila, it worked now.

northwestwitch:
Im closing this pull request and I'll created another with the re-based code

PR #800, opened 2018-05-18
Case report tool

northwestwitch:
Created a case report page, accessible by a link on the sidebar at the case level. It collects useful info on the case, gene panels used in the analysis, and variants falling in these  categories: causatives, pinned, ACMG-classified, commented, manual ranked and dismissed. It's possible to create a pdf report for the view by clicking on the link on the upper right corner of the report page.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,6 +16,8 @@ Flask-Babel
 livereload
 python-dateutil
 pymongo>=3.6.0
+flask_weasyprint
+
 
 pathlib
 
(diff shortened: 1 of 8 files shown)


northwestwitch:
That's weird. I just pushed some changes not related with testing (the test for my pull request passes) but now Travis fails over a test named "test_update_caseid", that has to do with a disrupted real_oldcase_database fixture apparently. Could it be that this happens because we stopped ( and removed? ) the migrate-cases.sh script ?? @ingkebil @patrikgrenfeldt @moonso

northwestwitch:
It's ok, I'll find the problem and I'll fix it. So please wait until then for reviewing the pull request!

PR #806, opened 2018-05-24
Cosmic ID link additon

hassanfa:


scout/parse/variant/variant.py
```diff
@@ -175,6 +176,12 @@ def parse_variant(variant, case, variant_type='clinical',
             dbsnp_ids.add(dbsnp)
         for cosmic in parsed_transcript.get('cosmic', []):
             cosmic_ids.add(cosmic)
+    
+    # The COSMIC tag in INFO is added via VEP and/or bcftools annotate
+
+    COSMIC_tag = variant.INFO.get('COSMIC')
```
moonso: Please don't use capitals in variable names, it is not done anywhere else.
(I have to make some remark 😈 )

scout/server/blueprints/variants/controllers.py
```diff
@@ -462,6 +463,19 @@ def swegen_link(variant_obj):
                     "{this[position]}-{this[reference]}-{this[alternative]}")
     return url_template.format(this=variant_obj)
 
+def cosmic_link(variant_obj):
+    """Compose link to COSMIC Database."""
```
moonso: Could you create a proper docstring here? I know it is probably lacking on the other functions here but let's work together and improve all code. 
It should look something like:
```python
def cosmic_link(variant_obj):
    """Compose link to COSMIC Database.

    Args:
        variant_obj(scout.models.Variant)

    Returns:
        url_template(str): Link to COSMIIC database if cosmic id is present
    """
```

docs/user-guide/annotations.md
```diff
@@ -48,11 +48,19 @@ The [PHASTcons][phastcons] conservation string.
 
 The [phylop][phylop] 100 way predicted conservation string.
 
+## External ID
```
moonso: Great that you added some documentation 🎆

moonso:
👍 👍

dnil:
Agree: :+1:

PR #807, opened 2018-05-25
Clinvar2

northwestwitch:
The same code as the old PR, but rebased to master.


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-- Added an analysis report page (html and PDF format) containing phenotype, gene panels and variants that are relevant to solve a case.
```
moonso: Why is this line not in the rebased version? This line comes from your earlier PR, right? Then it should be included if you have rebased this PR...

northwestwitch: It was included but I removed it at the conflict resolve. I'll put it back!

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,104 @@
...
+LOG = logging.getLogger(__name__)
+
+class ClinVarHandler(object):
+    """Class to handle clinvar submissions for the mongo adapter"""
+
+    def add_clinvar_submission(self, submission_obj, user, institute, case_id):
+        """Add a clinvar submission for a case. A submission consists of one or more variants added to the clinvar collection.
+
+            Args:
+                submission_obj(str)
```
moonso: The submission object is not a string, please change and describe more.
Is is just a dictionary with variants or is there any extra information?

scout/adapter/mongo/clinvar.py
```diff
@@ -0,0 +1,104 @@
...
+                user(str)
+                institute(str)
+                case_id(str)
+
+            Returns:
```
moonso: It looks like it can also return 0 and -1, what does that mean?

scout/parse/clinvar.py
```diff
@@ -0,0 +1,202 @@
...
+                    csv_line.append('""') # empty cell
+
+        if csv_line not in csv_lines and len(csv_line)>0:
+            csv_lines.append(csv_line)
+
+    return csv_lines
+
+
+def extract_submission_csv_lines(clinvars_dictlist):
+    """Parses a list of clinvar submission objects (variants) and creates the lines for printing
```
moonso: There are great docstrings above! Could you add the same information here?

scout/parse/clinvar.py
```diff
@@ -0,0 +1,202 @@
...
+                temp_cd_fields = []
+                for cd_field in casedata_header:
+                    temp_cd_fields.append('"'+case[cd_field.replace(' ','_')]+'"')
+                casedata_lines.append(','.join(temp_cd_fields))
+
+    return '"'+'","'.join(variants_header)+'"', '"'+'","'.join(casedata_header)+'"', clinvar_lines, casedata_lines
+
+
+def create_clinvar_submission_dict(variant_header, variant_lines, casedata_header, casedata_lines, variant_types):
+    """
```
moonso: Also nice with a lot of information here, please add `Args` and `Returns`.
I mean you could still keep this detailed information and add Args and Returns afterwards. It is nice if all docstrings has the same structure

scout/server/blueprints/variants/controllers.py
```diff
@@ -732,6 +741,35 @@ def cancer_variants(store, request_args, institute_id, case_name):
     )
     return data
 
+def clinvar_export(store, institute_id, case_name, variant_id):
+
```
moonso: What about docstrings here? Could you add that as well?

tests/adapter/mongo/test_clinvar_handler.py
```diff
@@ -0,0 +1,107 @@
+import logging
+import pymongo
+from numbers import Number
+
+LOG = logging.getLogger(__name__)
+
+def get_test_submission_object():
```
moonso: Great work on the tests 🌮

northwestwitch:
I've fixed the code according your comments @moonso. OMG the comments in the mongo adapter were sooo bad!! :)

PR #808, opened 2018-05-28
fixed small visualization issue in case report

northwestwitch:
Fixed a visualization issue. Nothing serious but it looks nicer now when a case doesn't have comments.


northwestwitch:
@moonso could you check and merge this before the parse-exons tomorrow?

northwestwitch:
Thanks!

PR #811, opened 2018-05-30
Make cases limit changeable.

### LINKED ISSUE #810: Showing more cases in list
I would like to be able to browse through more cases..

dnil:
Fixes #810


scout/server/blueprints/cases/views.py
```diff
@@ -32,9 +32,14 @@ def cases(institute_id):
     """Display a list of cases for an institute."""
     institute_obj = institute_and_case(store, institute_id)
     query = request.args.get('query')
+
+    limit = 100
```
moonso: 👍

PR #820, opened 2018-06-11
fixed missing key for SV ensembl link

northwestwitch:
Fixed the broken link to ensembl for genes in structural variants

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -89,6 +89,16 @@ def sv_variant(store, institute_id, case_name, variant_id):
     overlapping_snvs = (parse_variant(store, institute_obj, case_obj, variant) for variant in
                         store.overlapping(variant_obj))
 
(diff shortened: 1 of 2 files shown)


northwestwitch:
@moonso a lot of annotations are missing for genes in SVs, including the ensembl link. This should fix just that missing key. I don't think we need all the other gene annotations for now..

moonso:
Hi @northwestwitch I think we should have a closer look at this problem. To me it seems like we have two controller functions that does almost the same thing (variant and sv-variant) Maybe we can solve the problem my merging the two functions instead.
We can sit down together and solve this someday this week?

northwestwitch:
Not this week (Tuesday-Thursday I'm out of town for a course) and not for this next scout release (this version if tested this week by our team and next week by the clinicians). But I agree, the code could be optimized, even if I think it is fine to keep SNVs and SVs separate, especially if they are distinct pages with distinct content in the browser.

northwestwitch:
We could merge this one as well, then we'll think how to improve the code in a later moment

PR #824, opened 2018-06-12
fixed disappeared transcripts in case report

northwestwitch:
Bug fix for the disappeared transcripts for genes in the case report. Realized that they were missing from this page as well!!


northwestwitch:
This fix is needed for the next release, so I'll merge and update scout stage.

PR #827, opened 2018-06-13
bug fix #825

### LINKED ISSUE #825: List comprehensions produce different types
Hi @northwestwitch ! I found a strange thing in `scout/server/blueprints/cases/controllers.py`.

The function looks like this (cropped)
```python
def case(store, institute_obj, case_obj):
    """Preprocess a single case."""
    case_obj['individual_ids'] = []
    for individual in case_obj['individuals']:
        try:
            sex = int(individual.get('sex', 0))
        except ValueError as err:
            sex = 0
        individual['sex_human'] = SEX_MAP[sex]
        individual['phenotype_human'] = PHENOTYPE_MAP.get(individual['phenotype'])
        case_obj['individual_ids'].append(individual['individual_id'])

    case_obj['assignees'] = [store.user(user_email) for user_email in
                             case_obj.get('assignees', [])]
    suspects = [store.variant(variant_id) or variant_id for variant_id in
                case_obj.get('suspects', [])]
    causatives = [store.variant(variant_id) or variant_id for variant_id in
                  case_obj.get('causatives', [])]
```
When creating `suspects` and `causatives` above it says 
```python
store.variant(variant_id) or variant_id
```
This means that we either get a list with variants (dictionaries) or variant_ids (strings) or a mix of those.
The problem arises later in the code when the result is sent to the function `variants_filter_by_field` in
`scout/server/blueprints/variants/controllers.py`. That one looks like:

```python
def variants_filter_by_field(store, variants, field, case_obj, institute_obj):
    """Given a list of variant objects return only those that have a key specified
      ....
    """
    filtered_variants = []
    # Check if the variants have information if "field"
    for var in variants:
        if var.get(field):
            # Add more details to the variant
            if var['category'] == 'snv':
                var_object = variant(store, institute_obj, case_obj, var['_id'])
            else:
                var_object = sv_variant(store, institute_obj['_id'], case_obj['display_name'], var['_id'])

            filtered_variants.append(var_object['variant'])

    return filtered_variants
```
It loops over `variants` and expects them to be dictionaries since it does `.get`. This fails when the objects are strings.

Could you fix?

northwestwitch:
Fixed code for bug when variant passed is a string (variant_id) and not a dictionary

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -122,7 +122,7 @@ def case_report_content(store, institute_obj, case_obj):
     data.update({'causatives_detailed': causatives})
 
 
-    pinned_raw = data['suspects']
+    pinned_raw = data['suspects'] # This should be a list of dictionaries, but might contain elements that are strings (variant_ids)
 
     #remove any element which is not a dictionary from the list above
     pinned_raw[:] = [var for var in pinned_raw if isinstance(var, dict)]
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -17,7 +17,7 @@ STATUS_MAP = {'solved': 'bg-success', 'archived': 'bg-warning'}
 
 def cases(store, case_query, limit=100):
     """Preprocess case objects."""
-    
+
     case_groups = {status: [] for status in CASE_STATUSES}
     for case_obj in case_query.limit(limit):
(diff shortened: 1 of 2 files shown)


northwestwitch:
@moonso this should fix #825

moonso:
Hi, it is better if you make sure that the lists only contains dictionaries. Then the function will be more reusable and it will be easier to understand what happens.

northwestwitch:
@moonso what about this solution?

PR #828, opened 2018-06-13
Multiqc

moonso:
Rebased and opened to `master` instead of `3.3.0`. It would be really nice to have this feature in scout.


moonso:
Fixed!

PR #842, opened 2018-06-15
Use POST requests to transfer queries. Proof-of-concept with clinical…

dnil:
…_filter moved.

Not well tested, but shows the concept.


dnil:
Tested with large gene lists.
Remaining get-calls of primary interest:
- [ ] Next Page button
- [ ] HPO-terms menu item (should by the way not reset the entire filter, just add the HPO panel?!)
- [ ] SV variants...

moonso:
Where are we with this PR @dnil ? Are you ready for merge?

dnil:
..and unfortunately that HPO panel had some unusual interactions with the way the query adaptor forms mongo queries. Only good to get that fixed, but will not be finished today either. Next week then! Glad midsommar!

dnil:
Ok, I think this is ready for review now. The SV part is still pending, but ideally that should be very straight forward and not use anything conceptually different.

moonso:
I will review as soon as I get some time...

dnil:
No worries - I'm almost through with the SV-part as well. Found an issue with the panel-upload as well; it keeps wanting to get rather than POST. 😠

northwestwitch:
So @dnil we wait for you to finish with the SV part?

moonso:
Hi @dnil ! Where are we with this one?

dnil:
Unsurprisingly in conflict it seems! The post part works just fine; I fixed the SV as well on my local, but it was unexpectedly hard to integrate with the upload-gene-list-on-the-fly fix from before. I can give you a a version of this without the latter, if you wish, otherwise I must find a couple of hours of quiet to get to the bottom of how WTF handles return of data from multiple forms on one page, specifically how to preserve the state of the filter form after uploading a gene list.

moonso:
Ok, I can not really decide which one is the best. Since there seems to be no rush we can wait till you get some time.

moonso:
Easiest thing for me would be that you work as a collaborator and open pull requests directly to the scout repo. If you do like that I can handle the merge conflicts.

moonso:
It would be really nice to merge this, it will probably be harder and harder with time

dnil:
Right, so like #945 - does that look ok? If so, I'll give this a few h tomorrow.

dnil:
Ok, so closing this now and continuing in  #950.

PR #852, opened 2018-06-15
Fix evaluated variants

moonso:


moonso:
Fixed! Can you approve @northwestwitch ?

PR #854, opened 2018-06-15
bug fix #853

### LINKED ISSUE #853: Clinvar submission form crash
For variant https://scout-stage.scilifelab.se/cust000/15078-miptest/df781c32e54aeb0d0c79106464536401

northwestwitch:
Shortening as well variant names when the simple_name is longer than 25 characters. Fixed a typo as well.


northwestwitch:
Yes, that worked. The problem was due to the fact that whenever the simple_name of a variant was too long the program would use the chromosome + cytoband start + cytoband stop. I didn't know that some variants don't have these keys and that would trigger an error.

Issue #845, opened 2018-06-15
Add if a variant has sanger ordered in the case report


northwestwitch:

northwestwitch:
Fixed

Issue #844, opened 2018-06-15
Show just the default panel in case report


northwestwitch:

northwestwitch:
Or those checked in the list, even better

northwestwitch:
Fixed

PR #855, opened 2018-06-18
Add sanger status and only default panels to case report

northwestwitch:
Included info on Sanger validation for variants in case report (issue #845) and adding to report only default gene panels (issue #844)


northwestwitch:
This shouldn't really affect anything so I'll just merge

PR #860, opened 2018-06-18
Readd missing load report

patrikgrenfeldt:


moonso:
Hi there are several things here that are outdated so this will not work. I can work on fixing these

PR #867, opened 2018-06-20
Readd missing load report

patrikgrenfeldt:
Also ready for merge with beta as far as I understand


scout/demo/643594.config.yaml
```diff
@@ -47,6 +49,8 @@ peddy_ped: scout/demo/643594.peddy.ped
 peddy_ped_check: scout/demo/643594.ped_check.csv
 peddy_sex_check: scout/demo/643594.sex_check.csv
 
+delivery_report: scout/demo/delivery_report.html
```
patrikgrenfeldt: This path should be a full path, how can we solve that?

moonso: Hi, when parsing the yaml file you can do something like
```python
import os

full_path = os.path.abspath(config['delivery_report'])
```

northwestwitch:
Shall we merge?

ingkebil:
The code looks solid to me. I'm a bit hesitant to say go ahead as this is
not my code base. If Team Scout this is solid as well, yeah, :shipit:

wkr,
Kenny
--
Composed and sent on the road

scout/adapter/mongo/case.py
```diff
@@ -325,12 +328,37 @@ def update_case(self, case_obj):
...
             },
-            return_document = pymongo.ReturnDocument.AFTER
+            return_document=pymongo.ReturnDocument.AFTER
         )
 
         LOG.info("Case updated")
         return updated_case
-        ##TODO Add event for updating case?
+
+    def replace_case(self, case_obj):
```
moonso: I think it is a good idea to have both `replace_case` and `update_case`. Perhaps this might change in the future

scout/exceptions/database.py
```diff
@@ -2,23 +2,33 @@
...
+    pass
+
+
+class DataError(DatabaseError):
+    """Exception raised for errors that are due to problems with the processed data like division by
+     zero, numeric value out of range, etc. It must be a subclass of DatabaseError."""
+    pass
+
+
+class DataNotFoundError(DataError):
```
moonso: 👍 should be used in more places

scout/load/report.py
```diff
@@ -1,33 +1,53 @@
...
+    """Load a delivery report into the database
+
+    If the report already exists the function will exit.
+    If the user want to load a report that is already in the database
+    'update' has to be 'True'
 
-def load_report(adapter, case_id, report_path, update=False):
-    """Add the path to a report to a case
-    
     Args:
```
moonso: Please use [napoleon](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) style on docstrings which is used through the whole package. Also mentioned in [development](https://github.com/Clinical-Genomics/development/blob/6942b35731c169b49f67500a57bd75afa050cb0b/documentation/README.md)

tests/load/test_load_case.py
```diff
@@ -37,3 +38,19 @@ def test_load_case_rank_model_version(case_obj, panel_database):
 
     assert loaded_case['rank_model_version'] == case_obj['rank_model_version']
 
+
+def test_load_case_delivery_report(case_obj, panel_database):
```
moonso: 👍

scout/load/report.py
patrikgrenfeldt: I get your general idea, use napoleon style on docstrings. But what specifically is it that you want? I copied this docstring from another method, was it napoleon or should it have been napoleon, I don't follow.

scout/adapter/mongo/case.py
patrikgrenfeldt: ok, we'll leave it like that for now then

scout/demo/643594.config.yaml
patrikgrenfeldt: Do you think we should change the implementation in order to satisfy the demo-case-config delivery report? I guess we could skip it also

scout/exceptions/database.py
patrikgrenfeldt: You are welcome to use it wherever :)

scout/load/report.py
ingkebil: I presume Mans means to remove the `:param` from the statements, e.g.

`:param report_path: (string) path to report ` -> `report_path (string): path to report`

patrikgrenfeldt: Fixed :)

PR #875, opened 2018-06-22
pin flask-pymongo 0.5.2

dcrosta:
In about the next month, Flask-PyMongo 2.0 is going to be released, which is not (necessarily) compatible with 0.x versions. This change will prevent accidental updates, but I'd also be happy to work with you on upgrading safely to 2.0, which will be the base version going forward that will receive compatibility and bug fix improvements, etc.

In particular, it looks like you are configuring with MONGO_HOST, MONGO_DBNAME, etc. To use Flask-PyMongo 2.0, you should convert to using MONGO_URI. It should be pretty straight forward, and I'll be happy to lend a hand if it helps.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 
 Flask>=0.10
 Flask-DebugToolbar
-Flask-PyMongo>=0.5.1
+Flask-PyMongo==0.5.2
 Flask-Bootstrap
 path.py
 Flask-OAuthlib


northwestwitch:
Hi, and thanks for the tip!
I've checked which version of Flask-PyMongo we are using in our production environment and I see that we have the 0.5.1. In our testing environment (named "stage", for reference) we are using version 0.5.2 instead and it's working just fine.
Our plan with Scout is releasing a new version of the software immediately after summer, so it would be great to fix the connections to use PyMongo 2.0. But considering the approaching holidays and all there isn't so much time to do that at them moment, so for the time being I'll merge you pull request and open a dedicated issue. If you feel like you have the time to help us with this it'd be great, otherwise someone will pick it after summer. The important is that Scout doesn't crash after the update, but we have it covered.

Thanks again!!! :)

dcrosta:
Cool -- I'll follow #876, please let me know how I can help!

PR #877, opened 2018-06-25
show gene HGNC id in gene, genes, and gene panels pages

northwestwitch:
Mostly useful when, from a case view, you want to check the coverage for a certain gene of a panel. It can be used as a reference for the customized search on the coverage report. Addresses issue #865

### DIFF
--- a/scout/server/blueprints/genes/templates/genes/genes.html
+++ b/scout/server/blueprints/genes/templates/genes/genes.html
@@ -29,6 +29,7 @@
               <td>{{ gene.hgnc_id }}</td>
             </tr>
           {% endfor %}
+          </tbody>
         </table>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/genes/templates/genes/genes.html
```diff
@@ -17,15 +17,19 @@
...
-            <li class="list-group-item">
-              <a href="{{ url_for('genes.gene', hgnc_symbol=gene.hgnc_id) }}">
-                {{ gene.hgnc_symbol }}
-              </a>
-            </li>
+            <tr>
+              <td><a href="{{ url_for('genes.gene', hgnc_symbol=gene.hgnc_id) }}">{{ gene.hgnc_symbol }}</a></td>
+              <td>{{ gene.hgnc_id }}</td>
+            </tr>
           {% endfor %}
```
dnil: End Table body? 😉

northwestwitch: Oh yes, thanks!

PR #878, opened 2018-06-28
Introduced a BETA IGV alignments visualization tool alongside the Pileup one

northwestwitch:
It's probably safer testing it for a while before replacing the current Pileup viewer with this one.


northwestwitch:
I don't understand, all test pass on my local computer..

moonso:
Hi @northwestwitch , nice work! I will try to have a look at why tests fails.
Good idea to keep pileup for a while until we are sure that this works

northwestwitch:
Thanks! I'll be also trying to understand what's wrong at the tests level. And I'll update the changelog.

moonso:
@northwestwitch it works fine on my local instance as well!! 
I added a line to handle MT variants.
You should fetch the branch to rasta and test on stage as well. Great job 👌!

northwestwitch:
Thanks, but we have to pass the Travis CI first :(

moonso:
I think we should lock all versions in `requirements.txt` to use the versions that work in our local installation. The error could be that travis downloads the latest versions of some packages like `pymongo` etc

northwestwitch:
yes, I think that might be the case. I've tried to restart some old builds (once passing without problems) and everything seems to fail on Travis now.

northwestwitch:
Python 3.7 was released yesterday. Do you believe in coincidences??? :)

northwestwitch:
Fixed in #c21cf13

northwestwitch:
Fixed by freezing the pymongo dependency to 3.6. I've patched master as well. @patrikgrenfeldt you might want to pull or fix the dependencies file in your pull request.

northwestwitch:
I think I'm done with this feature. I'll need to test it on stage tho.

northwestwitch:
I'll merge and update the code of Scout stage with this PR, so we can test it there. I think it would be nice to have the GV viewer in the newest Scout release in August. To be on the safe side I've introduced the link to the viewer with the label "BETA"

henrikstranneheim:
That would be fantastic!

northwestwitch:
I've updated the code but when I click on the igv viewer link I get a "BAD REQUEST" error. I'll fix it!

PR #881, opened 2018-07-02
Igv viewer

northwestwitch:
Added the code to send chunks of files to IGV viewer


northwestwitch:
Fixed in #882 . The IGV viewer works in scout stage, at least for snvs. Tomorrow I'll check (and eventually fix) the SVs.

northwestwitch:
Available reference genomes:
https://github.com/igvteam/igv/blob/master/genomes/genomes.txt

PR #882, opened 2018-07-02
Modified reference name to match the one in case collection

northwestwitch:
Demo version has case.genome_build = '37' while in scout database the reference is named 'GRCh37'


northwestwitch:
Should hopefully help in fixing the igv viewer

PR #892, opened 2018-07-09
fix #891

### LINKED ISSUE #891: SCOUT-STAGE: global variant comments are not shown globally
Example: variant X_67414426_C_T_clinical has a global comment in cust002 cases that does not show up in a cust002 case.
[cust002 case 1](https://scout-stage.scilifelab.se/cust002/F0015268/5f1a00e07c8e84db619883bbc1f30a9d)
[cust002 case 2](https://scout-stage.scilifelab.se/cust002/F0015456/7f870a4f94db24ef34b14bc04edc99e7)
[cust003 case](https://scout-stage.scilifelab.se/cust003/126/4ce6af6a0d8e0849ffd438a5d86ec560)

Expected behavior: the global comment from the cust002 cases should show up in the cust003 case as well.

northwestwitch:
This PR modifies how variant comments are collected:
* Global comments for a variant should be shown across cases and institutes
* Case-specific variant comments should be shown just in that specific case of that institute


scout/adapter/mongo/event.py
```diff
@@ -79,26 +79,45 @@ def events(self, institute, case=None, variant_id=None, level=None,
...
-        # If no variant_id or panel we know that it is a case level comment
+            if comments:
+                # If it's comment-related event collect global and variant-specific comment events
+                query = {
+                    '$or': [
+                        {
+                            'category' : 'variant',
+                            'variant_id' : variant_id,
+                            'verb' : 'comment',
+                            'level' : 'global'
```
moonso: Hi! I don't understand, you never use the `level` parameter so this function will always return all comments for a variant...
I think you should add some tests for this to be sure that the behaviour is as expected. This one is a bit tricky so ot would be nice to have tests for the future.

northwestwitch: Why don't use 'level' ? If it's specified in the query isn't it used? 
The way the query is written should make it return:
 * global comments for variants with that variant_id (regardless of case and institute) 
 * specific comments for that variant_id, from that case and that institute.

What it DOES NOT RETURN are specific comments for that variant id but for other cases.

I'll write some tests as soon as I have some time, I'm trying to fix the aligners bug today.

moonso: Aha great! I misunderstood, you are right.

northwestwitch:
I think I'm done with this PR, if it passes the Travis CI we're good.

moonso:
Looks nice @northwestwitch , is it ok if I do some modifications to your test before merging?

northwestwitch:
Yeah, go on! :)

scout/adapter/mongo/event.py
```diff
@@ -79,26 +79,45 @@ def events(self, institute, case=None, variant_id=None, level=None,
...
 
-            if case:
-                query['case'] = case['_id']
+                if case:
+                    query['case'] = case['_id']
 
-        if comments:
-            query['verb'] = 'comment'
 
         return self.event_collection.find(query).sort('created_at', pymongo.DESCENDING)
```
moonso: @northwestwitch is this the query that takes to long when building the report?

northwestwitch: It might be, but we could be able to fix the issue by increasing the timeout of gunicorn. But if you feel like optimizing the query (by adding an index here?) please do!

moonso: I will at least give it a thought:)

northwestwitch:
Hi @moonso how is it going with those modifications you said you were doing? Because it'd be nice if we could merge this when they fix the prod environment issue and update the software..

moonso:
Hi @northwestwitch, i will sit down with this today

northwestwitch:
Nice, let me know if you want to discuss about it!

scout/adapter/mongo/event.py
northwestwitch: This was not an issue for the creation of the case report. We solved that already, so this query is fine!

PR #904, opened 2018-07-12
fix #903

### LINKED ISSUE #903: Don't show link to alignment view for SVs if bam files are not available
The link for the alignment viewers is shown for SVs even if bam files for the sample are not available. the links lead to error pages (of course!)

northwestwitch:
I don't know if we call SVs in MT DNA, but I adjusted the code to take care of this as well.


henrikstranneheim:
Excellent! We do via Tiddit!

PR #905, opened 2018-07-13
moved igv and pileup in a common subfolder

northwestwitch:
If we decide to keep both IGV and Pileup, since that are very similar and even share common subroutines it makes a lot of sense to have them in the same subfolder.


northwestwitch:
This should be fixed to display the alignments for the breakpoints of SVs, as in #914

dnil:
Oh, I didn't see this one in time before merging #914. Feel free to back that up a step if the reverse merge order would have been easier. Otw the new sv breakpoint links now need an overhaul here.

northwestwitch:
It's OK, it is easy to fix! :)

PR #914, opened 2018-08-13
re-introduced breakpoint alignments for SVs

### LINKED ISSUE #910: Display again breakpoint alignments for SV
Third point of #909

northwestwitch:
I removed the alignments view for the breakpoints by mistake when I introduced the IGV viewer! Fix #910 

Since the line was a bit busy I've introduced a left and right breakpoint line. Or should I name them breakpoint 1 and breakpoint 2 instead?

image


dnil:
Thanks! 1 and 2 or first and second sounds more to the point.

PR #918, opened 2018-08-20
adding jinja2 'do' extension

### LINKED ISSUE #917: Report visualisation fails on current production 4.0.0??
Another one interaction issue, somewhere between a production issue and a potential scout bug. 

While everything looked fine on stage-scout we seem to have an issue with reports on the production version? Anyone around to check logs when @ingkebil is away? Takes a long time, so could be a timeout, but as this was fairly quick on stage I'm leaning toward something else going amiss? 

Cheers!

northwestwitch:
The idea is to fix #917.
It is weird that jinja2 extensions are found in scout stage but not in scout production..!

### DIFF
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -31,6 +31,7 @@ def create_app(config_file=None, config=None):
     """Flask app factory function."""
     app = Flask(__name__)
     app.config.from_pyfile('config.py')
+    app.jinja_env.add_extension('jinja2.ext.do')
     if config:
         app.config.update(config)
     if config_file:


dnil:
Nicely fixed! But yes, weird. Is this something one could load manually outside / global config for the site perhaps?

northwestwitch:
I have no idea. But globally is good, I think I'm using the 'do' thing somewhere else as well. Let's check if it works!

northwestwitch:
@ingkebil we need to update scout from master after this fix. Do I run "bash update-scout-prod.sh" from the usual place?

ingkebil:
Yup, on clinical-db, usual place.

wkr,
Kenny
--
Composed and sent on the road

northwestwitch:
Thanks! :)

Issue #922, opened 2018-08-22
Generate pdf from gene panel


dnil:
We would like to be able to export gene panels (of a given version) to pdf, for easy retrieval and linking from answer reports. Add a weasy-pdf view for the panels?

northwestwitch:
I'm working on this. Would you like to have the HGNC ids of the PDF report to be linked to some other external pages, like HGNC database, NCBI .., or the scout gene page?

dnil:
Thank you! Yep, external pages is the way to go: the beneficiaries would not have scout access and HGNC/genenames, ensembl, Ncbi, Omim or panelApp should all be nice sources.

northwestwitch:
HGNC/genenames, ensembl, Ncbi, Omim etc is all contained in the genenames archive, so for now I'll just add a link to that, then we can discuss about other useful links to include

moonso:
Closed by #932

PR #932, opened 2018-08-24
Export gene panels to PDF

northwestwitch:
A super simple export function to save gene panels to PDF.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ $scout update hpo
 - Fix problem with parsing gnomad from VEP
 
 ### Added
+- Added a PDF export function for gene panels
 - Added a "Filter and export" button to export custom-filtered SNVs to CSV file
 - Dismiss SVs
 - Added IGV alignments viewer
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -22,10 +22,9 @@ LOG = logging.getLogger(__name__)
 
 class PanelHandler(object):
 
-    def load_panel(self, path, institute, panel_id, date, panel_type='clinical', version=1.0, 
+    def load_panel(self, path, institute, panel_id, date, panel_type='clinical', version=1.0,
                    display_name=None):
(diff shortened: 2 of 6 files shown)


northwestwitch:
image

northwestwitch:
image

northwestwitch:
Relative to issue #922

northwestwitch:
The HTML code used to build the PDF looks a bit nicer but I think that the PDF is not that ugly

henrikstranneheim:
Nice!

dnil:
Well done! Thats solves the issue. Do you think it would get too sluggish or visually busy showing what other panels the gene is on, like on the variant page?

northwestwitch:
We could try! I wouldn't like to print all panels the gene belongs to on
the page because it would be a bit messy, but maybe a link that brings out
a pop-up when you touch it or something..

northwestwitch:
If we include the html version I mean, otherwise we could put links to the
panels on the pdf. The thing is: how useful would be to know that a gene is
part of a gene panel namething short (otherwise for the genen in many
panels the info won't fit)?

northwestwitch:
I meant "named something short", I'm writing from my cell phone and I don't
have saved my github password here to edit stuff, sorry! :)

dnil:
The number of panels it is on might also be nice enough, but most users could guess these acronyms. It is just a bonus feature: the main issue being publishing what genes were part of the test. But if I were the doctor choosing between a few partly overlapping panels to order, it would help to be able to quickly see which panel has the best complement of genes.

Actually, why don't we just accept and merge this, and make that an update issue?

northwestwitch:
I know! We could write the acronyms and we explain them at the end of the
report!

northwestwitch:
I've introduced another field in the export table to show all panels a gene belongs to, with a footer explaining it. Take a look at it when you have time @dnil!

moonso:
Looks great, is this ready @northwestwitch ?

northwestwitch:
Ready!

PR #941, opened 2018-09-10
Fix view cases

moonso:


northwestwitch:
This run was forbidden, even! :D

PR #945, opened 2018-09-12
Fix #924. Default panel for SV and cancer views.

### LINKED ISSUE #924: SV variants should default to default panel
Default filter is not enabled on start on SV-variants filter.

If I remember, I'll fix this in #842, whenever, but just in case.

dnil:
Mostly testing to use the central repo. But does fix #924.

### DIFF
--- a/docs/user-guide/variants.md
+++ b/docs/user-guide/variants.md
@@ -52,8 +52,41 @@ This is the region of the most severe functional impact
 
 ### GT Call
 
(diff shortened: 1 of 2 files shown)


dnil:
Ok, I'll merge this and make a separate branch for the POST fix to make it a bit easier to follow.

moonso:
Good stuff!

PR #950, opened 2018-09-13
Post long lists

dnil:
Use POST with variants pages to enable long gene lists. Updated branch; replaces #842.
Not ready for merge - needs testing. And ideally fixing that upload-gene-list option: perhaps quickest just by including that in the main form, much like the other additional submit buttons, instead of dealing with the arcane arts of multiple WTFs.


moonso:
👍

moonso:
Ready for review?

dnil:
If you move the file upload into the main form as well; otherwise wait until I’ve done it. I don’t want to merge something that breaks a nice feature.

dnil:
Ok, bring it on. 🙈 Should be ready for review, assuming travis agrees..

moonso:
@dnil you have to checkout #951 !!

moonso:
Wow this is a big one, need some time and help from @northwestwitch to review and test ...

northwestwitch:
Thanks! I'll be a bit busy with an upcoming event in Uppsala on the next
two days, but I'll look into it as soon as I have some time, I swear!

dnil:
Well, @moonso promised to handle all conflicts, so no rush here! 😉

northwestwitch:
I'm reviewing this today!

northwestwitch:
Hi @dnil! Regarding your explanation I could use a bit more of info, since I have never used this type of filtering before. 
Question: is 'action -> HPO gene list' filtering the variants to show only those overlapping with the gene of the HPO phenotypes defined at the case level?

-  **add a HPO term, e.g. Ataxia, and check that you can use it in one of several ways.** .. You mean add a phenotype to the case? Which ways should I use the term, I don't understand really.

- **note a couple of genes with variants from the demo case, e.g. POT1, CACNA1G and test uploading**. Uploading of what and from where?

- **make a really large gene panel by e.g. adding several broad HPO terms and make sure you can re-filter after changing something else in the filter, or manually removing a gene from the symbol list**.  ??

I think I can answer all these question by looking at what the code does but maybe it's faster if we discuss about it in person.. When are you around?

dnil:
If you wish, you could focus on the code, and hope that my testing plus existing tests are good enough for the nitty gritty. I’ll drop by towards end of next week again, but could meet up for a bit earlier in the week as well.

- HPO: Search using e.g. HPO panel in the filter panel dropdown, directly from the case page using that button or from the “Actions” menu on the variants page.
- Upload as in upload a custom list from the upload box next to the gene symbol filter text field. Has moved down from the dropdown “Actions” menu.
- Large panel: on master, you will find that attempting to GET on a filter query with a long list comprising 100s of genes (a few kb of text) will fail miserably with some version of buffer issues, either on browser, proxy or server.

Cheers!

scout/adapter/mongo/query.py
```diff
@@ -55,13 +55,15 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
+        gene_query = []
 
         if query.get('hgnc_symbols') and query.get('gene_panels'):
-            gene_query = [
-                        {'hgnc_symbols': {'$in': query['hgnc_symbols']}},
-                        {'panels': {'$in': query['gene_panels']}}
-                    ]
+            gene_query.append({'hgnc_symbols': {'$in': query['hgnc_symbols']}})
+            gene_query.append({'panels': {'$in': query['gene_panels']}})
             mongo_query['$or']=gene_query
```
northwestwitch: Question: if you need to query by hgnc symbols AND panels, why there is OR in the query code?

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -183,18 +177,23 @@ <h4 class="panel-title">
   <div class="container-fluid">
     <div class="form-group text-center">
       {% if more_variants %}
-        <a class="btn btn-default" href="{{ url_for('variants.variants', institute_id=institute._id, case_name=case.display_name, page=(page + 1), **form.data) }}">
-          Next page
-        </a>
+        <div class "row">
```
northwestwitch: = is missing here I think

northwestwitch:
I have a suggestion regarding the scout/adapter/mongo/query **build_query**. This function is quite long and hard to follow, plus there are the 3 distinct objects: **query**, **major query** and **minor query**. I understand that you need to return just one query object from it, but maybe it's worth to split it into 3 distinct functions. 
My suggestion is: one functions that is called from the views and handles the overall building of the query + a function that builds the major query and one that builds the minor one. These can be called by the main function. What do you think?

scout/adapter/mongo/query.py
dnil: That semantics was never quite intuitive - but blame goes to the logicians and philosophers rather than me I think. If you want to use both a panel and add to it an additional free text set of symbols, OR is the operator to use. If we need both to be true at the same time, AND is correct.

scout/server/blueprints/variants/templates/variants/variants.html
dnil: Thank you!

dnil:
It will cement the current solution, but probably more readable! I can give it a try!
I would love to have a more modular structure, where new filter options can dock in their query requirements in a more general fashion. But there are more pressing things to cover first be for reworking it.

northwestwitch:
Ok, but that answers my question. And I was not blaming (actually I'm just
waiting for the right occasion to use the dedicated "BLAME" button of git,
but this one doesn't qualify I'm afraid!) :)

moonso:
Good suggestion, I will have a look at that

dnil:
@moonso @northwestwitch Should we split the build_query on this PR, or make a new issue for that perhaps?

moonso:
Let's test this on stage in the beginning of next week so it can be merged soon...

dnil:
🥁🎉😉

PR #951, opened 2018-09-14
Adds functionality to load PanelApp genes

### LINKED ISSUE #784: PanelApp again
It's panel update season again, and we are doing a lot of comparing to PanelApp. It is maturing well, and may be time to add import to Scout. We tentatively think "level 3" aggregation is a good panel level, and important to be able to pull only "Green" reviewed gene symbols.

API instructions available [here](https://panelapp.genomicsengland.co.uk/#!Webservices).

moonso:
Fix #784

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ pip-log.txt
 nosetests.xml
 .cache/
 .pytest_cache/
+pytest.ini
 
 # si
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [4.0.1b]
 
 ### Added
+- Command to download all panel app gene panels `scout load panel --panel-app`
 - Links to genenames.org and omim on gene page
 - Popup on gene at variants page with gene information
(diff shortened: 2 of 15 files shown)


moonso:
Not ready but feel free to test `scout load panel --panel-app` 🛩

dnil:
Ooooh! This is good stuff indeed!

dnil:
Ok, I've tested! Great start! I think it is already very useful. Updates seem possible, trusting PanelApp versioning, right?

Therese a lot of panels obviously. To make it effective, we shall need to be able to group panels since many of these are more specific than would be practical with the level of detail we typically get on a referral.

Perhaps the best path is to reintroduce a panel-group, where a panel(-group) could consist not only of gene symbols, but also of references to other panels? This will also help us with adding some coarse-grained structure to e.g. the neuromuscular disorders where as you may recall with some pain contained some 16 different sub panels at one point.

It may also be practical to mirror some of the PanelApp hierarchy as well, e.g. introducing a "parent" for panels. It could be just descriptive for starters, keeping the levels readable in scout, making it easier to pick the right ones to add?

moonso:
Ok I'm gonna merge this for now and then we can elaborate on improvements later

Issue #901, opened 2018-07-12
Feedback from clinvar


northwestwitch:
@dnil has just received a feedback from Clinvar after the first submission attempt (which dates back to before Easter 2018!) with files produced by Scout Clinvar submission tool. We should make a couple of changes to it, specifically:

* Variant tab/file: Reference sequence. A version number for each reference sequence must be provided in order to confirm the specific variant. i.e., NM_001604.5. What we provided with the available info from Scout was just 'NM_001604'.
 
* CaseData tab/file: Clinical features. This column is used to report observed clinical features.  We do not expect that these are the same as the condition or disease for which you assessed the variant’s clinical significance. Basically do not include phenotype for case data when the field is exactly the same as the one provided for the variant. 

The second one is easy to fix, while the first a bit more tricky..

I'm thinking about refactoring the Clinvar submission tool to allow users to add variants from one or more cases to a submission list that might grow over time, so that when they finally do the submission it will be many variants in one go. This would be the best approach, given also the months it takes for Clinvar curators to do the review.

northwestwitch:
I'm working on a new version of clinvar integration on scout which will allow the creation of submission files with an unlimited number of variants and case data from different cases. I'll be able to fix the second problem described above but I haven't figures out yet how to solve the first.

northwestwitch:
Handling first problem in #974

PR #956, opened 2018-09-14
Clinvar multi case submission

northwestwitch:
I rewrote almost completely the Clinvar pages, improving code and functionalities.

These changes make it possible to create submission objects with variants from different cases, allowing 
 to submit an unlimited number of variants and cases in one go.

I also addressed #901. So far I was able to fix the HPO issue in case data. The other issue (refseq version)  requires a bit of thinking and possibly collecting this info from VCF files.


northwestwitch:
@moonso we could test this one on stage next!

PR #962, opened 2018-09-18
fix #960

### LINKED ISSUE #960: Link to the case page doesn't work from Sanger modal in cases view.
This issue occurs in stage but affects potentially production as well. Should be fixed asap

northwestwitch:
Hopefully fixes the bug. Must be tested on stage!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Adds hgvs information on cancer variants page
 
 ### Fixed
-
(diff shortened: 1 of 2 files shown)


northwestwitch:
Keep in mind that you have to fix this if you want to show the new "show sanger sequenced but not yet eveluated" feature today.

northwestwitch:
Not yet, I do it today!

northwestwitch:
Just tried it and it has some issue, I'll fix it!

northwestwitch:
I was finally able to fix it. I had to take into account the possibility that there are removed cases with variants having sanger ordered. This way it works!

PR #970, opened 2018-09-26
Display gene info

moonso:


dnil:
Nice! Could there be a link-out to genenames.org from the gene_symbol as well?

moonso:
Sure! Do you mean from the popup or the gene page?

dnil:
Gene page!

PR #972, opened 2018-09-28
Download vcf2cytosure: Fix #714

### LINKED ISSUE #714: Add download vcf2cytosure data feature
MIP now produced the ".cgh" data for SV using vcf2cytosure. Let's discuss if these files could be downloaded via Scout's interface as they are connected to the corresponding SV variants and need to be updated for reruns and easily accessible when analysing a case. File size is roughly 94 MB for 30x WGS. This file is currently not splitted in clinical test and research.

Another approach is to add them to the delivery on Ceasar.

dnil:
Fix #714. Assumes vcf2cytosure fields per individual in config.yaml - issue for CG / HK rewrapping should follow, as this key is currently to be found in the MIP qc_sample_info.yaml. 
screen shot 2018-09-28 at 18 37 41

### DIFF
--- a/scout/build/individual.py
+++ b/scout/build/individual.py
@@ -14,19 +14,20 @@ def build_individual(ind):
 
         Returns:
             ind_obj (dict): A Individual object
-    
+
         dict(
             individual_id = str, # required
             display_name = str,
-            sex = str, 
-            phenotype = int, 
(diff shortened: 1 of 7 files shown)


moonso:
Also I think you should open an issue on cg and refer to this

PR #973, opened 2018-10-01
Update pymongo

moonso:


moonso:
This is done but we should wait with update until we have deployed current scout in production

northwestwitch:
Ok!

moonso:
These will have to be redone in another PR

PR #1009, opened 2018-10-12
Updates connection handling

moonso:


scout/commands/base.py
```diff
@@ -93,18 +90,18 @@ def cli(context, mongodb, username, password, authdb, host, port, loglevel, conf
...
-            username=mongo_config['username'],
-            password=mongo_config['password'],
-            authdb=mongo_config['authdb'],
-        )
-
-        LOG.info("Test if mongod is running")
-        if not valid_connection:
-            LOG.warning("Connection could not be established")
-            context.abort()
+        # valid_connection = check_connection(
```
northwestwitch: If you don't need these lines any more you could maybe delete them? Or?

northwestwitch:
I've testes this on stage and it works

Issue #1007, opened 2018-10-12
Issue regarding General Report - clearer boundaries between variants


KickiLagerstedt:
Would it be possible to get "thicker" lines regarding boxes including the different variants? It would clarify what "facts" belong to what variant.

The green/blue/yellow/red lines regarding the different classes could also be thicker...

northwestwitch:
I'll try!

moonso:
Fixed in #1011

Issue #1006, opened 2018-10-12
Report issue - disease transcript


AnnHam:
Missing information in the report

It's easy to see which transcript is the canonical transcript in the report, but we need to know which transcript is the disease transcript (according to HGNC/RefSeq). 

Could it be possible to highlight the disease transcript like in the interface of the variant?

northwestwitch:
I'll work on it!

moonso:
Fixed in #1011

PR #1011, opened 2018-10-12
Improve case report

northwestwitch:
Improved report code and aesthetics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,15 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 
 - Logs uri without pwd when connecting 
+- Disease-causing transcripts in case report
+- Thicker lines in case report
+- Supports HPO search for cases, both terms or if described in synopsis
 
 ### Fixed
 
 - Use db name instead of **auth** as default for authentification
 
 ## [4.1.1]
-- Disease-causing transcripts in case report
-- Thicker lines in case report
 - Fix problem with institute authentication flash message in utils
 - Fix problem with comments
 - Fix problem with ensembl link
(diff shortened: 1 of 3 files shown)


northwestwitch:
I don't think I will solve the slow report creation issue today. But this PR covers #1006 and #1007. We could probably review and merge this for the time being!

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 Add new stuff here
 
 ## [4.1.1]
+- Disease-causing transcripts in case report
```
moonso: These should be added to 4.1.2

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -280,15 +282,16 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
...
+<div class="card border-info mb-3" style="border-width: 5px;">
+  <div class="card-header bg-info text-white">
     <strong>Pinned Variants</strong>
   </div>
   <div class="card-body">
     {% set duplicated_variants = [] %}
     {% if suspects_detailed %}
       {% for pinned in suspects_detailed|sort(attribute='variant_rank') %}
-        {% if pinned['_id'] not in causatives_detailed | map(attribute='_id') %}
+        {% if pinned['_id'] not in printed_vars %}
```
moonso: We should try to move all of this logic so that it is done in the controllers instead of templates. But lets do that when we work on the performance stuff together

CHANGELOG.md
moonso: I fixed this so will merge

PR #1012, opened 2018-10-12
Cases with hpo

dnil:
Friday afternoon fun with Scout. Adds some rather handy candy to the cases search: list cases matching a phenotype, or with some free text word in the case synopsis.


dnil:
See docs for description, but basically just try "HP:0002664" or "synopsis:epilepsy". 😺

henrikstranneheim:
Awesome 👏🏻

Issue #984, opened 2018-10-08
Can not "Download PDF" Report


KickiLagerstedt:
https://scout.scilifelab.se/cust002/F0013129/case_report

"504 Gateway Time-out"

I have tried x3

northwestwitch:
I'll look into it. For the time being if the HTML version of the report is working (for me it woks but the page is slow to load), you might save a PDF of the report by pressing "print" and then "save as pdf" from the browser page. Not optimal I know, but it should work!

northwestwitch:
The query optimization in #985 is still not fixing the issue, so I reopen it and I'll find a real solution!

northwestwitch:
The 2 query optimizations I make didn't make much difference so I was thinking that we could try to do something at the level of the server configuration. Could we try to increase the Gateway timeout in the conf file (ngix?, I don't understand much of this stuff I'm afraid) @ingkebil ?

northwestwitch:
When there are a lot of variants involved like in the case above the server takes more time then 5 minutes. I can see if I find some other place where the code could be improved, I'll let you know!

KickiLagerstedt:
504 Gateway Time-out
https://scout.scilifelab.se/cust002/F0026810/case_report

When I try to print the report from the browser page suggested by Chiara @northwestwitch I do not get the whole report - only a few pages... ; everything BUT the commented variants (only two pages out of X pages) that I am interested in- Possibly also only two pages from dismissed variants.

northwestwitch:
@moonso what is slowing down the queries and the case report creating might be the search of the sanger sequenced variants. 
I'm testing something else and a stupid query like this:
sanger_vars = db.variant.find({ 'case_id': 'blablabla', sanger_ordered:true }) takes ages. 
Could we index the field?? To be certain we should create the same report with the Sanger sequenced info and see if it's faster!

PR #1014, opened 2018-10-15
Tries to make fetching of report variants more efficient

moonso:


northwestwitch:
Nice! Have you tested in on stage already?

scout/server/blueprints/cases/controllers.py
```diff
@@ -105,41 +134,87 @@ def case(store, institute_obj, case_obj):
...
-    data['commented_detailed'] = variants_filter_by_field(store, evaluated_variants,
-                                                        'is_commented', case_obj, institute_obj)
+    for var_obj in store.evaluated_variants(case_id=case_obj['_id']):
+        # Check which category it belongs to
+        for vt in variant_types:
+            keyword = variant_types[vt]
+            # When found we add it to the categpry
+            if keyword in var_obj:
+                evaluated_variants[vt].append(var_obj)
+                break
```
northwestwitch: This break takes care of not having the same variant in two categories of the evaluated variants, right? I think we should instead add to the evaluated category all those that belong to them, regardless. That's because for the clinicians might be important to know that a variant that for instance is ACMG classified is already printed among the pinned, for instance. Otherwise they'll not find variants that have actions associated and might wonder why.

northwestwitch:
Fixes #984

PR #1015, opened 2018-10-15
fixed stupid bug 

northwestwitch:
Instead of appending to a list I was printing it, in jinja! :)


northwestwitch:
That's the ninja? :D

moonso:
No it's a person fencing 😂

PR #1016, opened 2018-10-16
Add sanger dashboard

moonso:


scout/server/blueprints/dashboard/views.py
```diff
@@ -36,35 +36,37 @@ def index():
...
-        if doc_id in var_ids:
-            continue
-        var_ids.add(doc_id)
-
-        validation = variant_obj.get('validation', 'not_evaluated')
-
-        # Check that the variant is not evaluated
-        if validation in ['True positive', 'False positive']:
-            nr_evaluated += 1
+        # variant_id = event['variant_id']
```
dnil: # Ahem? #

moonso: Hmm this code does not correspond to my local...

moonso:
It should be correct now @dnil

moonso:
So @dnil check this out 🚀

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -0,0 +1,254 @@
...
+    if nr_ordered:
+        variants = [
+            {
+                'title': 'Validation ordered',
+                'count': nr_ordered
+            },
+            {
+                'title': 'Validated True Positive',
+                'count': len(validated_tp),
+                'percent': len(validated_tp) / nr_ordered,
```
dnil: I still think this should be tp / (tp + fp), in particular as not all clinics (*ahem* cust002 *ahem*) use the verification email extensively yet.

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -0,0 +1,254 @@
...
+            },
+            {
+                'title': 'Validated False Positive',
+                'count': len(validated_fp),
+                'percent': len(validated_fp) / nr_ordered,
```
dnil: ..and the same here obviously.

scout/constants/indexes.py
```diff
@@ -51,16 +51,6 @@
             name="caseid_variantid",
             background=True,
             ),
-        IndexModel([
```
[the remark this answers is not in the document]
moonso: Just wanted to remove that when I looked at the file...

scout/server/blueprints/dashboard/controllers.py
moonso: Sure I'll fix that

PR #1017, opened 2018-10-16
Use gnomad filter

moonso:
- Replaces 1000G and ExAC with gnomad filter
- Use gnomad in clinical filter

### DIFF
--- a/scout/server/blueprints/variants/forms.py
+++ b/scout/server/blueprints/variants/forms.py
@@ -65,8 +65,7 @@ class FiltersForm(FlaskForm):
     clinsig_confident_always_returned = BooleanField('CLINSIG Confident')
     spidex_human = SelectMultipleField('SPIDEX', choices=SPIDEX_CHOICES)
 
-    thousand_genomes_frequency = BetterDecimalField('1000 Genomes', places=2)
-    exac_frequency = BetterDecimalField('ExAC', places=2)
+    gnomad_frequency = BetterDecimalField('Gnomad', places=2)
     chrom = TextField('Chromosome')
     local_obs = IntegerField('Local obs. (archive)')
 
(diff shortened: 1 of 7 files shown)


scout/server/blueprints/variants/forms.py
```diff
@@ -65,8 +65,7 @@ class FiltersForm(FlaskForm):
     clinsig_confident_always_returned = BooleanField('CLINSIG Confident')
     spidex_human = SelectMultipleField('SPIDEX', choices=SPIDEX_CHOICES)
 
-    thousand_genomes_frequency = BetterDecimalField('1000 Genomes', places=2)
-    exac_frequency = BetterDecimalField('ExAC', places=2)
+    gnomad_frequency = BetterDecimalField('Gnomad', places=2)
```
dnil: Case mismatch: should render as "gnomAD".

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -94,13 +94,9 @@
...
     <div class="form-group">
       <div class="row">
-        <div class="col-xs-3">
-          {{ form.thousand_genomes_frequency.label(class="control-label") }}
-          {{ form.thousand_genomes_frequency(class="form-control") }}
-        </div>
-        <div class="col-xs-3">
-          {{ form.exac_frequency.label(class="control-label") }}
-          {{ form.exac_frequency(class="form-control") }}
+        <div class="col-xs-4">
```
dnil: You remove two xs-3, and add one xs-4, which skews the layout. Perhaps keep the frequency fields small, and allow some space to add more submit buttons later? We may for instance want one for "custom filters" later?

moonso: I think it is ok to have them slightly skewed for now. As you say it will be better when we add something else.

moonso:
I have a MIP6 case without SVs. Updating demo data is a hassle but I guess this is the time to do it...

moonso:
I think its better to update demo data in another PR, it's to much to include here. Please approve @dnil I have changed according to your suggestions

dnil:
I’m of course ok with holding off the demo data to another PR. It will take a little fiddling to get that right.
But for this:
- update tests to reflect the new expected adapter outcome
- align at least the buttons: there is no need to leave them askew, is there? Just make sure the row sum is 12 and you are home free! 😉

moonso:
Sure. Should be fixed now @dnil !

PR #1028, opened 2018-10-22
Institute page

moonso:


northwestwitch:
Do you need a review?

moonso:
Sure, I just wanted to get a skeleton in place for future stuff

scout/server/app.py
```diff
@@ -100,6 +101,7 @@ def register_blueprints(app):
     app.register_blueprint(api.api_bp)
     app.register_blueprint(alignviewers.alignviewers_bp)
     app.register_blueprint(phenotypes.hpo_bp)
+    app.register_blueprint(institutes.institutesov_bp)
```
northwestwitch: Don't know how final the name of this page is, but I don't like much 'institutesov'. Stands for institute overview right? what about just overview?

scout/server/blueprints/institutes/views.py
```diff
@@ -0,0 +1,49 @@
...
+        for user_mail in ins_obj.get('sanger_recipients',[]):
+            user_obj = store.user(user_mail)
+            if not user_obj:
+                continue
+            sanger_recipients.append(user_obj['name'])
+        institutes.append(
+            {
+                'display_name': ins_obj['display_name'],
+                'internal_id': ins_obj['_id'],
+                'coverage_cutoff': ins_obj['coverage_cutoff'],
```
northwestwitch: This is causing a key error while testing with the demo data:
'coverage_cutoff': ins_obj['coverage_cutoff'],
This is causing a key error if you test with the demo data:
``` 
'coverage_cutoff': ins_obj['coverage_cutoff'],
KeyError: 'coverage_cutoff'
``` 
We should either fix the demo institute to have a cutoff key or this code to use ins_obj.get('coverage_cutoff') instead

moonso: 👍 I'll fix that

moonso:
Ok I have fixed according to your suggestions. Can you approve @northwestwitch ?

northwestwitch:
One last thing. You named the html template in institutes/templates/overview 'index.html'. I know it's just a name but I think that to be consistent with the rest of the app it should be renamed to 'institutes.html'.

scout/server/blueprints/institutes/views.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+from pprint import pprint as pp
+
+from flask import (abort, Blueprint, current_app, redirect, render_template,
+                   request, url_for, send_from_directory, jsonify, Response, flash)
+from flask_login import current_user
+
+from dateutil.parser import parse as parse_date
+from scout.constants import PHENOTYPE_GROUPS
```
northwestwitch: You don't need most of these imports right now

northwestwitch: Not PHENOTYPE_GROUPS, but those flask ones above

PR #1034, opened 2018-10-26
Fixes so that all primary transcripts are displayed correct

moonso:


moonso:
Could I get a quick review of this please @northwestwitch or @dnil ? 🙏

PR #1036, opened 2018-10-29
Better panel pdf export

### LINKED ISSUE #1031: PDF-exported gene panels look bad when their genes are shared by too many other panels
Example:

image

northwestwitch:
Fix #1031

- Not visualizing the co-occurrence of genes from archived panels

- Line breaks when showing co-occurrence of genes from other panels, so that max 10 co-occurrences per line are shown in PDF report (see picture)

image


northwestwitch:
Shall we merge?

PR #1037, opened 2018-10-29
Pass case name to chanjo report to add it to downloaded pdf

### LINKED ISSUE #821: Provide family specific default filenames for generated pdfs
If a user downloads more than one a mixup risk ensues. Better to suggest naming according to family / sample name in addition to covereage, delivery, analysis etc.

northwestwitch:
Case name in downloaded coverage PDF report:
fix #821
Goes together with this code update of chanjo-report: https://github.com/Clinical-Genomics/chanjo-report/pull/2


northwestwitch:
This can be tested on stage!

dnil:
Looks good, but stage seems to be down / with error at the moment?

northwestwitch:
I'll restart it and let you know when it works!

moonso:
Is this one done?

northwestwitch:
This has become old. Plus we are going to re-write the coverage part. So i'd close it, what do you think @moonso ?

PR #1038, opened 2018-10-29
Download of filtered SVs

### LINKED ISSUE #952: Gene symbol list file upload for SV variants
Should be the same as for the SNVs, really. 😊

northwestwitch:
- Fixed broken download of filtered SNVs
- Created download of filtered SVs (n. max = 500)
- introduced an underscore in field "position+change" in the exported CSV file

fix #952

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -275,7 +275,7 @@ def variant_export_lines(store, case_obj, variants_query):
         variant_line.append(variant['chromosome'])
         variant_line.append(position)
         variant_line.append(change)
-        variant_line.append('_'.join([ str(position), change ]))
+        variant_line.append('_'.join([str(position), change]))
 
         # gather gene info:
         gene_list = variant.get('genes') #this is a list of gene objects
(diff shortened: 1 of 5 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -275,7 +275,7 @@ def variant_export_lines(store, case_obj, variants_query):
         variant_line.append(variant['chromosome'])
         variant_line.append(position)
         variant_line.append(change)
-        variant_line.append(str(position)+change)
+        variant_line.append('_'.join([ str(position), change ]))
```
moonso:
```suggestion
        variant_line.append('_'.join([str(position), change]))
```

scout/server/blueprints/variants/views.py
```diff
@@ -255,8 +255,27 @@ def sv_variants(institute_id, case_name):
...
+        # Return max 500 variants
+        export_lines = controllers.variant_export_lines(store, case_obj, variants_query.limit(500))
+
+        def generate(header, lines):
+            yield header + '\n'
+            for line in lines:
+                yield line + '\n'
+
+        headers = Headers()
+        headers.add('Content-Disposition','attachment', filename=str(case_obj['_id'])+'-filtered_sv-variants.csv')
```
moonso: Shouldn't it be `case_obj['display_name']` here? The customers have no idea what the internal case ids mean..

northwestwitch: You're right. I'll check and modify also the name of the downloaded file for SNVs if the code is like this one!

PR #1039, opened 2018-10-30
Fixes bug where build should be genome_build

moonso:


moonso:
Could you just do a quick review @northwestwitch ?

scout/server/blueprints/variants/controllers.py
```diff
@@ -275,7 +284,7 @@ def variant_export_lines(store, case_obj, variants_query):
         variant_line.append(variant['chromosome'])
         variant_line.append(position)
         variant_line.append(change)
-        variant_line.append('_'.join([str(position), change]))
+        variant_line.append(str(position)+change)
```
northwestwitch: Why did you change this?
```variant_line.append('_'.join([str(position), change]))```
is correct

moonso: Didn't touch that...

moonso: Fixed now?

PR #1045, opened 2018-10-31
Minify dismissed variants in report

northwestwitch:
Rearrange dismissed variants in report so that it fits in the report whenever a variant is dismissed because inherited from unaffected parents. 
HTML was not affected, just PDF.
The whole "Dimissed description" column should now be more clear

Previously looked like this:
image

After fix:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 Add new stuff here
 
 
+## [4.2.1]
+
+### Fixed
+- Better PDF rendering for excluded variants in report
+
+
 ## [4.2.0]
 
 ### Added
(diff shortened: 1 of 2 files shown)


dnil:
Would be lovely to see this one on stage at some point - much easier to check a few reports that way. Nice with width% and great to use more of the space on the page rather than have a huge left margin.

northwestwitch:
Ok, I'll put it on stage as soon as fix it, and after the chanjo one!

northwestwitch:
I'm having some issues with updating chanjo report on stage, so I've put this on stage for now. Ready top be tested!

dnil:
Yep, looks all good now! One might make the "gene" column slightly bigger so a "(downstream)" fits on on row in the pdf - that looks a little funky right now.

northwestwitch:
I've removed the <ul> element before the genes because that takes quite a space. Included the gene annotation in a badge to make it more compact. I'll update stage so it shows there!

northwestwitch:
Stage is updated!

dnil:
Looks great!

PR #1047, opened 2018-11-02
Query overlapping variants by matching genes instead of coordinates

### LINKED ISSUE #1046: Overlapping SVs hover misinterprets BNDs
BNDs appear to have a very large size (approximation for infinite). They should for the purpose of overlapping only occur in a gene they hit with their start / end coord - size does not matter here.

northwestwitch:
Fix #1046

I don't know if I'm naive and missing something here, but I was reasoning that instead of looping through the genes of the query variant to establish the largest genomic interval (given by the start of the first gene and the end of the last), you could simply get all variants that occur in the same genes as the query one (it's basically the same, isn't it?). This way the query should be faster and you also skip the passage where you have to create a gene object for each gene in the query variant.

It fixes the BND issue because BNDs that don't have that gene are not collected.

I also realized that BNDs still had visualizing issues in the sv_variant page and in the report. I've fixed them now!

Fixed also the template so it doesn't print the length of the variant if it's a huge rearrangement.

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -340,8 +340,8 @@ class VariantHandler(VariantLoader):
     def overlapping(self, variant_obj):
         """Return overlapping variants.
 
-        Look at the genes that a variant overlaps to, to get coordinates.
-        Then return all variants that overlap these coordinates.
+        Look at the genes that a variant overlaps to.
+        Then return all variants that overlap these genes.
 
         If variant_obj is sv it will return the overlapping snvs and oposite
         There is a problem when SVs are huge since there are to many overlapping variants.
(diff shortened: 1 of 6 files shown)


moonso:
Hi, have you tried this on stage? This would probably need a new index on strictural variants to work. One of the problems with this approach is that large structural variants can span 100s of genes which will make the query a bit difficult to perform and the index would probably be huge as well.
Also bear in mind that the overlapping function works both ways so it is used to fetch snvs when accessing a structural variant.
Please try it and check the performance. It if does not work fine I have another idea.
I'll do the complete review when it's tested with a large database

northwestwitch:
I'm going to update stage with this branch so we see how it performs

PR #1050, opened 2018-11-07
Use get when checking if panel is default

moonso:


moonso:
Tested on stage and it works. It's a one line review @northwestwitch 😂

northwestwitch:
Yes but conceptually important! :)

PR #1051, opened 2018-11-07
Mosaicism

moonso:
- Adds a couple of mosaicism options on variant level
- Highlight mosaic variants in variants view
- Show mosaicism on pinned variants

NOT IMPLEMENTED
- Display mosaic variants in case report
- Show is variant has been marked mosaic in other cases


vwirta:
How do we identify a variant as mosaic? What caller do we use for this right now?

northwestwitch:
Hi, we don't call the mosaicism, but users can assign this tag manually at the variant page, after validation I guess..

henrikstranneheim:
Usually, _de novo_ variants and manual inspection in IGV and then Sanger

scout/server/blueprints/variants/views.py
```diff
@@ -322,19 +322,31 @@ def variant_update(institute_id, case_name, variant_id):
...
         flash("Reset variant dismissal: {}".format(variant_obj.get('dismiss_variant')), 'info')
 
+    mosaic_tags = request.form.getlist('mosaic_tags')
+    if mosaic_tags:
+        store.update_mosaic_tags(institute_obj, case_obj, user_obj, link, variant_obj,
+                                     mosaic_tags)
+        if new_dismiss:
+            flash("Added mosaic tags: {}".format(mosaic_tags), 'info')
+
+    if variant_obj.get('mosaic_tags') and not mosaic_tags:
```
dnil: ..and this little shortcut won't work gracefully with multiple save buttons. I'll get back with a fix..

dnil:
Ok, so here goes a rough fix for the multiple buttons cross-resetting issue. Sorry to barge in - HTH!

moonso:
Thank you @dnil ! So I have updated the branch and rebased on master.
Regarding your comments @northwestwitch 
1. I have moved the tag to inheritance column, good suggestion!
2. I don't agree with this comment. Just like the dismiss variant options this will show in the select tags field. That is fine I think

dnil:
I'll assume you have tested setting and resetting different user tags after merge! Could you also write a line or two in the documentation for the users that aren't already in the loop on this? Thanks!

Issue #1054, opened 2018-11-07
Die gracefully and show error message when gene panel is not formatted correctly


northwestwitch:
App shouldn't crash on file parsing error

northwestwitch:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 28212: invalid start byte

Could it be that a line is terminated by \r instead of \n?

northwestwitch:
This should be also fixed in the new release so I close this issue

Issue #1052, opened 2018-11-07
Gene edit in panels doesn't show all required features


northwestwitch:
- Inheritance should be of manual type (AR, AD, [AD,AR], XL, XR, digenic)

- Entry version

northwestwitch:
This has to be fixed after #1055

northwestwitch:
I know, I did it on purpose because it should reflect a panel version and panel versions are floating point numbers..

4WGH:
If I do an upload It takes the text though....


this is helpfull for me as some genes are shared by some panel but wehre entered in different versions.


I'm open to alternative suggestions for dealing with this?


Thank for all the work till now 😊


________________________________
Från: Chiara Rasi <notifications@github.com>
Skickat: den 9 november 2018 15:38:52
Till: Clinical-Genomics/scout
Kopia: Michela Barbaro; Comment
Ämne: Re: [Clinical-Genomics/scout] Gene edit in panels doesn't show all required features (#1052)


I know, I did it on purpose because it should reflect a panel version and panel versions are floating point numbers..

PR #1058, opened 2018-11-07
fix UnicodeEncodeError when parsing gene panel csv file

northwestwitch:
Takes care of #1054.

I think it might be an invisible formatting character in the file that could probably be skipped anyway. It prevents the app from crashing when any type of decoding error occurs.

### DIFF
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -407,8 +407,8 @@ class PanelHandler(object):
                 gene_obj['reduced_penetrance'] = info['reduced_penetrance']
             if info.get('mosaicism'):
                 gene_obj['mosaicism'] = info['mosaicism']
-            if info.get('entry_version'):
-                gene_obj['database_entry_version'] = info['entry_version']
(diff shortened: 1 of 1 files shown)


northwestwitch:
Included support for edit of comments and entry version in gene panel "edit gene". Partly covers #1052

scout/adapter/mongo/panel.py
```diff
@@ -344,7 +344,7 @@ def add_pending(self, panel_obj, hgnc_gene, action, info=None):
             updated_panel(dict):
 
         """
-        
+
```
moonso: You can move indentation for the code block after `if action ...` back to make it more flat. We know at that stage that the action is valid

northwestwitch: I'm not sure that this change won't result in a bug, have you tested it?

northwestwitch:
Wait, don't merge, it's not working any more!

northwestwitch:
I think I'm done here!

PR #1060, opened 2018-11-08
Replace variants

moonso:
- Adds functionality to replace SV variants from CLI


scout/build/variant/variant.py
```diff
@@ -333,6 +333,8 @@ def build_variant(variant, institute_id, gene_to_panels = None,
         variant_obj['clingen_ngi'] = frequencies['clingen_ngi']
     if frequencies.get('decipher'):
         variant_obj['decipher'] = frequencies['decipher']
+    elif frequencies.get('decipherAF'):
```
dnil: Add a little comment or doc about that perhaps, as it is ... a little weird? 😅

scout/server/blueprints/variants/forms.py
```diff
@@ -93,6 +93,7 @@ class SvFiltersForm(FiltersForm):
     size = TextField('Length')
     size_inclusive = BooleanField('Length inclusive')
     svtype = SelectMultipleField('SVType', choices=SV_TYPE_CHOICES)
+    decipher = BooleanField('Decipher')
```
dnil:
```suggestion
    decipher = BooleanField('Decipher population')
```

moonso:
Hi this should be fixed and is tested on stage, you can check https://scout-stage.scilifelab.se/cust000 and search for case 643594-300M. Please approve if you think it's fine @dnil

PR #1071, opened 2018-11-15
Export single case with --case-id option fixed

adrosenbaum:
Fixed small bug when exporting single case with 'scout export cases --case-id ... '. On master, .count() method is used on a python list without any arguments, if --case-id option is given.

I also believe that it should be
```python
if not models.count():
    LOG.info("No cases could be found")
```
instead of 
```python
if models.count():
    LOG.info("No cases could be found")
```


scout/commands/export/case.py
```diff
@@ -56,21 +56,20 @@ def cases(context, case_id, institute, reruns, finished, causatives, research_re
...
+            LOG.info("No case with id {}".format(case_id))
 
     else:
         models = adapter.cases(collaborator=institute, reruns=reruns,
                            finished=finished, has_causatives=causatives,
                            research_requested=research_requested,
                            is_research=is_research, status=status)
-    if models.count():
-        LOG.info("No cases could be found")
+        if not models.count():
```
moonso:
```suggestion
        models = [case_obj for case_obj in models]
        if len(models) == 0:
```
Could you change to this so that it will work with PyMongo 3.7+ ?

PR #1075, opened 2018-11-15
Export causative variants from a case, with '--case-id' option

adrosenbaum:
If a case-id is given, i.e. 'scout export variants --case-id ...' this will print more complete vcf entries.

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -241,38 +241,36 @@ class VariantHandler(VariantLoader):
                                                variant_obj['position'])
         return variant_obj
 
-    def get_causatives(self, institute_id):
+    def get_causatives(self, institute_id, case_id=None):
         """Return all causative variants for an institute
 
             Args:
                 institute_id(str)
+                case_id(str)
 
             Yields:
                 str: variant document id
         """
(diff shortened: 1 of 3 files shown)


scout/adapter/mongo/variant.py
```diff
@@ -257,6 +257,24 @@ def get_causatives(self, institute_id):
         ])
         return [item['_id'] for item in query]
 
+    def get_case_causatives(self, case_id):
```
moonso: There is a function called `get_causatives` in adapter.variant. Please add the argument `case_id=None` to that function and add the logic there

scout/commands/export/variant.py
```diff
@@ -15,39 +15,103 @@
...
-            variant_obj['reference'],
-            variant_obj['alternative'],
-            variant_obj['quality'],
-            ';'.join(variant_obj['filters']),
-            '.'
-        ))
\ No newline at end of file
+    #If case_id is given, print more complete vcf entries, with INFO,
+    #and genotypes
+    if case_id:
```
moonso: Could you try to merge these so that as little code is duplicated as possible? Ask me if you need help or don't understand what I mean

PR #1085, opened 2018-11-22
Fix clinvar filter

### LINKED ISSUE #1079: Clinvar filter does not seem to work
Tried filtering for Benign in this case:

https://scout.scilifelab.se/cust003/18232/371c47ef74b1ff6f9b77262aba7cb482

northwestwitch:
This one should work for real. fix #1079


northwestwitch:
I'll install the branch on stage as soon as we finish testing another PR. For now I've tested on my local db by manually modifying the clinsig field to include strings along with the integer default values from the demo data. I'll let you know when it's on stage!

moonso:
Could you tell when its tested on stage @northwestwitch ?

northwestwitch:
Sure!

dnil:
I checked around quickly at klingen, and it appears to be a "known" that clinvar causatives are not found anymore. I got two recent cases from the top of AH and ES heads where the causative variant was not found when searching. So, flagging etc is called for - perhaps good to do together with the Decipher=1 cases?

northwestwitch:
That's too bad that nobody said anything before, we could have fixed sooner! I'm not sure about the decipher thing. We could discuss about it tomorrow (if you're around!)

northwestwitch:
It's on stage (https://scout-stage.scilifelab.se)!

northwestwitch:
I've realized that this PR can't be tested on stage if we don't make a newer backup of the data. All variants in scout stage database are in fact of the old type (clinsig is in numerical format). What I tested on stage is that the query for old cases still works and then I tested a simulated new case (with clinsig in string values) on my local computer. It looks like it's working so I think I'll merge.

Issue #793, opened 2018-05-04
mtDNA variant report


1ctw:
Enligt överenskommelse på ett möte 2018-03-29 mellan Nicole och Chiara skapar jag nu en issue för önskemålet att få ut en rapport med mtDNA-varianter från Scout.

Bifogar en excelfil med som visar vilka delar vi behöver få ut i rapporten.

Vi önskar en rapport i excelformat.
Titel på rapporten ska innehålla "Sample name" från Scout.
Icke-kodande delar får bli tomma i kolumnen för "Gene".
Icke-proteinkodande gener får bli tomma i kolumnen för "HGVS Description".
I kolumnen "HGVS Description" ska det bara vara p.
Det ska vara en rapport per prov även om det är en trio som körts.

[Report from Scout of variants in mtDNA.xlsx](https://github.com/Clinical-Genomics/scout/files/1973540/Report.from.Scout.of.variants.in.mtDNA.xlsx)

northwestwitch:
AD Reference and AD Alternative is the number of reads with with either the reference of the alternative allele?

1ctw:
I mean the part shown under GT call in Scout, number of reads of Reference or Alternative allele. From this we can quantify the level of heteroplasmy of a heteroplasmic mutation.

![bild](https://user-images.githubusercontent.com/15033593/39690317-edf00a98-51d9-11e8-95e9-aefd0616bacf.png)

I will get back to you regarding your first question after discussion with Nicole. I will meet her after lunch today.

1ctw:
Please also see my other issue #792. We will need all variants in mtDNA to be loaded in Scout to get a report with all information about mtDNA.

northwestwitch:
Ok, thanks for the explanation! When we had that meeting on the 2018-03-29 I remember I was told that all mtDNA variants are already loaded into scout now, so that would not be an issue!

1ctw:
Sorry but that is an issue! Nicole had missed that they are not. We have got all variants in some test cases manually from Anders but they are not in Scout. The variants in the D-loop region of mtDNA are missing in Scout. The D-loop region contains essential transcription and replications elements that could be of importance, but this part does not have transcripts and are not picked up.

The missing variants are for sure in the regions m.1-576 and m.16024-16569 (D-loop parts).

Probably variants in the small parts of mtDNA which are non-coding also are missing, for example m.7515-7517, m.8270-8294 or other small parts, but that is harder for me to see.

[...]

PR #1086, opened 2018-11-23
Exporting all MT variants into excel files (one per sample) + test

northwestwitch:
Created a CLI for this functionality. I'll work on the web app code next. 
This is addressing #793

- Creates an excel document for each sample in a case. Saves files in a specified outfolder or in the working directory


northwestwitch:
What?? They deprecate methods you are using until the day before (or very recently at least)! @moonso could you help me fix this error? I know for a fact that you've been busy specifically with this stuff: https://github.com/mongomock/mongomock/pull/4549

moonso:
Hi @northwestwitch what do you mean here? Maybe you can lock the mongomock version if it breaks?

northwestwitch:
Or I'll update the method with the new syntax, thanks!

northwestwitch:
It's not!

northwestwitch:
I'm working on a similar export function from the web app. I'll test them both on stage and get back to you when this is done, thanks!

northwestwitch:
I think I'm done with this PR. I've installed the branch on stage and it looks like it's working. (asked also the clinicians to test it, so I fix it if anything is off). 

For the CLI I've installed it on rasta stage and the installation works, but then there are the usual problems related to the db connection and I couldn't really test it there. I think it would be better tested on your local computer and then in production (it doesn't really affect anything else I think).

PR #1091, opened 2018-11-26
Sv length shorter

### LINKED ISSUE #1073: Filter for variant size range (min and max size)
This is important for structural variants

dnil:
Fix #1073. Replaces length inclusive option with a "shorter than" boolean, that is also inclusive to non-existing lengths. This extends the interface with the sought function, while preserving the old to all practical purposes.


moonso:
Hi @dnil , will this mean that you will get all small variants as well? I thought one wanted to specify a range like 500-20000?

dnil:
Hm, that's not on the request, but not difficult of course. 
Perhaps we should rather do a per-caller filter that allows deselecting e.g. Manta and Delly? ;-)

Issue #1067, opened 2018-11-12
panel export as pdf failed


4WGH:
I've tried to export the IEMv13 panel and for the following error 

504 Gateway Time-out

Michela

northwestwitch:
Hi! I think the problem is that for each gene it's searching if the gene is part of any other panels (those numbers on the right), and if there are thousands of genes it has to search for a long time, therefore the time out. We could try to optimize the algorithm but I don't know how much it will work when it comes to huge panels. 
I don't see any attachment, try a screenshot instead! Thanks! :)

4WGH:
I can not attach or paste any file....
An issue in the issue?

o_O

northwestwitch:
This should be fixed by the new release so I'm closing this issue

PR #1093, opened 2018-11-27
Removed reference to other panels in gene panel export

northwestwitch:
It takes too much time to check in which other panels each gene of a panel is. I'm removing this option for preventing bug #1067 . 
What we could do instead is improving the gene view, listing there all gene panels the gene belongs to.

### DIFF
--- a/scout/server/blueprints/panels/controllers.py
+++ b/scout/server/blueprints/panels/controllers.py
@@ -144,27 +144,4 @@ def panel_export(store, panel_obj):
     full_name = "{}({})".format(panel_obj['display_name'], panel_obj['version'])
     panel_obj['name_and_version'] = full_name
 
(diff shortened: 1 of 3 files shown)


northwestwitch:
@moonso  could you take a look at this so we include it in today's release?

scout/server/blueprints/panels/controllers.py
```diff
@@ -144,27 +144,4 @@ def panel_export(store, panel_obj):
     full_name = "{}({})".format(panel_obj['display_name'], panel_obj['version'])
     panel_obj['name_and_version'] = full_name
 
-    panel_explained = {} # a disctionary with display_name(version) as key and an increasing index as value
```
moonso: Is it ok to just remove this? Where is is/was it used?

northwestwitch:
Yes we don't need that line any more

PR #1101, opened 2018-12-07
Improve mt export

northwestwitch:
Changed the MT export name to "mtDNA report" and fixed the weird representation of the aminoacids in the report


northwestwitch:
Thanks!

PR #1118, opened 2019-01-07
fix #1116

### LINKED ISSUE #1116: Report - missing phenotype and sex
Phenotype and gender are missing on the generated general report. "Phenotype overview" window.

northwestwitch:
After the fix it should look like this:
image

With A and NA labels which stand for AFFECTED and NOT AFFECTED. Sex is also displayed now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Add new stuff here
 
 ### Fixed
 - Visualization of PDF-exported gene panels
+- Sex and affected status is correctly displayed on general report
 
 ## [4.3.0]
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 # everything the developer needs in addition to the production requirements
 
 # testing
-pytest==3.3.0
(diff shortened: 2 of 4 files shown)


northwestwitch:
I don't know if there was any specific reason why pytest was frozen to version 3.3, but it wasn't working any more. By updating it to 3.6 the tests pass!

dnil:
Nice, thank you! NA somehow gives us doubts about if it was typed or is NotAvailable? Not sure how to best convey that it is "unaffected" - make it "NAD" ("NothingAbnormal Detected"), leave those blank, use "U" and explain somewhere?

northwestwitch:
Oh yes, you're right! What about changing the header into 'affected' and on the individual level I write yes or no?

dnil:
Sounds good!

northwestwitch:
It looks like this now:
image

dnil:
Code looks good to me, but I'm no expert on the ped-file data and what to expect in the db there.

northwestwitch:
But your comment was really helpful, thanks! :)

Issue #1115, opened 2018-12-21
Sanger order


4WGH:
The exon or intron number is not appearing any longer after the transcript

EXemple from today: 
Case 18215: 2_166019184_G_C_clinical 
•Variant type: SNV (SNV) 
•Breakpoint 1: 2:166019184 
•Breakpoint 2: 2:166019184 
•HGNC symbols: SCN3A 
•Gene panels: OMIM-AUTO, EP 
•GT call 
•18215-I-1A: 0/1
•18215-II-2U: 0/0 
•Amino acid changes 
•SCN3A:NM_006922-ENST00000283254:c.849C>G:p.Ser283Arg
•SCN3A:NM_001081676-ENST00000283254:c.849C>G:p.Ser283Arg
•SCN3A:NM_001081677-ENST00000360093:c.849C>G:p.Ser283Arg 
•Ordered by: b'Helene Bruhn' 

Exemple from last week (before update)
Case 18238: 4_6303626_G_A_clinical 
•Variant type: SNV (SNV) 
•Breakpoint 1: 4:6303626 
•Breakpoint 2: 4:6303626 
•HGNC symbols: WFS1 
•Gene panels: OMIM-AUTO, IEM, ATX, Ataxi, ENDO 
•GT call 
•18238-I-2A: 1/1 
•Amino acid changes 
•WFS1:NM_001145853:exon8:c.2104G>A:p.Gly702Ser 
•Ordered by: b'Nicole Lesko'

northwestwitch:
I'll take a look at this today!

PR #1119, opened 2019-01-07
add intron/exon number or intergenic to sanger email

northwestwitch:
This is addressing #1115.

After the fix the verification email displays also exon/intron number or "intergenic".

Should be tested on stage before merging!


moonso:
Could you just add a small note in the CHANGELOG for traceability?

PR #1120, opened 2019-01-07
Number check in variant size filter

northwestwitch:
This way users are forced to input only numbers in "size" field of the SV filter!


northwestwitch:
Thanks!

dnil:
Interesting... Well solved.

PR #1121, opened 2019-01-07
added instructions to install cairo and pango separately

northwestwitch:
I'm referring to the page of the library we use. Otherwise it's messy because installation of the external libraries is platform-specific..


northwestwitch:
Sure!

PR #1130, opened 2019-01-21
fix #1125

### LINKED ISSUE #1125: Conservation values have gone bye-bye?
It appears we have lost touch with evolutionary conservation. We wonder where and when?

northwestwitch:
Took me a while to understand that variants could have multiple scores for either GERP, PHAST and PhyloP and that multiple values where tuples..!

### DIFF
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -9,7 +9,20 @@ CONSEQUENCE = (
     'unknown'
 )
 
-CONSERVATION = ('NotConserved', 'Conserved')
+CONSERVATION = {
+    'dbNSFP_GERP___RS' : {
+        'conserved_min' : 2,
+        'conserved_max' : 10
+    },
+    'dbNSFP_phastCons100way_vertebrate': {
+        'conserved_min' : 0.8,
+        'conserved_max' : 100
+    },
+    'dbNSFP_phyloP100way_vertebrate' : {
+        'conserved_min' : 2.5,
+        'conserved_max' : 100
+    }
+}
 
 FEATURE_TYPES = (
   'exonic',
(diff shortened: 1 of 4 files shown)


scout/constants/variant_tags.py
```diff
@@ -9,7 +9,20 @@
...
 
-CONSERVATION = ('NotConserved', 'Conserved')
+CONSERVATION = {
+    'dbNSFP_GERP___RS' : {
+        'conserved_min' : 2,
+        'conserved_max' : 10
+    },
+    'dbNSFP_phastCons100way_vertebrate': {
+        'conserved_min' : 0.8,
+        'conserved_max' : 100
```
dnil: I'll assume you have checked these values, or that they are the same as the previous terms: as far as I remembered, phastCons used to give a P-val? Is it scaled or the -log P perhaps to be more similar to the others?

scout/parse/variant/conservation.py
```diff
@@ -1,47 +1,52 @@
...
             conservations(dict): A dictionary with the conservations
     """
     conservations = {}
-    
+
     conservations['gerp'] = parse_conservation(
-                                            variant, 
-                                            'GERP++_RS_prediction_term'
+                                            variant,
+                                            'dbNSFP_GERP___RS'
```
dnil: Weird name, but that's not our doing. 😄

scout/constants/variant_tags.py
northwestwitch: I took these values from here: https://github.com/Clinical-Genomics/MIP/blob/master/templates/rank_model_cmms_-v1.20-.ini

as also @henrikstranneheim  suggested. Do they make any sense to you?

northwestwitch:
Numbers have been there the whole time, strings and numbers have been both there until we updated to the latest MIP version. This code won't break because numbers will be found also for old samples, even if they have the string terms. Your suggestion is good if numeric values have changed over time though.

Issue #1126, opened 2019-01-16
See all variant callers


sofievonlanthen:
We used to be able to see all variant callers, even if they haven't called the variant, is it possible to bring it back? For example saying "not called" or "filtered", to be reminded of the different callers that are used, both for SNVs and SVs.

northwestwitch:
Is anyone working on this, otherwise I can take it!

northwestwitch:
Ok I see that this happened after #1069. It's easy to fix but I was just wondering why the calls of "BCFtools" are returned as calls of "SAMtools" instead
 
![image](https://user-images.githubusercontent.com/28093618/51472936-40642a80-1d7b-11e9-9838-6550d15cde06.png)

dnil:
I guess since SAMtools, BCFtools, HTSlib: same difference nowadays, but used to be (m)pileup or samtools, so someone stuck with that?

dnil:
Solved by #1132 - will appear on a server near to you in the not too distant future!

PR #1132, opened 2019-01-21
show calls from all available callers

northwestwitch:
I changed the name of the bcftools constant into "Bcftools" otherwise under the gt calls "SAMtools" gets printed twice.

### DIFF
--- a/scout/constants/__init__.py
+++ b/scout/constants/__init__.py
@@ -79,8 +79,8 @@ CALLERS = {
         },
         {
             'id': 'bcftools',
-            'name': 'SAMtools',
-            }
+            'name': 'Bcftools',
+        }
     ],
     'cancer': [{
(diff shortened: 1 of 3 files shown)


northwestwitch:
fix #1126

scout/constants/__init__.py
```diff
@@ -79,8 +79,8 @@
         },
         {
             'id': 'bcftools',
-            'name': 'SAMtools',
-            }
+            'name': 'Bcftools',
+        }
```
dnil: Perhaps leave that as SAMtools for historical reasons?

northwestwitch:
The problem if we leave bcftools with name : "SAMtools" is that SAMtools caller tag is then showed twice, like this:
image

northwestwitch:
Ok, thanks for the review!

PR #1134, opened 2019-01-24
Fix #1124 - use POST for overview.

### LINKED ISSUE #1124: Coverage overview generates an address that is too long
414 Request-URI Too Large is the message

dnil:
It actually seems very possible to use POST for chanjo overview nowadays. Please test with a more complete chanjo db / setup: my local is very demo.


northwestwitch:
Tested with a local implementation of chanjo-report (same as yours?) and works. 
Unfortunately we don't have chanjo and chanjo-report installed in the stage environment so it's not possible to test it there. But by looking at the code I don't think it could cause any harm, so I think it's fine to merge!

PR #1140, opened 2019-02-06
Count variants

moonso:


scout/commands/view/case.py
```diff
@@ -14,8 +14,16 @@
 @click.option('-c', '--case-id',
     help="Search for a specific case"
 )
+@click.option('--nr-variants',
+    is_flag=True,
+    help="Show number of clinical and research variants"
+)
+@click.option('--variants-treshold',
```
northwestwitch: I'm testing this new options. I see that if I use the --variants-treshold option without the --nr-variants I get this error: 
"'--variants-treshold' can only be used together with '--nr-variants'!"

I think it would be better to assume that if one uses the threshold it's ok to assume that --nr-variants == True, no?

moonso: Ok, I fixed that

northwestwitch: I saw and approved! :)

Issue #1139, opened 2019-02-01
Create code to view and download the variants contributing to the dashboard stats.


northwestwitch:
For instance validated true positive : 279. Include also how many from customer institution and the possibility to download them (just from their own institution) as a list.

This list should include: 
1) Gene name, region and function
2) The HGVS description
3) All information which comes in the GT call box including which variant caller has called the variant
4) Information in the summary box in Scout (ie rank score, CADD score etc)

dnil:
And this is basically done, isn't it?

northwestwitch:
Yes, done, thanks!

PR #1141, opened 2019-02-06
Export validated variants

northwestwitch:
- Created a cli to export validated variants for a single institute. One excel file is created.

- Created a button to export validated variants for all institutes of a user. One excel file per institute is created and download returns a zip file with these excel files. This button is in the dashboard, under the variant stats:
image

- Test for checking that variant query and document line creation works fine


northwestwitch:
fix #1139

northwestwitch:
No, is stage available? It would be nice to test if from there you can produce on excel sheet per institute

moonso:
Should be fine to test on stage as long as you talk to @dnil first

northwestwitch:
> Should be fine to test on stage as long as you talk to @dnil first

Yes I asked him first. Thanks for the review :)

northwestwitch:
Testing on stage and doesn't work. I guess the query takes too much time to run. I'll investigate

moonso:
Didn't you get the mail from scout-browser? I think there was a bug in the code with institutes on case_obj. Guess it should be collaborators instead?

northwestwitch:
I didn't get it, could you post the content here? Thanks

northwestwitch:
Thanks @moonso for your suggestion to use the variant events, it's super fast now!!

I've updated scout-stage with this PR and it works. In the beginning I was a bit surprised that when I tried to download the validated variants and I just got those from one institute, but it's the intended behavior I guess, so keep in mind that if you are an admin you don't get to download everything but just stuff from your institute.

If we don't like this we could change at a later stage.

PR #1146, opened 2019-02-07
MatchMaker patients add/update/remove

northwestwitch:
I've been thinking a lot on how to integrate scout and MatchMaker. I've decided to split this relatively big addition of code into 2 pull requests, mostly thinking about the poor soul who is going to do the review.

This PR contains the code that adds/updates or removes patients from patientMatcher. I will create another PR that takes care of visualizing patient matches.

Instructions to test this PR:

1. Install patientMatcher(https://github.com/Clinical-Genomics/patientMatcher). There's no need to create username and password for its database for testing purposes, so skip Prerequisites.

2. Using the patientMatcher cli, add a client for scout as described here: https://github.com/Clinical-Genomics/patientMatcher#adding-a-client-to-the-database. For instance you could use this command:
```bash
pmatcher add client -id scout_test -token test_token -url scout_test_url
```

3. Run patientMatcher on a different port than the one you use to run scout. For instance:
```bash
pmatcher run -p 9020
```

4. Make sure that you don't see anything matchmaker-related in the case page of scout.

5. Add a 'mme_submitter' role to your scout test user:
image

6. Edit the scout config file(in server/config.py) adding these params:
#MatchMaker related parameters
MME_ACCEPTS = 'application/vnd.ga4gh.matchmaker.v1.0+json'
MME_URL = 'http://localhost:9020'
MME_TOKEN = 'test_token' #this should be the same of when you created the client in step 2

7. Go to the case page and see that now you can submit max 3 pinned variants to matchmaker (genes only or complete variants). You can optionally submit patient sex, hpo terms and OMIM diagnoses. The form should look like this:
image

8. Once you submit the patient you can update your submission by changing pinned variants, or the other options described above. You should also be able to delete the patient from matchmaker. The form should look like this:
image

8. Then end, for now!


northwestwitch:
I'm closing this because I've created a unique PR for MatchMaker integration in Scout

PR #1149, opened 2019-02-13
Showing matchmaker matches for a case

northwestwitch:
This should be tested after #1146 !!!!

Functionalities:

- Click on the "matches" button:
image

- And see matchmaker matches for each affected subject in a case. There is a view with 3 windows.
First is matchmaker submission details, second is matches against external MME nodes, third is matches against other scout samples in patientMatcher:

image

image

- Included a test for the mme_matches controller


northwestwitch:
I'm closing this because I've created a unique PR for MatchMaker integration in Scout

PR #1157, opened 2019-02-25
Expand validated report

### LINKED ISSUE #1156: Add more info to validated variants export
Fix excel files so that:

1. All individuals are on a separated row
2. Sample genotype (ref and alt) are in distinct columns 
3. Include genotype quality (GQ)
4. Include functional annotation (ex: frameshift_variant)

northwestwitch:
fix #1156

### DIFF
--- a/scout/constants/variants_export.py
+++ b/scout/constants/variants_export.py
@@ -29,22 +29,24 @@ VCF_HEADER = [
 ]
 
 VERIFIED_VARIANTS_HEADER= [
+    'Institute',
     'Database id',
     'Variant category',
     'Variant type',
     'Display name',
+    'Local link',
     'Verified state',
     'Case name',
-    'Institute',
+    'Sample',
     'Position',
     'Change',
-    'Canonical_transcript_HGVS'
+    'Canonical_transcript_HGVS',
+    'Functional annotations',
     'Genes', # This is going to be a list flattened to a string
-    'Protein effect',
-    'Variant callers', # This is going to be a list flattened to a string
     'Rank score',
     'CADD score',
-    'Samples',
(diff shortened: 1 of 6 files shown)


northwestwitch:
This one is easy to test on a local computer. 
The steps are the following:

1) Before actually testing the branch see what the actual downloaded document looks like from master branch. 
To do so you'll need to add the following parameters to the config file:

ADMINS = ['a_valid_email@email.com']
MAIL_USERNAME = 'a_valid_email@email.com'
MAIL_PASSWORD = 'a_valid_password_for_the_mail_above'

Then you validate a number of variants (doesn't matter if SNVs or SVs).
From the case page you can choose if these variants are True or False positives.
After this you can download the excel report from this button at the end of the dashboard page:

image


2) Checkout to the expand_validated_report branch and compare the new exported excel file. It should be like this:

* All individuals are on a separated row
* Sample genotype (ref and alt) are in distinct columns
* Include genotype quality (GQ)
* Include functional annotation (ex: frameshift_variant)

dnil:
Nice, and with a test at that! Works well, especially for SNVs. One would need to add a few more things to make it fully useful for the SVs (subtype, coordinates, chr_band,  size), but I'm kind of guessing no-one has requested that yet?

northwestwitch:
Do you want to add the SV-related fields?

dnil:
So we have two updates to dashboard; any preferred merge order @northwestwitch?

northwestwitch:
Which one is the other? The one that is more ready to me merged I guess. I think let's go for the export as it is then we can add other fields in another moment.

northwestwitch:
Ok I merge this

PR #1159, opened 2019-03-01
Skip loading when there are no variants. Fix #1158

### LINKED ISSUE #1158: Do not crash when importing variants from an empty VCF file

moonso:


northwestwitch:
@dnil if you have time review this one first, because it's needed in production!

dnil:
Looks good! Wouldn't it be  possible to add a test as well here? I find it ideal to have such old-bug-tests around making sure a bug can never pop back in again.

moonso:
I'm working on it!

northwestwitch:
> Looks good! Wouldn't it be possible to add a test as well here? I find it ideal to have such old-bug-tests around making sure a bug can never pop back in again.

You kno

dnil:
I made a header only file for testing; could upload that, but I kind of guess you already did the same, but even closer to the original problem? 😄

dnil:
> You know that you don't have tests on yours? ;)

The irony hasn't escaped me. But mine is an graphical interface oriented feature one.. ;-)

moonso:
Tests added 🌟

Issue #1162, opened 2019-03-04
Split adapter. build_query into smaller and simpler queries


northwestwitch:
And write a test for each one of them. 
Bug #1160 has been there for months and nobody noticed before. We need to have a better control of the variant filter functionality.

northwestwitch:
I'm starting with this. Wish me good luck 😒

northwestwitch:
Done, so I'll close this

Issue #695, opened 2017-12-21
Parsing CLNACC from clinvar in Scout


patrikgrenfeldt:
"Here is a heads up for parsing CLNACC from clinvar in Scout:

- Clnvar has removed the CLNACC and replaced it with a variant id in the vcf ID column. When MIP downloads clinvar it will preprocess the clinvar vcf as follows: 
   1. Relay the clinvar variant id from the clinvar vcf ID column as a vcf info key in the same file.
   2. Add a new vcf INFO key pair and header as; '##INFO=<ID=CLNVID,Number=1,Type=Integer,Description="ClinVar Variation ID”>'

i.e I am replacing the CLNACC (which has been deprecated by clinvar) with CLNVID.

Also now it is structured as one line per alt. allele, which is a great improvement."

Contact: @henrikstranneheim

moonso:
When will this change happen in the VCFs?

henrikstranneheim:
Next update of MIP (v6.0.0). We cannot update this outside of MIP now since there is a major change of format.

henrikstranneheim:
Here are the header from clinvar:
```
##INFO=<ID=CLNREVSTAT,Number=.,Type=String,Description="ClinVar review status for the Variation ID">
##INFO=<ID=CLNSIG,Number=.,Type=String,Description="Clinical significance for this single variant">
##INFO=<ID=CLNVID,Number=1,Type=Integer,Description="ClinVar Variation ID">
```
and some vcf info fields:
```
CLNREVSTAT=criteria_provided,_single_submitter;CLNSIG=Uncertain_significance;CLNVID=296280
CLNREVSTAT=criteria_provided,_multiple_submitters,_no_conflicts;CLNSIG=Benign/Likely_benign;CLNVID=262791
```

Here is an address for linking to the hompage:
https://www.ncbi.nlm.nih.gov/clinvar/variation/95677/

where: 95677 is the CLNVID.

moonso:
I am sorry but these fields does not really make sense to me...
What should we actually display? Is it enough with a link to clinvar?

How should we display the above example?

or how should we display:

```
CLNREVSTAT=criteria_provided,_single_submitter;CLNSIG=Likely_benign;CLNVID=295823
```
?

PR #1163, opened 2019-03-04
fixed the clinsig filter and added tests with real variants

### LINKED ISSUE #1160: Clinsig filter not working if variant has multiple values
If a variant has multiple clinsig values, for instance "Pathogenic/Likely pathogenic", it cannot be found using the CLINSIG-filter.

Multiple clinsig.values are stored in the database as a comma-separated string, like this: "Pathogenic, Likelypathogenic":
https://github.com/Clinical-Genomics/scout/blob/e531c3b15769672b0b6ad92e5d75ecd0ac9c5f8a/scout/parse/variant/clnsig.py#L42

When the filter is applied (with Pathogenomic and/or Likely pathogenic checked) it matches the string clinsig.value "Pathogenic, Likely pathogenic" against the list of selected values ["Pathogenic", "Likely pathogenic"], which will not match.

Responsible code:
https://github.com/Clinical-Genomics/scout/blob/e531c3b15769672b0b6ad92e5d75ecd0ac9c5f8a/scout/adapter/mongo/query.py#L292-L296

I'd be happy to fix this myself but I would need some input!

My first attempt was to store the values in the database as a list, but that breaks things in other places. Alternatively, I guess one could do some weird regex matching in the mongo query, but that seems suboptimal...

northwestwitch:
fix #1160. 
So we can have it fixed in the new release!


northwestwitch:
This fix would be partly retroactive because:
1.  All variants with a single clinsig value will be found by clinsig filter
2. All new cases with variants containing multiple clinsig values as in #695 will be annotated correctly: multiple clinsigs, one for each term in the string value.
3. All old cases with clnsig.value like this 'Pathogenic, Likely pathogenic', won't be found by the current clinsig filter.

Suggestions on how to fix point 3?

northwestwitch:
The last commit should take care of point 3. I've included a test with a variant having value == 'Pathogenic, Likely Pathogenic' and it looks like the filter is finding it. I don't know how well it will perform in a real database. Should be tested on stage!

northwestwitch:
This is an ugly fix to a functionality that doesn't work. In the next future we should rewrite the build_query function because as it is now it's too long and convoluted (see #1162)

northwestwitch:
I'll work on it!

> Well done! Would it be possible to add a test for the parsing as well?

I'll work on it!

northwestwitch:
Added the test to check parsing clnsig combination of values from different submitters.

Issue #1170, opened 2019-03-08
Admin-level slice-dash access


northwestwitch:

dnil:
Solved in #1173.

PR #1173, opened 2019-03-11
Fix gene variants

dnil:
Fixes two known aestethic issues with the gene_variants view.

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -476,6 +476,7 @@ def gene_variants(store, variants_query, page=1, per_page=50):
                     if not hgnc_gene:
                         continue
                     gene_obj['hgnc_symbol'] = hgnc_gene['hgnc_symbol']
+                    gene_obj['description'] = hgnc_gene['description']
 
         # Populate variant HGVS and predictions
         gene_ids = []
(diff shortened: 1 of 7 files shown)


northwestwitch:
Nice! If you are done with this we can test it on stage

dnil:
Feel free to make it more elegant: one could either geralise the owner query to take an array for $in, or actually solve the None to "None" conversion issue. But this addresses the issue.

dnil:
And the last one is to fix #1170 (super groovy admin access to dash) ofcourse.. 🤷‍♂️

northwestwitch:
I'm thinking, with the latest commit to this PR, users which are not admins will not be able to see general stats for all institutes. Maybe this is not what we want. We still want to be able to see all general stats whenever there is not a specific query above which could be revealing sensitive stuff.

For instance testing with test user Stella Gibson I see this:
image

But what if I want to see general stats without querying for specific genes in all cases (also not in Stella's institutes?) I'll work on this!

northwestwitch:
I've added a simple javascript function to make sure that it is not possible to have a slice query on all institutes from the web form

northwestwitch:
I've also updated scout stage with this last commit. Didn't test it yet

dnil:
Dashboard looks solid now. Whats wrong with the ”users” page though? Something strange with the stage: looks fine on a demo install.

northwestwitch:
The error is the following:
File "/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/server/blueprints/login/controllers.py", line 25, in users
user_obj['institutes'] = [store.institute(inst_id) for inst_id in user_obj['institutes']]
KeyError: 'institutes'

I'll fix it and test it on stage again

northwestwitch:
I'm going to merge this. Stop me now or never!

PR #1176, opened 2019-03-12
Fix #1175 second part

### LINKED ISSUE #1175: Slice dash interaction with variant level export
- [x] variant export broken for (at least) cust002: using sanger_ordered as a proxy for validation is not enough as variants can be marked true / false at the pin level without a sanger_order.
- [ ] totals on the variants is for a slice. Actually use a sliced query anyway for the export? That is best given the page as it is today, otherwise let's make a more clear boundary btw the sliced and unsliced part.

Sorry, I didn't pay enough attention to this when the two branches merged - and the tests were presumably not geared towards finding this issue.

dnil:
Here is a python translation for the part of this that did not deal with export. 😄 

I will leave the export for another day, but I would tentatively suggest considering passing the slice query also to the VariantHandler.verified(). Or making even more clear that the button will export ALL your variants. We can ask tomorrow which is preferred!

### DIFF
--- a/scout/server/blueprints/dashboard/controllers.py
+++ b/scout/server/blueprints/dashboard/controllers.py
@@ -66,6 +66,8 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
     validated_cases = set()
 
     # Variant level information
+    our_validated_positives = 0
+    our_validated_negatives = 0
     validated_tp = set()
     validated_fp = set()
 
--- a/scout/server/blueprints/dashboard/controllers.py
+++ b/scout/server/blueprints/dashboard/controllers.py
@@ -43,20 +43,41 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
     data['analysis_types'] = get_analysis_types(adapter, total_cases,
                                 institute_id=institute_id, slice_query=slice_query)
 
+    # VARIANTS
+
+    # Here we disregard any slice query.
+    general_info = get_general_case_info(adapter, institute_id=institute_id)
+
+    total_cases = general_info['total_cases']
     # Fetch variant information
     LOG.info("Fetch sanger variants")
-    validation_query = {'sanger_ordered': True}
+    order_query = {'sanger_ordered': True}
+    # NOTE: this will not find all verified variants. Any pinned variant can
+    #       be tagged true or false positive. Generally speaking all
+    #       causatives were true positives.
+
+    # validation_query = {'validation': {'$in'}} #unused
+    # Also actually query for TPs and FPs
+    tp_query = {'validation': "True positive"}
+    fp_query = {'validation': "False positive"}
 
     # Case level information
-    validation_cases = set()
+    ordered_validation_cases = set()
     validated_cases = set()
 
     # Variant level information
     validated_tp = set()
     validated_fp = set()
 
-    LOG.info("Find all validated variants with query {}".format(validation_query))
-    validation_ordered = adapter.variant_collection.find(validation_query)
+    LOG.info("Find all validated variants with query {}".format(order_query))
+    validation_ordered = adapter.variant_collection.find(order_query)
+
+# If you want the full collection:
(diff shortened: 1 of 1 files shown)


dnil:
Yes, and for GitHub: this will partially Fix #1175.

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -43,20 +43,43 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
...
+    # VARIANTS
+
+    # Here we disregard any slice query.
+    general_info = get_general_case_info(adapter, institute_id=institute_id)
+
+    total_cases = general_info['total_cases']
     # Fetch variant information
     LOG.info("Fetch sanger variants")
-    validation_query = {'sanger_ordered': True}
+    order_query = {'sanger_ordered': True}
```
northwestwitch: Would it be possible here to have a query which combines 'sanger_ordered', and 'validation' in ['True positive', 'False positive]? Then you could loop on the results just once instead of having 3 different loops.

dnil:
The issue I had was that this got slow with my combined queries - not quite an order of magnitude but a larger constant factor than 3, and I didn’t want to require a new index.
I would be most grateful for an optimized query if you can think of/test one!

northwestwitch:
Could you use the events collection? For instance I've created this adapter.variant.verified which returns all verfied variants for an institute. I'm using it to do the excel export of the verified vars.

dnil:
Yes: I was hesitant since that part seemed broken yesterday, but optimistic now! I am not sure which will be faster, presumably your suggestion. But lets try this on stage: if it is good enough.. 😄

northwestwitch:
Stage updated!

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -43,20 +43,41 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
     data['analysis_types'] = get_analysis_types(adapter, total_cases,
                                 institute_id=institute_id, slice_query=slice_query)
 
+    # VARIANTS
+
+    # Here we disregard any slice query.
+    general_info = get_general_case_info(adapter, institute_id=institute_id)
```
northwestwitch: Question: why do you want to disregard the filter from here? Then it will happen that some stats will be based on all cases available and some other on the slice_query. Wouldn't it be messy then to understand where the numbers come from?

dnil: Yes, that is non-obvious. Personally, I don't see the use for fraction of verifying variants in any of the current slice categories. If we have subsets like "low quality variants", "tissue", "cancer" or such it suddenly makes a lot of sense!

Couldn't we just make two distinct panels on the page: one with cases, which is sliceable. And then one with variants, which is so far not sliceable, and has the download button grouped with it?

scout/server/blueprints/dashboard/controllers.py
dnil: But with your event query instead, obviously!

scout/server/blueprints/dashboard/controllers.py
northwestwitch: It should be quite easy to encapsulate the distinct parts in 2 panels, this way it will be more clear the data in which the slice query has an effect. I can draft this

northwestwitch:
Something like this? 
image


image

northwestwitch:
Well it's actually 3 panels that I made. The first 2 which are sliceable, and the variants that isn't for now. Suggestions?

dnil:
Nice screenshots - me like!

northwestwitch:
I didn't have much time to work on this today, but it's high on my to-do list!

northwestwitch:
Hi, I think I'm done with this fix. @dnil do you prefer if I apply those changes to this PR or if I open a new?

dnil:
Cool! That is entirely up to you: whichever is more practical!

PR #1184, opened 2019-03-15
Hpo tree

moonso:
This adds some basic functionality to search for phenotypically similar cases. It is rudimentary at the moment, but it shows how this could be done.
One may test the functionality by running `scout view cases --case-id 'some id' --similar`


moonso:
Tested on stage so feel free to review whenever you have time

dnil:
Awesome! ✌️ 

I guess by looking at the checks you can anticipate at least one review comment?

moonso:
👀 tests 👀

northwestwitch:
@moonso it could be useful to use one of the existing algorithms to perform ontology-based similarity searches: look [here](#1214).

northwestwitch:
I've been taking a look at this PR and you are indeed using one of the semantic similarity search algorithms. That's nice @moonso ! So, tests I guess :)

scout/utils/algorithms.py
```diff
@@ -0,0 +1,30 @@
...
+        set_1, set_2 (set(str))
+    
+    Returns:
+        ui_score (float)
+    """
+    LOG.debug("Set 1: %s", ', '.join(set_1))
+    LOG.debug("Set 2: %s", ', '.join(set_2))
+    if not (set_1 and set_2):
+        return 0
+    ui_score = len(set_1.intersection(set_2)) / len(set_1.union(set_2))
```
dnil: Neat!

northwestwitch:
Hi @moonso , as you can see I've made an attempt to write tests for this PR. The problem is that when I test for similarity against 2 cases (one very similar and one that shouldn't be similar at all) I get the same score. Could you take a look when you have time? Any other suggest for testing it?

moonso:
Hi! Thank you for finishing this. Sure I will have a look at it

dnil:
Great! I’ve made a sketch of how to call this from a general case name_query. It works with the caveat that I haven’t figured out a good way to preserve the score order all the way out to the case view. Any thoughts? Can one find a natural cutoff threshold for similarity, so that all hits above it would be relevant? Or simply a separate view?

moonso:
Hi! So this should be fixed now. I hadn't done just one but two classic misstakes :D 
1. Forgot to reset variable in a loop. (in this case the set of phenotypes just got bigger in the loop)
1. Sorted the resulting list on lowest scores first instead of highest 😸 

Nice with tests 👍.

moonso:
@dnil would it be wise to put that in another PR? I will have a look at it but it would be nice to merge this first

dnil:
Yes, go ahead and merge this! I can open a new one with a sketch afterwards, so we can be a bit more specific!

tests/adapter/mongo/test_case_handling.py
```diff
@@ -415,7 +415,7 @@ def test_get_similar_cases(hpo_database, test_hpo_terms, case_obj):
     assert adapter.case_collection.find().count() == 1
 
     # Add another case with slightly different phenotype
-    case_2 = copy.copy(case_obj)
+    case_2 = copy.deepcopy(case_obj)
```
northwestwitch: Oh yes, deepcopy!

northwestwitch:
We should probably re-test it on stage

northwestwitch:
It's on stage now. Looks like it's working.

dnil:
Lets merge; it shouldn't break anything (famous last words). ;-)

PR #1187, opened 2019-03-16
Add more server level tests.

dnil:
I thought I’d start with just accessing each page and find it viewable. Not finished but I guess mergeable if it should get stuck. 😸


dnil:
Sure, good suggestion! I'll give it a shot!

northwestwitch:
Nice!! 👍

dnil:
60.4%; but still doesn't show on the badge. Need 80% for orange next. 😆 CLI tests will help, and there are a couple of main pages to just reload. If we keep adding a test or three for existing functions once in a while between PRs, we will get there

henrikstranneheim:
👍

PR #1188, opened 2019-03-18
Fix #1175 query events

### LINKED ISSUE #1175: Slice dash interaction with variant level export

northwestwitch:
1. Collect validated and validable variants using event collection
2. Splitted the dashboard into 3 tabs
3. Variant tab is showing panels even if there are no validated variants yet


northwestwitch:
This branch is on stage now

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -23,130 +23,143 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
...
+        }
+    )
+
+    # taking into account that var_valid_orders might be 0:
+    percent_validable = 0
+    percent_validated = 0
+
+    if var_valid_orders:
+        percent_validable = len(validated_tp) / var_valid_orders
+        percent_validated = len(validated_fp) / var_valid_orders
```
dnil: Could we change the name of these? percent_validated_tp, percent_validated_fp perhaps?

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -23,130 +23,143 @@ def get_dashboard_info(adapter, institute_id=None, slice_query=None):
...
+    # append
+    overview.append(
+    {
+        'title': 'Validation ordered',
+        'count': n_validation_cases,
+        'percent': n_validation_cases / total_sliced_cases,
+    })
+    overview.append(
+    {
+        'title': 'Validated cases',
```
dnil: Hm, might be misunderstood as nr of true positives. How about 'Validated cases (TP + FP)'?

scout/server/blueprints/dashboard/controllers.py
northwestwitch: Sure, didn't realize how bad I named them

northwestwitch:
Thanks, I'll merge it!

PR #1189, opened 2019-03-19
Hide not causatives

### LINKED ISSUE #1178: Remove not causative buttons from Causatives page

northwestwitch:
I've included the variant category, rank score, zygosity and inheritance models. It's not printing variants which are not in the cases.causatives now

image

fix #1178


dnil:
Looks really nice! Removing the "non-causatives" entirely was a bit further than I gathered from the meeting - I for one only thought to remove the label - but it does seem the more prudent approach. We can always re-enable it later after some discussion.

northwestwitch:
Yes, they're there, but just hidden!

northwestwitch:
Merging now, thanks for the review!

PR #1190, opened 2019-03-20
Split query builder

northwestwitch:
Splits the monster long query builder into a smaller function that calls other smaller functions (when needed). These other functions are:

- gene_filter query builder
- coordinate filter query builder
- a clinsig filter query builder
- a secondary filter query builder ( ex minor query )

I've also moved the previous comments here and there, plus I've added some more.

I've left untouched the tests on purpose, because I think the re-organized code should pass the tests from before without the need to change anything. == > The final query product MUST be the same.

PS: this PR drove me crazy at first, but now it feels like I've finally understood the whole prioritization of the query params. So a very useful exercise!! :)


northwestwitch:
fix #1162

scout/adapter/mongo/query.py
```diff
@@ -163,7 +358,7 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
                         ]
                     }
                 )
-            logger.debug("Adding gnomad_frequency to query")
+            LOG.debug("Adding gnomad_frequency to query")
```
dnil: This function gets fairly long as well. Would splitting into a few broader categories according to type; e.g. frequency, function, sv specific, cancer specific make it better?

scout/constants/query_terms.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+# These terms are used by the function that creates a variant query dictionary in adapter
+
+SECONDARY_CRITERIA = [
```
dnil: :+1: Perhaps also a 0-level for gene, genlist etc for clarity?

scout/adapter/mongo/query.py
```diff
@@ -114,33 +314,28 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
-            chromosome = query['chrom']
-            mongo_query['chromosome'] = chromosome
-            #Only check coordinates if there is a chromosome
-            if (query.get('start') and query.get('end')):
-                mongo_query['position'] = {'$lte': int(query['end'])}
+        return gene_query
 
-                mongo_query['end'] = {'$gte': int(query['start'])}
 
-        mongo_query_minor = []
```
dnil: minor = secondary, major = primary. Got it! :wink:

scout/constants/query_terms.py
dnil: E.g. one could argue that inheritance pattern might often be used as a 0th level excluding criteria. Users are typically surprised to find variants not following inheritance pattern after hitting this, and it is not used in the clinical filter. Im not saying we should change it now, but perhaps convenient to have it explicit!

scout/adapter/mongo/query.py
northwestwitch: Yes, that's the first thing I should have written in the comment, probably!

scout/adapter/mongo/query.py
```diff
@@ -89,20 +91,218 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
+        # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency
+        secondary_terms = False
+
+        # check if any of the primary criteria was specified in the query
+        for term in PRIMARY_CRITERIA:
+            if query.get(term):
+                primary_terms = True
+
+        # check if any of the secondary criteria was specified in the query:
+        for term in SECONDARY_CRITERIA:
```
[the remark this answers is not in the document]
northwestwitch: Ok, then I leave it as it is?

scout/adapter/mongo/query.py
```diff
@@ -89,20 +91,218 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
+            if query.get(term):
+                secondary_terms = True
+
+        if primary_terms is True:
+            clising_filter = self.clinsig_query(query, mongo_query)
```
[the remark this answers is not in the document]
northwestwitch: Sure, didn't see the typo!

scout/adapter/mongo/query.py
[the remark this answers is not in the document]
dnil: To me, the nicest way would be to use three categories: inclusive, hard exclusive and soft exclusive (with whatever names you like). But, we currently have no need for searching a list for the hard criteria, so I guess leaving them in a comment is good enough.

scout/adapter/mongo/query.py
northwestwitch: What about looping through the list of secondary criteria and grouping them by filter construction? I have given it a try and the function is a bit shorter now. See if you like it, otherwise I can change!

scout/adapter/mongo/query.py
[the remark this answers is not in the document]
northwestwitch: clisign_filter now

scout/adapter/mongo/query.py
```diff
@@ -89,20 +91,218 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
+
+        elif gene_query: # no primary or secondary filters provided
+            mongo_query['$and'] = [{ '$or': gene_query }]
+
+        LOG.info("mongo query: %s", mongo_query)
+
+        return mongo_query
+
+
+    def clinsig_query(self, query, mongo_query, ):
```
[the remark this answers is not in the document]
northwestwitch: Done!

scout/adapter/mongo/query.py
```diff
@@ -114,33 +314,28 @@ def build_query(self, case_id, query=None, variant_ids=None, category='snv'):
...
+                query(dict): a dictionary of query filters specified by the users
+                mongo_query(dict): the query that is going to be submitted to the database
+
+            Returns:
+                mongo_query_minor(list): a dictionary with secondary query parameters
+
+        """
+        LOG.debug('Creating a query object with secondary parameters')
 
         mongo_query_minor = []
```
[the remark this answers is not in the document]
northwestwitch: Sure, mongo_secondary_query now

scout/constants/query_terms.py
northwestwitch: I think we have to discuss in person about this because I don't understand. Same for the soft, hard excluding criteria. : -/

northwestwitch:
@dnil when you have time take a look at the last 2 commitments. I think we might have it right now

dnil:
Yep, this works for me! Tried the latest on stage? If so, I think merge! 👍

northwestwitch:
Testing on stage and I get the following error: 
 File "/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/adapter/mongo/query.py", line 160, in build_query
    secondary_filter = self.secondary_query(query, mongo_query)
  File "/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/adapter/mongo/query.py", line 366, in secondary_query
    'gnomad_frequency': {'$lt': float(gnomad)}
TypeError: float() argument must be a string or a number, not 'NoneType'

It needs some fixing!

northwestwitch:
It was a stupid mistake, fixed now. I've just tested on stage and it looks like it's working. I'll merge it!

northwestwitch:
Thanks for the review! :)

PR #1193, opened 2019-03-25
provide collaborator when querying for cases in cases views

### LINKED ISSUE #1192: Shared cases no longer visible on institute cases
BUG: Cases view no longer shows shared cases. Presumably something goes awry with asking for owner rather than collaborator, very likely as a consequence of an only partly thought through tightening of the query for the dashboard view.

northwestwitch:
This should fix #1192, hopefully.


northwestwitch:
fix #1194 as well

PR #1195, opened 2019-03-25
simpler fix for #1192 and #1194

### LINKED ISSUE #1192: Shared cases no longer visible on institute cases

### LINKED ISSUE #1194: Can't display some case pages
Example: https://scout.scilifelab.se/cust002?query=F0017016

This is very likely caused by the fact among the saved collaborators there is an unwanted item coming from the HTML form:

image

northwestwitch:
fix #1192 and fix #1194


northwestwitch:
This should work for real. The branch is already testable on stage!

scout/server/blueprints/cases/templates/cases/actionbar.html
```diff
@@ -278,7 +278,7 @@ <h5>
         <form action="{{ url_for('cases.share', institute_id=institute._id, case_name=case.display_name) }}" method="POST">
           <div class="input-group">
             <select class="form-control" name="collaborator">
-              <option>Select institute</option>
+              <option class="placeholder" selected disabled value="">Select institute</option>
```
dnil: Very good - that was the cause of it, and your solution should be applied to all selects!

scout/server/blueprints/cases/controllers.py
```diff
@@ -124,8 +124,11 @@ def case(store, institute_obj, case_obj):
                                 .format(hpo_term['phenotype_id']))
 
     # other collaborators than the owner of the case
-    o_collaborators = [store.institute(collab_id) for collab_id in case_obj['collaborators'] if
-                       collab_id != case_obj['owner']]
+    o_collaborators = []
+    for collab_id in case_obj['collaborators']:
+        if collab_id != case_obj['owner'] and store.institute(collab_id):
```
dnil: Better safe than sorry!  👍

scout/adapter/mongo/case.py
```diff
@@ -53,6 +53,10 @@ def cases(self, owner=None, collaborator=None, query=None, skip_assigned=False,
         LOG.debug("Fetch all cases")
         query = query or {}
 
+        # Prioritize when both owner and collaborator params are present
+        if collaborator and owner:
+            collaborator = None
```
dnil: Elegant!

dnil:
Verified on stage. Suggest we think about deploying before someone else misses a case?

scout/server/blueprints/cases/controllers.py
northwestwitch: That was adding None things to the array of collab_ids, because it was using "Select institite" for ids

northwestwitch:
We can do a new patch release tomorrow!

scout/server/blueprints/cases/controllers.py
dnil: Yep: when I saw your picture, and recalled Anna clicking "Share" without selecting an institute it was all clear.. 😄

PR #1197, opened 2019-03-27
Release 4.3.3

northwestwitch:
Releasing a new version with these specifics:

### Added
- Different individual table for cancer cases

### Fixed
- Dashboard collects validated variants from verification events instead of using 'sanger' field
- Cases shared with collaborators are visible again in cases page
- Force users to select a real institute to share cases with (actionbar select fix)


northwestwitch:
Thanks!

northwestwitch:
@moonso could you update pypy when you have time?

northwestwitch:
Updated scout stage and stage production on clinical-db. No problems so far!

dnil:
Looking good! :+1:

PR #1201, opened 2019-04-02
Add case_id where global comment was added

bjhall:
For global comments: show the case ID for the case where the comment was originally added.


northwestwitch:
I'm thinking about this PR. I like the idea of give an indication of where the global comment on one variant comes from but I don't think we can introduce this suggestion (at least as it is now). There are 2 main reasons for this:

- First reason is security related. If you share the id of a case then any users will know that a certain case (also from other institutions) has that variant. I don't think most of our users are allowed by law to share this information with other scout users. What is good with global comments is that you can give feedback on a mutation without disclosing the subject that harbors it.

- Second reason is a practical one: case id is a string that is used in the database but it's not really how cases are named on scout interface, all users see if cases.display_name. So nobody (at least among our clinicians) would recognize samples by their id, and those names wouldn't really add useful information. Even if you grab display names for samples and you show those instead, then the users wouldn't know which institute that samples belongs to. And I'm afraid if you are too specific you get into the problem I mentioned above

bjhall:
I understand. It has always been unclear to me what global comments should be used for.  We use it mainly for writing comments that we want to see the next time we find the same variant, not necessarily sharing it with other institutes (we only really have one institute in Lund though). It feels like there should be a way to add institute-specific global comments as well...


Anyway, how about this solution:
* If the global comment was written in a case of the same institute, as the current show the name of the case (like the current PR, but with display_name)
* Otherwise show the name/institute of the user who wrote the comment (like you suggested)

If you think that's acceptable I could try to update the PR with these changes as soon as I find the time.


Related: The _display_name_ and __id_ are always the same for us. How come? We are using YAML files to load the cases and specifying the "family" field. This value ends up being used for both _display_name_ and __id_.

dnil:
Sounds like a plan! In summary, I'd suggest always showing user and institute, then case names only for institutes the user belongs to.

For case loading, I would defer to @moonso, but usually there are subtle differences regarding what happens with data from pedigree files (.ped) and what has been specified in the load config yaml..

moonso:
Hi @bjhall , I think this new suggestion from you and @dnil sounds good. 

For the loading part I think that this has been messy for a long time now, it is hard to follow in the code what is going on and the documentation is not really up to date. There are historical reasons for this and I hope we will be able to fix it soon. It is fine to use the `family` field like you're doing at the moment. If you want to have a display name that differs from the case id you can add a `family_name` field in the yaml load config.

I'm really happy that you use and develop scout, think we will both benefit a lot from this🤠

moonso:
So should we close this or try to merge?

northwestwitch:
> So should we close this or try to merge?

We can't merge it as it is, I suggest we close it and we re-open it if we need to.

moonso:
Ok sound good. Shout if you want something else @bjhall 🛩

bjhall:
No problems! Perhaps I'll try to implement it in the way discussed above the next time I have time for scout development (hopefully in a couple of months).

PR #1202, opened 2019-04-03
Label pinned variants in variant list and list overlapping gene panels…

bjhall:
Adds a couple of things to the variant list which were often requested by users here in Lund:

* Display a label for pinned variants in the variant list (ideally I suppose we should use some icon for this, but for now I just added a label with the word "Pin!")

* For variants that are in any of the cases gene panels, show a badge with the number of ovelapping gene panels. On hover it shows the IDs of the panels (maybe this should be the display names)

Here's a screenshot showing both these changes:
image

BTW, I won't be offended if you reject any of these pull requests! :) I've implemented these features because they've been requested by our users, and thought I'd give you the chance to evaluate if you want to merge any of them.


henrikstranneheim:
This is great Björn! This is really how we would like to work in the future - contributing together to solve the common problems that we all face.

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -151,6 +151,26 @@ <h4 class="panel-title">
       <span class="glyphicon glyphicon-comment"></span>
     </a>
   {% endif %}
+
+
+  {% if variant.panels|count > 0 %}
```
dnil: Great for the research view! For clinical use, we essentially always see variants in multiple panels - e.g. the primary analysis panel, OMIM and OMIM-AUTO? Do you see a need for the clinical side as well? Otherwise perhaps check if this is a research view and then display?

dnil:
```suggestion
  {% if form.variant_type.data == 'research' %}
```

bjhall: In Lund we're not really using the research/clinical feature of Scout, we're always loading all variants. Thus, this suggestion won't really work for us (since we're never in "research mode").

I see that it becomes cluttered when every variant has the badge, but it still has some value to be able to quickly see what panels the gene is in without clicking on the variant (at least this is what our users are after).

Maybe one could consider  a different, less intrusive way of showing this?

dnil: I see.. You may want to reconsider this at some point as you start amassing whole genomes, but let's try to keep things compatible. How about adding it to the gene hover? The gene hover currently only shows a full name / description for the gene, and the panel info might be rather logical there without increasing clutter?

bjhall:
I updated the branch and moved the panel info to the gene column as @dnil suggested, but it is still in a always-visible badge (now only showing the count).  Completely hiding it in the hover text would unfortunately not satisfy our users' request.  See screenshot below.

If you still think this is to intrusive I'm completely OK to keep maintaining this feature specifically in our fork. 

varlist_enhancements_rev2

dnil:
I don't think it either slows down or introduces too much clutter. Still, it looks a little stupid after filtering, but I can live with that. Pins are good! Let's merge this and see if there are many objections. If so we can alter it a bit later. Thank you for contributing!

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -10,13 +11,29 @@
...
+          {{ gene.hgnc_symbol or gene.hgnc_id }}
+        </a>
+      </div>
+   
+    {% endfor %}
+  </div>
+  {% if variant.panels|count > 0 %}
+    <div
+       class="badge pull-right"
+       data-toggle="popover"
```
dnil: Actually, when I think about it: why not use the tooltip framework as for the other hovers? That way you will get the same background color etc?

northwestwitch:
Shall we merge this one???

scout/server/blueprints/variants/templates/variants/components.html
bjhall: The tooltips/popovers are already inconsistently used in the variants view. Both the compounds and comments hovers use popover (supposedly since they contain table elements?), and the others use tooltips. I guess the main difference (apart from the look) is that you can click popovers to make them stick.

Any guidelines on when to use one or the other? (if not, we should probably avoid bikeshedding this issue :))

dnil: Thanks; I didn't notice.. :blush: It seems the intended use is popover for "new content" and tooltip for short remarks that really don't add anything. I guess that qualifies for gene names, not so sure about the frequency. At any rate, do go ahead and use popover, and we can later change the styles for either or all if we want it to look consistent!

PR #1203, opened 2019-04-03
Mark primary transcript in verification email

bjhall:
Simply adds the the word "primary" to the refseq primary transcript in the verification email.

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -946,7 +946,7 @@ def variant_verification(store, mail, institute_obj, case_obj, user_obj, variant
                         transcript_line.append(urllib.parse.unquote(tx_obj['protein_sequence_name']))
                     else:
                         transcript_line.append('')
-                    if refseq_id in gene_obj['common']['primary_transcripts']:
+                    if refseq_id in gene_obj.get('common',{}).get('primary_transcripts',{}):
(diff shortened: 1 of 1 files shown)


northwestwitch:
When it has to do with transcripts @moonso is the expert here ;)

scout/server/blueprints/variants/controllers.py
```diff
@@ -946,6 +946,10 @@ def variant_verification(store, mail, institute_obj, case_obj, user_obj, variant
                         transcript_line.append(urllib.parse.unquote(tx_obj['protein_sequence_name']))
                     else:
                         transcript_line.append('')
+                    if refseq_id in gene_obj['common']['primary_transcripts']:
```
northwestwitch: I think it's fine but I remember we had some bug before because not all gene_obj had the 'common' key. I think it won't hurt here if we check this first, something like:
```
if  gene_obj.get('common') and if refseq_id in gene_obj['common']['primary_transcripts']:
```

moonso: what about:

```python
if refs_id in gene_obj.get('common',{}).get('primary_transcripts',{}):
    ...
```

bjhall: I've updated the branch according to moonsos suggestion.

northwestwitch:
Nice, I'll merge! :)

PR #1206, opened 2019-04-05
Converts the CLI to use the app context and introduces tests for all its functions

northwestwitch:
Coverage will be increased by 10% (basically the whole CLI was not tested before).

### Don't be scared, it's a lot of changes but always the same ones!

Now **every function in the CLI is tested except the command scout setup demo**

Scout should still work with the 2 provided config files (one for the cli and one for the app), but the cli is not reading connection params from its config file, it basically lives in the context of the app (so the app config file is used for these params). 

The only param which is taken and used from the cli config file is the omim_api_key. If we decide to add it to the app config file then we completely get rid if the cli config, which would be good. But to do so you have to rewrite how you launch the app (the gunicorn commands in server repo), and since this is already a huge PR I'd leave it for later!

I can recommend to the reviewer of this PR to use the split view when watching the changed code, since it makes it much more easy to follow the changes.


dnil:
+10% coverage is worth a lot of reviewing.. But I think @moonso or @patrikgrenfeldt would be more appropriate? Thanks!

scout/commands/base.py
```diff
@@ -28,97 +24,46 @@
...
 @click.option('--demo', is_flag=True, help="If the demo database should be used")
-@click.version_option(__version__)
-@click.pass_context
-def cli(context, mongodb, username, password, authdb, host, port, loglevel, config, demo):
-    """scout: manage interactions with a scout instance."""
-    # log_format = "%(message)s" if sys.stdout.isatty() else None
+@click.option('-c', '--config', type=click.Path(exists=True))
+@with_appcontext
+def app_cli(loglevel, demo, config):
+    """Entry point of Scout CLI"""
```
dnil: I don't think I would have understood either version when I ran "scout" for the first time. How about at least merging them:

```suggestion
    """Scout CLI:  manage interactions with a scout instance."""
```
Ah, and the corresponding test will need a change then! :blush:

scout/commands/convert.py
```diff
@@ -2,32 +2,34 @@
...
 
 log = logging.getLogger(__name__)
 
 
 @click.command('convert', short_help='Convert gene panels')
 @click.argument('panel', type=click.File('r'))
-@click.pass_context
-def convert(context, panel):
+@with_appcontext
+def convert(panel):
```
dnil: :+1:

scout/commands/export/export_command.py
```diff
@@ -27,12 +27,12 @@
...
 
 @click.group()
-@click.pass_context
-def export(ctx):
+def export():
     """
     Export objects from the mongo database.
     """
     LOG.info("Running scout export")
+    pass
```
dnil: Yeah, that seems to have been missing! 😸

scout/commands/export/variant.py
```diff
@@ -51,7 +51,12 @@ def verified(context, collaborator, test, outpath=None):
         LOG.warning('There are no verified variants for institute {} in database!'.format(collaborator))
         return None
 
-    document_lines = export_verified_variants(verified_vars)
+    unique_callers = set()
+    for var_type, var_callers in CALLERS.items():
+        for caller in var_callers:
+            unique_callers.add(caller.get('id'))
+
+    document_lines = export_verified_variants(verified_vars, unique_callers)
```
dnil: A little bit OT I suppose? Well, you are the expert here @chiara: looks good to me, if you confirm this is intentional.

scout/commands/query/query_command.py
```diff
@@ -1,10 +1,19 @@
 import logging
-
+from flask.cli import with_appcontext
+from scout.server.extensions import store
 import click
 
 log = logging.getLogger(__name__)
 
-@click.command('hgnc', short_help='Check if a gene exist')
+@click.group()
```
dnil: Yes, me like! No dependencies on this I guess? Or was the grouping implicit before?

scout/commands/query/query_command.py
```diff
@@ -17,46 +26,39 @@
...
+
     else:
         click.echo("#hgnc_id\thgnc_symbol\taliases\ttranscripts")
         for gene in result:
             click.echo("{0}\t{1}\t{2}\t{3}".format(
                 gene['hgnc_id'],
                 gene['hgnc_symbol'],
                 ', '.join(gene['aliases']),
-                ', '.join(tx['ensembl_transcript_id'] for tx in gene['transcripts']),
+                ', '.join(tx['ensembl_transcript_id'] for tx in gene['transcripts']) if gene.get('transcripts') else '',
```
dnil: Good catch!

scout/commands/update/case.py
```diff
@@ -29,44 +31,41 @@
...
     """
     Update a case in the database
     """
-    adapter = context.obj['adapter']
+    adapter = store
     if not case_id:
         if not (case_name and institute):
             LOG.info("Please specify which case to update.")
-            context.abort
+            return
```
dnil: Why no raise click.Abort() in here?

scout/commands/view/intervals.py
```diff
@@ -1,22 +1,30 @@
...
-def intervals(context, build):
-    """Show all indexes in the database"""
+@with_appcontext
+def intervals(build):
+    """Show all coding intervals in the database"""
     LOG.info("Running scout view index")
-    adapter = context.obj['adapter']
+    adapter = store
 
+    if adapter.hgnc_collection.find().count() == 0:
```
dnil: 👍

scout/demo/__init__.py
```diff
@@ -28,4 +28,7 @@
 research_snv_path = pkg_resources.resource_filename('scout', research_snv_file)
 research_sv_path = pkg_resources.resource_filename('scout', research_sv_file)
 
-empty_sv_clinical_path = pkg_resources.resource_filename('scout', empty_sv_file)
\ No newline at end of file
+empty_sv_clinical_path = pkg_resources.resource_filename('scout', empty_sv_file)
+
+delivery_report_file = 'demo/delivery_report.html'
```
dnil: Ehr, yes! 😸

setup.py
```diff
@@ -119,12 +119,12 @@ def run(self):
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Topic :: Internet :: WWW/HTTP :: Dynamic Content',
-        'Topic :: Software Development :: Libraries',
+        'Topic :: Software Development :: Librariesxs',
```
dnil: Nope. @moonso would not approve - it appears you have been using emacs. 🤣

tests/commands/delete/test_delete_cmd.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Test the CLI that removes it by giving a wrong version
+    result =  runner.invoke(app_cli, ['delete', 'panel',
+        '--panel-id', db_panel['panel_name'],
+        '-v', 5.0 # db_panel version is 1.0
+        ])
+    assert 'No panels found' in result.output
+
+    # Test the CLI by using panel name without version
+    result =  runner.invoke(app_cli, ['delete', 'panel',
```
dnil: Ah, this is really neat! Good job!

tests/commands/test_base_command.py
```diff
@@ -1,10 +1,24 @@
...
 
-from scout.commands import cli
-from click.testing import CliRunner
+from scout.commands import app_cli
+from flask.cli import current_app
 
 #Sanity check for cli
-def test_version():
-    runner = CliRunner()
-    result = runner.invoke(cli, ['--version'])
```
dnil: Absolutely better on almost every level, but why not still have a ```--version```?

tests/commands/view/test_view_whitelist_cmd.py
```diff
@@ -0,0 +1,26 @@
...
+    assert 'INFO Running scout view users' in result.output
+
+    # insert mock obj in whitelist collection
+    store.whitelist_collection.insert({'_id' : 'fake_whitelist_id'})
+
+    # Test CLI
+    result =  runner.invoke(app_cli, ['view', 'whitelist'])
+    assert result.exit_code == 0
+    # mock obj should be returned
+    assert 'fake_whitelist_id' in result.output
```
dnil: I feel the CLI is now well tested. 🥇

northwestwitch: Thanks for the review, I really appreciate. I'll address your comments as soon as I can! :)

northwestwitch:
I'm testing this on stage (clinical-db), after installing the branch and moving to the stage environment (us). I get connection refused error:
pymongo.errors.ServerSelectionTimeoutError: localhost:27017: [Errno 111] Connection refused

It shouldn't even be using that default mongo port.

northwestwitch:
Ok the 'us' command to get to the stage environment is not the right one on clinical-db because it moves to an environment named 'stage', while scout stage is installed by the bash script in an environment named prod-stage.

northwestwitch:
Ok this is hell. I've just realized that the programs which use scout as s dependency will not be able to import the commands using the cli.commands because now it's renamed app_cli. I need to do changes before this PR is usable.

dnil:
Hmpf.. Python needs a model for what - and how - interfaces are exposed to other applications. When someone said they were going to use the CLI interface, I kind of assumed they would run the commands via e.g. an interpreter.. 😊

northwestwitch:
I've just tested this on clinical-db stage and **BOTH THE CLI AND THE WEB APP WORK!!!!**.

Next steps before and after merging this, to be done in the right order:


1. update **changelog**
1. create environment variable in prod environment like the one now available on stage, see this PR: https://github.com/Clinical-Genomics/servers/pull/123
1. put the above **envvar** in production (restart the server thing)
1. Create e new **major release** of scout (of the type that breaks the code) named 5.0
1. scout cli accepts an optional argument that is not used any more (config). This is because scout cli is invoked by our servers (rasta, hasta, clinical-db) using an **alias 'scout'**.
This alias looks like this: 'scout --conf path_to_conf_file'
Since there is no need of this param any more do this:
  1. change these aliases (stage and prod) on servers repo to just execute scout.
  1. Modify scout cli base command removing this parameter.

scout/commands/base.py
```diff
@@ -32,38 +32,30 @@
 @click.option('--loglevel', default='INFO', type=click.Choice(LOG_LEVELS),
               help="Set the level of log output.", show_default=True)
 @click.option('--demo', is_flag=True, help="If the demo database should be used")
-@click.option('-c', '--config', type=click.Path(exists=True))
 @click.option('-v', 'vers_num', is_flag=True, help="display version of Scout")
+@click.option('-c', '--config', type=click.Path(exists=True))
 @with_appcontext
-def app_cli(loglevel, demo, config, vers_num):
+def cli(loglevel, demo, vers_num, config):
```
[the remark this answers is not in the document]
northwestwitch: You mean the config argument that is not called afterwards? There is a reason to keep it for now unfortunately. See point 5 of my comment under this.

scout/commands/base.py
```diff
@@ -28,97 +24,46 @@
...
-@click.option('-db', '--mongodb', help='Name of mongo database [scout]')
-@click.option('-u', '--username')
-@click.option('-p', '--password')
-@click.option('-a', '--authdb', help='database to use for authentication')
-@click.option('-port', '--port', help="Specify on what port to listen for the mongod")
-@click.option('-h', '--host', help="Specify the host for the mongo database.")
-@click.option('-c', '--config', type=click.Path(exists=True),
-              help="Specify the path to a config file with database info.")
 @click.option('--demo', is_flag=True, help="If the demo database should be used")
-@click.version_option(__version__)
```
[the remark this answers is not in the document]
dnil: :+1:

scout/commands/base.py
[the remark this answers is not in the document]
dnil: Ok, I *almost* understand.. If you have tested and thought it through I think it is ok.

northwestwitch:
> We should like have a discussion of this before going ahead and releasing a new major release. This will trigger revalidation in the sense of accreditation unless specified why not.

I agree. I was reasoning in terms of new major release because this change of code is likely to break how you now run the cli (by providing a config file). There will be no config file for the CLI when we do this change.
In the meanwhile I could think of a workaround (maybe using the cli config file to initialize the app) so we can postpone the major release for later..?

henrikstranneheim:
Regarding MIP 7.0, I have my doubts that it will be in production before summer. Basically, MIP 7.0 is done and should be validated by next week. However, getting it into cg is another matter entirely.

northwestwitch:
I've modified this PR not to break anything (No major release involved)
It works like this: 

- when you invoke the cli without yaml config file the app is created using the default config file.
- when you invoke the cli with the --demo param the app object will be created and uses the 'scout-demo' db name
- when you provide the config file as a cli command, the app will be created using the params specified in the file.

northwestwitch:
TESTED ON CLINICAL-DB STAGE (webapp and CLI) and works

northwestwitch:
Tested again on clinical-db stage (webapp and CLI) and works

northwestwitch:
Looks like it's working as it should on scout stage and rasta cli stage. After talking with @dnil this morning, that agrees I think I'm going to merge this. It will become a new minor release. 

Should it break anything I take responsibility and I'll fix what there is to fix.

PR #1210, opened 2019-04-08
Fix #1209. Add a flag to hide all non-research cases. 

### LINKED ISSUE #1209: New list of cases "research"
Would it be possible to add another list on top of the "active", "solved" and "archived", for cases that are still active, but more in a research mode. We have several cases that are quite old, that we are still working on, but not actively in a clinical setting. Cases we can't really let go, so want to dig deaper.

dnil:
Fix #1209 by adding a flag to hide all non-research cases. Reuses (seemingly unused) existing query option is_research.


northwestwitch:
Nice @dnil! Could it be possible to fix this PR so you can filter by case status, using the search cases form in cases? The easy way would be writing **status=active** in the textfield.
It's something @sofievonlanthen pointed out and I think it will be useful for others as well. Tried to write just that (status=active) but then I see that the code enters the last else in the if name_query and it's gotten wrong.

dnil:
Yep, can sure fix that: my initial thought was that might be more to my liking than the users, so I skipped it. Most happy to oblige! 😺

northwestwitch:
@dnil I've got some time to work on this today. If you want I can fix it!

dnil:
Thanks; I think I already have a working version of that! As soon as I have a few minutes I can check why its not here - I think I might have started something else in addition to that.. :)

northwestwitch:
Ok but then I wait!

northwestwitch:
Are you ready for a review?

dnil:
Yeah, should be fine - I guess tests, though. Hm, you have a nice one on your search empty query..

northwestwitch:
We could merge  #1211 and write the same type of tests for this one!

northwestwitch:
I'll put it on stage

northwestwitch:
It's on stage and nothing is working.!! Not my PR, which is merged, nor this one. But don't worry, because I've just found out that the script that is restarting the server after installing from the branch is not working. We had it fixed this morning and it's deployed but maybe we have to restart the server for it to take effect. I'll try to figure out!

dnil:
Yeah, but with some text changed, so pretty confusing! Good luck!

northwestwitch:
I know, but I think it's the same thing that happens when you modify the HTML and the backend in Flask : you see the changes in the HTML immediately but you have to restart the server to see the changes in the python code

dnil:
Unless you use livereload, in which case it will occasionally work without restarting, depending on arcane arts. Oh, btw we should probably put the name search explanation in a tooltip or such; its getting big..

northwestwitch:
Stage is working again. I've tested it briefly and looks good. If you are satisfied feel free to merge!

Issue #1198, opened 2019-03-30
List cases with no HPO-terms


henrikstranneheim:
As a scout user I would like to easily get a list of cases lacking important metrics such as HPO-terms or cohort tag. Then I can add the relevant metrics to these cases and improve both the analysis in Scout and the statistics.

PR #1211, opened 2019-04-11
List cases with missing HPO terms, pheno groups and blank synopsis

northwestwitch:
In cases page, write 'synopsis:', 'HP:' or 'PG:' to get all cases with missing info for these fields

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/cases.html
+++ b/scout/server/blueprints/cases/templates/cases/cases.html
@@ -119,7 +119,8 @@
       </div>
     </div>
     <div class="bottom-align-text">
-      <br>Examples: case_id=<strong>18201</strong>, HPO-term=<strong>HP:0001166</strong>, synopsis=<strong>synopsis:epilepsy</strong>, panel=<strong>panel:NMD</strong>, phenotype group=<strong>PG:0100022</strong>, cohort=<strong>cohort:pedhep</strong>.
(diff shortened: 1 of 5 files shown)


northwestwitch:
fix #1198. It's on stage.
In the beginning I was thinking that the code was not working, then I realized that A LOT of those test cases don't have phenotype terms, and that's why I kept getting the same list even if I used the filter!

scout/server/blueprints/cases/templates/cases/cases.html
```diff
@@ -119,7 +119,8 @@ <h5 class="modal-title" id="sangerModalLabel">Sanger validations to evaluate:</h
       </div>
     </div>
     <div class="bottom-align-text">
-      <br>Examples: case_id=<strong>18201</strong>, HPO-term=<strong>HP:0001166</strong>, synopsis=<strong>synopsis:epilepsy</strong>, panel=<strong>panel:NMD</strong>, phenotype group=<strong>PG:0100022</strong>, cohort=<strong>cohort:pedhep</strong>.
+      <br>Examples: case_id=<strong>18201</strong>, HPO-term=<strong>HP:0001166</strong>, synopsis=<strong>synopsis:epilepsy</strong>, panel=<strong>panel:NMD</strong>, phenotype group=<strong>PG:0100022</strong>, cohort=<strong>cohort:pedhep</strong>.<br>
+      Leave blank values after key to search for cases with missing info (<strong>HP:</strong>,<strong>PG:</strong>,<strong>synopsis:</strong>).
```
dnil:
```suggestion
      Use key without value to search for cases with missing info (<strong>HP:</strong>,<strong>PG:</strong>,<strong>synopsis:</strong>).
```

dnil:
What do you think: do we need to test this (and the other name search PR) on stage separately? Perhaps we could merge them and then test deployment?

northwestwitch:
Yes, we can test them togethers I'd say, but maybe write a test for #1210 ?

PR #1213, opened 2019-04-16
Fixes broken load delivery_report command

patrikgrenfeldt:
This PR fixes the broken load delivery_report command (by removing unused extranous parameter)

How to test:
1. install on scout stage
1. us
1. scout load delivery-report [CASE-ID] [REPORT-PATH]

Expected outcome: 
Message about upload successful or already uploaded

### DIFF
--- a/scout/commands/load/report.py
+++ b/scout/commands/load/report.py
@@ -6,12 +6,12 @@ from scout.load.report import load_delivery_report
 LOG = logging.getLogger(__name__)
 
 
-@click.command()
+@click.command('delivery-report')
 @click.argument('case_id')
 @click.argument('report_path', type=click.Path(exists=True))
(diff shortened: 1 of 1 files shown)


scout/commands/load/report.py
```diff
@@ -6,21 +6,20 @@
 LOG = logging.getLogger(__name__)
 
 
-@click.command()
+@click.command('delivery-report')
 @click.argument('case_id')
-@click.argument('analysis_date')
 @click.argument('report_path', type=click.Path(exists=True))
 @click.argument('update', required=False)
```
northwestwitch: Nice! What about this, here:
@click.option('-update', '--update', is_flag=True, help='update delivery report for a sample')

PR #1228, opened 2019-05-20
Safer cli base code

northwestwitch:
Avoid key error whenever scout is invoked without a command (--help option).

To reproduce this error (that is not really affecting the cli, but it's annoying) simply type this:
'scout --help'

This fix makes the code safer.


northwestwitch:
Tested on clinical-db stage

scout/commands/base.py
```diff
@@ -52,20 +52,20 @@ def get_app(ctx):
...
+    if options.params.get("config"):
         with open(options.params["config"], 'r') as in_handle:
             cli_config = yaml.load(in_handle, Loader=yaml.FullLoader)
 
-    if options.params["demo"]:
+    if options.params.get("demo"):
         cli_config['demo'] = 'scout-demo'
 
     app = create_app(config=dict(
         DEBUG=True,
```
dnil: Hm, should we really do that? But, anyways, hasn't changed.. :)

northwestwitch: Should do what?

dnil: DEBUG=True in --demo? Often enough the case, I guess.. :)

northwestwitch: The app is initiated with that log level but after it is created the log of the cli is set to what the user indicates in the option. See the body of the 'cli' function. 
I'm testing that this works in a test named test_base_command.py

northwestwitch:
Merges, thanks @dnil!!

northwestwitch:
I've just found out that there is a -d (--debug) option when you run 'scout serve' command. So no need to have this annoying debug option as a default. I'll remove it in the next PR. Thanks for noticing!

PR #1235, opened 2019-06-03
Include script to migrate server paths in the scripts folder

northwestwitch:
This is the script I've used to migrate the file paths from one server to another (rasta to hasta). It's not really needed by scout (cli or web app) but maybe it's good to have. 
I think there might be some official document that describes how we did the migration that is going to reference this PR. 

But this is not an urgent thing, so no rush with the review. Ah, and the code has been already used for the migration, so I can't really change it a lot after the review.


northwestwitch:
@dnil @moonso could someone approve this? It doesn't require any review, it's just a script that has been used and we have to include it in the scripts folder. Thanks!

northwestwitch:
No problem, thanks!
You mean if there's a possibility of not including the scripts folder? I don't know.

But of course we could include a description of what the scripts do in the documentation (under the admin part). I'll include a description of this script in this PR!

requirements.txt
```diff
@@ -41,3 +41,6 @@ python-dateutil
 
 # Requesting resources
 pybiomart
+
+# Required by scripts
+tabulate
```
dnil: I mean this! 😄

northwestwitch: Aaah OK!

dnil: I'm always afraid of inflating the requirements - dependencies can be a pain.

northwestwitch: I think it's safe not include the script dependencies in the requirements. I don't think anybody except us will ever use this script. And I think if we'll use it again it will be in a long time

PR #1246, opened 2019-06-12
Case names on mt report

### LINKED ISSUE #1236: MT reports should display case display name instead of ID
So it's easier to understand which individual is which for scout users

northwestwitch:
fix #1236.

Mitochondial reports downloaded from case page contain an excel file per case individual. 
Each file was named like this:  **caseDisplayName_individualID_date**.

This fix renames the files to: **caseDisplayName_individualDisplayName_date**

- [x] review by @moonso 
- [x] verified by @northwestwitch


northwestwitch:
Oh no, coverage went down. I'll write a test for the function!

northwestwitch:
@moonso an easy one, when you have time, thanks!

PR #1247, opened 2019-06-12
Tissue type

moonso:


tests/build/test_build_individual.py
```diff
@@ -155,3 +154,113 @@ def test_build_individuals(parsed_case):
         assert ind_obj['individual_id'] == ind_info['individual_id']
         assert ind_obj['display_name'] == ind_info['display_name']
         assert ind_obj['capture_kits'] == ind_info['capture_kits']
+
+# Cancer stuff
+
+def test_build_individual_tmb():
+    ## GIVEN information about a individual
+    ind_info = {
```
northwestwitch: It's a lot of ind_info dictionaries here. Could it be moved into the fixtures and imported from there? 
An even cooler alternative would be to read from the real demo conf file and do the case load as one would do from the cli. But I don't know if this would add a lot of complexity. In that case nevermind!

moonso: They are so different so I do not see any value of using a fixture...

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -157,13 +157,20 @@
...
+            <th class="col-xs-1" title="Sample sex">Sex</th>
+            <th class="col-xs-1" title="Phenotype of sample">Phenotype</th>
+            <th data-toggle='tooltip' data-container='body' class="col-xs-1" 
+             title="Type of sequencing (e.g wes, wgs)">Sequencing</th>
+            <th data-toggle='tooltip' data-container='body' class="col-xs-1" 
+             title="Ancestry prediction from peddy">Ancestry (pred.)</th>
+            <th data-toggle='tooltip' data-container='body' class="col-xs-1" 
+             title="If parenthood is confirmed by peddy">Parenthood</th>
+            <th data-toggle='tooltip' data-container='body' class="col-xs-1" 
+             title="Downloadable cytosure file ">CGH</th>
```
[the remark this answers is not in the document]
moonso: Fixed

PR #1248, opened 2019-06-13
Fix broken update case command

northwestwitch:
@moonso caught you! You can't escape my amazing CLI tests now :D 
Next time write a PR, I know it's a small change, but it broke the function.


moonso:
Haha SORRY!!

PR #1249, opened 2019-06-13
External cases

moonso:


scout/server/blueprints/cases/templates/cases/cases.html
```diff
@@ -222,6 +222,10 @@ <h5 class="modal-title" id="sangerModalLabel">Sanger validations to evaluate:</h
           {% set label_class = 'info' %}
         {% elif analysis_type == 'panel' %}
           {% set label_class = 'info' %}
+        {% elif analysis_type == 'mixed' %}
+          {% set label_class = 'danger' %}
```
northwestwitch:
```suggestion
          {% set label_class = 'mixed' %}
```

northwestwitch: And add this code in styles.css:

/* Create a purple label to be used for spans, buttons etc. */
.label-mixed {
  background-color:#800080;
}

Doesn't have to be purple, but maybe it's nice if it's different from the others

moonso: Nice suggestion, I fixed this and it works fine:)

moonso:
Hi @northwestwitch I updated according to your suggestions

northwestwitch:
I'm thinking about the tests for the UI score. Those tests compare sets of chars like 'a', 'b' etc for 2 samples. I'm not so much into the algorithm but for what I've understood it should measure how similar terms are in a structured ontology, like OMIM or HPO. 
And in your code, when you calculate the score, you do take into account the ancestors of the terms. So why are you using chars here? Wouldn't be better to use real HPO terms? 
What about using/ expanding the test in scout/tests/adapter/mongo/test_case_handling.py, like 400?

moonso:
The algorithm only compares sets of strings so it wouldn't change anything if we use strings that are HPO terms. When we add the ancestors they are just another string so for testing the algorithm only it is not necessary. Then it is a better idea to do as you propose, to expand the functionality tests in `scout/tests/adapter/mongo/test_case_handling.py`, but not in this PR

northwestwitch:
> When we add the ancestors they are just another string

Fair enough!

PR #1251, opened 2019-06-17
Causatives complete info

### LINKED ISSUE #1169: Display rank score and filter by gene on causatives page

northwestwitch:
fix #1169.

Added ACMG classification and filter by gene. Causatives page now looks like this: 

image

Note how first variant has ACMG classification and second one doesn't.

How to test:
1. Assign any ACMG classification to a variant and mark it as causative.
1. Variant should be displayed among the causatives http://scout-url/custXXX/causatives (local implementation of scout: http://localhost:5000/cust000/causatives)
1. Make sure that the ACMG classification is displayed and the tooltip is displayed when the mouse hovers over it: 
image
1. Mark another variant as causative.
1. Go to causatives (same page as above) and check that 2 variants are displayed.
1. Filter by gene using the gene of one of the 2 variants (variants should be in different genes)
1. Just one variant should now be displayed (the one falling in the queried gene)

- [x] Review by @dnil 
- [x] Verified by @dnil

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/causatives.html
+++ b/scout/server/blueprints/cases/templates/cases/causatives.html
@@ -96,7 +96,7 @@
                   {% endfor %}
                 </div>
                 <div class="col-xs-8 col-md-1">
-                  ACMG:
+                  <a href="#" data-toggle="tooltip" title="ACMG classification assigned by Scout users (not Clinvar)" style="text-decoration: none; color: #000;">ACMG:</a>
                   {% if 'acmg_classification' in variant %}
                     <span class="label label-{{variant.acmg_classification.color}}">{{variant.acmg_classification.short}}</span>
                   {% else %}
@@ -128,6 +128,11 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-3-typeahead/4.0.2/bootstrap3-typeahead.min.js"></script>
 
   <script>
+
+    $(document).ready(function(){
+      $('[data-toggle="tooltip"]').tooltip();
+    });
+
     $(function () {
       function getTerms(query, process) {
         $.get("{{ url_for('genes.api_genes') }}", {query: query}, function(data) {
--- a/scout/server/blueprints/cases/templates/cases/causatives.html
+++ b/scout/server/blueprints/cases/templates/cases/causatives.html
@@ -96,12 +96,13 @@
                   {% endfor %}
                 </div>
                 <div class="col-xs-8 col-md-1">
-                  <a href="#" data-toggle="tooltip" title="ACMG classification assigned by Scout users (not Clinvar)" style="text-decoration: none; color: #000;">ACMG:</a>
+                  <a href="#" data-toggle="tooltip" title="ACMG classification assigned by Scout users (not Clinvar)" style="text-decoration: none; color: #000;">ACMG:
                   {% if 'acmg_classification' in variant %}
                     <span class="label label-{{variant.acmg_classification.color}}">{{variant.acmg_classification.short}}</span>
                   {% else %}
                   -
                   {% endif %}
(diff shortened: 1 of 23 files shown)


dnil:
It will be useful - and perhaps necessary to test this on stage as well, with a long list of variants.

scout/server/blueprints/cases/templates/cases/causatives.html
```diff
@@ -99,3 +123,28 @@
     </div>
   {% endfor %}
 {% endblock %}
+
+{% block scripts %}
+  {{ super() }}
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-3-typeahead/4.0.2/bootstrap3-typeahead.min.js"></script>
```
dnil: 😄

northwestwitch:
I agree. Are you able to install stuff on stage, I can't :(

scout/server/blueprints/cases/views.py
```diff
@@ -316,7 +316,13 @@ def matchmaker_delete(institute_id, case_name):
 @templated('cases/causatives.html')
 def causatives(institute_id):
     institute_obj = institute_and_case(store, institute_id)
-    variants = store.check_causatives(institute_obj=institute_obj)
+    query = request.args.get('query', '')
+    hgnc_id = None
+    if '|' in query:
+        # filter accepts an array of IDs. Provide an array with one ID element
+        hgnc_id = [int(query.split(' | ', 1)[0])]
```
dnil: crashes if a misguided user uses a | in the query.

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -89,21 +89,58 @@ def test_case(app, user_obj, case_obj, institute_obj):
             # THEN it should return a page
             assert resp.status_code == 200
 
-def test_causatives(app, user_obj, institute_obj):
+def test_causatives(app, user_obj, institute_obj, case_obj):
     # GIVEN an initialized app
     # GIVEN a valid user and institute
+    # There should be no causative variants for test case:
+    assert 'causatives' not in case_obj
+    var1_id = '4c7d5c70d955875504db72ef8e1abe77' # in POT1 gene
```
dnil: May they never change! 😃

dnil: > I agree. Are you able to install stuff on stage, I can't :(

I haven't tried yet: a little distracted with the conference, and haven't done it before. I'll see if I can get a hold of @moonso and try it out, otherwise I'd prefer sitting down with you Wednesday/Thursday to have someone look over my shoulder..

northwestwitch: I'm afraid also you and @moonso don't have the privileges yet. I'll ask at the bioinfo/operations meeeting today!

northwestwitch: > May they never change! 😃

I accept them as a dogma, they can't change! :)

scout/server/blueprints/cases/views.py
northwestwitch: Fixed, thanks! I've fixed it also where the I've taken the original code: in file scout/server/blueprints/genes/views.py

northwestwitch:
I've spent quite some time on the causative variants sorting. It still doesn't work for all variant because I think their keys are heterogeneous: some of the are missing the sorting key.

I think I'll use the sorting key that works best for now (sorts well most of the variants on scout-stage) and is also defined in the current version of scout (variant.hgnc_symbols).

If you try sorting instead by variant.gene.hgnc_symbol then you'll miss most oldish variants that don't have this key and the sorting is more messy.

Merging this, since I've just tested on stage and it works (except for the sorting thing)

Issue #1215, opened 2019-04-17
Update key names for ClinGen CGH (benign and pathogenic) 


henrikstranneheim:
Svdb has been updated in MIP 7 and to accomodate svdb changes and to clarify what has been annotated the input vcf INFO keys for ClinGen CGH has been renamed:
```
clingen_cgh_benign -> clingen_cgh_benignOCC
clingen_cgh_pathogenic -> clingen_cgh_pathogenicOCC
```

moonso:
Hi @henrikstranneheim , how about `clingy_ngi`?

henrikstranneheim:
I have had some discussion with Daniel N and Jesper and forgot to update this issue. We decided to use:
```
##INFO=<ID=clingen_cgh_benign,Number=1,Type=Float,Description="The frequency of the event in the database">
##INFO=<ID=clingen_cgh_pathogenic,Number=1,Type=Float,Description="The frequency of the event in the database">
##INFO=<ID=clingen_ngiAF,Number=1,Type=Float,Description="The frequency of the event in the database">
##INFO=<ID=clingen_ngiOCC,Number=1,Type=Integer,Description="The number of occurances of the event in the database">
##INFO=<ID=clinical_genomics_mipAF,Number=1,Type=Float,Description="The frequency of the event in the database">
##INFO=<ID=clinical_genomics_mipOCC,Number=1,Type=Integer,Description="The number of occurances of the event in the database">
```

@dnil and @J35P312 If you object please say so.

dnil:
Accept for the general polite grinding of teeth about the need for it, I think we are good. 

On that note I also had a conversation w the aCGH ppl. We will likely be able to get an update procedure in place on our side, but the pathogenic ref file will still have to be treated privately, until there is some sort of access review board for clinical reference data in place. I guess we could give contact info in the download instruction.

// Daniel

moonso:
Why do we use both OCC and AF? We do not treat any other frequency like this. I think we should choose to show one of them whenever there is an alternative

dnil:
There is no real need to have both: the cohort total number of individuals (or chromosomes, depending) would be a more informative number.

As soon as the cohorts are large enough (>200?) it is fine to default to AF. 

For another cautionary side note, at some point the clingen_cgh_benign was used - at least in scoring - as a binary item, much like the old Decipher population track. I take it the aCGH files have the normal AF and OCC instead?

henrikstranneheim:
[...]

Issue #1186, opened 2019-03-15
gnomAD-SV 


genecracker:
See the pre-print on gnomAD-structural variants, based on WGS data from 14,891 individuals

https://www.biorxiv.org/content/10.1101/578674v1

vcf with data is downloadable

dnil:
Thank you Bianca! See issue on MIP, but we missed the VCF!

dnil:
Add keys to frequencies as annotated with new version of SVDB.

moonso:
Ok could anyone write here exactly what the keys will look like and what values they will have (eg float, one value or multiple etc)? @dnil @J35P312 @henrikstranneheim

henrikstranneheim:
```
##INFO=<ID=gnomad_svAF,Number=1,Type=Float,Description="The frequency of the event in the database">

In INFO field: gnomad_svAF=0.010375
```

Issue #1068, opened 2018-11-12
Display local SV freq dump in SV view


henrikstranneheim:
Live from MIP 7.0 Clinical-Genomics/MIP/issues/523
```
##INFO=<ID=clinical_genomics_mipAF,Number=1,Type=Float,Description="The frequency of the event in the database">
```

dnil:
Solved in #1253.

Issue #868, opened 2018-06-20
Notation on gene level regarding existance of pseudogenes?


KickiLagerstedt:
Regarding some genes / some regions of genes annotation could be false positive / difficult due to the presence of duplicated region. Would it be possible to annotate this on the gene page or something like that?

For inspiration see list here
https://blueprintgenetics.com/pseudogene/

// Kicki

dnil:
One solution would be to view a mappability score, e.g. one of 
http://rohsdb.cmb.usc.edu/GBshape/cgi-bin/hgFileUi?db=hg19&g=wgEncodeMapability

Another perhaps just to instruct chanjo / sambamba not to count reads with MAPQ == 0 towards coverage, I e value non-uniquely mapping reads as little as not existing ones.

dnil:
..or we could provide a link to the good old seg_dup track from the Eichler lab - which appears to be exactly what BluePrint Genetics has intersected with.

moonso:
@KickiLagerstedt @dnil Could you elaborate on this? Where would the information be to help analysis? If chanjo we should open the issue there instead...

moonso:
Ok so please close this or make it more clear what should be done

dnil:
How about we settle for flagging the Eichler segmental duplication regions for now: that together with the Chanjo coverage should give a decent picture? The UCSC genomicSuperDups that is. @henrikstranneheim is that annotated somewhere already, or should we try to query automatically? BluePrint settled on a static page with a list of regions, but personally I would prefer a flag on the variant page..

henrikstranneheim:
Available in MIP 7.0 as:
VCF key in CSQ field: "genomic_superdups_frac_match". For a variant found in overlapping multiple segdups the annotation uses "&" to split each annotation record within the field.

Should be displayed at the variant level even though the information is present in each transcript in the CSQ field.

dnil:
We will consider this solved with #1253 for now. It is not unlikely that someone will wish for aggregate lists from this feature, perhaps in combination with average (expected) mapped coverage for genes on a panel, but that may be better solved by e g Chanjo development.

PR #1253, opened 2019-06-19
Mip7 annotations

moonso:
- [x] #1215 (SV cgh frequencies)
- [x] #1186 (gnomad-SV)
- [x] #1068 (local SV frequencies)
- [x] #868 (GenomicSuperDups)


northwestwitch:
@henrikstranneheim could you write an example of a vcf line with GenomicSuperDups annotation? I know it's a string of items separated by '&',  but what type of items?

northwestwitch:
@moonso I finished this PR. Could you check that it's fine? I've added tests for the new keys you've introduces before summer and I've added the GenomicSuperDups. These latter should be checked with care because I (kind of) have no idea what I'm doing with the transcript parsing here :)

scout/parse/variant/frequency.py
```diff
@@ -20,7 +20,8 @@ def parse_frequencies(variant, transcripts):
     exac_keys = ['EXACAF']
     exac_max_keys = ['ExAC_MAX_AF', 'EXAC_MAX_AF']
 
-    gnomad_keys = ['GNOMADAF', 'GNOMAD_AF']
+    # Gnomad have both snv and sv frequencies
+    gnomad_keys = ['GNOMADAF', 'GNOMAD_AF', 'gnomad_svAF']
```
moonso: Could you check that this is being displayed in the SV variants view?

moonso:
And since this is originally my PR I can't approve... @northwestwitch or @dnil you have to do a review

dnil:
I could have a look tomorrow, but feel free if you get there before!

scout/parse/variant/frequency.py
northwestwitch: I haven't tried with such a VCF. We could maybe pull the branch on stage and see if uploads a test case with these keys.

northwestwitch: Is anyone using hasta stage?

northwestwitch:
I've managed to re-upload SNV and SV variants for stage case 
https://scout-stage.scilifelab.se/cust002/15026-validation using this branch.  I don't know what is wrong with stage but it takes forever to open a variant. I'll try to check later

northwestwitch:
SV are much faster to load. clingen mip frequency shows:
image

moonso:
@northwestwitch I wonder if it get's slow since you make one request to entrez for each transcript with a refseq id every time a variant page is loaded. In general I think we should avoid making requests in the pages that one expects to load fast, we will be dependent on how fast there is a reply or if the service is down etc. Perhaps we can find another solution to how the transcript versions are added, like update ALL transcripts at some point. Or you only fetch this information when producing the ClinVar submission at this point.

northwestwitch:
> I wonder if it get's slow since you make one request to entrez for each transcript with a refseq id every time a variant page is loaded.

I've been thinking about it. I'll try to modify this PR commenting the entrez request so we can see if that was the issue!

northwestwitch:
Yes, that was the problem. I've fixed it in this PR in a way that refseq versions are collected only when calling the clinvar page. And now the variants are loaded fast on stage! But the PR went down in coverage unfortunately. Adding a test now!

northwestwitch:
Example of a SV variant with updated annotations: 
https://scout-stage.scilifelab.se/cust002/15026-validation/sv/variants/e9a550af102743993ebe851d9189b89d
image

Checked in the database and unfortunately none of the 355 SNV variants loaded for this case has the superdups_fracmatches. (queried this way: {case_id: 'cuddlyoryx','genes.transcripts': {$exists: true}} )

@henrikstranneheim shall I try loading other cases among these: 


And hope to get one with the superdups?

northwestwitch:
I've tried to load to scout all those test cases from the screenshot above but checking in the database I couldn't find any with the superdups_fracmatches in the transcript info. I don't know if it's because this PR doesn't work and the key is not captured or because the cases don't actually have variants in the tandem repeats. What should I do, should we trust the unittest (tests/parse/test_parse_transcripts.py) and that's it?

moonso:
Have you tested to grep on the vcfs to see if they have superdups? We might need some help from @henrikstranneheim here

northwestwitch:
> Have you tested to grep on the vcfs to see if they have superdups? We might need some help from @henrikstranneheim here

They have the info field at least but variants in the dups, I don't know

henrikstranneheim:
It looks like this from case justhuskey:
```
1       10492   rs55998931      C       T ... INFO => T|downstream_gene_variant|MODIFIER|WASH7P|ENSG00000227232|Transcript|ENST00000488147|unprocessed_pseudogene|||||||||||3912|-1||HGNC|38034||||||||||Ensembl|C|C||||||||||||||||-11.244|2.005|-8.082|-9.238|0.435|-3.175|-2.739|-2.739||||0.982898&0.981582&0.992904&0.987856&0.986907&0.994232
```
GenomSuperDups here is: **0.982898&0.981582&0.992904&0.987856&0.986907&0.994232**
They are there but usually in upstream, downstream or in regulatory region or pseudogenes

northwestwitch:
Still working on this. I'm having problem loading the gene region of WASH7P into scout:

> scout load region --case-id justhusky --hgnc-id 38034

image

henrikstranneheim:
Looks like the index file is missing

northwestwitch:
Index is there so I don't know why I get the error.

I've tried another approach, loading research variants. If I grep WASH7P in the research variants file (/home/proj/long-term-stage/MIP7.1-dev/housekeeper-bundles/justhusky/2019-08-23/justhusky_gatkcomb_rhocall_vt_frqf_cadd_vep_parsed_snpeff_ranked.vcf.gz) I see that the transcripts do have the superdups style -> float&float&etc

But  the if I open scout these variants are not there, I think because they're below the rank threshold, that I see in the code is 8 :(

northwestwitch:
Ah wait, the index is there just for snvs and svs, not the strs..

northwestwitch:
Yess, I managed to upload research variants in that gene using this command: 
```
scout load variants --hgnc-symbol WASH7P justhusky --snv-research --rank-treshold -20 --force
```
And the variant mentioned by @henrikstranneheim is in the database but It doesn't show superdups:
https://scout-stage.scilifelab.se/cust000/SWEDAC-2019-HG-05-Trio-NovaSeq/fa342e72621f6dd522d71c93c469d26f

So there's an error somewhere in the code. I'll fix it!

northwestwitch:
Mappability is collected from annotations and showed on variants page now: 
image

As soon as travis finished I think I'll approve and merge

dnil:
Nice! Could we figure out how many digits are significant and round accordingly?

northwestwitch:
> Nice! Could we figure out how many digits are significant and round accordingly?

The score is the fraction of matching bases, so what about 3 numbers after comma ?

dnil:
Sounds about right; they are all >1000 bp at least. In actual practice, having the seg_dup feature show at all on a variant will constitute a warning flag.

northwestwitch:
There's another issue I see. Entrez e-utilities are down so the refseq version test I wrote for the previous PR is failing

dnil:
😞 Ok. So, that would again speak strongly for seeking a solution where we dl all of the transcript versions at once, wouldn't it?

northwestwitch:
> Ok. So, that would again speak strongly for seeking a solution where we dl all of the transcript versions at once, wouldn't it?

That would be the best solution. Yes

northwestwitch:
All good. Entrez is working and the test passes it looks. This is the final looks of the superdups:
image

@dnil do I merge?

northwestwitch:
NML has just confirmed that the eutilities were down yesterday, so we don't require an API key as I was starting to suspect:

```
Dear Colleague,

The E-Utilities were experiencing server outages for a period this morning.  The issue should be resolved.  Please let us know if you experience additional errors.

Apologies for the inconvenience.

Best regards,

User Services
NCBI | NLM | NIH
```

I'm going to merge now!

dnil:
Good! The transcript number part is stuff for another issue/PR anyway, eventually, so don't worry about that here. Seg_dup is now fine. 

What is ClinGen MIP? One of the research cohorts, captured with SVDB, or  loqusDB? At any rate, we need to document the f sources, and if possible add a cohort size to each. Also time for a fred hover, but both of these are for another PR as well.. 😅

henrikstranneheim:
What is ClinGen MIP? 
It is the one of svdb merge of roughly 2000 WGS SV (grch37).

PR #1257, opened 2019-06-20
Pr template

northwestwitch:
Pull request template. I don't know if it will work out of the box or we should change some settings for the repo.


northwestwitch:
Thanks @dnil!

PR #1263, opened 2019-06-26
Refactor: case.parse_individual(sample) <parse>

mikaell:
Remove if-clauses when building ind_info dict. Make sure no None
value is added by callin removeNoneValues(ind_info) in Return.

This PR is refactoring. And preparation for later bigger commits.

**How to test**:
1. how to install it
`git checkout feature_igv-tracks_190626`

`pytest tests/parse/test_parse_case.py`

**Expected outcome**:
Tests should pass.

**Review:**
- [ ] tests executed by

### DIFF
--- a/tests/parse/test_parse_case.py
+++ b/tests/parse/test_parse_case.py
@@ -267,15 +267,17 @@ def test_wrong_relations():
 
 
 def test_removeNoneValues():
+    # WHEN a dict *not* containing a value which is None
     d = {"a":"1", "b":2, "c":3}
 
+    # THEN calling removeNoneValues(dict) will not change dict
     assert d == removeNoneValues(d)
 
 
 def test_removeNoneValues():
+    # WHEN a dict containing a value which is None
     d = {"a":"1", "b":2, "c":None}
-
-    d_ = {"a":"1", "b":2}
     
-    assert d_ == removeNoneValues(d)
-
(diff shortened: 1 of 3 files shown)


northwestwitch:
I think Måns knows better this part of the code, so if you don't mind and @moonso has time he could review this PR. Otherwise I'll do it, just let me know!

tests/parse/test_parse_case.py
```diff
@@ -261,3 +262,20 @@ def test_wrong_relations():
     # THEN a PedigreeError should be raised
     with pytest.raises(PedigreeError):
         parse_individuals(samples)
+
+
+
+
+def test_removeNoneValues():
+    d = {"a":"1", "b":2, "c":3}
```
moonso: Could you follow the GIVEN, WHEN, THEN like in most other tests? I know it's a bit cumbersome for small tests like these but it is nice if the code follows the same patterns. Also it makes the tests really easy to understand

northwestwitch:
We can test it with real case data tomorrow!

tests/parse/test_parse_case.py
mikaell: Done! In commit "fb3f0695".

northwestwitch:
Don't forget to update the changelog (also including the change of the previous PR) Then I think we can merge! Nice! 👍

dnil:
:+1:

Issue #576, opened 2017-08-24
Local observation case display should be based on user access to cases


dnil:
When viewing a shared case, the local observations only reflect the owner institute, not user. Would be very nice to be able to see case ids from the users own institute and ideally other cases shared with the users institute as well.

moonso:
Yes we'll see what we can do, right @robinandeer ?

moonso:
I think we should be able to fix this one @dnil

PR #1275, opened 2019-08-05
Display observations 

### LINKED ISSUE #1274: Local observations doesn't seem to show links to variants belonging to the same institute
Example: 
https://scout.scilifelab.se/cust004/518701400112fam/b0bfd12e6ce4de2caa660de490c1fbdf

3 observations in the old archive and 15 observations in the new.

3 of the 15 observations in the new archive belong to the same institute, cust004 (417701600625fam, 319701400073fam and 518701400112fam which is the very institute where you are in the first place), but the link to these other cases/variants is not shown.

Look at **line 664 of scout/server/blueprints/variants/templates/variants/variant.html**. I bet if there's a problem is some value missing there

northwestwitch:
This PR should fix #1274 

**How to test**:
1. Make sure scout stage has installed the current master branch.
1. Go to this variant : https://scout-stage.scilifelab.se/cust002/F0027817/aba7ac625e06534d71ed16e3766ffc59 
and make sure that the local observations table looks like this:
image

1. install this branch on clinical-db stage using this command:  
```
bash update-scout-stage.sh display_observations
```

Go to the link above and make sure that now the page displays a link to another case's variant (the other observation) and the link works and points to the other variant:

image

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -636,11 +636,13 @@ def observations(store, loqusdb, case_obj, variant_obj):
     obs_data['cases'] = []
     institute_id = variant_obj['institute']
     for case_id in obs_data.get('families', []):
-        LOG.debug('IN OBSERVATIONS: --> case id: {}'.format(case_id))
-        if case_id != variant_obj['case_id'] and case_id.startswith(institute_id):
-            other_variant = store.variant(variant_obj['variant_id'], case_id=case_id)
+        if case_id != variant_obj['case_id']:
+            # other case might belong to same institute, collaborators or other institutes
             other_case = store.case(case_id)
-            obs_data['cases'].append(dict(case=other_case, variant=other_variant))
+            # display only variants which belong to the same institute or collaborators
+            if other_case and (other_case.get('owner') == institute_id or institute_id in other_case.get('collaborators', [])):
+                other_variant = store.variant(variant_obj['variant_id'], case_id=case_id)
+                obs_data['cases'].append(dict(case=other_case, variant=other_variant))
 
     return obs_data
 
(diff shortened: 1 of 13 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -636,10 +636,13 @@ def observations(store, loqusdb, case_obj, variant_obj):
...
     institute_id = variant_obj['institute']
     for case_id in obs_data.get('families', []):
-        if case_id != variant_obj['case_id'] and case_id.startswith(institute_id):
-            other_variant = store.variant(variant_obj['variant_id'], case_id=case_id)
+        if case_id != variant_obj['case_id']:
+            # other case might belong to same institute, collaborators or other institutes
             other_case = store.case(case_id)
-            obs_data['cases'].append(dict(case=other_case, variant=other_variant))
+            # display only variants which belong to the same institute or collaborators
+            if other_case and (other_case.get('owner') == institute_id or institute_id in other_case.get('collaborators', [])):
```
dnil: Yep, that looks like the old functionality. Now, it would be really nice to also pay attention to what institutes the current user has access to. It is not unreasonable that you can use also "your own" patient data to determine if the variant is of interest. The actual user institute should be available from store.user() something, shouldn't it?

northwestwitch: Good point, and of course we should include the option in this PR. I don't want to touch it again later!

dnil:
I think you can now say you also solve Issue #576 as well: it was so familiar somehow.. 😄

northwestwitch:
> I think you can now say you also solve Issue #576 as well: it was so familiar somehow.. 😄

Nice! Still not working but I found the error. Almost done with this PR

Issue #1276, opened 2019-08-06
Grch38 contig name for export gene panel


henrikstranneheim:
Currently Scout produce this gene panel export for genome build 38:
```
##genome_build=38
##gene_panel=MIT,version=12.0,updated_at=2018-02-05,display_name=Mitochondrial nuclear
##gene_panel=mtDNA,version=8.0,updated_at=2015-11-30,display_name=Mitochondrial DNA
##gene_panel=IEM,version=12.0,updated_at=2018-02-05,display_name=Inborn Errors of Metabolism
##contig=1
##contig=2
##contig=3
##contig=4
##contig=5
##contig=6
##contig=7
##contig=8
##contig=9
##contig=10
##contig=11
##contig=12
##contig=13
##contig=14
##contig=15
##contig=16
##contig=17
##contig=18
##contig=19
##contig=20
##contig=21
##contig=22
##contig=X
##contig=MT
```
However, grch38 should have a chromosome prefix = "chr" and MT is now called chrM, see: ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.15_GRCh38/seqs_for_alignment_pipelines.ucsc_ids/GCA_000001405.15_GRCh38_full_analysis_set.fna.fai

We need to add the chromosome prefix and rename MT to M when exporting using grch38.

northwestwitch:
Closing this.

PR #1277, opened 2019-08-07
Export gene and gene panels to chr build GRCh38

northwestwitch:
This PR adds the functionality to export genes and panels to build GRCh38. It's addressing #1276

**How to test**:

** 1) Testing the gene export**
- from a local installation of scout, master branch run the following command:
```
scout --demo export genes -b 38 
```
You should get a list of gene coordinates where the first column is a chromosome such as 1,2..X,Y,**MT**
- switch to this branch (git checkout export_grch38)
- run now the following command
```
scout --demo export genes -b GRCh38
```
Expected outcome is list of gene coordinates where the first column is a chromosome formatted like this:
chr1, chr2... chrX, chrY, **chrM**

The formatting above should be valid also using the --json option 

** 1) Testing the gene panels export**
- from master branch run the following command:
```
scout --demo export panel panel1 -b 38 --bed
```
- You should see a list of gene coordinates where the first column is a chromosome such as 1,2..X,Y,**MT**. If you look up at the beginning of what is printed (contig list) the following contigs should be printed:

##genome_build=38
##gene_panel=panel1,version=1.0,updated_at=2019-08-06,display_name=Test panel
##contig=1
##contig=2
##contig=3
##contig=4
##contig=5
##contig=6
##contig=7
##contig=8
##contig=9
##contig=10
##contig=11
##contig=12
##contig=13
##contig=14
##contig=15
##contig=16
##contig=17
##contig=18
##contig=19
##contig=20
##contig=21
##contig=22
##contig=X
##contig=Y
##contig=MT

- checkout to this branch again and run the following command:
```
scout --demo export panel panel1 -b GRCh38 --bed
```
- expected outcome is that contigs in the result are formatted like this: chr1, chr2... chrX, chrY, **chrM** and the contig list above is the following:
##genome_build=GRCh38
##gene_panel=panel1,version=1.0,updated_at=2019-08-06,display_name=Test panel
##contig=chr1
##contig=chr2
##contig=chr3
##contig=chr4
##contig=chr5
##contig=chr6
##contig=chr7
##contig=chr8
##contig=chr9
##contig=chr10
##contig=chr11
##contig=chr12
##contig=chr13
##contig=chr14
##contig=chr15
##contig=chr16
##contig=chr17
##contig=chr18
##contig=chr19
##contig=chr20
##contig=chr21
##contig=chr22
##contig=chrX
##contig=chrY
##contig=chrM

**Review:**
- [ ] code approved by
- [x] tests executed by

### DIFF
--- a/scout/commands/export/gene.py
+++ b/scout/commands/export/gene.py
@@ -1,9 +1,7 @@
 import logging
 
 import click
-
 from bson.json_util import dumps
-
 from flask.cli import with_appcontext
 
 from scout.commands.utils import builds_option
@@ -22,15 +20,22 @@ def genes(build, json):
     """Export all genes from a build"""
     LOG.info("Running scout export genes")
     adapter = store
-
     result = adapter.all_genes(build=build)
+    gene_list = list(result)
+    if build == 'GRCh38':
+        for gene_obj in gene_list:
+            if gene_obj['chromosome'] == 'MT':
+                gene_obj['chromosome'] = 'M'
+            gene_obj['chromosome'] = ''.join(['chr',gene_obj['chromosome']])
+
     if json:
-        click.echo(dumps(result))
+        click.echo(dumps(gene_list))
         return
 
+
     gene_string = ("{0}\t{1}\t{2}\t{3}\t{4}")
-    click.echo("#Chromosom\tStart\tEnd\tHgnc_id\tHgnc_symbol")
-    for gene_obj in result:
+    click.echo("#Chromosome\tStart\tEnd\tHgnc_id\tHgnc_symbol")
+    for gene_obj in gene_list:
         click.echo(gene_string.format(
             gene_obj['chromosome'],
             gene_obj['start'],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ## [4.x.x]
 
 ### Added
+- Export genes and gene panels in build GRCh38
 
 ### Fixed
 
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -188,6 +188,9 @@ class GeneHandler(object):
             Returns:
                 result()
         """
+        if build == 'GRCh38':
+            build = '38'
+
         LOG.info("Fetching all genes")
         return self.hgnc_collection.find({'build': build}).sort('chromosome', 1)
 
--- a/scout/commands/export/utils.py
+++ b/scout/commands/export/utils.py
@@ -7,9 +7,8 @@ json_option = click.option('--json',
     help='Output result in json format'
 )
 
-build_option = click.option('-b', '--build', 
-    default='37', 
+build_option = click.option('-b', '--build',
+    default='37',
     type=click.Choice(BUILDS),
     show_default=True
 )
-
--- a/scout/constants/__init__.py
+++ b/scout/constants/__init__.py
@@ -40,7 +40,7 @@ COLLECTIONS = [
     'acmg',
 ]
 
-BUILDS = ['37', '38']
+BUILDS = ['37', '38', 'GRCh38']
 
 CYTOBANDS = parse_cytoband(cytobands_handle)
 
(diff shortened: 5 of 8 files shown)


scout/commands/export/gene.py
```diff
@@ -22,15 +20,22 @@ def genes(build, json):
     """Export all genes from a build"""
     LOG.info("Running scout export genes")
     adapter = store
-
     result = adapter.all_genes(build=build)
+    gene_list = list(result)
+    if build == 'GRCh38':
```
dnil: Why only for 'GRCh38' and not for '38'? That makes it a fairly hidden function. Do we need the old, broken '38' format for anything?

scout/commands/export/gene.py
```diff
@@ -22,15 +20,22 @@ def genes(build, json):
...
     if json:
-        click.echo(dumps(result))
+        click.echo(dumps(gene_list))
         return
 
+
     gene_string = ("{0}\t{1}\t{2}\t{3}\t{4}")
-    click.echo("#Chromosom\tStart\tEnd\tHgnc_id\tHgnc_symbol")
-    for gene_obj in result:
+    click.echo("#Chromosome\tStart\tEnd\tHgnc_id\tHgnc_symbol")
```
dnil: :+1:

scout/commands/export/gene.py
```diff
@@ -22,15 +20,22 @@ def genes(build, json):
     """Export all genes from a build"""
     LOG.info("Running scout export genes")
     adapter = store
-
     result = adapter.all_genes(build=build)
+    gene_list = list(result)
```
dnil: :+1:

scout/commands/export/gene.py
northwestwitch: I don't know. The default was that '37' and '38' behave in the same way and I didn't want to change it because I'm not sure if '38' is used for anything or in some other software..

dnil: @henrikstranneheim: are you aware of anything downstream requiring the *broken* version of hg38 output? I'm not, and would prefer to replace it rather than risk confusion. Its quite ok to have both 38 and GRCh38 (and why not hg38) as options, but I think they should do the same thing.

dnil: hg19 vs GRCh19 was another thing, but I hope we are beyond that at the 38 stage.. 😸

henrikstranneheim: Nothing I am aware of. To me it is fine to replace.
Any thoughts on upper case vs lower case in GRCh (or grch)?
Perfectly fine with hg38 as an alias. In MIP we are going to use grch38 internally (trying to avoid mixed case)

henrikstranneheim: grch is an acronym of the official name I think and hg is just short for human genome

northwestwitch: the acronym is going to be used in the CLI, while in the database the genome build field is named '38'. 
I could remove completely GRCh38 and leave only '38', as in was before. But it will export in the format 'chr1', 'chr1', 'chrX', 'chrM'. What do you think?

dnil: @northwestwitch I say we like that, in principle! Just to be sure, what would happen if we re-import/upload the exported panel file (or import it into another instance of scout)?

@henrikstranneheim Yep, GRCh38 is Genome Reference Consortium human38. Generally, I try to preserve the case for established acronyms - some people go bonkers if you "misspell" their pet mixed-case software - but as long as the options are clearly stated in the docs everything is fine.

northwestwitch: > @northwestwitch I say we like that, in principle! Just to be sure, what would happen if we re-import/upload the exported panel file (or import it into another instance of scout)?
> 
In theory, if you export it using '38' you should be able to re upload it as it is. If you export a panel as GRCh38 then you have to reconvert it back to '38' ---> 'chrX' -> 'X' , 'chrM' -> 'M' otherwise you will get an error when you upload it (I guess)
I think we should leave the 38 thingy for now along with whatever we call the 'GRCh38' and then after we have fixed all the rest we can remove the 38 option.

henrikstranneheim: Sounds good!

northwestwitch: Ok, so what should I name this option, is 'GRCh38' fine?

dnil: Yes, fine for now, then it can disappear when import behaviour has been fixed!

northwestwitch:
> I guess that means we take the PR as is, and file a new related issue to generalise panel upload before touching this again!

Thanks for the reviews, this and the other!

PR #1280, opened 2019-08-08
Remove MT variants with AD alt=0 from MT report

### LINKED ISSUE #1278: mtDNA report
Hi Chiara!

We are finally going to use the mtDNA report as originally planned, but I now noticed that there is still a problem with the report. Can you please help me fix this.

The report shows all variants detected in that family, i.e. the variants from the father are also in the report of the child and vice versa. When the variants are from the father they show up as AD Alternative=0 in the report of the child. Could you please make the report only showing the variants that are in the individual it is representing.

A good example is case 19169 where the report for the child 19169-I-2A ends up with 8 variants from the father mixed up with the true variants from the patient.

Thanks!
Helene

northwestwitch:
fix #1278

**How to test**:
1. Use the scout-demo database and modify the following het variant: {_id: '123c73baa49053a95b8f6058164a32e6'} so that it becomes homozygous (0/0) and with 0 reads in the in one of the 3 individuals: 
image

1. On scout demo it will look like this:
image

1. Using the master branch (not this one) make sure that when you export the MT record and open the specific excel sheet for that individual there will be 3 lines. One of these lines should have the variant above with AD alternative = 0

1. Checkout to this branch and re-export the the MT report.

**Expected outcome**:
The line with AD alternative=0 should disappear:
image

**Review:**
- [x] code approved by MM
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Export genes and gene panels in build GRCh38
 
 ### Fixed
-
+- Fixed bug affecting variant observations in other cases
 
 ## [4.6.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fixed bug affecting variant observations in other cases
-- MT report only shows variants occurring in the specific individual of the excel sheet 
+- Fixed a bug that showed wrong gene coverage in general panel PDF export
+- MT report only shows variants occurring in the specific individual of the excel sheet
 
 
 ## [4.6.1]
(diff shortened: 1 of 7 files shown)


northwestwitch:
@moonso it's a quick review, so we have it in place in next release. Thanks!

moonso:
Why should this variant disappear? It needs more documenting I think otherwise this looks fishy... Does it mean that if the mother doesn't have the variant it should be removed? If the father has 0 alternative alleles it follows the interesting pattern it should not be removed, right?

northwestwitch:
> Why should this variant disappear? It needs more documenting I think otherwise this looks fishy... Does it mean that if the mother doesn't have the variant it should be removed? If the father has 0 alternative alleles it follows the interesting pattern it should not be removed, right?

It doesn't actually disappear. It disappears only in the report of the individual that is wild type. The variant is present in the other individuals of the case if they have at least a read with the alternate allele. Look at customer request: #1278.

northwestwitch:
Thanks! :)

Issue #1065, opened 2018-11-09
Extend cases filter


dnil:
The search function on the cases page should also include 
- [x] phenotype group
- [ ] gene of variants in causative or pinned collections
- [x] default gene panel (if we can figure out a smooth way to do that..)

moonso:
@dnil do you have time for this?

moonso:
Ping @dnil

Issue #1064, opened 2018-11-08
Generalise causatives view


dnil:
Causatives view can show case-variants. Add slice filter and allow searching the full variants collection, with restrictions on institute. Tentatively search with gene, rank, CADD, HPO term, phenotype group and functional_annotation. Indexing and optimisation and will be needed.

dnil:
Closed by #1281.

PR #1281, opened 2019-08-13
Wui similar case

dnil:
I'm going to PR this now so I don't get tempted to add more features.
This adds three features as it is now:
- Solves the last remaining item in #1065 by allowing case queries for cases with pinned or causative variants in a particular genes. 
- Allows local case searches for phenotypically similar cases using the HPO-tree queries previously available from scout CLI.
- Solves the remaining suggested items in #1064 by allowing limitation of case_variants searches to HPO-terms, phenogroups and cohorts - and for good measure similar cases as well, which didn't exist when the issue was written.

Not ready for merge, obviously - automated tests to follow.

**How to test**:
1. Pin variants on case and ensure you can search with pinned gene and find these cases in the case view.
2. Mark variants on case causative and ensure you can search with causative:gene and find them on the case view. 
3. Construct phenotypically similar cases and search in case view, ensuring they can be found.
4. Ensure you can find variants from said newly marked phenotypically similar cases in the gene_variants view, and only these variants, when searching for variants in a gene, e.g. POT1. 

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 - Export genes and gene panels in build GRCh38
+- Search for cases with variants pinned or marked causative in a given gene.
+- Search for cases phenotypically similiar to a case also from WUI.
+- Case variant searches can be limited to similar cases, matching HPO-terms,
+  phenogroups and cohorts.
 
 ### Fixed
 - Fixed bug affecting variant observations in other cases
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -19,15 +19,15 @@ LOG = logging.getLogger(__name__)
 
 class CaseHandler(object):
     """Part of the pymongo adapter that handles cases and institutes"""
-    
+
     def get_similar_cases(self, case_obj):
         """Take a case obj and return a iterable with the most phenotypically similar cases
-        
+
         Args:
             case_obj(models.Case)
-        
+
         Returns:
-            scores(list(tuple)): Returns a list of tuples like (case_id, score) with the most 
+            scores(list(tuple)): Returns a list of tuples like (case_id, score) with the most
                                  similar case first
         """
         scores = {}
@@ -91,6 +91,7 @@ class CaseHandler(object):
         """
         LOG.debug("Fetch all cases")
         query = query or {}
+        order = None
 
         # Prioritize when both owner and collaborator params are present
         if collaborator and owner:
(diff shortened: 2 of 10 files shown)


northwestwitch:
Nice!! First thing tomorrow!

northwestwitch:
Starting to test the options.

1. Pinned variant in gene TBX1 and searched for cases with variants pinned in this gene:
image
works!

1. Marked causative POT1 and looked for cases with causatives in POT1:
image
works!

While testing these 2 cases it occurred to me that it's not really obvious for the normal user which options there are for searching stuff in the cases filter form. There is this popup with examples 
image 
(We must add the examples to search for causatives or pinned at least in the popup btw.)

but since the number of keywords and options is growing maybe it's worth to improve the search by introducing some pre-determined field that the user can select. What about a select input field with the following keywords: case_id, HPO term, synopsis, etc and the then the input field that the user fills in? That would make the search easier.

3. You enclosed the code for searching for similar cases inside the gene_variants page and I think it's supercool. Same for cohort and pheno groups and I think this should totally be there. 
Let's say I've these 2 similar cases on scout stage: https://scout-stage.scilifelab.se/cust000/F0007437-validation and https://scout-stage.scilifelab.se/cust000/F0007437-mip7.

Tried to search for cases similar to the first using the second but it doesn't work (or I don't know how to use the search), could you elaborate? 
image

dnil:
Agreed on the growing number of search options! Personally, I enjoy these little tags, but it must be a pain for a new user. I can try to add a select, like next to the search text  input and see how it looks. 

For 3, you also need to pick a gene or five to display results for. We don’t do list all variants - (yet). Actually picking a gene panel there wouldn’t be half bad.

Also, when the news of cases with overlapping phenotypes grow, we may need a threshold value for what is a similar case in the first place. Any suggestions of such a cutoff? User selectable?

northwestwitch:
> For 3, you also need to pick a gene or five to display results for.

AAAHHH I see now! It works if I choose the gene as well. But then I think we should expand and (somewhere and not in this PR) we should allow users to search for similar cases, regardless of the genes. Just like the CLI works. Because what if you have the phenotype and not suspected causatives?

About the threshold thing we could order the similar cases by inverted similarity score maybe. That would be the best way I think.

dnil:
You can do that on the cases page, with another tag keyword... 😊
> About the threshold thing we could order the similar cases by inverted similarity score maybe. That would be the best way I think.
> 
Mm, probably displaying the score as well somewhere - otherwise one can’t differentiate between a bunch of very similar cases or just gadzillions of almost-not-similar-at-all ones.

// Daniel

northwestwitch:
> Mm, probably displaying the score as well somewhere - otherwise one can’t differentiate between a bunch of very similar cases or just gadzillions of almost-not-similar-at-all ones.

Right?

dnil:
Ok, lets go with this. Everything is not tested yet, but at least test coverage is up.
Other tests are complicated by hgnc_symbols population seeming sketchy: perhaps I should adapt the search to go via hgnc_id instead.

PR #1286, opened 2019-08-20
Except KeyError in Load Command

mikaell:
Hi,
This PR is to:
* Get feedback
* Push some work so I don't loose it 
* ...or Delete the branch if there is no need 

The PR is a small part of an idea I had, to make some configuration parameters mandatory and raise an exception if they were not present. We had a small discussion on the matter in the begging of summer, but then I forgot. For example index files would be mandatory to BAM files. 

This PR adds a functionality.

**How to test**:
1. how to install it

`git co fix-load_cmd-190820`

a . Edit `config.yaml` and comment out `sample_id`
b. Try to load the configuration (I run in demo mode)
`scout --demo load case <path.to.config.yaml> -u`
c. Restore `config.yaml`, load the configuration -it should now load

**Expected outcome**:
KeyError is caught and is printed as Error. Stack trace is printed as debug. It might look like this:

`2019-08-20 12:55:07 mikaellmbp scout.commands.load.case[60571] ERROR KEYERROR 'sample_id' missing when loading 'scout/demo/643594.config.yaml'`

**Review:**
- [ ] code approved by
- [ ] tests executed by


mikaell:
Added some tests! Coverage is up.

One is a bit of a hack. In order to force a KeyError I call a sed command from inside the test. Let me know if you can think of a better way.

northwestwitch:
Tested by remove one line (sample ID:) from one of the 3 samples in demo config file: 
output loading the case:
image

tests/commands/load/test_load_case_cmd.py
```diff
@@ -21,6 +22,54 @@ def test_load_case(mock_app, institute_obj, case_obj):
...
     # Test command to upload case using demo resources:
-    result =  runner.invoke(cli, ['load', 'case', load_path ])
+    result = runner.invoke(cli, ['load', 'case', load_path ])
     assert result.exit_code == 0
     assert store.case_collection.find().count() == 1
+
+
+
+
+def test_load_case_KeyError(mock_app, institute_obj, case_obj):
```
northwestwitch: Could it be that the sed command creates an extra file at the end of the test. After pytest on this file I end up with an extra config file: 
image

Maybe a better way to handle editing the config file would be making a copy of it on a temporary file?
See here: https://docs.python.org/3/library/tempfile.html and here: https://stackoverflow.com/questions/6587516/how-to-concisely-create-a-temporary-file-that-is-a-copy-of-another-file-in-pytho

tests/commands/load/test_load_case_cmd.py
```diff
@@ -21,6 +22,54 @@ def test_load_case(mock_app, institute_obj, case_obj):
...
+    # WHEN: config is loaded
+    result = runner.invoke(cli, ['load', 'case', load_path ])
+    # THEN KeyError is caught and exit value is non-zero
+    assert result.exit_code != 0
+
+    sed_restore = "sed -i -e 's/SAMPLE_ID/sample_id/g' "    # restore config
+    os.system(sed_restore + load_path)    
+
+    
+def test_load_case_NoConf(mock_app, institute_obj, case_obj):
```
northwestwitch: 👍

mikaell:
> I don't like so much how the conf file is edited using sed command. Could you instead create a temporary copy from it and edit that instead?

Yeah, that was a botch hack. I'll make a test conf.

tests/commands/load/test_load_case_cmd.py
mikaell: Added use of temp file instead:

```
    temp_conf = os.path.join(tempfile.gettempdir(), 'temp.conf')
    shutil.copy2(load_path, temp_conf)
    sed_change = "sed -i -e 's/sample_id/SAMPLE_ID/g' "
    os.system(sed_change + temp_conf)
```

mikaell:
> After I run tests I end up with this extra conf file named 643594.config.yaml-e. A sub-process of sed maybe? What happens if you remove the line in the conf file in some other way instead?

Strange, I'll have a look.

northwestwitch:
Could this be the reason? https://stackoverflow.com/questions/34533893/sed-command-creating-unwanted-duplicates-of-file-with-e-extension

Issue #1289, opened 2019-08-26
Unable to create General report (dowloading of PDF)


KickiLagerstedt:
I tried yesterday and today and I get the same error message:

"Internal Server Error

The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application."

Something wrong with my computer or something general??

northwestwitch:
We've received some error messages from the server at this regard, I'll take a look as soon as I'm at the office.

northwestwitch:
@ingkebil we are having some problems with scout that might be due to server settings, here's the error:

Traceback (most recent call last):
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/urllib3/connectionpool.py", line 603, in urlopen
    chunked=chunked)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/urllib3/connectionpool.py", line 344, in _make_request
    self._validate_conn(conn)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/urllib3/connectionpool.py", line 843, in _validate_conn
    conn.connect()
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/urllib3/connection.py", line 370, in connect
    ssl_context=context)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/urllib3/util/ssl_.py", line 355, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/ssl.py", line 401, in wrap_socket
    _context=self, _session=session)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/ssl.py", line 808, in __init__
    self.do_handshake()
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/ssl.py", line 1061, in do_handshake
    self._sslobj.do_handshake()
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/ssl.py", line 683, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1290, opened 2019-08-26
Disable ssl

northwestwitch:
This PR fixes the SSL: CERTIFICATE_VERIFY_FAILED error in #1289.

This is a workaround before changing the settings in servers (see here: https://github.com/Clinical-Genomics/servers/issues/179)

**How to test**:
1. It's not possible to download PDF reports from the 'general report' page. You can test on any case on scout-stage
1. How to install the branch on clinical-db stage:
```
ssh your.user@clinical-db.scilifelab.se
sudo -iu hiseq.clinical
cd servers/resources/clinical-dbscilifelab.se
bash update-scout-stage.sh disable_ssl
```

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fixed bug affecting variant observations in other cases
 - Fixed a bug that showed wrong gene coverage in general panel PDF export
 - MT report only shows variants occurring in the specific individual of the excel sheet
+- Disable SSL certifcate verification in requests to chanjo
 
 ## [4.6.1]
 
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -303,7 +303,7 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
     request_data['level'] = institute_obj.get('coverage_cutoff', 15)
 
     #send get request to chanjo report
-    # Disable default certificate verification
+    #disable default certificate verification
     resp = requests.post(base_url+'reports/report', data=request_data, verify=False)
 
     #read response content
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -7,7 +7,6 @@ import datetime
 import logging
 
 from bs4 import BeautifulSoup
-import ssl
 from xlsxwriter import Workbook
 from flask import url_for, current_app
 from flask_mail import Message
@@ -305,8 +304,7 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
 
     #send get request to chanjo report
     # Disable default certificate verification
-    ssl._create_default_https_context = ssl._create_unverified_context
-    resp = requests.post(base_url+'reports/report', data=request_data)
+    resp = requests.post(base_url+'reports/report', data=request_data, verify=False)
 
     #read response content
     soup = BeautifulSoup(resp.text)
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -7,6 +7,7 @@ import datetime
 import logging
 
 from bs4 import BeautifulSoup
+import ssl
 from xlsxwriter import Workbook
 from flask import url_for, current_app
 from flask_mail import Message
@@ -303,6 +304,8 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
     request_data['level'] = institute_obj.get('coverage_cutoff', 15)
 
     #send get request to chanjo report
+    # Disable default certificate verification
+    ssl._create_default_https_context = ssl._create_unverified_context
     resp = requests.post(base_url+'reports/report', data=request_data)
 
     #read response content


dnil:
I don't quite agree that the proper response to an updated certificate is to disable checks, but there is nothing wrong the code, and in practice a "man-in-the-middle" or such on one server is going to be quite difficult.

northwestwitch:
> I don't quite agree that the proper response to an updated certificate is to disable checks,

I know. Let's wait to merge this because we should change the settings. Another problem is that to fix production we should release but we are not quite there yet with the other PRs

northwestwitch:
We fixed the settings so I'm closing this now

northwestwitch:
I'm reopening this because apparently IT can't fix the SSL issue permanently and every time the SSL key expires we'll have this problem. Their fix of 2 days ago was reverting to the old SSL key, but this can't be done in the future. 
I think it's this fine with this PR, because the request to chanjo report for the html report doesn't use SSL anyway, it's just the PDF report one that verifies.

northwestwitch:
I think I'm going to merge this. You have 5 minutes to stop me ... :)

dnil:
Sure, go ahead. It should be safe enough when requesting locally. 

Still, this is treating symptoms. It doesn’t really explain why the new certificates didn’t quite work like the old ones? In particular, we did loose access to scout and scout-stage completely while the new certain where first installed. Then after a first fix (scout reverted?) we where back, but couldn’t access channel over ssl, which was fixed by a full revert.

northwestwitch:
> Still, this is treating symptoms. It doesn’t really explain why the new certificates didn’t quite work like the old ones? 

Apparently the most probable explanation is that there are some settings in the server (not the scout config files, but nginx or the dotfiles of the 'server' repo, I don't know) that associate scout to the old certificate key. So if you update the key it doesn't match and you get the error.

dnil:
I don’t have access to the server repo, but I guess I can get to the nginx settings etc directly on clinical-db. I’ll see if I can’t find time in the afternoon.

northwestwitch:
> I don’t have access to the server repo, but I guess I can get to the nginx settings etc directly on clinical-db. I’ll see if I can’t find time in the afternoon.

Maybe these 2 files?
ssl_certificate /etc/pki/tls/ssl-bundle.crt;
ssl_certificate_key /etc/pki/tls/private/myserver.key;

I don't have the permission to see the key but I do have access to the budle. I don't a thing about certificates and SSL tho! @ingkebil ?

ingkebil:
This is the problem: Scout is serving content over https which is not needed. Scout is behind a proxy, which serves scout's content over https. There is no need to double encrypt the content.

The reason why scout stopped working is because scout itself does not have access to the SSL certificates, only the webserver nginx has. To circumvent that, we at one time, copied the certificates from the webserver to a location that scout could access. So, when IT updated the certificates, they did not update the ones we had copied as they were not aware of that.

Anyway. There is no need for scout (or chanjo-report) to serve encrypted content. The solution is not to disable the checks but to provide the option of serving the content over http (instead of https).

Does this PR make scout able to serve content over http or does it disable certificate checking? Because the former is what I want and the latter is really bad practice.

northwestwitch:
> Does this PR make scout able to serve content over http or does it disable certificate checking? Because the former is what I want and the latter is really bad practice.

Ahem.. it was somehow complicated because the code doesn't have access to the app stuff (including the url_for method to create the link to chanjo) and I didn't want to hard code the url to chanjo either. So we are having the bad practice for now, but it's easily fixable!

ingkebil:
You've got 3 weeks to fix it without resorting to bad practice. Go! You can make it :)

ingkebil:
So guys, this PR does not address the issue that Scout _itself_ is serving https content with soon to be expiring certificates. Scout should be able to serve content over http to solve this issue.

northwestwitch:
> So guys, this PR does not address the issue that Scout _itself_ is serving https content with soon to be expiring certificates. Scout should be able to serve content over http to solve this issue.

Yes, I know how to fix it, I'll do it as soon as I have time

ingkebil:
Alright! That wasn't clear from the posts above. Good luck!

dnil:
Good, both of you!
Afaik there is no strict requirement on http or https serving in the app itself. When testing I usually run scout over localhost http. 

Communication with local apps (eg Chanjo, MME) is another matter, but could go via http if we are sure this is contained locally. Most of the issue - certainly a 502 Bad Gateway on the login page - will be a http server / reverse-proxy config matter, but I guess that is what you refer to. 😅

ingkebil:
OK, if it would help, I can just configure the proxy to ask for the http content on scout-stage. Do you want me to do that?
Do you need to change anything in the scout code?

northwestwitch:
> OK, if it would help, I can just configure the proxy to ask for the http content on scout-stage. Do you want me to do that?
> Do you need to change anything in the scout code?

I'm not sure. Why on scout stage only?

ingkebil:
I don't understand the question, so I'll give a really straithforward answer: You need to test these changes on scout-stage first. If that works, then we can apply the same changes to scout production.

I'll change the proxy now to point to the http version of scout.

northwestwitch:
> I'll change the proxy now to point to the http version of scout.

Didn't have time to work on this yet but could it be that the proxy change is causing a 502 error on scout stage now?

ingkebil:
Yes, this is causing a 502 because Scout is serving pages on https. You need to make Scout serve pages on http.

northwestwitch:
> Yes, this is causing a 502 because Scout is serving pages on https. You need to make Scout serve pages on http.

I'm able to fix the SSL issue we had when scout communicates with chanjo, but I have no idea how to fix this. I'll google

northwestwitch:
> Yes, this is causing a 502 because Scout is serving pages on https. You need to make Scout serve pages on http.

I don't understand. We didn't have a problem with scout after the IT updated the certificate, we had the issue just with that request sent to chanjo. And that I can easily fix. I don't know about proxies and certificates and server settings so I'm not able to fix stage. Could you please change back the proxy as it was, because now stage is broken and we need it to test other PRs. Thanks

ingkebil:
Of course I can put the proxy back the way it was.

When do you want me to tell IT to update the certificates?

northwestwitch:
> Of course I can put the proxy back the way it was.
> 
> When do you want me to tell IT to update the certificates?

As the things are now if they update the certificates in this moment scout works if we make a new scout release (because of this PR). So I think you can close the AM deviation  (at least the scout issues ) as soon as we release. 
Scout was working also after the certificate update, it was just one page that gave error, and that is fixed now. I can fix it in a more elegant way and you said we have 3 weeks, right?

ingkebil:
I've reset the proxy and assigned the deviation to Henrik. Good luck!

northwestwitch:
Oh no, I just solved the problem on servers! ( I don't think so, but I'm trying)

PR #1295, opened 2019-08-29
Export exons

moonso:
This PR adds exons to Scout. This means that we can update and export files with exon information to use with Chanjo from both builds and for all genes.

**How to test**:
1. Install the branch stage or in local instance
1. Either load the small example files with exons with command:
```
scout load exons -b 37 -e scout/demo/resources/ensembl_exons_37_reduced.txt
and
scout load exons -b 38 -e scout/demo/resources/ensembl_exons_38_reduced.txt
```
or load all exons from ensembl with command
```
scout load exons
```
Sometimes the ensembl biomart does not work for build 38. In this case one has to manually make an mart export from the homepage and load this file. Instructions for this is in `docs/admin-guide/genes`

1. Export the exons with 
```
scout export exons
```

**Expected outcome**:
The functionality should be working
Export of exons should be on the bed like chanjo specific format like:
```
3	72055925	72056126	3-72055925-72056126	ENST00000490086	49517	UBE2Q2P9
8	87673667	87674150	8-87673667-87674150	ENST00000514854	49518	UBE2Q2P10
8	87673409	87673562	8-87673409-87673562	ENST00000514854	49518	UBE2Q2P10
8	87672753	87673014	8-87672753-87673014	ENST00000514854	49518	UBE2Q2P10
15	84924665	84924766	15-84924665-84924766	ENST00000560644	49521	UBE2Q2P8
15	84976434	84976535	15-84976434-84976535	ENST00000559691	49522	UBE2Q2P11
15	84967471	84967573	15-84967471-84967573	ENST00000559691	49522	UBE2Q2P11
15	85043729	85043830	15-85043729-85043830	ENST00000414573	49523	UBE2Q2P12
15	85034711	85034812	15-85034711-85034812	ENST00000414573	49523	UBE2Q2P12
1	112150758	112150940	1-112150758-112150940	NR_034126	49525	LINC01160
1	112147251	112147324	1-112147251-112147324	NR_034126	49525	LINC01160
1	112146991	112147122	1-112146991-112147122	NR_034126	49525	LINC01160
1	112144814	112144877	1-112144814-112144877	NR_034126	49525	LINC01160
```

**Review:**
- [ ] code approved by
- [ ] tests executed by
- [ ] "merge and deploy" approved by

PR #1306, opened 2019-09-08
Ensembl Rest API and Biomart handlers

northwestwitch:
This PR adds classes for retrieval of data from Ensembl via:
1. Ensembl Rest API (for targeted queries for instance genes, regions etc)
2. Downloading to file (tempfile? Depends on the users requirements) of chunks of information from Biomart (exons for instance). 

Both classes work for retrieving info for build 37 and 38

**How to test**:
1. Check unit tests
1. It is meant to be used from the CLI. 
1. It is meant to fix the issue of not being able of downloading data for build 38 in PR #1295

**Expected outcome**:
Use this code to download exons in build 37 and 38 and it should work!

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Tests passing. Good enough for now, we can fix that request test later. Thanks @moonso . Merging now!

PR #1307, opened 2019-09-09
Fix #1273 - string replace STR var names.

### LINKED ISSUE #1273: STR alleles to copy number
@vwirta pointed out that there is a difference between how allele sizes and reference sizes are shown. Let’s parse the allele <STRn> for n, ideally on load, and use this instead.

dnil:
This PR fixes the display of STR var names, issue #1273.

**How to test**:
1. Look at the STR page for a case, specifically at the "Estimated size " column. 

**Expected outcome**:
The functionality should be working. Alt alleles should show as numbers only: ```123``` - not ```<STR123>```.

Screenshot 2019-09-09 at 11 08 45

**Review:**
- [x] code approved by
- [x] tests executed by


scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -41,13 +41,14 @@
       <thead>
         <tr>
           <th class="col-xs-1" title="Rank position">Rank</th>
```
northwestwitch: You are using 11 cols out of 12, you could use the last one to give some more space to one of the existing cols? It's up to you!

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -64,13 +65,19 @@
...
             <td>{{ variant.str_repid }}</td>
 	          <td class="text-right">{{ variant.str_ru }}</td>
-            <td>{{ variant.alternative }}</td>
+            <td class="text-right">{{ variant.alternative|replace("STR", "")|replace("<", "")|replace(">", "") }}</td>
             <td class="text-right">{{ variant.str_ref }}</td>
             <td>{{ variant.str_status }}</td>
+            <td>{{ variant.chromosome }}</td>
+            <td>
+                {% if case.bam_files %}
+                  <a class="btn btn-default btn-sm" href="{{ url_for('alignviewers.igv', sample=case.sample_names, build=case.genome_build, bam=case.mt_bams, bai=case.mt_bais, contig=variant.chromosome, start=(variant.position - 50), stop=(variant.position + 50), center_guide="T") }}" target="_blank">{{ variant.position}}</a>
```
northwestwitch: Nice!

dnil: :blush: There was actually an error there - I apparently tried to load the mt_bams instead of bam_files. Otherwise nice. 😄

scout/server/blueprints/variants/templates/variants/str-variants.html
dnil: Ok!

dnil:
We should load a recent case with STRs run to stage. There might be one somewhere, but the current verification ones don't seem to have it in place. But, this is working just as well locally as the previous version. I think we can safely merge it.

northwestwitch:
I'll use the stage cli to upload str variants from a recent case. So we can compare before and after the PR

northwestwitch:
If we test it on stage we can see if the igv link works as well

northwestwitch:
STRs uploaded for case https://scout-stage.scilifelab.se/cust000/SWEDAC-2019-HG-05-Trio-NovaSeq.

Test STRs page before: 
image

Test STRs page after:
image

I've noticed that the pileup viewer doesn't display anything. igv works!

vwirta:
What do you think about the column header repeat Id? Should we call it gene symbol instead?

dnil:
Hm, almost there.. I'll fix the link out from the str-variants page, but have to finish something else first!

dnil:
@vwirta: its not always precisely a gene symbol, unfortunately. Repeat locus perhaps, but I doubt anyone else will actually read it.. 😉

dnil:
And for good measure: now with IGV viewer button (case info loaded from controller) and genotype directly on the variants view. So far so good. Coveralls apparently fell since only 16/17 new lines were covered. 🤕

PR #1283, opened 2019-08-16
Bootstrap4 ready-sh

northwestwitch:
This PR changes the layout of Scout and introduces nice graphs in the dashboard pages

**How to test**:
1. install the branch on clinical-db so that the changes will be in scout-stage:
```
bash update-scout-stage.sh bootstrap4_ready
 ```
1. Extensive tests should be performed by checking all web pages available.
1. Colors and objects are not fixed and can (and should) be changed. **I'm just suggesting a style** but we can discuss about it together to improve the portal.

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil, MMA, CR etc

PR #1310, opened 2019-09-09
Fix #594 by manual addition of genes to dynamic panel.

### LINKED ISSUE #594: Clinical filter - own gene panel -- not the other way around
We have an issue regarding filtering using "clinical filter" and an "home made" gene panel (HGNC Symbols (case sensitive) filter):

What we do now:
1) filter with "clinical filter"
2) add home made gene filter
3) see variants

What we would like to do:
1) add home made gene filter
2) see all variants regarding these genes

bild

3) add clinical filter  - 
-  What happens now is that we see ALL variants - not only those from HGNC Symbols (case sensitive) filter

bild

Then we have to add the add home made gene filter again and press "filter variants"

bild

This  is an risk if the added gene list is long - could be analyzing genes not been asked for so we would appreciate if this "bug" could be fixed.

// Kicki & Anna

dnil:
This PR gives the functionality to manually add genes to dynamic "HPO" gene panels, thereby fixing #594. A user can now construct a panel of her choosing, then go directly to the variants view via the HPO/dynamic panel button, without having to first view any of the default panel genes - typically OMIM in this scenario.

Screenshot 2019-10-02 at 20 33 31

**How to test**:
1. add genes to gene panel manually, e.g. POT1.
1. check that a generated HPO panel can be used in filtering on the variants page. Use a narrow term that still yields a few clinical variants, e.g. HP:0000763 on the demo case.
1. check that clinical filtering works on the variants page (for default panel)
1. check that toggling the "use HPO panel in clinical filter" sets "clinical filter" to use the HPO panel for clinical filter, via checking "HPO" in the gene list dropdown on the filter page or pressing the corresponding clinical filter button. **Edit**: *After a bit of consideration I reverted from changing the behaviour of the Clinical HPO SNV panel direct link. Its fine as it is*.

**Expected outcome**:
The functionality should be working.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


dnil:
Not quite finished; just adding to test the idea and keep track of it.

dnil:
:metal: I'll try and update once finished. This now works, but I realised while testing the "Clinical HPO" shortcut does in fact return a clinical (non-research) HPO list, but not clinical filter of HPO terms as the customer thought. I'll investigate, but not now.. 😪

dnil:
When testing, I'd appreciate an opinion on wether its ok that the page returns to the top of the page upon submit of adding genes, or toggling the checkbox. If this is annoying, perhaps it would be possible to solve with some book-mark anchors on the page.

dnil:
Ok, this should be ready for review now; anyone who has time!

northwestwitch:
@dnil this PR is heavy on web stuff so same here, I'd like to merge first #1283. I can suggest the changes to the code to make it compliant in the review later.

dnil:
Oki, good exercise - I concur.

northwestwitch:
I'm going to review this now!

dnil:
I could also have a look at the merged version first, but may not have time until tomorrow. As you have time!

northwestwitch:
Testing this, while working with fixing the case page I've realized that the typeahead javascript doesn't work any more. Not your fault, it's caused by the new bootstrap code. I'll try to fix it in this PR

northwestwitch:
> Testing this, while working with fixing the case page I've realized that the typeahead javascript doesn't work any more. Not your fault, it's caused by the new bootstrap code. I'll try to fix it in this PR

This is easily fixed by changing the bootstrap 4 template (bootstrap4.html) to use the the follosing version of jquery: <script src="https://code.jquery.com/jquery-3.1.1.min.js"></script>
Instead of the slim version.

northwestwitch:
Would it be a lot of work to include an additional test to cover the changes in adapter.update_dynamic_gene_list? 

image

dnil:
Hi, in particular @moonso: I'm more or less satisfied with this one, but then I tried to add a test as per suggestion from @northwestwitch. No luck with finding gene by alias. I suspect I have to populate the hgnc collection in some more way than just adding my gene object, but I can't find any isolated test examples for this. Suggestions?

moonso:
Have you tried to use a gene database fixture?

northwestwitch:
gene_obj has chromosome build == '38', genes_by_alias doesn't find it if
you don't specify the build

dnil:
Thank you! I tried to make sure there were build:'37' defaults all the way through, but did not  check gene_obj. That was probably it; it worked with a gene db fixture - which is clearly in 37, after removing some other test things.

northwestwitch:
I'll take a look at this in the afternoon, tomorrow morning at the latest!

scout/adapter/mongo/case.py
```diff
@@ -289,18 +289,25 @@ def update_dynamic_gene_list(self, case, hgnc_symbols=None, hgnc_ids=None,
             updated_case(dict)
         """
         dynamic_gene_list = []
+        if add_only:
+            dynamic_gene_list_cursor = self.case_collection.find({ '_id': case['_id'] },
```
northwestwitch: From what I understand from the code if you are only interested in the first result returned from the query. Why don't you use find_one() instead of find() and the dynamic_gene_list_cursor[0] to get the first element of the cursor?

tests/adapter/mongo/test_case_handling.py
```diff
@@ -356,6 +356,26 @@ def test_update_case_collaborators(adapter, case_obj):
     # THEN assert all collaborators where added
     assert set(res['collaborators']) == set([coll_1, coll_2, coll_3])
 
+def test_update_dynamic_gene_list(gene_database, case_obj):
+    adapter = gene_database
+    # GIVEN an populated gene database,
```
northwestwitch: a populated gene database, move it to the top of the function?

tests/adapter/mongo/test_case_handling.py
```diff
@@ -356,6 +356,26 @@ def test_update_case_collaborators(adapter, case_obj):
...
+
+    # GIVEN a new cases with an empty dynamic_gene_list
```
northwestwitch: a new case

tests/adapter/mongo/test_case_handling.py
```diff
@@ -356,6 +356,26 @@ def test_update_case_collaborators(adapter, case_obj):
...
+    adapter.case_collection.insert_one(case_obj)
+    assert adapter.case_collection.find_one()
+    assert len(adapter.case(case_obj['_id'])['dynamic_gene_list']) == 0
+
+    # GIVEN a gene with a gene symobl
+    gene_obj = gene_database.hgnc_collection.find_one({'build': '37'})
```
northwestwitch: :)

scout/adapter/mongo/hgnc.py
```diff
@@ -265,9 +265,11 @@ def gene_by_alias(self, symbol, build='37'):
         Returns:
             res(pymongo.Cursor(dict))
         """
-        res = self.hgnc_collection.find({'hgnc_symbol': symbol, 'build':build})
+        LOG.debug("Fetch gene by symbol if possible: {}".format(symbol))
+        res = self.hgnc_collection.find({'hgnc_symbol': symbol, 'build': build})
         if res.count() == 0:
```
northwestwitch: Coud you change the code here so you are not using res.count ? You could use the sum function as in the tests

moonso:
Can you sign off and merge @northwestwitch ?

PR #1324, opened 2019-09-19
Display earlier ACMG classifications in the variants list

bjhall:
This PR shows earlier ACMG evalautions (ie in other cases) to the variants list, to allow quick identification of previously evalauated variants.

Not sure how this affects performance on large databases. Do you have tests for such things? I see no noticable slow-down on our installation.

See screenshot below.  I used "label-primary" to visually differentiate it from ACMG-classifications made in the specific case, but I could change that to something else.
image

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -48,6 +48,14 @@ def variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50
     for variant_obj in variant_res:
         overlapping_svs = [sv for sv in store.overlapping(variant_obj)]
         variant_obj['overlapping'] = overlapping_svs or None
+
+        # Get all previous ACMG evalautions of the variant
+        evaluations = []
+        for evaluation_obj in store.get_evaluations(variant_obj):
+            evaluation_obj['classification'] = ACMG_COMPLETE_MAP[evaluation_obj['classification']]
+            evaluations.append(evaluation_obj)
+        variant_obj['evaluations'] = evaluations
+            
         variants.append(parse_variant(store, institute_obj, case_obj, variant_obj,
                         update=True, genome_build=genome_build))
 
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -48,6 +48,14 @@ def variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50
     for variant_obj in variant_res:
         overlapping_svs = [sv for sv in store.overlapping(variant_obj)]
         variant_obj['overlapping'] = overlapping_svs or None
+
+        # Get all previous ACMG evalautions of the variant
+        evaluations = []
+        for evaluation_obj in store.get_evaluations(variant_obj):
```
moonso: If it is slow we could add an index, lets use it for a while and see

dnil: I don't think it should slow, but that *would* be a very straightforward index otherwise! Wonderful incentive to classify variants! Let's just document / legend this well so that we don't confuse anyone.

bjhall: Waiting for bootstrap 4 sounds like a good idea, as it would allow for better styling of the badges. Any ideas on how to style it to avoid misunderstandings? The tooltip and different color is maybe not enough?

northwestwitch:
Hi, I don't have time to fix the bootstrap 4 PR on these days. But go on and merge this one, it's easy to convert the code to the latest bootstrap later!

moonso:
All right @bjhall , could you update `CHANGELOG.md` and then I'll merge!

bjhall:
@moonso The changelog has been updated. Sorry for the delay.

northwestwitch:
I can merge, shall I?

northwestwitch:
> I can merge, shall I?

I'll merge!

PR #1333, opened 2019-09-26
Store simple acmg

moonso:

**How to test**:
1. Fetch the branch and install
1. Add a "simple" ACMG annotation, go back to variants list and check that it shows up(has to be P/LP). Now go back to variant page and unbox the annotation. See that it disappears.

**Expected outcome**:
This is described in issue 

**Review:**
- [ ] code approved by
- [ ] tests executed by


bjhall:
This has some unexpected behaviours in combination with https://github.com/Clinical-Genomics/scout/pull/1324

1. If you have the same variant in two cases and do a quick evaluation in one case (P or LP), the variant will not be marked in the variant list of the other case.

2. If you do a quick evalation of VUS/LB/B in a case, it shows as an tiny empty badge in the variant list of the same page. I expected it to show VUS/LB/B as before in the same case, but nothing in other cases with the same variant. @dnil, correct me if I misinterpreted your suggestion.

moonso:
Hi @bjhall , I have fixed those things. Could you have another look?

moonso:
Oh and this is ready for review whenever you have time @dnil , @northwestwitch , @bjhall

bjhall:
I'll try to have a look at it tonight (I'm paternity leave this week and next)

scout/demo/643595.config.yaml
```diff
@@ -0,0 +1,61 @@
+---
+
+owner: cust000
+
+family: 'internal_id_2'
```
northwestwitch: Why did the case id change here?

dnil: Since you want to test that the labels appear on *another* case?

dnil:
👍 And agreed @bjhall: show all classifications in the same case, but show LP/P classification from other cases with a distinctly different set of labels.

And I realised I missed one aspect before: how about the secondary classifications, eg variant tags? They would be relevant for e g SVs..

moonso:
What do the rest of you say @bjhall , @dnil is there a need for different badges depending on the type of acmg evaluation?

dnil:
Nah, not really for me, unless there is a technical issue that you would end up on an unfilled evaluation page? I had it confused with variant tags, which should probably also be shown in a similar manner, with a slightly different label..

bjhall:
Not a problem for me either!

scout/demo/643595.config.yaml
moonso: Yes, it is nice to have two test cases with the same variants to test behaviours like this one

northwestwitch:
Fine by me with same badge as well then. Try to add the test when you have time!

moonso:
@northwestwitch there is to much work for me to automate a test for this, can we settle with the fact that 4 people have manually tried it and have seen that it works as expected?

northwestwitch:
> @northwestwitch there is to much work for me to automate a test for this, can we settle with the fact that 4 people have manually tried it and have seen that it works as expected?

Fine 👍

northwestwitch:
Merging this as soon as it passes the tests

PR #1385, opened 2019-10-11
Gene manual inheritance edit in gene panel

### LINKED ISSUE #1335: editing the panel at the gene level
I want ot change the manually curated inheritande model in a gene but it is not possible. I gent the wrong type of inheritance model type , see atttached picture

bild

northwestwitch:
fix #1335 and a bug

**How to test**:
1. Install the branch on stage and go to a gene panel. 
1. Go to the list of included genes and click on 'edit'
1. You should reach the gene-edit panel:
image
1. It should be possible to add custom inheritance models for this gene
1. The available inheritance models should be different from those that are now on prod. These models are specified in #1335
1. Try to add one or more inheritance models from the select as well as one or more custom inheritance models (comma separated) for this gene
1. Update the panel and make sure that all the inheritance models are shown in the gene list of the panel page, like this:
image

**Expected outcome**:

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil


scout/constants/gene_tags.py
```diff
@@ -0,0 +1,12 @@
+INHERITANCE_MODELS = (
```
dnil: There appears to be a constant in the panel controller with the same name? Perhaps this should be CUSTOM_INHERITANCE_MODELS?

dnil: Also perhaps a couple of comments describing the difference to GENETIC_MODELS and matching OMIM inheritance models?

northwestwitch: You are right, I'll change to constant name into GENE_CUSTOM_INHERITANCE_MODELS!

northwestwitch: I don't have any official reference to OMIM for these models. The idea is that the should be custom, and in fact this is a list compiled by @4WGH. I don't know if there is an official documentation in OMIM and most importantly if it corresponds to these models.

northwestwitch:
Renamed models variable and added a tentative comment!

scout/constants/gene_tags.py
```diff
@@ -0,0 +1,16 @@
+# These inheritance models are distinct from the official OMIM models of inheritance for variants
+# which are specified by GENETIC_MODELS (in variant_tags.py).
+# The following models are used while describing inheritance of genes in gene panels
+# It's a custom-compiled list of values
```
dnil: :+1: Thanks!

PR #1389, opened 2019-10-15
Toggle date sort order - fix #1299.

### LINKED ISSUE #1299: Case sort order cycling
When you select a case sort order, selecting the same order multiple times should cycle through forward and reverse sort order on that key.

dnil:

**How to test**:
1. Install in an environment with more than one case.
1. Click sort of track, status and date to see that date sorting reverses sort order on toggle and icon colour changes to indicate which order is active.

Screenshot 2019-10-15 at 12 44 29

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Dynamic gene panels get link out to chanjo-report for coverage report
 - Load all clinvar variants with clinvar Pathogenic, Likely Pathogenic and Conflicting pathogenic
 - Show transcripts with exon numbers for structural variants
+- Case date sort order can now be reversed (toggle old->new or new->old).
 
 ### fixed
 - Fixed missing import for variants with comments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Dynamic gene panels get link out to chanjo-report for coverage report
 - Load all clinvar variants with clinvar Pathogenic, Likely Pathogenic and Conflicting pathogenic
 - Show transcripts with exon numbers for structural variants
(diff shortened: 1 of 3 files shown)


northwestwitch:
I'm reviewing this PR. I think you could sort by all 3 options and using a lot less code if you use jinja in the cases page. Can I try some changes on this PR?

dnil:
Sure thing: if so, perhaps include the sort order from case similarity searches? That should fit well with a jinja sort rather than mongo!

northwestwitch:
> Sure thing: if so, perhaps include the sort order from case similarity searches? That should fit well with a jinja sort rather than mongo!

Let's start from the 3 sorting options and we move from there. If I put too much untested stuff I'm afraid I'm going to mess up!

northwestwitch:
I have no idea if it works. Should be tested on stage!

dnil:
Looks good from the code at least. Didn't you like the graying out for when other sort category is active? 😉

northwestwitch:
> Looks good from the code at least. Didn't you like the graying out for when other sort category is active? 😉

Let's reintroduce it, I wasn't thinking about it actually! It's on stage now!

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Dynamic gene panels get link out to chanjo-report for coverage report
 - Load all clinvar variants with clinvar Pathogenic, Likely Pathogenic and Conflicting pathogenic
 - Show transcripts with exon numbers for structural variants
+- Case date sort order can now be reversed (toggle old->new or new->old).
```
northwestwitch: After testing that everything works you could update this to include case date, case name and track type

dnil:
Right, this is still using the mongo sort order, so lets keep the case similarity for a later PR.

PR #1392, opened 2019-10-16
Fix #1391 - STR back link (and column layout) broken.

### LINKED ISSUE #1391: Wrong link on STR varianT page
When entering an STR variant the link back to variant list leads to SNV/Indel and not STRs

dnil:
This PR fixes a bug / adds a missing function.

**How to test**:
1. click an STR variant (rank link). 
1. see that the variant page shows a link back to the STR variants list, and that it works.

Screenshot 2019-10-16 at 14 22 16

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Reintroduced missing coverage report
 - Fixed a bug preventing loading samples using the command line
 - Better inheritance models customization for genes in gene panels
+- STR variant page back to list button now does its one job.
 
 
 ## [4.7.3]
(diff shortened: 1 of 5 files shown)


dnil:
Interesting! That means something else messed up the column widths. The first column was super wide before, and ok after. I'll check!

dnil:
Ah, ok. Either one has to explicitly enable d-flex on all rows (and use the col-N notation rather than col-xs-N) or use a style width (as on the cases table). I'll update.
See e.g. https://stackoverflow.com/questions/25385289/how-to-set-the-size-of-a-column-in-a-bootstrap-responsive-table 😊

PR #1400, opened 2019-10-18
Fix #1395. Inactivate rerun cases only on re-upload.

### LINKED ISSUE #1395: Inactive cases not shown
My understanding from the last Scout user meeting was that inactive cases should always show up after prioritized (if it fits within the number of cases shown) before active and archieved cases, not to miss new inactive cases. 
I just realised that doesn't work. We have asked for a re-run of a case, which has now been placed in inactive. However I can't see the inactive cases unless I change the number of cases shown to 200. 
Also, the re-run case in inactive, has not yet been re-run from what I can tell on the analysis date. should it not move to inactive when the re-run is done, rather than when the re-run has been ordered? Just want to have a simpel way to see if a re-run case is ready. 

/Sofie

dnil:
This PR fixes a bug.

**How to test**:
1. Activate (or archive) a case.
1. Request rerun, and ensure that case does not become inactive.
1. See that the case gets a rerun-pending badge on cases list.
1. Re-upload case, e.g.  ```scout --demo load case --update scout/demo/643594.config.yaml```
and ensure that it now becomes inactive, a rerun badge is shown but no pending badge.

**Expected outcome**:
The functionality should be working.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -516,7 +516,7 @@ class CaseHandler(object):
             {'_id': case_obj['_id']}
         )
 
-        # if case is updataed and was archived or active - make it inactive
+        # if case is updated and was archived or active - make it inactive
         updated_status = old_case.get('status')
         if old_case.get('status') in ['archived', 'active']:
             updated_status = 'inactive'
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -515,6 +515,12 @@ class CaseHandler(object):
         old_case = self.case_collection.find_one(
             {'_id': case_obj['_id']}
         )
+
+        # if case is updataed and was archived or active - make it inactive
+        updated_status = old_case.get('status')
+        if old_case.get('status') in ['archived', 'active']:
+            updated_status = 'inactive'
+
         updated_case = self.case_collection.find_one_and_update(
             {'_id': case_obj['_id']},
             {
@@ -544,7 +550,7 @@ class CaseHandler(object):
                     'research_requested': case_obj.get('research_requested', False),
                     'multiqc': case_obj.get('multiqc'),
                     'mme_submission': case_obj.get('mme_submission'),
-                    'status': old_case.get('status') if  old_case.get('status') != 'archived' else 'inactive'
+                    'status': updated_status
                 }
             },
             return_document=pymongo.ReturnDocument.AFTER
--- a/scout/adapter/mongo/case_events.py
+++ b/scout/adapter/mongo/case_events.py
@@ -294,8 +294,7 @@ class CaseEventHandler(object):
             {'_id': case['_id']},
             {
                 '$set': {
-                    'rerun_requested': True,
-                    'status': case.get('status') if case.get('status') != 'archived' else 'inactive'
+                    'rerun_requested': True
                     }
             },
             return_document=pymongo.ReturnDocument.AFTER
--- a/scout/server/blueprints/cases/templates/cases/cases.html
+++ b/scout/server/blueprints/cases/templates/cases/cases.html
@@ -219,9 +219,9 @@
     <td>
       {{ case.analysis_date.date() }}
       {% if case.rerun_requested %}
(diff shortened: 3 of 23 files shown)


dnil:
Ready for review. One notes that TICKET_SYSTEM_EMAIL config option is rather undocumented. I made a little bypass for it for now, but perhaps that should stay, at least until we document that feature a bit more.

northwestwitch:
Reviewing this now!

northwestwitch:
Tested on a local implementation (after adding MAIL_USERNAME, MAIL_PASSWORD, ADMINS) to the config file.

After clicking on the rerun an active case doesn't get inactivated and gets the rerun pending badge:
image

After reuploading the case it becomes inactive and the rerun pending badge disappears:

image

scout/adapter/mongo/case.py
```diff
@@ -515,6 +515,12 @@ def update_case(self, case_obj):
         old_case = self.case_collection.find_one(
             {'_id': case_obj['_id']}
         )
+
+        # if case is updataed and was archived or active - make it inactive
```
northwestwitch:
```suggestion
        # if case is updated and was archived or active - make it inactive
```

scout/adapter/mongo/case_events.py
```diff
@@ -294,8 +294,7 @@ def request_rerun(self, institute, case, user, link):
             {'_id': case['_id']},
             {
                 '$set': {
-                    'rerun_requested': True,
-                    'status': case.get('status') if case.get('status') != 'archived' else 'inactive'
+                    'rerun_requested': True
```
northwestwitch: 👍

scout/server/blueprints/cases/templates/cases/cases.html
```diff
@@ -218,7 +218,9 @@ <h5 class="modal-title" id="sangerModalLabel">Sanger validations to evaluate:</h
     </td>
     <td>
       {{ case.analysis_date.date() }}
-      {% if case.is_rerun %}
+      {% if case.rerun_requested %}
+          <span class="badge badge-secondary pull-right">rerun pending</span>
```
northwestwitch: pull-right is deprecated in Bootstrap 4 and that's why the barge is aligned weirdly.

You could add **class="d-flex justify-content-between align-items-center"** to the td parent element instead

scout/server/blueprints/cases/templates/cases/cases.html
```diff
@@ -218,7 +218,9 @@ <h5 class="modal-title" id="sangerModalLabel">Sanger validations to evaluate:</h
...
+      {% elif case.is_rerun %}
         <span class="badge badge-secondary pull-right">rerun</span>
```
northwestwitch: Same here, it will be fixed by changing the class of the td as well

tests/adapter/mongo/test_case_handling.py
```diff
@@ -447,11 +446,10 @@ def test_update_case_rerun_status(adapter, case_obj, institute_obj, user_obj, ):
     # request rerun for test case
     adapter.request_rerun(institute_obj, res, user_obj, 'blank')
     res = adapter.case(case_obj['_id'])
-
+    # THEN rerun_reuquested is flagged
```
northwestwitch: requested

scout/server/blueprints/cases/templates/cases/cases.html
dnil: Looks like float-* is going to remain supported, so for simplicity I switched to that instead!

northwestwitch:
I'm testing the latest changes (float-right instead of pull-right). Unfortunately the bagde alignment looks still off:

image

dnil:
Right you are: pull or float won't affect the vertical align! Updated!

northwestwitch:
While testing this I noticed a super-ugly thing that is totally my fault:
image

I'll fix it in this PR!

northwestwitch:
Now it's better:
image

PR #1402, opened 2019-10-19
Skip omim genes

moonso:
This PR fixes so that it is possible to setup scout without an OMIM api key

**How to test**:
1. Install this branch
1. Setup scout in the usual way but without an api key:
    1. `scout -db gene-test setup database`
    1. `scout -db gene-test load panel scout/demo/panel_1.txt` 
    1. `scout -db gene-test load case scout/demo/643594.config.yaml`
    1. `scout -db gene-test serve`

**Expected outcome**:
Everything should work as usual but there will not be any OMIM information. This means reduced gene information and no disease terms.

**Comments**

Don't forget to apply for a api-key from OMIM: https://omim.org/downloads/
And when you have it run:

`scout update genes --api-key <your key>`
`scout update diseases --api-key <your key>`

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil

### DIFF
--- a/scout/load/hgnc_gene.py
+++ b/scout/load/hgnc_gene.py
@@ -106,7 +106,7 @@ def load_hgnc_genes(adapter, genes = None, ensembl_lines=None, hgnc_lines=None, 
             exac_lines=exac_lines,
             hpo_lines=hpo_lines,
             mim2gene_lines=mim2gene_lines,
-            genemap_lines=genemap_lines,
+            genemap_lines=genemap_lines
         )
 
     non_existing = 0
--- a/scout/load/hgnc_gene.py
+++ b/scout/load/hgnc_gene.py
@@ -89,10 +89,11 @@ def load_hgnc_genes(adapter, genes = None, ensembl_lines=None, hgnc_lines=None, 
         exac_lines = exac_lines or fetch_exac_constraint()
         if not (mim2gene_lines and genemap_lines):
             if not omim_api_key:
-                raise SyntaxError("Need to provide omim api key")
-            mim_files = fetch_mim_files(omim_api_key, mim2genes=True, genemap2=True)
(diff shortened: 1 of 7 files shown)


scout/load/hgnc_gene.py
```diff
@@ -103,9 +104,9 @@ def load_hgnc_genes(adapter, genes = None, ensembl_lines=None, hgnc_lines=None,
             ensembl_lines=ensembl_lines,
             hgnc_lines=hgnc_lines,
             exac_lines=exac_lines,
+            hpo_lines=hpo_lines,
             mim2gene_lines=mim2gene_lines,
             genemap_lines=genemap_lines,
```
dnil: Trailing commas are scary. Could we just not have them? 😟

scout/utils/link.py
```diff
@@ -161,8 +164,8 @@ def get_correct_ids(hgnc_symbol, alias_genes):
             return set(hgnc_id_info['ids'])
     return hgnc_ids
 
-def link_genes(ensembl_lines, hgnc_lines, exac_lines, mim2gene_lines,
-               genemap_lines, hpo_lines):
+def link_genes(ensembl_lines, hgnc_lines, exac_lines, hpo_lines, mim2gene_lines=None, 
+               genemap_lines=None):
```
dnil: :+1:

scout/load/hgnc_gene.py
dnil: I known theory they are allowed here, even being practical in some sense, but since they are not allowed in JSON and carry some unfortunate meaning with tuples, I get nervous. 😊

moonso: Hehe, in python code I think they have the opposite effect. If you always include them it is more likely that nothing goes wrong.

PR #1409, opened 2019-10-21
Removes flask version from cli

hassanfa:
This PR adds a fixes a bug at cli where `--version` shows twice, and one is from flask.

**How to test**:
```
scout --help
```

**Expected outcome**:
There should be one `--version` in output.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

This is a patch bump, as it is clearing up help and an inactive command line option,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - STR variant page back to list button now does its one job.
 - Allows to setup scout without a omim api key
 - Fixed error causing "favicon not found" flash messages
+- Removed flask --version from base cli
 
 
 ## [4.7.3]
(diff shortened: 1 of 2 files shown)


hassanfa:
Some context, i noticed the PR text is not clear enough:

**Before:**
image

**After:**
image

northwestwitch:
Oh I was thinking that I was reviewing a vogue PR, but it's scout. Before I merge, could you update the Changelog fix field? Thanks!

hassanfa:
> Oh I was thinking that I was reviewing a vogue PR, but it's scout. Before I merge, could you update the Changelog fix field? Thanks!

I was updating vogue so I thought I should fix scout as well. Why not fix it in two places if it takes double amount of time fixing two?

PR #1413, opened 2019-10-22
Improve variant page

moonso:
This PR refactors the backend code for the single variant views. This is a mayor code refactoring, many lines have been splitted into own functions and some functions have been merged. In particular there is only one variant controller for all types of variants 🗡 . Hopefully it will increase the loading time for a single variant...

## NOT READY TO MERGE BUT READY TO TEST AND REVIEW ##
I'm about to add more tests

**How to test**:
1. Install this branch on stage
1. Browse all different types of variants

**Expected outcome**:
The functionality should be better

**Review:**
- [x] code approved by dNil
- [ ] tests executed by


moonso:
I feel done! Please have a look @dnil @northwestwitch

dnil:
Coverall begs to differ! 😅

scout/server/blueprints/cases/controllers.py
```diff
@@ -252,26 +251,17 @@ def case_report_content(store, institute_obj, case_obj):
...
-                        variant_obj=var_obj,
-                        add_case=False,
-                        add_other=False,
-                        get_overlapping=False
-                    )
-            else:
-                decorated_info = sv_variant(
+            decorated_info = variant_decorator(
                     store=store,
-                    institute_id=institute_obj['_id'],
```
dnil: 👀

scout/server/blueprints/variant/templates/variant/sv-variant.html
```diff
@@ -228,7 +228,7 @@
 
   <div class="row">
     <div class="col-md-12">
-      {{ overlapping(overlapping_snvs, variant.rank_score) }}
+      {{ overlapping(overlapping_vars, variant.rank_score) }}
```
dnil: :+1:

dnil: No reason the second var cannot also be another SV, just more rare.

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
+import logging
+
+from scout.constants import (CLINSIG_MAP, CALLERS, ACMG_COMPLETE_MAP)
+from scout.server.links import (add_gene_links, ensembl, add_tx_links)
+
+LOG = logging.getLogger(__name__)
+
+def add_panel_specific_gene_info(panel_info, gene_obj):
```
dnil: Yes! That!

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+        # those
+        transcript['refseq_identifiers'] = hgnc_transcript.get('refseq_identifiers',[])
+
+        transcript['change_str'] = transcript_str(transcript, hgnc_symbol)
+    
+    variant_gene['primary_transcripts'] = refseq_transcripts
+        
+
+def add_gene_info(store, variant_obj, gene_panels=None, genome_build=None):
+    """Add extra information about genes from gene panels.
```
dnil: Could you be a bit more specific: it's extra how? Eg manually added to panels, imported from other dbs than hgnc, ...

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+        # Add omim information if gene is annotated to have incomplete penetrance
+        if hgnc_gene.get('incomplete_penetrance'):
+            variant_gene['omim_penetrance'] = True
+
+        ############# PANEL SPECIFIC INFORMATION #############
+        # Panels can have extra information about genes and transcripts
+        panel_info = add_panel_specific_gene_info(extra_info.get(hgnc_id, []), variant_gene)
+        variant_gene.update(panel_info)
+        
+        update_transcripts_information(variant_gene, hgnc_gene, variant_obj)
```
dnil: This is also not so clear: why a transcript update function here, then a section of code with transcripts being updated afterwards? Perhaps merge or make one more update function? Or just document it.

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+        frequencies(list(tuple)): A list of frequencies to display
+    """
+    
+    # Mandatory frequencies
+    
+    frequencies = [
+        ('GnomAD', variant_obj.get('gnomad_frequency')),
+    ]
+
+    if 'clingen_cgh_benign' in variant_obj:
```
dnil: A dict and a loop instead perhaps?

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+    
+    Args:
+        variant_obj(scout.models.Variant)
+    
+    Returns:
+        str in ['common','uncommon','rare']
+    """
+    most_common_frequency = max(variant_obj.get('thousand_genomes_frequency') or 0,
+                                variant_obj.get('exac_frequency') or 0,
+                                variant_obj.get('gnomad_frequency') or 0
```
dnil: For the SVs I would include the largest available local one - gnomAD aint too good yet. I suppose that's clingen_mip currently.

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+    """Calculate end position for a variant.
+    
+    Args:
+        variant_obj(scout.models.Variant)
+    
+    Returns:
+        end_position(int)
+    """
+    alt_bases = len(variant_obj['alternative'])
+    num_bases = max(len(variant_obj['reference']), alt_bases)
```
dnil: No, can't follow. What if both len(ref) > 1 and len(alt)> 1 as in a classical indel?

dnil: Also not compatible with SVs obviously, but I suppose its never called for them.. 😊

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,453 @@
...
+        if not panel_obj:
+            LOG.warning("Panel {0} version {1} could not be found".format(
+                panel['panel_name'], panel.get('version')))
+            continue
+        default_panels.append(panel_obj)
+    return default_panels
+
+def update_hgncsymbol(variant_obj):
+    """Check if the HGNC symbols have changed since the variant was loaded"""
+    pass
```
dnil: 👀

scout/server/blueprints/variants/controllers.py
```diff
@@ -262,23 +124,22 @@ def parse_variant(store, institute_obj, case_obj, variant_obj, update=False, gen
                                           key=lambda compound: -compound['combined_score'])
 
     # Update the hgnc symbols if they are incorrect
-    variant_genes = variant_obj.get('genes')
-    if variant_genes is not None:
```
dnil: Ok, the default [] return is more neat.

northwestwitch:
I'm also taking a look at this!

scout/server/blueprints/variants/controllers.py
```diff
@@ -993,7 +506,7 @@ def variant_verification(store, mail, institute_obj, case_obj, user_obj, variant
                     tx_changes.append("<li>{}</li>".format(':'.join(transcript_line)))
 
     else: #SV
```
dnil: Or STR! And what about cancer?

dnil:
Cancer variant page fails on stage.

dnil:
STR variant page fails on stage.

dnil:
SV frequencies suddenly missing on stage (older case).
Screenshot 2019-10-23 at 10 07 35

dnil:
SNV variant page fails on stage.

dnil:
Ok, I think you now have enough to work with for a bit @moonso! Really good work, and not strange that some things are not perfect yet.

northwestwitch:
Wow that looks like a success @moonso :D , just kidding, I really appreciate the effort and I understand the difficulty of the task!

dnil:
Also, how current is it? I had a feeling it was missing some stuff from the current master - notably the favicon fix, which is only annoying, but presumably some more important db changing things like partial_causatives may mess with testing on stage?

moonso:
Thanks a lot for the review @dnil , I'll try to fix those

moonso:
@dnil I have rebased on latest master, will try to see if some things have disappeared due to the fact that I did `git rm <file>`, realised now I should have done `git mv` ...

Get back when done!

scout/adapter/mongo/variant.py
```diff
@@ -246,18 +127,17 @@ def variant(self, document_id=None, gene_panels=None, case_id=None, simple_id=No
             query['_id'] = document_id
 
         variant_obj = self.variant_collection.find_one(query)
-        if variant_obj:
-            variant_obj = self.add_gene_info(variant_obj, gene_panels)
-            if variant_obj['chromosome'] in ['X', 'Y']:
-                ## TODO add the build here
-                variant_obj['is_par'] = is_par(variant_obj['chromosome'],
-                                               variant_obj['position'])
+        if not variant_obj:
```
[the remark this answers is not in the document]
moonso: You love forced indentation @dnil , just admit it!

scout/server/app.py
```diff
@@ -103,6 +103,7 @@ def register_blueprints(app):
     app.register_blueprint(genes.genes_bp)
     app.register_blueprint(cases.cases_bp)
     app.register_blueprint(login.login_bp)
+    app.register_blueprint(variant.variant_bp)
```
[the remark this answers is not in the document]
moonso: Probably but not in this PR!

scout/adapter/mongo/variant.py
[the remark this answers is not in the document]
dnil: I just ... can't. It messes with my anarchist self image. ;-)

scout/server/blueprints/cases/controllers.py
moonso: 😱

moonso: Nive find, I fixed that

scout/adapter/mongo/variant.py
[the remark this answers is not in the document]
dnil: But, I will admit as much as that if it wasn't so unnecessarily enforced, I would still indent everything, and correct improperly balanced indents of others.. ;-)

scout/server/blueprints/variant/utils.py
moonso: 😄

scout/server/blueprints/variant/utils.py
moonso: Sure

moonso: extended the doctoring

scout/server/blueprints/variant/utils.py
moonso: Good idea

moonso: AAAHHHhh it doesn't work! This informatin is added on the variant_obj and is collected from the panels. There is no nice way of solving this I'm afraid...

scout/server/blueprints/variant/utils.py
moonso: Fixed that, good idea. This will make it easy to expand with more frequencies

scout/server/blueprints/variant/utils.py
moonso: That would be new functionality, can we do that in another PR?

scout/server/blueprints/variant/utils.py
moonso: This is not used for SVs. Since all variants are normalised when loaded that situation should not happen

scout/server/blueprints/variant/utils.py
moonso: 🗽

scout/server/blueprints/variant/utils.py
dnil: I may just be slow: take it step wise! Normalise would to me only mean the alt and ref are as short as can be, and correctly adjusted to one side? C f https://genome.sph.umich.edu/wiki/Variant_Normalization; the first blue CAT->TGC is normalised but both alleles have size >1.

moonso: Aha.. and in that case one would like to have length 3?

dnil: Right; which still works I guess? And the next step would be to imagine two alleles with different >1 lengths...

But more to the point is perhaps that I don't exactly see why the variant end position would not always just be the variant start pos plus length of ref allele? Even if the alt allele is a km long, I right now don't see how that could affect the position on the reference?

moonso:
Can you try again on stage @northwestwitch @dnil ?

scout/server/app.py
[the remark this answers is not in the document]
dnil: :+1:

scout/server/blueprints/variant/controllers.py
```diff
@@ -0,0 +1,228 @@
...
+            # Case could have been removed
+            LOG.debug("Case %s could not be found in database", case_id)
+            continue
+        other_institutes = set([other_case.get('owner')])
+        other_institutes.update(set(other_case.get('collaborators', [])))
+        
+        if user_institutes_ids.isdisjoint(other_institutes):
+            # If the user does not have access to the information we skip it
+            continue
+        document_id = parse_document_id(chrom, str(pos), ref, alt, var_type, case_id)
```
dnil: :+1:

scout/server/blueprints/variant/utils.py
```diff
@@ -0,0 +1,463 @@
...
+    This function converts the raw annotations to something better to visualize.
+    GnomAD is mandatory and will always be shown.
+    
+    Args:
+        variant_obj(scout.models.Variant)
+    
+    Returns:
+        frequencies(list(tuple)): A list of frequencies to display
+    """
+    sv_freqs = {
```
dnil: :+1:

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -0,0 +1,122 @@
...
+        
+        var_id = variant_obj['_id']
+    
+        ## WHEN fetching the variant from the controller
+        data = variant(store, institute_id, case_name, variant_id=var_id, add_case=True,
+                add_other=True, get_overlapping=True, add_compounds=True, variant_type=category)
+    
+    ## THEN assert the data is on correct format
+    
+    assert 'overlapping_vars' in data
```
dnil: 🛑 No newline!

PR #1418, opened 2019-10-23
fix #1416

### LINKED ISSUE #1416: Weird variant lines in variants template
Look at the rank col!

image

northwestwitch:
fix #1416

**How to test**:
1. git checkout to this branch
1. Make sure that rank cell in variants page doesn't look like #1416

**Expected outcome**:

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fixed error causing "favicon not found" flash messages
 - Removed flask --version from base cli
 - Request rerun no longer changes case status. Active or archived cases inactivate on upload.
+- Fixed weird Rank cell in variants page
 
 
 ## [4.7.3]
--- a/scout/server/blueprints/variants/templates/variants/variants.html
+++ b/scout/server/blueprints/variants/templates/variants/variants.html
@@ -71,7 +71,7 @@
               {% else %}
                 <tr>
               {% endif %}
-                <td class="d-flex justify-content-between align-items-center">{{ cell_rank(variant) }}</td>
+                <td class="text-left">{{ cell_rank(variant) }}</td>
                 <td class="text-right">{{ variant.rank_score|int }}</td>
                 <td>{{ variant.chromosome }}</td>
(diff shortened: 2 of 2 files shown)


dnil:
When more exactly did this behaviour appear? Mine look nicely justified and centered, so don't know where to look for a change? Stacking all the different things that might appear in that cell in a jumble to the left doesn't quite sound like the solution, but maybe if I knew what drove it there I would understand better?

Screenshot 2019-10-23 at 10 31 32

northwestwitch:
> When more exactly did this behaviour appear? Mine look nicely justified and centered, so don't know where to look for a change? Stacking all the different things that might appear in that cell in a jumble to the left doesn't quite sound like the solution, but maybe if I knew what drove it there I would understand better?
> 
> Screenshot 2019-10-23 at 10 31 32

For instance here:
https://scout-stage.scilifelab.se/cust000/WNB128N/variants?variant_type=clinical&gene_panels=OMIM-AUTO 

I think it happens when the content of a line doesn't fit the line (for instance many gene annotations or Func. annotation)..

northwestwitch:
Otherwise of course it's better with the official col- thingy

northwestwitch:
The double badge thing might be solved by a pop-up explaining where the classificstion comes from?

dnil:
Nah, lets ignore it for the present case. It's like if the matching pathogenic would include the current case. Simply not elegant. It used to (and should still) have a distinct appearance - different kind of badge - that would indicate that it is not from the same case at any rate.

dnil:
But ok, lets leave it with tooltips - otherwise we have to touch the variant controller again... Lets revisit that at another time.

dnil:
How about this compromise then? Not the most elegant, but good for now perhaps?
Screenshot 2019-10-23 at 17 07 40

northwestwitch:
Ok! Why there are badges with numbers?

dnil:
Thats the manual ranks. Cant say I quite like that representation, but we've had that for a long time. At some point we should make badgeable abbreviations for them as well.

PR #1427, opened 2019-10-27
Persistent filters

dnil:
This PR adds a functionality: users can save filter settings that can be accessed by the institute

Screenshot 2019-10-27 at 21 33 20

I'm very tempted to use these saved filters to document different analysis sub-tasks. E.g. to log into the report that the user has checked SNVs for AR, de novo and reduced penetrance AD, and STRs but not yet SVs. But, that is likely another PR.

Also, a filter list with remove options should be present on the upcoming institute page.

**How to test**:
For each of the SNV, SV and Cancer variants filters,
1. Edit the filter settings and hit save. 
1. Ensure that a prompt asks for a filter name
1. Make sure you can load the filter, e.g. by hitting Reset filters, selecting your saved filter and hitting Load filter again.

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


henrikstranneheim:
Very nice functionality! 👍

dnil:
Thanks - credits to @bjhall for suggesting that we do it way-back-when!

This is now ready for review.

northwestwitch:
> Thanks - credits to @bjhall for suggesting that we do it way-back-when!
> 
> This is now ready for review.

I can review it!

northwestwitch:
Testing on a local branch: variants filter button 'Reset filters' occupies 2 lines even if there's a lot of space:

image

dnil:
You need a wider screen! 😜 

But, sure, the text is the longest in the set: let's make it two columns wide instead, to give it some margin.

northwestwitch:
I've been testing with 3 filters, first is empty and the other 2 contain params. It creates these database entries:
image

Question, what's the array object named 'filters' ?

dnil:
Thanks - redundant it is. And good idea to test with an empty filter - to be fair, it had a gene panel selected, but it seems to work with an "empty" form as well!

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+            filter_dict[element] = value
+
+        result = self.filter_collection.insert_one(filter_dict)
+
+        filter_id = result.inserted_id
+
+        # log event
+        subject = institute_obj['display_name']
+        # clever link - eg url_for load filter?
+        link = "filter/" + str(filter_id)
```
northwestwitch: In general links point to variants or cases. Even if you create a smart link you re not using it right now, so maybe an empty string, if link is required? Or even better, change the create event to accept link=None?

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+        filter_obj = self.filter_collection.find_one({'_id': ObjectId(filter_id)})
+
+        LOG.info("User {} deleting filter {} for institute {}.".format(
+            user_id, filter_obj.get('display_name'), institute_id))
+
+        result = self.filter_collection.delete_one({'_id': ObjectId(filter_id)})
+
+        return result
+
+    def filters(self, institute_id, category="snv"):
```
northwestwitch: 👍

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+        filter_id = result.inserted_id
+
+        # log event
+        subject = institute_obj['display_name']
+        # clever link - eg url_for load filter?
+        link = "filter/" + str(filter_id)
+        # snv, sv, str..
+
+        self.create_event(institute=institute_obj, case=case_obj, user=user_obj,
+            link=link, category='case', verb='filter_stash', subject=subject,
```
northwestwitch: verb="filter_save"

scout/server/blueprints/variants/templates/variants/sv-variants.html
```diff
@@ -222,6 +222,11 @@
       return true;
     }
 
+    function promptFilterName(suggestion) {
```
northwestwitch: Here also, it would be better to do it with a modal. A modal placed in utils.html which will be shared by all 3 types of variants perhaps?

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -247,6 +247,11 @@
       return true;
     }
 
+    function promptFilterName(suggestion) {
```
northwestwitch: Modal here as well!

scout/server/blueprints/variants/views.py
```diff
@@ -62,15 +59,32 @@ def variants(institute_id, case_name):
             'gene_panels': clinical_filter_panels
              })
 
+    user_obj = store.user(current_user.email)
     if(request.method == "POST"):
+        # If special filter buttons were selected:
         if bool(request.form.get('clinical_filter')):
             form = FiltersForm(clinical_filter)
             form.csrf_token = request.args.get('csrf_token')
+        elif bool(request.form.get('save_filter')):
```
northwestwitch: What about a completely different endpoint which is dedicated to save filters? So you can use it from 3 types of variants without repeating the same code in each of them. It could save the filter and return to the request refererrer (variants page)

scout/server/blueprints/variants/views.py
```diff
@@ -62,15 +59,32 @@ def variants(institute_id, case_name):
...
+            # The form should be applied and remain set the page after saving
+            form = FiltersForm(request.form)
+            # Stash the filter to db to make available for this institute
+            filter_obj = request.form
+            store.stash_filter(filter_obj, institute_obj, case_obj, user_obj, category='snv')
+        elif bool(request.form.get('load_filter')):
+            filter_display_name = request.form.get('filters')
+            filter_obj = store.retrieve_filter(filter_display_name)
```
northwestwitch: What if you have more than one filter with the same display name?

scout/server/blueprints/variants/views.py
```diff
@@ -80,6 +94,14 @@ def variants(institute_id, case_name):
 
     form.gene_panels.choices = panel_choices
 
+    # update status of case if vistited for the first time
```
northwestwitch: visited

scout/server/blueprints/variants/views.py
```diff
@@ -229,24 +248,53 @@ def sv_variants(institute_id, case_name):
             'gene_panels': default_panels
              })
 
+    user_obj = store.user(current_user.email)
     if(request.method == "POST"):
         if bool(request.form.get('clinical_filter')):
             form = SvFiltersForm(clinical_filter)
             form.csrf_token = request.args.get('csrf_token')
+        elif bool(request.form.get('save_filter')):
```
northwestwitch: Manage in another endpoint, same one as the one used by the variants filters

scout/server/blueprints/variants/views.py
```diff
@@ -229,24 +248,53 @@ def sv_variants(institute_id, case_name):
...
     else:
         form = SvFiltersForm(request.args)
 
+    # populate filters dropdown
+    available_filters = store.filters(institute_id, category='sv')
+    form.filters.choices = [(filter.get('_id'), filter.get('display_name'))
+        for filter in available_filters]
+
+    # redundant?
     form.variant_type.data = variant_type
```
northwestwitch: Dunno..?

scout/server/blueprints/variants/views.py
```diff
@@ -329,13 +367,30 @@ def cancer_variants(institute_id, case_name):
 
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
 
+    user_obj = store.user(current_user.email)
     if(request.method == "POST"):
-        form = CancerFiltersForm(request.form)
         page = int(request.form.get('page', 1))
+
+        if bool(request.form.get('save_filter')):
```
northwestwitch: Same as before, better handled in a different endpoint

tests/adapter/mongo/test_event_handling.py
```diff
@@ -512,6 +512,46 @@ def test_update_clinical_filter_hpo(adapter, institute_obj, case_obj, user_obj):
     assert sum(1 for i in adapter.event_collection.find()) == 1
 
 
+def test_filter_stash(adapter, institute_obj, case_obj, user_obj, filter_obj):
```
northwestwitch: rename to test_filter_save?

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+            level='global')
+
+        return filter_id
+
+    def delete_filter(self, filter_id, institute_id, user_id):
```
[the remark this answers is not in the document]
dnil: I had a comment about that somewhere; the idea was to show them on the anticipated "institute" page. But, you are right: it will be annoying to have to go to another page if you just want to make a slight change. I'll make more buttons.. X-)

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+                filter_obj(dict)
+        """
+        LOG.debug("Retrieve filter {}".format(filter_id))
+        filter_obj = self.filter_collection.find_one({'_id': ObjectId(filter_id)})
+
+        # use _id to preselect the currently loaded filter, and drop it while we are at it
+        filter_obj.update([('filters', filter_obj.pop('_id', None))])
+        return filter_obj
+
+    def stash_filter(self, filter_obj, institute_obj, case_obj, user_obj, category='snv'):
```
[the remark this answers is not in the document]
dnil: So the reason for the esoteric names was that we have this "load" functionality, implying file reading. The way I made it now they are only visible to the developers, not the end-users, and should not be to strange to us.. I'm not super enthusiastic about changing it, but it's just a name so will do if you feel strongly about it.

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+        for (element, value) in filter_obj.lists():
+            if value == ['']:
+                continue
+            if element in ['save_filter', 'filters', 'csrf_token']:
+                continue
+            if element == 'filter_display_name':
+                # filter display_name if given
+                filter_dict['display_name'] = value[0]
+                continue
+            filter_dict[element] = value
```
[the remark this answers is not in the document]
dnil: Yep, do. As far as I have tested, it seems to work, but dict and MultiDict interactions do not commute.

scout/server/blueprints/variants/views.py
dnil: Thanks, thats a bad old variable name. The functions actually use a unique id for retrieval. Fixed!

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+import logging
+from bson.objectid import ObjectId
+
+# needed?
+from scout import __version__
```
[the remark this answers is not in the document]
dnil: Right, concur. I never got around to actually using it anyway. I suppose the code itself is relatively resilient to changes, as the Form(MultiDict) would just drop extra keys, and there may be scarier things happening when one attempts to look at old, not rerun cases.

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+LOG = logging.getLogger(__name__)
+
+class FilterHandler(object):
+    """Class to handle persistent variant filters in the mongo adapter"""
+
+    def retrieve_filter(self, filter_id):
+        """Retrieve a known stored filter object from the db
+
+            Arguments:
+                filter_display_id(str)
```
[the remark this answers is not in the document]
dnil: Right, sorry; the docstring is outdate. Me fix!

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+        filter_obj.update([('filters', filter_obj.pop('_id', None))])
+        return filter_obj
+
+    def stash_filter(self, filter_obj, institute_obj, case_obj, user_obj, category='snv'):
+        """Store away filter settings for later use.
+
+            Arguments:
+                filter_obj(MultiDict)
+                institute_obj(str)
+                user_obj(str)
```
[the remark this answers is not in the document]
dnil: Yep, same there. Fixed!

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+                category(str): in ['cancer', 'snv', 'str', 'sv'].
+
+            Returns:
+                filter_id(str)
+        """
+        filter_interface_version = __version__
+
+        LOG.info("Stashing filter for user '%s' and institute %s, version %s",
```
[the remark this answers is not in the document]
dnil: Yes, thanks. That is easier.

scout/adapter/mongo/filter.py
```diff
@@ -0,0 +1,101 @@
...
+                'display_name') + "-" + case_obj.get('display_name')
+
+        for (element, value) in filter_obj.lists():
+            if value == ['']:
+                continue
+            if element in ['save_filter', 'filters', 'csrf_token']:
+                continue
+            if element == 'filter_display_name':
+                # filter display_name if given
+                filter_dict['display_name'] = value[0]
```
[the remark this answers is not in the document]
dnil: MultiDict to list will give an array, but the rest of the code expects a simple string. I'll add a comment. Do you know if there is a function with a transparent name to do that?

scout/adapter/mongo/filter.py
dnil: Thanks; I've updated to link back to the view where the filter was created.

scout/server/blueprints/variants/views.py
dnil: I have a feeling @moonso is itching to deal with that in a little bit, so I'll leave it as a cliff-hanger.

scout/server/blueprints/variants/views.py
dnil: https://twitter.com/wistikent/status/660560676595847168

northwestwitch: Sorry, didn't know it was a citation! :D

dnil: 😸 Freshly, and post-hoc, invented verb!

scout/server/blueprints/variants/views.py
dnil: Hm, I did start that at some point(😅), but ditched it since it gets a little involved with passing the info between different forms. It would be possible I guess, but I see it as rather complex whereas this is straight to the point. It does make the veeeery long variants view longer: I think I'll try breaking it out into the controller instead.

northwestwitch: That's a very good idea. It will make the variants code in views much more readable!

scout/adapter/mongo/filter.py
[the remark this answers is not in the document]
northwestwitch: Fine!

scout/server/blueprints/variants/templates/variants/sv-variants.html
dnil: Sure, that has merit. It is a bit more code in total, but elegant in terms of final simplicity.

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -174,87 +175,15 @@
...
-        </div>
-      </div>
-    </div>
-{% endmacro %}
-
 {% block scripts %}
   {{ super() }}
   <script src="https://cdn.jsdelivr.net/npm/bootstrap-select@1.13.9/dist/js/bootstrap-select.min.js"></script>
   <script>
+    function promptFilterName(suggestion) {
```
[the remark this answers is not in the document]
dnil: I suppose there is quite a bit of javascript magic happening to show that bs4/aria modal, but lets not be picky about the details.. ;-) I will agree that "prompt" function isn't exactly classy.

dnil:
Ok, I think this just about wraps the previous review comments. Ready for more! 😺

scout/adapter/mongo/filter.py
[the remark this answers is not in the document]
northwestwitch: Multiselects work fine!

scout/adapter/mongo/filter.py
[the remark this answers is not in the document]
northwestwitch: Nope, so this way is fine by me!

northwestwitch:
> Ok, I think this just about wraps the previous review comments. Ready for more! 😺

I think you addressed pretty much everything, but I do have a stylistic suggestion, can I push it here?

dnil:
Hm, I like the star, and the grouping of buttons that use the filter select! 😅
Could we just do that without the gap? I suppose that separation wont work on the cancer view, if we don’t make another line for it. There should be more connected things coming in, like marking a filter checked for a case, that would likely cause a  rearrangement of the area again.

dnil:
> Works fine and it's a really useful functionality. The only thing that might cause a bit of trouble is that a user can save different filters with exactly the same name (at least the name displayed in the select). But I guess if you create different filters then you name them differently, right?

The internal ids are unique. I was toying with dates but in the end that would look kind of silly both for something that sticks around and something you just created.
It is relatively painless to rename filters by loading, deleting and saving.

northwestwitch:
> Hm, I like the star, and the grouping of buttons that use the filter select! 😅
> Could we just do that without the gap? I suppose that separation wont work on the cancer view, if we don’t make another line for it. There should be more connected things coming in, like marking a filter checked for a case, that would likely cause a rearrangement of the area again.

What gap? I was thinking to make a new line for the cancer filters:
image

northwestwitch:
Shall I push my suggestion and you fix it?

dnil:
Dont know about fixing, but I can make another suggestion! 😜

dnil:
(But please push, and I can take it from there!)

dnil:
This is perhaps a tad silly; it doesn't really change the function, but nice order/groupings of things in the filters will help us(ers) I suppose.

Anyway, like so:
Screenshot 2019-10-30 at 10 13 43
Screenshot 2019-10-30 at 10 13 33

northwestwitch:
> This is perhaps a tad silly; it doesn't really change the function, but nice order/groupings of things in the filters will help us(ers) I suppose.
> 
> Anyway, like so:
> Screenshot 2019-10-30 at 10 13 43
> Screenshot 2019-10-30 at 10 13 33

I like it! 👍

PR #1434, opened 2019-10-31
Fix #822 - avoid highlighting matching causatives in unaffected.

### LINKED ISSUE #822: Matching causatives from other cases
to discuss at the user meeting:

Should we restrict this to appear for variants in affected cases? in order to avoid to see patogenic variants in unaffected parental samples?

dnil:

**How to test**:
1. With the current master, mark a somewhat common variant causative on stage, say https://scout-stage.scilifelab.se/cust002/F0017223/0501ab658fa97ff11e8522e9d3587ddf
1. view a case where only an unaffected individual carries the variant, say 
https://scout-stage.scilifelab.se/cust002/F0011840/91f873d8f31a7fa5ed37385c777ca1e2
1. note that the variant shows as a "Matching causative" on the case page of the latter case. Screenshot 2019-10-31 at 13 03 40
1. now update to the PR branch on stage
1. and see that the variant no longer displays as a "Matching causative" on the case where only the unaffected parent carries it.Screenshot 2019-10-31 at 13 19 00
1. Finally, check that a case with true matching causative to other cases is still flagged as such, e.g. https://scout-stage.scilifelab.se/cust002/F0015258

**Expected outcome**:

**Review:**


dnil:
Ok, that should be enough. Ready for review!

moonso:
Nice @dnil , test works perfect

PR #1441, opened 2019-11-01
bugfix: Case loading crashes if CLNACC is an integer

bjhall:
We have CLNACC as an integer in our VCFs. Since 59559116901fe189fa7b46e7f7bdeb3bed6e95a6 this causes case loading to crash with this error:

```
  File "/opt/scout/scout/parse/variant/clnsig.py", line 45, in parse_clnsig
    if acc.isdigit():
AttributeError: 'int' object has no attribute 'isdigit'
```

Which makes sense, since isdigit() is a string method. 

This PR fixes this error. :)

- [x] code review dNil
- [x] tests by bjhall

### DIFF
--- a/scout/parse/variant/clnsig.py
+++ b/scout/parse/variant/clnsig.py
@@ -42,7 +42,7 @@ def parse_clnsig(variant, transcripts=None):
         return clnsig_accsessions
 
     # There are some versions where clinvar uses integers to represent terms
-    if acc.isdigit():
+    if isinstance(acc, int) or acc.isdigit():
         revstat_groups = []
         if revstat:
             revstat_groups = [rev.lstrip('_') for rev in revstat.split(',')]


bjhall:
As usual I've forgotten to update CHANGELOG. But I think this falls under "various bugfixes"

dnil:
Thank you - nicely found! :+1: A test and a CHANGELOG entry perhaps? :wink:

dnil:
> As usual I've forgotten to update CHANGELOG. But I think this falls under "various bugfixes"

Hehe, ok!

dnil:
Actually, since you have ints in your VCFs, I'll assume it is tested, right?
It does not break current loading tests, and really just fixes the intention of that statement anyway.

bjhall:
Yes it works fine for our vcfs now. Thanks for merging, you're quick!

PR #1446, opened 2019-11-05
Fix screen resize 191105

mikaell:
This PR fixes a/two bug/s.

1. Screen sizes for Chromograph picture columns
2. x/y pictures disappeared when resizing

Tested on resolutions:
* 2560 x 1440 (Thunderbolt display)
* 2048 x 1152 (Thunderbolt display)
* 1600 x 900 (Thunderbolt display)
* 2560 x 1600 (13" MacBook Pro)

**How to test**:
1. how to install it

`git pull fix-screen_resize-191105`

Test by dragging browser window borders to size.

**Expected outcome**:
Column configuration is changed from 2 to 3 and vice versa on window resize without pushing the left-side controller panel upwards or downwards.

Sex chromosomes don't disappear when resizing.

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil

Screenshot 2019-11-05 at 09 41 55

Screenshot 2019-11-05 at 09 42 53


dnil:
Works, and is better than before, so let's merge it!

Now the transition between two and three columns is smooth, but I still have issues in the interval just below two columns of images, while the other cards still fold fairly neatly.

Screenshot 2019-11-05 at 19 46 59

..and as a side note, on installation I would really suggest ```git fetch; git checkout``` to avoid issues with ```git pull``` into current branch.

mikaell:
> Better and better! Still some issues in the smallest window size range, around when one might consider going to a scrolled single column or hiding the image altogether.

Made hiding images to a separate issue -#1510- assigned it to me.

PR #1453, opened 2019-11-05
Check dismissed variant code when printing report

### LINKED ISSUE #1444: Can not open General report
It is not possible to open General report in case 19200. It seems to work for other cases I have tried. Internal Server Error?

northwestwitch:
Bug fix #1444 

Looking at the files changes it looks like I changed a lot of stuff, but it's not this big deal, I've just added 2 lines of code:

- line 448: {% if reason is number %}
- line 569 {% endif %}

**How to test**:
1. To recreate this bug on a local instance of scout create the following key value in a variant object:
image
1. Using master branch notice that it is not possible to open the general report for the demo case
1. Switch to this branch and make sure that general report is working and the variant is among the dismissed variants in the report. The reason why the variant is dismissed should be empty.

**Expected outcome**:

**Review:**
- [x] code approved by dNil
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Improves loading speed for variant page
 - Problem with updating variant rank when no variants
 - Improved Clinvar submission form
+- General report crashing when dismissed variant has no valid dismiss code
 
 
 ## [4.7.3]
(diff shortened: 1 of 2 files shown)


dnil:
I wonder how it got in there? The "save" for dismiss isn't supposed to do that, is it now?

northwestwitch:
> I wonder how it got in there? The "save" for dismiss isn't supposed to do that, is it now?

I think it's fixed in the bootstrap release, but the current release still suffers from this bug

dnil:
Oki, time doesn't fly always! I thought that was a long time ago.. :+1: Will test after lunch - or feel free to yourself.

northwestwitch:
Or maybe this variant was dismissed like that a quite some time ago, when there was still this bug

northwestwitch:
Testes and it works. I'm going to merge!

PR #1456, opened 2019-11-05
Show total number of available cases in cases search footer

### LINKED ISSUE #1445: Case count not operational
Case count is not accurate, but rather reflects nr of fetched cases (for sorting)?

Screenshot 2019-11-05 at 07 06 58

northwestwitch:
fix #1445

**Behavior before the fix**:
1. From a local instance of Scout load demo sample:
```
scout setup demo
```
1. Load another test sample as well:
```
scout --demo load case scout/demo/643595.config.yaml
```
1. You will now have these 2 samples:
image

1. If you filter for cases with the name: 643594
1. Then in the footer of the search form there will be written **Showing 1 / 1 cases**

**Behavior after the fix**
1. Checkout to this branch and perform the same search.
1. Then in the footer of the search form there will be written **Showing 1 / 2 cases**

**Review:**

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/cases.html
+++ b/scout/server/blueprints/cases/templates/cases/cases.html
@@ -32,7 +32,7 @@
       {{ search_form() }}
     </div>
     <div class="card-footer text-center">
-      Showing {{ limit if limit < found_cases else found_cases }} /  {{ found_cases }} cases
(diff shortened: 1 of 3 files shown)


northwestwitch:
This is also ready to be reviewed!

PR #1457, opened 2019-11-05
Search user email

moonso:
This PR fixes a problem when `_id != email` in the `user_objs`

**Expected outcome**:
Users can now be fetched even if the _id is not the email adress
Check tests in https://github.com/Clinical-Genomics/scout/blob/search-user-email/tests/adapter/mongo/test_user_handling.py

**Review:**
- [x] code approved by CR
- [x] tests executed by @travis, CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - export causative variants to Excel
 - Add support for ROH, WIG and chromosome PNGs in case-view
 
-### fixed
+### Fixed
 - Fixed missing import for variants with comments
 - Instructions on how to build docs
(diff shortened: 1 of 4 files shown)


northwestwitch:
We should be careful with this code. This pull request will cause the LDAP login to fail.
That's because of the code in https://github.com/Clinical-Genomics/scout/blob/master/scout/server/blueprints/login/views.py:
image

the LDAP username is not necessary an email, so this code will fail to find a user:
image
And the LDAP login will fail.

northwestwitch:
Maybe change store.user so it accepts a user _id or email?

northwestwitch:
I'd like to give it a try on stage but I'm sure it will work fine!

northwestwitch:
Works fine on stage, merging now!

PR #1470, opened 2019-11-10
Cancer SV

dnil:
This PR adds a functionality - cancer_sv!

Thank you @szilvajuhos! This is Good Stuff that we should try to get into the base repo. We should add some tests and streamline the views a little bit before merging, but you've done a really Good Thing already!

**How to test**:
1. Load a manta cancer sv track!
2. Note that a cancer sv button appears on the case
3. Look and click through functionality on the variants and variant pages for the cancer_sv to see that all is working as far as can be expected from current annotation status.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
One immediate thing is the addition of a test file, ideally from the previously existing test case, but anything fit for the public domain is fine really. @moonso is this something we have / can produce with small effort? I'm guessing the current bams are not complete enough that I could just fire up manta on and get something meaningful? Otherwise I'm sure @hassanfa or @szilvajuhos have test cases that we could use.

dnil:
Ok, this is why its good that you are now a collaborator. I guess I'll have to recast this as a two step process. On it..

PR #1474, opened 2019-11-11
Select preferably users with _id defined by an email

### LINKED ISSUE #1472: Scout users with more than one account might not see all their institutes
A quick(ish) fix would be to manually copy all the missing institutes from the complete account to the one missing institiutes.

A less quick fix is to modify the adapter.user query to return the user which has most institutes.

The best way to fix this is to go thru all the users in the database, add all institutes to the official user account and remove the extra accounts.

### LINKED ISSUE #1476: assign yourself
Hi!

I tried to assign myself to a case, and while I see that happening in the activity log (for as many times as I clicked on "assign me"), my name does not show up in the main list of cases. 

Kind regards,
Bianca

### LINKED ISSUE #1480: Assignees tag not working
Assigning to cases don't seem to work in the new Scout. We have all assigned ourselves to new cases, but can't see this in the list of cases, the assignees name won't show up (although I can see at the bottom of the page at the activity tracking that I assigned myself to a case).

/Sofie

### LINKED ISSUE #1481: Can't remove comments
In new Scout, we can't remove our own comments created on the case page. Is there a reason for this or a bug? Old comments can still be removed (as long as it is our own comment). 

/Sofie

northwestwitch:
This might (or might not ) fix a number of bugs. I don't have much time to test this right now. If someone wants they could take over!

fix #1472
fix #1481
fix #1480
fix #1476

**How to test**:
1. Test on Scout stage using a user with a double account: one with _id of type ObjectID and one with _id as an email address.
1. Make sure that the user object with most institutes is selected upon login (should be the one with _id = email address)

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
If we can avoid it I would prefer not to go down this road. Its not super clear. I'll give a "winter cleaning"-script a go, and if that runs into problems I'll get back to this.

scout/adapter/mongo/user.py
```diff
@@ -87,8 +87,12 @@ def user(self, email=None, user_id=None):
             query['_id'] = user_id
         else:
             LOG.info("Fetching user %s", email)
-            query['email'] = email
-            
+            query = {
+                '$or' : [
```
dnil: As it stands, this will not change anything, but the idea is right.
We could as for *$and: _id:email, email:email* and fall back to email:email if that doesn't help. I'll update!

dnil:
Like this, right? I'm guessing that was your idea all along, and agree, that would have reverted the status to pre-4.8, and left us more room to fix the events etc manually.

northwestwitch:
> Like this, right? I'm guessing that was your idea all along, and agree, that would have reverted the status to pre-4.8, and left us more room to fix the events etc manually.

Or you could just have _id = email and that's it-

dnil:
> Or you could just have _id = email and that's it

Ah, now that would break tests.. 😜

northwestwitch:
> > Or you could just have _id = email and that's it
> 
> Ah, now that would break tests.. 😜

Oh yes, I didn't think about it

dnil:
What do you say: should I test on stage to perhaps merge this now, or sleep on it? I doubt it makes things worse.. *famous last words*

northwestwitch:
You can sleep on it, i's not that we are deploying tonight, right?

dnil:
Well, Kenny was around just a while back - not impossible. Still, it hardly matters from a cleanup perspective: from the user side I guess its another couple of h of annoyance tomorrow morning, but most things are fine.

scout/adapter/mongo/user.py
```diff
@@ -82,14 +82,25 @@ def user(self, email=None, user_id=None):
             LOG.warning("No key provided to fetch user")
             return None
         query = {}
+        user_obj = None
         if user_id:
             LOG.info("Fetching user %s", user_id)
             query['_id'] = user_id
+            user_obj = self.user_collection.find_one(query)
```
northwestwitch: This query could be done also later in the code, maybe, once you have defined the query dictionary?

dnil: Hm, I don't see a simpler way of doing that, with two queries in the else block? Feel free to edit! 😄

dnil:
Ah, I see; stage install is bugged again. I'll definitely sleep on it then. 'later!

northwestwitch:
Bye!

northwestwitch:
I have an idea that might work: what about matching email = email and sorting by _id type?

According to the official documentation :

When comparing values of different BSON types, MongoDB uses the following comparison order, from lowest to highest:

MinKey (internal type)
Null
Numbers (ints, longs, doubles, decimals)
Symbol, String
Object
Array
BinData
ObjectId
Boolean
Date
Timestamp
Regular Expression
MaxKey (internal type)

https://docs.mongodb.com/manual/reference/method/cursor.sort/

dnil:
I guess we close this for now, and revisit if need be. Thanks for doing it anyway!

PR #1486, opened 2019-11-11
Avoid coverage report opening an extra blank page 

### LINKED ISSUE #1477: Clicking on coverage report opens up 2 tabs
One correct and one empty

northwestwitch:
fix #1477

**How to test**:
1. Clicking on coverage report on the sidebar now opens an extra blank page
1. This change should avoid this
1. To be tested on stage where there are actual coverage reports to open for cases.

**Expected outcome**:

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
+++ b/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
@@ -52,9 +52,9 @@
       </a>
     {% endif %}
     {% if config.SQLALCHEMY_DATABASE_URI %}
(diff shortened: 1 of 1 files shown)


northwestwitch:
Haven't tested this but it should work. Ready for review whenever Scout stage is available.

dnil:
Yes, agreed. Lets test, but I need some 💤 before looking mending that install script / environment again. It does not seem to work out of the box right now. It would be so much simpler if we could just separate the different tools into different environments..

dnil:
Ahem, no, the other way around. This now opens an extra tab and loads the coverage report in the original scout browser window. 😝

northwestwitch:
> Ahem, no, the other way around. This now opens an extra tab and loads the coverage report in the original scout browser window. 😝

Thanks for the fix and the review! :)

PR #1512, opened 2019-11-14
Fix #1487. Actually ignore same case previous evaluations.

### LINKED ISSUE #1487: Previously classified as Pathogenic
Pinned variant in gene XDH in case 19194 now has to symbols for ACMG classification in variant list. One for "Pathogenic" which is blue and one "Previously classified as Pathogenic" which is grey. A bug? We have not seen this variant in any other case.

bild

dnil:
This PR fixes a bug.
P/LP ACMG classifications in the same case would appear also as previous evaluation, due to loop logic.

**How to test**:
1. on master, classify some variants and note that they appear on the variants list both as ACMG-P and previous-P.
1. with this branch, repeat and note that only the current ACMP-P event is shown for the case/variant.
1. pick a reasonably common variant on one case, where you also have shared variants with other cases. Classify it in one of the cases and check that it is still shown as previously classified in the other.

**Expected outcome**:

**Review:**
- [x] tests executed by @dnil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
+- Same case evaluations are no longer shown as gray previous evaluations on the variants page
 - Bug when ordering sanger
 
 ## [4.8.2]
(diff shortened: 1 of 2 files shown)


moonso:
How about a test?

dnil:
Tested on stage:
Screenshot 2019-11-14 at 16 12 56
and same var in anoter case:
Screenshot 2019-11-14 at 16 13 34

moonso:
Could you update CHANGELOG and merge this?

dnil:
Thanks! An automated test would have been nice, but it would have to be a full app one I guess. I am also not quite convinced we have seen the last change to this feature. Next time perhaps.

PR #1520, opened 2019-11-16
Fix #1494: always show prioritized cases - redux

### LINKED ISSUE #1494: Prioritized cases again
Sorry @sofievonlanthen - we have managed to botch the prioritised cases first rule, presumably when re-enabling the case count again. Prioritised cases are currently only shown first, if shown at all.

https://scout-stage.scilifelab.se/cust002?query=&limit=100
Screenshot 2019-11-12 at 20 22 01

https://scout-stage.scilifelab.se/cust002?query=&limit=200
Screenshot 2019-11-12 at 20 22 22

dnil:
This PR adds a functionality / fixes a bug.

**How to test**:
1. prioritise a case you can't find on the limit searched (e.g. n>100, or search for something specific)
1.  ensure that the prioritised case is always visible regardless of sort order selection, case searches etc.

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


dnil:
Ahem, and add tests..

dnil:
Yes, only:
Screenshot 2019-11-19 at 14 06 00

"3.4". We could probably update, but...

northwestwitch:
> Yes, only:
> Screenshot 2019-11-19 at 14 06 00
> 
> "3.4". We could probably update, but...

Oh man! :(

dnil:
This would also solve the order issue for case similarity searches without much extra code. Me like, but, update first. So, lets keep this

northwestwitch:
Sooo.. Which one do we pick so solve #1494, this PR or #1531?? :D 
I suggest we need another reviewer. I suggest @moonso chooses!

dnil:
Reviewer #3! 😜

scout/server/blueprints/cases/controllers.py
```diff
@@ -66,6 +66,23 @@ def cases(store, case_query, limit=100):
         case_obj['display_track'] = TRACKS[case_obj.get('track', 'rare')]
         case_groups[case_obj['status']].append(case_obj)
 
+    if prioritized_cases_query:
+        extra_prioritized = 0
+        for case_obj in prioritized_cases_query:
+            if any(group_obj.get('display_name') == case_obj.get('display_name')
```
northwestwitch: This is because sets don't do well with dictionaries right?

scout/server/blueprints/cases/controllers.py
```diff
@@ -66,6 +66,23 @@ def cases(store, case_query, limit=100):
         case_obj['display_track'] = TRACKS[case_obj.get('track', 'rare')]
         case_groups[case_obj['status']].append(case_obj)
 
+    if prioritized_cases_query:
```
northwestwitch: I might be missing something but why instead of having 2 loops and collecting first all the info for the non-prio cases and then for the prio-case, you don't do 
```list(prioritized_cases_query) + list(case_query)```
and then you loop just once?  This way the prio case will be in the beginning anyway

File: scout/adapter/mongo/case.py [hunk 3]
```diff
@@ -246,6 +246,22 @@ def cases(self, owner=None, collaborator=None, query=None, skip_assigned=False,
         else:
             return self.case_collection.find(query).sort('updated_at', -1)
 
+    def prioritized_cases(self, institute_id=None):
```
northwestwitch: This calls for a nice test, but you already knew it! ;)

scout/server/blueprints/cases/controllers.py
```diff
@@ -34,20 +34,20 @@
...
-def cases(store, case_query, limit=100):
+def cases(store, case_query, prioritized_cases_query=None, limit=100):
     """Preprocess case objects.
 
     Add the necessary information to display the 'cases' view
 
     Args:
         store(adapter.MongoAdapter)
         case_query(pymongo.Cursor)
+        priritized_cases_query(pymongo.Cursor)
```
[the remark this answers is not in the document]
dnil: Thanks! :+1:

dnil:
> I've tested it in this scenario: 2 cases on my local instance, one is prio, limit to return just one,

This is current expected behaviour: There will be the 100 of whatever is asked for, plus any additional prioritised cases. One could change that and subtract the number of found prioritized from the limit, but.. 😅

northwestwitch:
But that's weird, if I limit to return 100 cases why should I see 107?
I have a suggestion, why don't we return 100 and flash a warning that not
all prio cases are currently visualized, so increase the limit to see them
all?

scout/server/blueprints/cases/controllers.py
dnil: The idea is likely to avoid duplicates. We could just use a set, I suppose, but I'm not super certain how that works on general objects. Lets see!

dnil: It will again cost a bit more memory, since the objects will first be instantiated in the set, then appended to their case groups. Maybe the case groups could be sets?

dnil:
> But that's weird, if I limit to return 100 cases why should I see 107?

Because someone marked 7 cases prioritized? 😜

scout/adapter/mongo/case.py
dnil: 😆 I will not reach 100% tonight, but here is a contribution at least!

scout/server/blueprints/cases/controllers.py
dnil: Ok, thanks, that saved me some writing, and might have been it - or just me chickening out on the whole deal of juggling cursors, lists, dicts, enumerate builtin tuple operation and sets - not to mention extra tuples - in a dynamically typed, but type sensitive language. 😆 

I added a local function instead. This may come across as somewhat procedural code. I for the time being choose to consider it functional, somewhat encapsulated and avoiding the main concern with repeating code. Also I'm a little tired. :wink:

northwestwitch:
I'm writing a test for the cases controllers, I'll add it to the PR if it's fine

scout/server/blueprints/cases/controllers.py
```diff
@@ -81,6 +81,7 @@ def populate_case_obj(case_obj):
                 extra_prioritized += 1
                 case_obj = populate_case_obj(case_obj)
                 case_groups[case_obj['status']].append(case_obj)
+        # extra prioritized cases are potentially shown in addition to the case query limit
```
northwestwitch: 👍

dnil:
This has to be one of the better reviewed pieces of code. I bet there are still a bunch of errors in there anyway, but it at least appears to solve the issue. Please do, I'm all for more tests. But I want it duly noted that travis and coveralls are super happy, and that there is an option to open new issues for anything that you realise needs fixing. 😜

northwestwitch:
Ok, I think I understood the test error

dnil:
😆I just walked over to CG to meet up with you..

northwestwitch:
> 😆I just walked over to CG to meet up with you..

Oh no!

northwestwitch:
> > 😆I just walked over to CG to meet up with you..
> 
> Oh no!

You can still sit at my desk, there will be more space today? :)

dnil:
Nice, that was a good test. Ok, so lets go then!

northwestwitch:
> Nice, that was a good test. Ok, so lets go then!

Of course, the cursors reach the end when you see how long they are! Thanks! 🥇

PR #1541, opened 2019-11-21
Version 4.9.0

northwestwitch:
New Scout release 4.9

### Added
- Improved MatchMaker pages, including visible patient contacts email address
- New badges for the github repo
- Links to [GENEMANIA](genemania.org)
- Sort gene panel list on case view.
- More automatic tests

### Fixed
- Fix error when a gene is added to an empty dynamic gene panel
- Fix crash when attempting to add genes on incorrect format to dynamic gene panel
- Manual rank variant tags could be saved in a "Select a tag"-state, a problem in the variants view.
- Same case evaluations are no longer shown as gray previous evaluations on the variants page
- Stay on research pages, even if reset, next first buttons are pressed..
- Overlapping variants will now be visible on variant page again
- Fix missing classification comments and links in evaluations page

**Review:**
- [x] tests executed by CR, MM, DN, Travis
- [x] "merge and deploy" approved by @ingkebil 

This is a **minor version** because it adds a few new functionalities and has serveral bugfixes.
It **doesn't affect production, only the web app** @emiliaol


CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [x.x.x]
 
+### Added
+
```
ingkebil: ;)

northwestwitch: It has always been like this. Maybe it's not super orthodox but this way we always have the new template!

ingkebil: aha alright, I see now that this is not the part on the new version. My apologies

northwestwitch:
No problem! :)

Issue #548, opened 2017-06-07
Classification of somatic variants


bjhall:
In the cancer variant view the variants should be classified according to the four-tiered system recently suggested by Association for Molecular Pathology et al. [1]. 

This is similar to the current "Priority" feature, but the classications should to be "global" for cases of the same indication. So if I classify a variant in a case with the indication "acute myeloid leukemia" the classification should apply for that variant in all cases with the same indication.

This would require an indication/diagnosis/whatever field in the sample document, which as far as I can tell does not presently exist?

[1] http://jmd.amjpathol.org/article/S1525-1578(16)30223-9/fulltext

moonso:
Ok, this is a new way of thinking for us, we need to think a bit on how to solve this.
Would it be possible for you to use a `cohort tag` on case level to group cases in "indication"?
What would be the different indications?

bjhall:
Sure, using the cohort  tag should be fine. Unless you think it is abusing that feature :)

The indications are different cancer types such as "lung cancer", "acute myeloid leukaemia" or "melanoma". I think it should be up to the user how fine-grained they want it to be. 

The reason this is important is that some variants are highly recurrent in cancer, so you wouldn't want to classify them each time. However the classification could be different depending on the indicatin.

moonso:
Ok, it is fine to use the cohort tag. Our idea about that tag is to have it institute specific, guess that would work in your setup as well?

moonso:
Could you also have a look at this @hassanfa ?

moonso:
@hassanfa could you read through this issue and see if you agree? I think we should implement a classification according to the article then store the classification for every variant and display if the variant shows up again in another case just like @bjhall suggests

hassanfa:
There was also a simple classification idea in BALSAMIC, which was inspired from MSK IMPACT: https://github.com/Clinical-Genomics/BALSAMIC/blob/develop/BALSAMIC/config/MSK_impact.json

I'll check out the article here and some other sources and get back to this issue. Lets keep it open.

dnil:
Do we have a full or partial auto-classification for this scheme now as well? Otw the manual version is now ready for review, just a couple of months later - or years, depending on how you view it.. 😅

hassanfa:
We do have a partial one, just some 🐍 script to do that. Ill put it somewhereon GitHub soon ™

hassanfa:
But if someone else a more stable one, please do share it as well. 🤣 My classification code is ancient

[...]

PR #1547, opened 2019-11-23
Add cancer tiers

dnil:
This PR adds a functionality - cancer tier classification. See #548.

**How to test**:
1. with a cancer case, pick a variant and assign a tier.
1. make sure the tier can be seen on the variants page and general report
1. de-assign the tier and make sure it disappears.

**Expected outcome**:

Screenshot 2019-11-23 at 12 20 28

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


dnil:
Fascinating - I 💛 having tests! Not writing them tho. :wink:

dnil:
Ready for review folks!

northwestwitch:
> Ready for review folks!

Nice!! I can take a look at it today after lunch!

scout/adapter/mongo/variant.py
```diff
@@ -535,6 +535,7 @@ def evaluated_variants(self, case_id):
                     '$or': [
                         {'acmg_classification': {'$exists': True}},
                         {'manual_rank': {'$exists': True}},
+                        {'cancer_tier': {'$exists': True}},
```
northwestwitch: 👍

scout/adapter/mongo/variant_events.py
```diff
@@ -559,8 +558,56 @@ def unmark_causative(self, institute, case, user, link, variant):
 
         return updated_case
 
-    def update_manual_rank(self, institute, case, user, link, variant,
-                           manual_rank):
+    def update_cancer_tier(self, institute, case, user, link, variant, cancer_tier):
```
northwestwitch: This function is exactly the same as update_manual_rank. What if we use a unique function for updating manual rank and cancer tier? As the last argument of the function we could pass a tuple maybe. 

- For manual rank:('manual_rank', manual_rank)
- For cancer tier: ('cancer_tier': cancer_tier)

And then element one could be used as event verb and both of them later in find one and update. Just an idea!

scout/constants/variant_tags.py
```diff
@@ -85,6 +85,32 @@
     'high'
 )
 
+CANCER_TIER_OPTIONS = {
```
northwestwitch: Nice!

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -508,3 +508,41 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
     </div>
   </div>
 {% endmacro %}
+
```
northwestwitch: 👍

scout/adapter/mongo/variant_events.py
northwestwitch: What about a test for this one, to increase the coverage?

dnil: Yes, the thought crossed my mind. In event terms it's more like a verb with an object really. But I couldn't decide if it would make things simpler or more complicated in the end; especially I worried that later additions of reading some version of this from file would be complicated by storing it with manual annotation events. It may be the lazy streak speaking, but how about we keep them separate, and have a more overall look at the event structure later? The event list does keep growing in a way that suggests it could use a refactor.

dnil: Yep, absolutely!

northwestwitch: Ok, I agree!

tests/adapter/mongo/test_variant_events.py
```diff
@@ -275,6 +275,80 @@ def test_dismiss_variant(adapter, institute_obj, case_obj, user_obj, variant_obj
     # THEN the variant should be dismissed
     assert updated_variant.get('dismiss_variant') == dismiss_reason
 
+
+def test_update_cancer_tier(adapter, institute_obj, case_obj, user_obj, variant_obj):
```
northwestwitch: 👍

northwestwitch:
I'm adding a couple of test for the variant view, which went also down in coverage. I guess I really don't like that red cross! :)

northwestwitch:
Ready to merge @dnil ?

PR #1549, opened 2019-11-25
support for CRAM alignments

northwestwitch:
This PR will introduce support for visualization of cram alignment files in scout (via igv.js)

**How to test**:
1. I've uploaded CRAM and CRAI files for this case on stage: https://scout-stage.scilifelab.se/cust000/livingox/

image

1. The case above, from master branch on stage, should show alignment buttons on the variant pages but the igv.js page shouldn't be able to load any sample track
1. Check out to this branch and check that alignments work for this sample (check alignments from all types of variants)
1. Make sure that samples with BAM files still work (You can use this old case for instance: https://scout-stage.scilifelab.se/cust002/F0027117)

**Expected outcome**:

**Review:**
- [x] code approved by dNil
- [x] tests executed by dNil


northwestwitch:
Ready for review!

northwestwitch:
Hello?

dnil:
😅No way I’ll review this while I’m working on another PR on the same part of the code! 

😉Hopefully I’ll get to it in the afternoon!

northwestwitch:
> 😅No way I’ll review this while I’m working on another PR on the same part of the code!
> 
> 😉Hopefully I’ll get to it in the afternoon!

Right? We can now officially start the race to who is merging first, because who's working on the other PR has to fix the conflicts! XD

scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
```diff
@@ -61,11 +61,12 @@
...
-                        name: '{{ track.name }}',
-                        url: '{{ url_for("alignviewers.remote_static", file=track.url) }}',
-                        indexURL: '{{ url_for("alignviewers.remote_static", file=track.indexURL) }}',
-                        sourceType: 'file',
+                        name: "{{ track.name }}",
                         type: "alignment",
+                        url: "{{ url_for('alignviewers.remote_static', file=track.url) }}",
+                        indexURL: "{{ url_for('alignviewers.remote_static', file=track.indexURL) }}",
+                        sourceType: "file",
+                        format: "{{ track.format }}",
```
dnil: That one line would have been enough, right? Asking for a friend who will have to merge conflicts.. 😆

scout/server/blueprints/alignviewers/views.py
```diff
@@ -18,7 +18,7 @@ def remote_static():
     file_path = request.args.get('file')
 
     range_header = request.headers.get('Range', None)
-    if not range_header and file_path.endswith('.bam'):
+    if not range_header and (file_path.endswith('.bam') or file_path.endswith('.cram')):
```
dnil: :+1:

scout/server/blueprints/alignviewers/views.py
```diff
@@ -88,15 +88,17 @@ def igv():
...
         # some samples might not have an associated bam file, take care if this
         if bam_files[counter]:
-            sample_tracks.append({ 'name' : sample, 'url' : bam_files[counter],
-                                   'indexURL' : bai_files[counter],
-                                   'height' : 700
-                                   })
+            sample_tracks.append({
+                'name' : sample,
+                'url' : bam_files[counter],
+                'format': bam_files[counter].split(".")[-1], # "bam" or "cram"
```
dnil: Same here.. Though this *is* neater.

scout/server/utils.py
```diff
@@ -134,24 +134,31 @@ def case_append_bam(case_obj):
...
-    return bai_file
+    index_file = None
+    if align_file.endswith('cram'):
+        index_file = align_file.replace('.cram', '.crai')
+        if not os.path.exists(index_file):
+            index_file = "{}.crai".format(align_file)
+    else:
+        index_file = align_file.replace('.bam', '.bai')
+        if not os.path.exists(index_file):
+            index_file = "{}.bai".format(align_file)
```
dnil: ..and that one never gets tested for existence. I know its the same as before really, but perhaps a good time to fix that and warning/error handling?

scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
northwestwitch: As you know well, that's the line that does the magic! ;)

scout/server/utils.py
northwestwitch: I'll try to handle it, thanks!

northwestwitch: But I'm not flashing messages from here, because this is called in a loop, for instance from the str variants page. I'll handle it from the view instead

northwestwitch: @dnil see if you like this way of handling missing files, otherwise I'm open to suggestions!

dnil: Ehr, am I reading this backwards, or isn't this is basically the same. There is a test in the view now, but index_file will always be set regardless `if os.path.exists(that_fallback_index_path), so the test will still pass even if an index here or there is missing?

dnil:
Anyway, tested to work just fine on stage!

northwestwitch:
Ah OK, now I understood, I'll fix the test as well!

dnil:
How's it going here? Ready to check again?

northwestwitch:
Oh yes, I forgot ti remind you, thanks!

scout/server/utils.py
dnil: The loophole is still there. Its not super important, but...

dnil: Well, I give in. It is working, I don't really think we promise to test user given paths anyway, and the test that is in place is better than none. As you say, worst case appears to be that igv.js fails/reports error.

northwestwitch: Thanks for the review!

PR #1554, opened 2019-11-27
debugging issue with validation bars on the case statistics tab.

mayabrandi:
Bugfix - fixing the fact that validation bars dont show up when all institutes are selected. Found two issues. 

1) Verification query doesn't include sanger validations.  
2) sliced_case_ids is empty when all institutes are selected.

**How to test**:
1. install on stage and go to dashboard view. 

**Expected outcome**:
Select all cases and see that validation bars show up. Try to add a sanger validation to a variant on a case with no previous validations. See that the **Validation ordered Cases** increase. Set the variant to True or False positive and see that the **Validated cases (TP + FP) Cases** increases.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


dnil:
The sliced_case_ids should contain a list of cases that matched the case query - a user can ask for stats on e.g. only a particular gene panel, cohort or phenotype.

The whole stats on verification thing was a little tricky, since not all Institutes really use the "order Sanger" feature, but just mark variants as verified or not.

northwestwitch:
I've never seen this option!! 
image
Weird! :D

mayabrandi:
> The sliced_case_ids should contain a list of cases that matched the case query - a user can ask for stats on e.g. only a particular gene panel, cohort or phenotype.
> 
And the customer should be able to ask such a query combined with selecting all institutes?

> The whole stats on verification thing was a little tricky, since not all Institutes really use the "order Sanger" feature, but just mark variants as verified or not.
> 
Think I need to discuss this in person.. I don’t see the problem really.

northwestwitch:
> Think I need to discuss this in person.. I don’t see the problem really.

I'll checkout to your branch and try to suggest some actions!

northwestwitch:
A good starting point would be to fix the problem when you have:
- one case (the demo one is good enough to test this)
- one sanger-ordered variant (snv) or one verification ordered variant (SV)
- one if them is validated (positive or negative, doesn't matter)

Then if you look at the stats, second tab you can see them if the institute is cust000:
image

But they disappear if institute select is 'All institutes':

image

northwestwitch:
List of all eventualities we have to cover in the fix:
- [ ] Sometimes users order Sanger and then they cancel it (could count as sanger ordered I guess!)
- [ ] Sometimes users validate a variant then they change their mind and reset its validation (should not count as validated)
- [ ] Verification orders (sanger) and validation should work when one institute is selected
- [ ] Verification orders (sanger) and validation should work when all institutes are selected
- [ ] All the above options should also work when other filters are specified (the input field on the left)

Did I miss anything?

northwestwitch:
I've pushed here a fix for #1560

mayabrandi:
Ill look into it. Had to work with other things in cg today.

northwestwitch:
> Ill look into it. Had to work with other things in cg today.

Point 2 works. I was checking variants but the stats should return the number of validated cases, and that is correct: case is not taken into account when its validated variants have validation resetted.

mayabrandi:
Did so now. Looks ok?

PR #1555, opened 2019-11-27
Find cases solved within timespan

adrosenbaum:
Query cases through CLI that have had variants marked as causative within a given
number of days.

This PR adds a functionality.

**How to test**:
1. continuous-integration in travis

**Expected outcome**:
Travis build should pass

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
Ensembl Rest API is down and we have some tests using it. Skipping those tests for the time being!

adrosenbaum:
Thanks @northwestwitch! One way to generalise this would simply be to let the user specify a timespan option AND a verb for the case-events. Then look for all cases that have had a case-event with that verb in the given timespan. What do you think of this?

northwestwitch:
> Thanks @northwestwitch! One way to generalise this would simply be to let the user specify a timespan option AND a verb for the case-events. Then look for all cases that have had a case-event with that verb in the given timespan. What do you think of this?

I think it would be a bit cumbersome for the user to remember exactly which verb is associated to each preference ( reruns, finished, causatives, research-requested, is-research, status)
What about assigning the verb according to the params provided to the case function?
image

northwestwitch:
> > Thanks @northwestwitch! One way to generalise this would simply be to let the user specify a timespan option AND a verb for the case-events. Then look for all cases that have had a case-event with that verb in the given timespan. What do you think of this?
> 
> I think it would be a bit cumbersome for the user to remember exactly which verb is associated to each preference ( reruns, finished, causatives, research-requested, is-research, status)
> What about assigning the verb according to the params provided in the cases function?
> image

How do we handle the possibility that they might ask for research requested cases with causatives in the last 100 days ??

adrosenbaum:
@northwestwitch tried to generalise a bit more. What do you think?

scout/adapter/mongo/case.py
```diff
@@ -237,6 +238,30 @@ def cases(self, owner=None, collaborator=None, query=None, skip_assigned=False,
                     {'individuals.display_name': {'$regex': name_query}},
                 ]
 
+        if within_days:
+            verbs = []
+            if finished or has_causatives:
```
northwestwitch: finished means archived or solved, but it is possible to archive cases without solving them (without having a causative) , you should use the "archive" key in this case

tests/adapter/mongo/test_case_handling.py
```diff
@@ -601,3 +601,19 @@ def test_get_cases_cohort(real_adapter, case_obj, user_obj):
     result = adapter.cases(name_query="cohort:{}".format(cohort_name))
     # THEN we should get the case returned
     assert sum(1 for i in result) == 1
+
+def test_get_cases_solved_since(real_adapter, case_obj, user_obj, institute_obj, variant_obj):
```
northwestwitch: I know it would be a bit of a pain, but could you include a small test for each of the options usable in the query that can be combined with the time frame?

scout/adapter/mongo/case.py
```diff
@@ -237,6 +238,34 @@ def cases(self, owner=None, collaborator=None, query=None, skip_assigned=False,
...
+
+            days_datetime = datetime.datetime.now() - datetime.timedelta(days=within_days)
+            # Look up 'mark_causative' events added since specified number days ago
+            event_query = {
+                'category': 'case',
+                'verb': {'$in': verbs},
+                'created_at': {'$gte': days_datetime}
+                }
+            recent_events = self.event_collection.find(event_query)
+            solved_cases = set()
```
northwestwitch: Maybe better to rename this to something else, since it's no more just the solved ones?

adrosenbaum:
No events with verb 'archive' seems to be created when a case is archived, although there exists a method 'adapter.archive_case' that does this...

dnil:
Hm, I thought it just made a 'status' event, and updated status on the case to 'archived'?

northwestwitch:
@adrosenbaum is right, It must be a bug: look at the left case view on the left (should be triggered by archive on the sidebar) and on the right is where the right event is created. 

image

I wonder why it's not used and since when. We should fix it!

northwestwitch:
Should be fixed now, tested it with the export case with days and it's working.

northwestwitch:
It's not passing the ensembl test, their services should still be down! :-/

Issue #791, opened 2018-04-24
Add ranking ranges to Rank score results


northwestwitch:
Suggestion from @KickiLagerstedt: 
At the variant level, Rank score results table, it would be better to know where the tank score numbers lay. For instance: Splicing: 3 (out of?), Consequence: 5 (range?). it would help to understand the weight that the scoring has on the ranking sum.

moonso:
Sure this is possible but then we need a way to transfer the rank models into scout so I would not label it `Easy`

dnil:
I think we are about to get/partly have that functionality; see  #494?

moonso:
This will be solved when the ranking module gets tightly coupled to Scout. I'm slowly working on it.

dnil:
Well done! Perhaps change the range dash to e.g. an ellipsis ... or something like < x < to make the -12 - 4 a bit easier to parse? 😊

dnil:
Yes, like on a row below/above the values perhaps? I like the fat vs plain.

fc-io:
![image](https://user-images.githubusercontent.com/757478/149571534-8b7ede5c-7be3-48ba-af19-a9396d0b1ea3.png)

🤔

fc-io:
any favorites? or new suggestions?

henrikstranneheim:
I vote for the rank strikes back

PR #1561, opened 2019-11-29
Rank pie chart

mayabrandi:
This PR adds a rank score pie chart to the variant view

**How to test**:
1. install on stage and go to a variant. 

**Expected outcome**:
There should be a pie chart under the summary, showing some details about the rank score

**Review:**
- [ ] code approved by
- [ ] tests executed by


mayabrandi:
I directly noticed that Clinical STRs are using the same html, resulting in an empty chart there. Ill look into how to handle that @moonso

moonso:
Nice Maya! One thing that came to my mind: what will happen if there is no rankscore result on the variant?

moonso:
Haha ok you are already on it.

mayabrandi:
That is not taken care of. Ill add a exeption for that

Skickat från min iPhone

mayabrandi:
Ready for review @moonso

mayabrandi:
@northwestwitch See what you think now!

northwestwitch:
Nice! I'll check as soon as I'm at the office!

northwestwitch:
Testing the PR, unfortunately I don't see the chart under the variant summary. There are some javascript errors:
image

mayabrandi:
Sry. I missed to push the views.py

mayabrandi:
@dnil @moonso @northwestwitch I pushed a commit with both pie and bar plot. Can you look and select which you prefer. (Ignore the weird borders in the plots. They should not be there )

moonso:
@mayabrandi Could you post screenshots as comments here? And maybe add a comment what you think

dnil:
I'm guessing this is a prototype, but just to be clear we probably need a full OSS solution? This appears semi-commercial, and rather expensive at that?
Screenshot 2019-12-03 at 13 40 45

mayabrandi:
I don’t really have an opinion.

mayabrandi:
Valtteri already approved on using Highcharts in Vogue. I tried some different packages and talked to Phill about it and he agreed on that Highcharts is a good choice. He is also using it in multiQC and megaQC if I remember correctly. Should be okej.

hassanfa:
Vega/vega-lit/Altair-viz offers a similar solution (and similar grammar), and it is free for commercial usage: https://github.com/vega/vega . It is not pretty to work with though.

mayabrandi:
Skärmavbild 2019-12-03 kl  13 44 35

dnil:
Neato! I much prefer the bar version. It is as good as a table. :wink: I'm also curious to see it with ranges as well - there it might be even better!

How does one go about getting a highcharts version that can be used locally, and not complicate things for other sites? I can't imagine multiQC etc shipping with a little ad from that company in it?

vwirta:
When we progress in use of highcharts we can look into the licensing. Non-commercial license is ok for universities, but we can still look into if a commercial developer’s license gives added value.

northwestwitch:
Vega is good. Otherwise I used Chart.js to create the charts present in the dashboard. @mayabrandi I don't think we should buy any library here. You are using a demo?

dnil:
@vwirta From their ad, I don't quite see how we could avoid a license, but perhaps there is more to it behind the scenes? 
Our research is formally governmental, and the clinical side is commercial, arguably non-profit, but that is usually a complicated area.

vwirta:
It is complicated… universities were explicitly mentioned as ok for non-commercial somewhere in the faq
But the clinic side is of course different…  let’s see where we land regarding choice of tools and then discuss strategy for licensing

dnil:
Ok. This is so far the only place its used. It would seem easier just to use another library or draw ourselves?

mayabrandi:
@dnil Absolutely. Not a problem. ill switch to something else

dnil:
Interesting! You are absolutely right Phil is using this, but you have to drill down all the way to the actual js files to find the highcharts license in the MultiQC package. Upfront it sells with https://github.com/ewels/MultiQC/blob/master/LICENSE which to this lay person appears legally incompatible with https://github.com/highcharts/highcharts and its www.highcharts.com/license?

dnil:
> @dnil Absolutely. Not a problem. ill switch to something else

Thanks @mayabrandi - sorry to be a bother.. :sweat:

dnil:
For the record we had a chat with Phil, who recommended not using HighCharts for new projects, but who also was not in any immediate rush to remove it from MultiQC, and when in the distant future this happens plans to move to plotly (MIT license).

mayabrandi:
Thanks for digging into this Daniel.

northwestwitch:
Does anybody mind if I take over this one?

dnil:
Check with @mayabrandi - but note that it is effectively blocked by #791, even if a different drawing package is used?

northwestwitch:
> Check with @mayabrandi - but note that it is effectively blocked by #791, even if a different drawing package is used?

Ok, let's wait then!

northwestwitch:
This PR is almost a year old and blocked by #791. I suggest we close and reopen when it can be addressed properly. Anyone disagrees?

mayabrandi:
Sure!

PR #1582, opened 2019-12-10
Dockerize

ninanorgren:
This PR adds dockerfile for building a simple docker image and running it as a container

**How to test**:
1. Set up the docker-compose file
2. Use the Makefile to build the images and run the container

**Expected outcome**:
Scout should be running in a container

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Good job @ninanorgren 👍 
We'll try to review ASAP!

adrosenbaum:
Very cool @ninanorgren, I'll have a look at this PR this week 😄

docker/Makefile
```diff
@@ -0,0 +1,18 @@
+build:
+	@docker build -t scout:latest ../
+	@docker build -t scout-setup:v1.0 scout-setup/
+	@docker build -t scout-import:v1.0 scout-import/
```
adrosenbaum: Is it necessary to have one container for each possible CLI command? Maybe all CLI stuff could be done using a single image, it seems the only thing that differs between them is the entrypoint.

docker/case-handler.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+
+parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(description = "Helper script to add or remove cases from database")
+group  = parser.add_mutually_exclusive_group(required=True)
+group.add_argument('--add',         action='store_true',   help ="Add samples to db")
+group.add_argument('--remove',      action='store_true',   help ="Removes samples from db")
+parser.add_argument("--sampleid",   metavar = "SAMPLEID",  type = str,   required = False,    help = "Sample ID for sample to remove (defined under 'family' in yaml file")
+parser.add_argument("--instid",     metavar = "INSTID",    type = str,   required = False,    help = "Institute ID for sample to remove")
+parser.add_argument("--folder",     metavar = "FOLDER",    type = str,   required = False,    help = "Folder where the sample that should be imported is stored")
```
adrosenbaum: so the --folder option decides what dir on the host that should be bound to /sample in the container, and --yaml is the path to the .yaml file relative to /sample. 

Here the tricky part is that some of the paths (e.g. the bam files) specified in the yaml-file should be visible to the container where the scout server is run. Also all paths given in the yaml-file must be relative to the /sample dir in the container. right?

docker/docker-compose.EXAMPLE.yml
```diff
@@ -0,0 +1,81 @@
+version: "3"
+services:
+  scout:
+    image: scout:latest
+    volumes:
+      - /home/scout/vault:/scout/vault:ro
```
adrosenbaum: I think there should be an additional volume here, binding the sample data to the container
e.g. ```- /home/scout/sampledata:/scout/sampledata:ro``` otherwise this container will not be able to access any of the files. Maybe this is why you had problems with the bam-files?

docker/user-handler.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
```
adrosenbaum: I agree there should be some wrapper scripts to interact with the containers, however would it be possible to generalise it to a single script? I'm thinking something like

```
#!/bin/bash
docker-compose run <scout_cli_container> "$@"
```

Once again, maybe it would be enough with one container managing all the CLI stuff, and its entrypoint could be something like 'scout -h $MONGOHOST -p $MONGOPORT'

requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Everything needed in production
 
+gunicorn==19.9.0
```
adrosenbaum: I think it would be better to move this installation to the Dockerfile instead

docker/Makefile
ninanorgren: It can definitely be done in a single image, just that image will be very complex. The more commands you add the more complex the image will be. But I agree, this setup might not be optimal either. If you have a good solution please feel free to change

docker/case-handler.py
ninanorgren: Yes for the first part. The comment regarding the bamfile path is exactly the problem I have also. There is something with the path of the bam files that doesn't work in the container. It can see the path to the vcf file, but it cannot find the bam files. I haven't yet had time to look more into this and how to fix it.

docker/docker-compose.EXAMPLE.yml
ninanorgren: the mounting of the sampledata folder is actually done inside the scout-import image, which builds on the scout image, and connects to the mongo-db image. I see now that I actually mounted this already in the docker-compose file, and then I mount it again in the case-handler script, I should only do it in the case-handler actually. The idea was that you could have your samples in any folder and just mount that specific folder using the case-handler script. Otherwise you'd have to hardcode the path to the folder where all the data is in, and also there's no point in mounting the entire data folder every time.

docker/user-handler.py
ninanorgren: I can think a bit about how to do this in a good way without making it too complicated. But I agree, it would be nice to make the user interaction easier.

requirements.txt
ninanorgren: Sure, I'll fix that

docker/docker-compose.EXAMPLE.yml
adrosenbaum: Yes, it works fine using the scout-import image to upload cases to the database if the files are mounted properly. However, let's say a user that connects to the server in the 'scout'-container, wants to have a nice igv-view around a variant, there will be no mount to the bam-file from that container. The mounts will only exist in the 'scout-import' container while the case is imported, not in the container where the server is running. This I think is the problem.

docker/Makefile
northwestwitch: Hi @ninanorgren I'm not really fond of this solution with different containers for different CLI commands. I'll to write a simpler Dockerfile for handling both the CLI and the server side of Scout, with suggestions on how to connect it to other services, like mongodb and nginx.

northwestwitch:
I'm going to close this PR since we have opted for a more compact Docker image, see #2173. The new Docker is super basic (it installs only the software and required libs on a lightweight linux distro). Then the idea is to use the same image from the cli and the web server just by invoking different cli commands.

Thanks for this @ninanorgren!

Issue #1578, opened 2019-12-05
Show arbitrary data from VCF on variant view


bjhall:
A feature request that I can implement myself, but I'd like your feedback first.

The most common request I get from our users is to add some random information to the variant view (for example if the variant exists in some database or overlaps something etc). At the moment it requires that I add parsing/loading, modify variant.html and deploy a new version of scout. And if it is not relevant for others I don't want to push it upstream, so I have to maintain a local branch with all our modifications.

Wouldn't it be nice to be able to add custom fields in the vcf that are loaded and displayed in a simple table om the variant page.

For example an INFO-field like this:

```
SCOUTCUSTOM=key1|value1,key2|value2
```

Gets parsed and loaded into the variant document like this:

```
custom: {
key1: 'value1',
key2: 'value2'
}
```

And then of course display the keys and values in a table in variant.html. For simplicity all values could be treated as strings.

Any comments? Is it a bad idea? It would simplify my job significantly...

dnil:
We have indeed discussed that before, last time partially at the beginning of this semester. But we have so far decided against it. I was initially very keen on having something like that too, or eventually a full blown definition yaml specifying whet values should be loaded how and shown where on the layout. The main argument against is probably that: in order to specify where to display the info above you have to write a fairly complex definition, which is not so dissimilar to just typing it into the code. 

Another argument is also something you touch upon: in general we find it is useful for all to have features added and live through some design conflicts to end up with something that is sharper for all. In the long term that saves effort for all. We always welcome PRs, so go ahead if you have a clear idea of how to show this, but I think we would benefit more from additional specific PRa ones rather than the general one!

bjhall:
Thanks for your comments! I mostly agree, and I think most things should be added "properly". But for more obscure things I think somthing like this could be useful. I'm planning on keeping it very simple.

I will make an attempt at implementing this, since I think I will need it anyway. I have very limited time for Scout developement and this would save me a lot of time.

I'll make a pull request when I'm done, which you are welcome to decline if you want to! :)

moonso:
Closed by https://github.com/Clinical-Genomics/scout/pull/1584

PR #1584, opened 2019-12-10
Custom annotation

bjhall:
This PR adds support for custom annotations from VCF, that do not need to be defined in Scout. Solves: https://github.com/Clinical-Genomics/scout/issues/1578

The data needs to be defined in the VCF in an INFO field called SCOUT_CUSTOM. It takes key-value pairs (both are always treated as strings). Pairs are separated with commas and the key and values are separated by pipes. Example:

```SCOUT_CUSTOM=Some key|Some value,Another key|Another value```

**How to test**:
1. Create a VCF which has the SCOUT_CUSTOM field above in at least one variant
1. Load the case 

**Expected outcome**:
The custom annotations should be shown in the variant overview.

Both for SNVs:
image

And SVs:
image

The table is only displayed if there are custom annotations available, so it doesn't change anything if the feature is not used. If you are at all interesting in merging this I can add tests as well. I'd also be happy to change anything regarding the implementation. If you do not want to merge it I'll just keep it in our fork, and I won't be offended :)


dnil:
I don't see why we couldn't keep this in here! I would prefer PRs on specific items, but why not! Anyone against - @moonso @northwestwitch?

Tests and a CHANGELOG update would be lovely. 😸

bjhall:
The reason I don't always want to make PRs for specific items is that we are getting fairly obscure requests from our users, such as: "Show if an SV is overlapping anything from this collections of variants that I've been manually curating in an excel file for the past 10 years". Since I don't see this ever being of interest to anyone else I don't want to litter the parsing code of Scout with such things.

I will still make specific PRs for things that are of potential interest to anyone else!

I'll update the PR with tests and CHANGELOG as soon as I find the time!

northwestwitch:
> The reason I don't always want to make PRs for specific items is that we are getting fairly obscure requests from our users, such as: "Show if an SV is overlapping anything from this collections of variants that I've been manually curating in an excel file for the past 10 years". Since I don't see this ever being of interest to anyone else I don't want to litter the parsing code of Scout with such things.
> 
> I will still make specific PRs for things that are of potential interest to anyone else!
> 
> I'll update the PR with tests and CHANGELOG as soon as I find the time!

It's OK, we get also a lot of custom request so we do understand! :)
I'm not against this PR, but I agree with Daniel: it would be nice to have tests and a changelog update.

moonso:
Sounds like a nice feature

bjhall:
@northwestwitch Tests and CHANGELOG entry added! Hope it's fine :)

northwestwitch:
@moonso you approved, it this ready to merge?

cogrho:
Hello,

Are custom annotations also available for cancer variants? Thanks!

northwestwitch:
> Hello,
> 
> Are custom annotations also available for cancer variants? Thanks!

Hello! I'm quite positive that they are parsed, but I'm not 100% certain that they are shown on cancer variants pages. Are you asking before you don't see your custom annotations?

cogrho:
Hi, @northwestwitch! 
Thank you for the quick response! I attempted to add the SCOUT_CUSTOM INFO tag to some somatic variants and then loaded the VCF file into Scout. However, it seems like the custom annotations are not being displayed on the cancer variants page.

northwestwitch:
> Hi, @northwestwitch!
> Thank you for the quick response! I attempted to add the SCOUT_CUSTOM INFO tag to some somatic variants and then loaded the VCF file into Scout. However, it seems like the custom annotations are not being displayed on the cancer variants page.

I see, what probably happened is that when we separated the template page for cancer variants from the general variant template we didn't include the code for displaying the custom annotations. We'll fix it, thanks for reporting the issue!

PR #1593, opened 2019-12-16
Include partial causatives in general report

### LINKED ISSUE #1587: Include partial causative variants in general report
They're not shown at the moment

northwestwitch:
fix #1587

**How to test**:
1. Mark one or more variants (SNVs, SVs) as partial causatives
1. Go to general report and check they're shown on it (also relative HPO and OMIM terms)
1. Make sure that also the PDF report is correct.

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -236,6 +236,7 @@ def case_report_content(store, institute_obj, case_obj):
     """
     variant_types = {
         'causatives_detailed': 'causatives',
+        'partial_causatives_detailed': 'partial_causatives',
         'suspects_detailed': 'suspects',
         'classified_detailed': 'acmg_classification',
         'tagged_detailed': 'manual_rank',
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/cases/controllers.py
```diff
@@ -263,16 +264,17 @@ def case_report_content(store, institute_obj, case_obj):
     data['report_created_at'] = datetime.datetime.now().strftime("%Y-%m-%d %H:%M")
 
     evaluated_variants = {vt:[] for vt in variant_types}
-    # We collect all causatives and suspected variants
+    # We collect all causatives (including the partial ones) and suspected variants
     # These are handeled in separate since they are on case level
-    for var_type in ['causatives', 'suspects']:
+    for var_type in ['causatives', 'suspects', 'partial_causatives']:
```
dnil: :+1:

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -857,240 +868,24 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 </div>
 {% endmacro %}
 
-{% macro sv_variant_content(variant,index) %} <!--The entire description of a structural variant, dynamic contentc -->
```
dnil: Eh, what happened there?!

northwestwitch: If you open master you'll see that there are 2 sv_variant_content macros on the same page. Of course only the last one was used in the end. I must have forgotten one when I touched this code last!

dnil: Ok! Looks good when looking at the whole file, but really weird on the diff! :blush:

northwestwitch:
Tested on stage and it works, both the HTML code and the PDF:

image

northwestwitch:
One thing that is actually not rendering well on the PDF version is the Font Awesome icons (above is the HTML page, below the PDF):

image

I'm going to convert them to text instead!

northwestwitch:
Now it's better:

image

PR #1603, opened 2020-01-07
Allow users to customize institute settings

### LINKED ISSUE #988: Create a view for institutes
This would simplify updates to institutes, e.g.

- [ ] Update institutes to share cases with
- [ ] Update phenotype groups
- [ ] Set rank score tresholds

northwestwitch:
fix #988 

This PR introduces a page where an institute might be customized at this level:
- Display name
- Sanger recipients list
- Coverage cutoff
- Frequency cutoff
- Phenotype groups
- Patients cohorts
- Collaborating institutes

**How to test**:
1. Probably a good idea to test this on clinical-db stage.
1. Install the branch and go to "Institutes" page. Note that an edit button is present for each institute the user has permission to change (user shouldn't be able to modify institutes he or she is not allowed to edit)
1. Edit institute-related fields

**Expected outcome**:
The functionality should be working and changes should be committed to database.

**Review:**
- [x] code approved by DN


northwestwitch:
So far I've managed to create this interface:

image

All fields are updatable so far, except:
- **patient cohorts**
- **SNVs_rank_threshold**
- **SVs_rank_threshold**

Questions: 

1. We now import **patient cohorts** from a file in the constants, should I remove them from there and allow every institute to create its own constants from the above interface. Should these be saved at the institute level in the database? I think the answer is yes, right?

1. Regarding SNVs_rank_threshold and SVs_rank_threshold, should they also be saved  be saved at the institute level in the database?

dnil:
1. I agree: one might use the constants or a yaml file to seed a new instance - eg local/develop or new site, but then it should go into the db.

2. My feeling is that those thresholds depend most of all on the application at hand - e.g. a small panel or deep whole genome would have different defaults. In practice I suppose only a handful of institutes run more than one application. Per institute is better than all having the same default.

Perhaps show them there, but don't allow the users to update them at will? I imagine one could argue this would mess with validation etc? I would rather that they were set after careful evaluation of the rank model - ie mutacc etc.

northwestwitch:
> Perhaps show them there, but don't allow the users to update them at will? I imagine one could argue this would mess with validation etc? I would rather that they were set after careful evaluation of the rank model - ie mutacc etc.

Shall I keep them out of this PR then? Because if users are not allowed to update them then I don't know where I should collect these numbers from ..

dnil:
Yes, I would say it is more of a backend thing. You are right: if one can't change it fills little purpose there. A user would then be more interested in displaying that on a case by case basis. Leave it out - if someone else has a good use case, we can add it in another PR!

northwestwitch:
This is ready for review now

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,148 @@
...
+    $('#cohort_tags').select2({
+    	tags: true,
+      data: ["endo","mito","ketogenic diet","pedhep","other"],
+      tokenSeparators: [','],
+      placeholder: "Add cohort tags",
+    });
+
+    $('#sanger_tags').select2({
+      tags: true,
+      tokenSeparators: [','],
```
dnil: :+1:

scout/server/blueprints/institutes/forms.py
```diff
@@ -0,0 +1,29 @@
...
+    for key in PHENOTYPE_GROUPS.keys():
+        option_name = ' '.join([key, ',', PHENOTYPE_GROUPS[key]['name'], '(', PHENOTYPE_GROUPS[key]['abbr'] ,')'])
+        hpo_tuples.append((option_name,option_name))
+
+    display_name = TextField('Institute display name', validators=[validators.InputRequired(),
+        validators.Length(min=2, max=100)])
+    sanger_emails = NonValidatingSelectMultipleField('Sanger recipients', validators=[validators.Optional()])
+    coverage_cutoff = IntegerField('Coverage cutoff', validators=[validators.Optional(),
+        validators.NumberRange(min=1)])
+    frequency_cutoff = DecimalField('Frequency cutoff', validators=[validators.Optional(),
```
dnil: :+1:

scout/adapter/mongo/institute.py
```diff
@@ -71,6 +77,15 @@ def update_institute(self, internal_id, sanger_recipient=None, coverage_cutoff=N
                      internal_id, sanger_recipient))
             updates['$push'] = {'sanger_recipients':sanger_recipient}
 
+        if sanger_recipients is not None:
+            for recipient in sanger_recipients:
+                user_obj = self.user(email=recipient)
+                if not user_obj:
+                    return "user {} does not exist in database".format(recipient)
```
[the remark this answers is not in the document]
northwestwitch: The current CLI code doesn't allow to add sanger recipients which are not database users. Of course you can add any user you want if you enter and edit directly the database field..

northwestwitch:
> As said before, really nice! A couple of requests:
> 
> * Tooltips for the fields, at the very least for the threshold values to avoid confusion.
> * Open also custom pheno groups for editing, just as for the cohort tags. At least I think we said so?

Good suggestions, working on including both!

scout/adapter/mongo/institute.py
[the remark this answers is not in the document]
dnil: :+1: Ok! Good, then this is consistent, but again perhaps a tooltip to indicate that only known user emails are expected?

northwestwitch:
> * Open also custom pheno groups for editing, just as for the cohort tags. At least I think we said

northwestwitch:
@dnil I've modified this PR according to your suggestions. I'm not really certain the tooltips have the best explanation ever, just take a look!

northwestwitch:
> @dnil I've modified this PR according to your suggestions. I'm not really certain the tooltips have the best explanation ever, just take a look!

I'll need to fix a test first!

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+              {{ form.display_name.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Name of the institite which will be displayed on Scout pages") }}
+              {{ form.display_name(class='form-control', value=institute.display_name) }}
+              {% for error in form.display_name.errors %}
+                <span style="color: red;">{{ error }}</span><br>
+              {% endfor %}
+            </div>
+          </div>
+          <div class="form-row mt-3 mb-3">
+            <div class="col-md-2">
+              {{form.sanger_emails.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Email addresses to send variant verification email to")}}
```
dnil: ...only accepts current Scout user emails: exceptions to this can be made by a db admin. Please ask your support!

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+                {% endfor %}
+              </select>
+              {% for error in form.sanger_emails.errors %}
+                <span style="color: red;">{{ error }}</span><br>
+              {% endfor %}
+            </div>
+          </div>
+          <div class="form-row mt-3 d-flex justify-content-between">
+            <div class="form-group col-md-5">
+              {{ form.coverage_cutoff.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Read depth considered reliable for displaying variants") }}
```
dnil: ...used in coverage reports.

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+              {{ form.coverage_cutoff(class='form-control', value=institute.coverage_cutoff) }}
+              {% for error in form.coverage_cutoff.errors %}
+                <span style="color: red;">{{ error }}</span><br>
+              {% endfor %}
+            </div>
+            <div class="form-group col-md-5">
+              {{ form.frequency_cutoff.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Lower frequency threshold to display variants") }}
```
dnil: NOTE : used as the institutet default in clinical filter.

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+              {{ form.frequency_cutoff(class='form-control', value=institute.frequency_cutoff) }}
+              {% for error in form.frequency_cutoff.errors %}
+                <span style="color: red;">{{ error }}</span><br>
+              {% endfor %}
+            </div>
+          </div>
+          <div class="form-row mt-3 d-flex justify-content-between">
+            <div class="form-group col-md-7">
+              {{ form.pheno_group.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Add a new phenotype group item the the list of phenotype shortcuts below") }}
```
dnil: the the -> to the?

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+              <input name="hpo_term" class="typeahead_hpo form-control" id="pheno_group" data-provide="typeahead" autocomplete="off" placeholder="Search HPO..">
+            </div>
+            <div class="d-block col-md-3">
+              {{ form.pheno_abbrev.label(class="control-label",data_toggle="tooltip", data_placement="top", title="A short name for this phenotype group") }}
+              <input name="pheno_abbrev" class="form-control" name="pheno_abbrev" id="pheno_abbrev" placeholder="ex: NEUR" pattern=".{2,10}">
```
dnil: :+1:

dnil: Oh, but perhaps "e.g."!

scout/server/blueprints/institutes/templates/overview/institute.html
```diff
@@ -0,0 +1,201 @@
...
+              <select class="form-control" id="cohort_tags" name="cohorts" multiple="true" style="width:100%;">
+                {% for cohort in institute.cohorts %}
+                  <option value="{{ cohort }}" selected>{{ cohort }}</option>
+                {% endfor %}
+              </select>
+            </div>
+          </div>
+          <div class="form-row mt-3 d-flex justify-content-center">
+            <div class="col-md-12 align-self-center">
+              {{ form.institutes.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Allow case sharing only with preselected Institutes") }}
```
dnil: "institutes"

tests/server/blueprints/institutes/test_institutes_views.py
```diff
@@ -15,3 +17,59 @@ def test_overview(app, user_obj, institute_obj):
 
         # THEN it should return a page
         assert resp.status_code == 200
+
+
+def test_institute(app, user_obj, institute_obj, monkeypatch):
```
dnil: 👍

scout/server/blueprints/institutes/templates/overview/institute.html
northwestwitch: Should I write also 'NOTE'?

scout/server/blueprints/institutes/templates/overview/institute.html
northwestwitch: I guess! :)

scout/server/blueprints/institutes/templates/overview/institute.html
dnil: If it fits the tooltip, why not - some kind of emphasis since it does change something rather crucial. Oh, and "institute" - I've had to unlearn that because KI.. :)

northwestwitch: Ah ok, I'll add it!

northwestwitch:
Thanks for the review and suggestions, my tooltips where really bad, enriched of spelling mistakes and more misleading than helping! :D

northwestwitch:
Merging this!

Issue #1591, opened 2019-12-13
ftplib problem


parlar:
The ```get_request()``` function in ```request.py``` does not alwas manage to download the 'hgnc_complete_set.txt' file from ebi during the initial database setup.

Sometimes it works, sometimes it doesn't, at least here within our hospital environment. At the university it works fine, so it seems to be some incompatibility with our network infrastructure but perhaps also with other corporate environments?

It turns out that urllib.request does not handle FTP requests itself, they are referred to ftplib. I have tried to find a solution for the problem with ftplib did not succeed. I have tried both python 3.6.8 and 3.8.0 and the problem persists in both these releases. However, I did try another ftp library (ftpretty) that worked flawlessly.

It's definately messier but should we use some other lib for ftp downloads?

northwestwitch:
Hi Pär! 
I don't know about the others, but I personally have anything against using another library instead  of ftplib. If you tested ftpretty and it works we could include that instead. If you have some working code already could you create a pull request? We'll try to review it this week!

northwestwitch:
Another feasible alternative would be changing the code to re-introduce the option to provide an already downloaded file. We had it like this before!

dnil:
The option to use existing files is not a bad idea: I sometimes feel embarrassed for re-downloading again while testing.

But I also agree on libraries - as far as I know we have no particular attachment to ftplib. It does appear sort of standard, and ftpretty a more recent development. Still, whatever works: if you already tried it, just make a PR for it we will review and test! :+1:

parlar:
I guess the possibility to use already downloaded files would solve also the other issue that I just entered: https://github.com/Clinical-Genomics/scout/issues/1592

On the other hand, making things simple and smooth for new users seems attractive too so they (we!) don't have to download everything manually first. I can implement both things if you think I should. It should only take minutes.

northwestwitch:
I've just spoken to @moonso and he offered to reintroduce the code to provide downloaded files (he already knows this part of the code well)

moonso:
Hi @parlar , I can implement to upload with static files. If you want to do the ftp-thing feel free to open a PR!

parlar:
Brilliant, thanks!

parlar:
Good morning, @moonso et al! Just so that I understand, changes have now been included to better handle when the Ensemble REST service is down. But is the question of using pre-downloaded data files for setup still open?

dnil:
[...]

PR #1605, opened 2020-01-08
Restore coveralls call from Travis CI and handle Biomart service errors

### LINKED ISSUE #1564: Handle better situation when Ensembl Rest services are down
The update genes command for instance crashed without any indication of the problem

### LINKED ISSUE #1604: Travis CI skips coverage check
Before there were 3 checks and now it looks like the coverage check is skipped. 

Maybe it's just some change in Travis CI, I don't know, but now the checks are always green even if the coverage goes 50% down.. I don't like it!

northwestwitch:
fix #1604 and fix #1564 

The ideal would be to fix the tests to take into account also when the Ensembl Rest API is down, but now it is working and I don't remember exactly what the page says when the service is down. I guess we could fix it some other time!

**How to test**:
1. Travis CI will run the tests

**Expected outcome**:
Automated tests should pass (3 tests, including coveralls call)

**Review:**
- [x] code approved by DN

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,13 @@ services:
 
 script:
     - py.test --cov=scout -rxs tests/
+    
 install:
     - pip install cython
     - pip install -r requirements.txt -r requirements-dev.txt -e .
 
 after_success:
-    - coveralls --ignore-errors
+    - coveralls
 
 notifications:
     email: false
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,7 @@
 # everything the developer needs in addition to the production requirements
 
 # testing
-pytest>=3.8
-pytest-flask
-pytest-cov
+pytest>=3.7.1
 mongomock<=3.12
 coveralls
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@
 
 # testing
 pytest>=3.7.1
+pytest-cov
 mongomock<=3.12
 coveralls
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@
 
 # testing
 pytest>=3.7.1
+pytest-flask
 pytest-cov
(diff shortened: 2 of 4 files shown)


requirements-dev.txt
```diff
@@ -1,11 +1,11 @@
...
 
 # testing
-pytest==3.8
-pytest-flask==0.11
+pytest>=3.7.1
+pytest-flask
 pytest-cov
 mongomock<=3.12
-python-coveralls
+coveralls
```
dnil: There is also coveralls-python. Reminds me, perhaps not without reason, of Monty Python. But, this seems official enough, and importantly to work so lets go with it!

.travis.yml
```diff
@@ -12,12 +12,13 @@ services:
...
 script:
     - py.test --cov=scout -rxs tests/
+    
 install:
     - pip install cython
     - pip install -r requirements.txt -r requirements-dev.txt -e .
 
 after_success:
-    - coveralls --ignore-errors
+    - coveralls
```
dnil: Good. This will also make @patrikgrenfeldt happy! :wink:

dnil:
Fixes #1604, but I think it kind of fixes #1564 rather than #1591, doesn't it?

northwestwitch:
> Fixes #1604, but I think it kind of fixes #1564 rather than #1591, doesn't it?

You are right, I might have tagged the wrong issue!

tests/utils/test_ensembl_rest_clients.py
```diff
@@ -106,8 +106,9 @@ def test_test_query_biomart_38_xml():
     ## THEN assert that the result is correct
     i = 0
     for i,line in enumerate(client,1):
-        assert 'ACTR3' in line
-    assert i > 0
+        ## THEN assert the correct gene is fetched or that an HTML page us returned is service is down
```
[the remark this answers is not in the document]
northwestwitch: Right!

Issue #1608, opened 2020-01-13
Conservation and REVEL scores annotations


henrikstranneheim:
MIP 8 have removed snpeff - now these annotations are calculated using VEP. As such they are now part of the CSQ field instead of the vcf INFO field.

```
GERP++_NR|GERP++_RS|REVEL_rankscore|phastCons100way_vertebrate|phyloP100way_vertebrate is
```

northwestwitch:
This is a bit more tricky because we still want to maintain backward compatibility so I guess we should allow for both types of parsing: from the INFO and the CSQ field, right?

PR #1623, opened 2020-01-14
Conservation values from CSQ field (vep annotation)

northwestwitch:
Partial fix to #1608

This PR allows parsing of CSQ fields for collecting the following conservation values: 
- GERP++_RS
- phastCons100way_vertebrate
- phyloP100way_vertebrate

It's an extension and does not replace the parsing of values from the VCF INFO field, where the same keys are named respectively:
- dbNSFP_GERP___RS
- dbNSFP_phastCons100way_vertebrate
- dbNSFP_phyloP100way_vertebrate

**How to test**:
This will be tested when uploading a test case from MIP 8.0.
Unit testing it for now.

**Expected outcome**:

**Review:**


scout/parse/variant/conservation.py
```diff
@@ -1,37 +1,61 @@
...
+        'csq_keys' : {
+            'gerp' : 'GERP++_RS',
+            'phast' : 'phastCons100way_vertebrate',
+            'phylop' : 'phyloP100way_vertebrate'
+        }
+    }
+    vep_info = variant.INFO.get('CSQ')
+
+    # if CSQ field contains any conservation key specified in conservation_keys['csq_keys']
+    if any(elem in list(conservation_keys['csq_keys'].values()) for elem in vep_header):
```
moonso: This will slow down the code. For every entry in the vep header it will create a list, for every variant in the VCF. I would rewrite this and check the quickest check first. We can look at it together tomorrow if you like.

northwestwitch: Right, I'll try write a solution, otherwise together, yes, thanks!

northwestwitch: I've made it faster!

scout/parse/variant/variant.py
```diff
@@ -280,7 +282,7 @@ def parse_variant(variant, case, variant_type='clinical',
         parsed_variant['spidex'] = float(spidex)
 
     ###################### Add conservation ######################
-    parsed_variant['conservation'] = parse_conservations(variant, vep_header)
+    parsed_variant['conservation'] = parse_conservations(variant, vep_conservation, raw_transcripts)
```
moonso: Unfortunately this will not work since `raw_transcripts` is a generator with will be exhausted in line 225. One idea could be to parse the information in the `parse_transcripts` function. I guess it is easier if we sit down together tomorrow

northwestwitch: Mmm, What if we convert raw_transcripts into a list?

northwestwitch: But yes, tomorrow!

northwestwitch: Made it into a list, see you tomorrow! :)

northwestwitch:
Done @moonso

scout/parse/variant/conservation.py
```diff
@@ -1,37 +1,58 @@
...
 
-def parse_conservations(variant):
+LOG = logging.getLogger(__name__)
+
+def parse_conservations(variant, parsed_transcripts=[]):
     """Parse the conservation predictors
 
         Args:
             variant(dict): A variant dictionary
+            raw_transcripts(generator): if provided, use transcript annotations defined in the CSQ
```
moonso:
```suggestion
            parsed_transcripts(list): if provided, use transcript annotations
```

scout/parse/variant/variant.py
```diff
@@ -47,6 +47,7 @@ def parse_variant(variant, case, variant_type='clinical',
     rank_results_header = rank_results_header or []
     # Vep information
     vep_header = vep_header or []
+    vep_conservation = False
```
moonso:
```suggestion
```

northwestwitch:
Oh no, I can't merge because the required statuses are not passing!

PR #1626, opened 2020-01-16
REVEL score from CSQ field (vep annotation)

### LINKED ISSUE #1608: Conservation and REVEL scores annotations

northwestwitch:
Final fix #1608

This PR allows parsing of CSQ fields for collecting variants [REVEL](https://sites.google.com/site/revelgenomics/about) scores 

**How to test**:
This will be tested when uploading a test case from MIP 8.0.
Unit testing it for now.

**Expected outcome**:

**Review:**
- [x] code approved by MM


northwestwitch:
> Hi @northwestwitch , nicely done. Don't forget to add the revel thing to scout/build/variant otherwise it will not be added to the database. Also I suggest that you add the REVEL entry to one variant in the test VCF (Or create a new test VCF with this information) So that you can see that it actually works.
> Perhaps adding the new conservations would be a good idea to.

Right! I think the conservations should already be added, because it's not a new feature, it's just collected from another field, but I'll check!

northwestwitch:
> > Hi @northwestwitch , nicely done. Don't forget to add the revel thing to scout/build/variant otherwise it will not be added to the database. Also I suggest that you add the REVEL entry to one variant in the test VCF (Or create a new test VCF with this information) So that you can see that it actually works.
> > Perhaps adding the new conservations would be a good idea to.
> 
> Right! I think the conservations should already be added, but it's a good idea to create an updated VCF with this type of fields

northwestwitch:
@moonso you had a good point when you asked to load from VCF. Doing so I've found an error in the code, which now should be fixed. I've updated code and tests!

scout/parse/variant/transcript.py
```diff
@@ -69,12 +69,15 @@ def parse_transcripts(raw_transcripts, allele=None):
...
+        if entry.get('REVEL_RANKSCORE'):
+            transcript['revel'] = float(entry.get('REVEL_RANKSCORE'))
+
         transcript['swiss_prot'] = entry.get('SWISSPROT') or 'unknown'
 
         # Check for conservation annotations
         transcript['gerp'] = entry.get('GERP++_RS')
-        transcript['phast'] = entry.get('phastCons100way_vertebrate')
-        transcript['phylop'] = entry.get('phyloP100way_vertebrate')
+        transcript['phast'] = entry.get('PHASTCONS100WAY_VERTEBRATE')
```
moonso: 👍

northwestwitch:
Hi @moonso, I've tried to reformat but I still get issues and it doesn't pass pylint. Since I don't want to spend any more time on this and you approved and tests are passing, could you merge? Thanks

PR #1630, opened 2020-01-17
Avoid flashing popovers

bjhall:
The comment popover:
image

...is flickering wildly and is unreadable for long comments (if wider than the browser window I think). This has been the case since the BS4 update.

This simple CSS change fixes it. And as far as I can see it doesn't have any other side effects.


northwestwitch:
Thanks @bjhall , we'll review ASAP!

PR #1651, opened 2020-01-23
Fix case inactivation on update

### LINKED ISSUE #1633: Form validation in age and tissue type case form
- [x] Validate age and tissue type form on case page.
- [x] Fix bug that inactivate case when values are submitted using this form 
- [x] The inactivate case feature also appears on other case level changes, e.g. matchmaker submission

Error message for the first part of the issue:
File "/home/hiseq.clinical/SERVER/apps/scout/git/scout/scout/server/blueprints/cases/views.py", line 446, in update_individual
    controllers.update_individuals(store, institute_obj, case_obj, user_obj, ind_id, age, tissue)
  File "/home/hiseq.clinical/SERVER/apps/scout/git/scout/scout/server/blueprints/cases/controllers.py", line 451, in update_individuals
    subject['age'] = round(float(age), 1)
ValueError: could not convert string to float:

northwestwitch:
Fix #1633

**How to test**:
1. Make sure that editing case age and tissue type (also with empty case age) doesn't cause crash
1. Make sure that if a case is active it doesn't become inactive upon change (for instance modifying the above age or tissue type)

**Expected outcome**:

**Review:**
- [x] code approved by DN, CR
- [x] tests executed by CR


scout/adapter/mongo/case.py
```diff
@@ -565,9 +565,9 @@ def update_case(self, case_obj):
             {'_id': case_obj['_id']}
         )
 
-        # if case is updated and was archived or active - make it inactive
+        # if case is updated and was archived make it inactive instead
         updated_status = old_case.get('status')
-        if old_case.get('status') in ['archived', 'active']:
+        if updated_status ==  'archived':
```
dnil: ..but then the inactivate-upon-rerun logic fails: rerun active cases will no longer be made inactive. Also, all existing comments would now flag as "old" since the case has indeed been updated. 

What do you think about silent (like minor) update mode for the update function that does not alter case status or modification date? We could then use that also for the report reuploads?

scout/server/blueprints/cases/controllers.py
```diff
@@ -448,8 +448,12 @@ def update_individuals(store, institute_obj, case_obj, user_obj, ind, age, tissu
     case_individuals = case_obj.get('individuals')
     for subject in case_individuals:
         if subject['individual_id'] == ind:
-            subject['age'] = round(float(age), 1)
-            subject['tissue_type'] = tissue
+            if age:
+                subject['age'] = round(float(age), 1)
+            else:
+                subject['age'] = None
+            if tissue:
```
dnil: Ah! o_O O_o This was a tricky one to deduce from the server error. Clearly safer now!

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fixed an issued with parsed conservation values from CSQ
 - Coverage calculated on Pull Request commits instead of last push commit
 - Clinvar submissions accessible to all users of an institute
+- Case inactivation upon update
```
dnil:
```suggestion
- Case should not always inactivate upon update
```

scout/adapter/mongo/case.py
```diff
@@ -565,9 +565,9 @@ def update_case(self, case_obj):
             {'_id': case_obj['_id']}
         )
```
dnil:
```suggestion

```
        rerun_requested = case.get('rerun_requested', False)
        if case.get('rerun_requested'):
            updated_status = old_case.get('status')
            if case.status in ['archived', 'active']:
                updated_status = 'inactive'
            rerun_requested = False
```

Comment (dnil) [reply scout/adapter/mongo/case.py]: and then ```rerun_requested``` in the ```$set ``` operation below obviously?

File: scout/adapter/mongo/case.py (hunk 1 above)
Comment (northwestwitch) [reply scout/adapter/mongo/case.py]: Right! A minor update mode or passing a flag to the function! Good point!

**Comment by dnil** [comment]
Actually I realised for the update part, we kind of have a flag. request rerun will normally be set. It is not waterproof though - the reanalysis could have come about through a ticket or other means.

File: scout/adapter/mongo/case.py (hunk 4 above)
Comment (northwestwitch) [reply scout/adapter/mongo/case.py]: Oh I see now that you've suggested the solution,thanks! Im going to look into it later today

Comment (dnil) [reply scout/adapter/mongo/case.py]: ..and it doesn't have to be today - you were planning not to work all day, remember? 😄

Comment (northwestwitch) [reply scout/adapter/mongo/case.py]: I need to do some matchmaker testing but it won't take too long, and this one is quite short as well!

Comment (northwestwitch) [reply scout/adapter/mongo/case.py]: But if you have time ypi could always take over this PR!

File: scout/load/case.py [hunk 5]
```diff
@@ -28,7 +28,7 @@ def load_case(adapter, case_obj, update=False):
 
     if existing_case:
         if update:
-            adapter.update_case(case_obj)
+            adapter.update_case(case_obj, update)
```
dnil: ..that was left from an early idea, right?

dnil:
The new one looks ok, but do consider my suggestion about the rerun_requested flag: False. Currently changing the case age will make the rerun flag lower as well. If it gets too complex one could just make an update_case_without_status_change function or something, that can only update things that should not change the status?

northwestwitch:
> The new one looks ok, but do consider my suggestion about the rerun_requested flag: False. Currently changing the case age will make the rerun flag lower as well. If it gets too complex one could just make an update_case_without_status_change function or something, that can only update things that should not change the status?

But as you said sometimes cases don't have the "rerun_requested" if the reruns are ordered by ticket. Those cases wouldn't work right?  But you are right, the rerun shouldn't be touched by updates other than rerun. I can fix this..

dnil:
Yep, true that; I do like setting status in the load function before calling update - that kind of solves the rerun to inactive part. It should also be responsible for lowering the rerun_requested flag, shouldn't it?

northwestwitch:
> Yep, true that; I do like setting status in the load function before calling update - that kind of solves the rerun to inactive part. It should also be responsible for lowering the rerun_requested flag, shouldn't it?

What about this?

scout/adapter/mongo/case.py
```diff
@@ -506,6 +506,9 @@ def load_case(self, config_data, update=False):
             LOG.warning(error)
 
         if existing_case and update:
+            if case_obj['status'] == 'active':
+                case_obj['status'] = 'inactive'
+                case_obj['research_requested'] = False
```
dnil: Exactly! Only I think both should also be done if the case has status archived.

northwestwitch: If the case is archived it will be inactivated anyway, because of line 573 in adapter.case.py

dnil: Yes, but not have its rerun flag lowered. Which I'm guessing you planned to not always set to false on update now?

dnil: Sorry for second guessing you on this - should I make a commit instead? 😊
You are just much faster coming up with a suggestion - I get stuck in potential implications and possible other things that it touches.. 🙄

northwestwitch: > Yes, but not have its rerun flag lowered. Which I'm guessing you planned to not always set to false on update now?

I'm just thinking that it shouldn't be necessary to specify `case_obj['research_requested'] = False` because there is some code somewhere that does it already

northwestwitch: Can I try locally, it's quite easy to test it

dnil: Oh, and I'm reading to sloppily: I was going about ```case_obj['rerun_requested']```.

northwestwitch: I'm also a bit confused, I think testing would be the best way to understand how all this works :-/

dnil: ```'research_requested'``` is already read from the case obj, so that should (probably) work. I'm now leaning towards that there should be a second function for just updating the case individual parameters without touching the status etc. But still, the load function should be responsible for setting case_obj['request_rerun'] to false, rather than assuming that was why the case was updated. 😬

northwestwitch: feel free to make commits if you find a solution!

dnil:
Reasons to load a case:
- first load (inactive)
- updating due to scout requested rerun (active, archive -> inactive, requested_rerun flag down)
- updating due to scout requested research upload (no clinical status change?, request_research flag down)
- updating from cli because of other reason (no status change?)
PENDING: - updating from cli for delivery report upload (no status change? no date change?)

Other reasons to run the case update function:
- updating because editing some things in scout (no status change)

Other status changes to consider that sidestep the update function:
- user clicks archive, prioritise button 
- user activates a case
- last user unassigns from case that is not prioritised

scout/adapter/mongo/case.py
dnil: I think I will find lunch first. Just leave it when you feel its time to drop out and I'll take it from there if there is anything left to do. Multiple parallell commits are too complicated. 😅

northwestwitch: You are right, the rerun_requested in adapter.update_case will cause a bug and has to be fixed!

northwestwitch:
> * n (active, archive -> inactive, requested_rerun flag down)



> Reasons to load a case:
> 
> * first load (inactive)
> * updating due to scout requested rerun (active, archive -> inactive, requested_rerun flag down)
> * updating due to scout requested research upload (no clinical status change?, request_research flag down)
> * updating from cli because of other reason (no status change?)
>   PENDING: - updating from cli for delivery report upload (no status change? no date change?)
> 
> Other reasons to run the case update function:
> 
> * updating because editing some things in scout (no status change)
> 
> Other status changes to consider that sidestep the update function:
> 
> * user clicks archive, prioritise button
> * user activates a case
> * last user unassigns from case that is not prioritised

This is a mess!

dnil:
Ah, no, to hasty again - I forgot about the date change. That will confuse everyone. Change the age of an individual and all comments suddenly go "Old".

northwestwitch:
> Ah, no, I was too hasty again - I forgot about the date change. That will confuse everyone. Change the age of an individual and all comments suddenly go "Old".

If this about the other issue with people seeing comments as old when they're newish?

dnil:
Nah, but *any* change will have that effect right now since that is in the ```$set``` operation of ``update_case```. Even e.g. submitting to MME.. I'll make a commit suggestion and you'll see!

northwestwitch:
> Nah, but _any_ change will have that effect right now since that is in the `$set` operation of ``update_case```. Even e.g. submitting to MME.. I'll make a commit suggestion and you'll see!

Ok, good because I understand the problem but not much how to fix it!

scout/commands/update/case.py
```diff
@@ -107,7 +107,7 @@ def case(context, case_id, case_name, institute, collaborator, vcf, vcf_sv,
         case_changed = True
 
     if case_changed:
-        adapter.update_case(case_obj)
+        adapter.update_case(case_obj, soft=False)
```
northwestwitch: What about a more self-explanatory name for this variant: `reset_updated `or something like this?

dnil: Yes! Will do!

scout/adapter/mongo/case.py
```diff
@@ -558,7 +558,7 @@ def update_case(self, case_obj, soft=False):
 
             Args:
                 case_obj(dict): The new case information
-                soft(boolean): The update is small and should not trigger a date change
+                keep_date(boolean): The update is small and should not trigger a date change
```
northwestwitch: 👍

scout/load/case.py
```diff
@@ -28,7 +28,7 @@ def load_case(adapter, case_obj, update=False):
 
     if existing_case:
         if update:
-            adapter.update_case(case_obj, update)
```
northwestwitch: Why did you remove update?

dnil:
It will now be easy also to add a cli flag ```--preserve_date``` for changes like uploading delivery reports that should not change date. But, another PR.

northwestwitch:
> It will now be easy also to add a cli flag `--preserve_date` for changes like uploading delivery reports that should not change date. But, another PR.

I agree

scout/adapter/mongo/case.py
```diff
@@ -506,7 +506,11 @@ def load_case(self, config_data, update=False):
             LOG.warning(error)
 
         if existing_case and update:
-            self.update_case(case_obj)
+            case_obj['rerun_requested'] = False
+            if case_obj['status'] in ['active', 'archived']:
+                case_obj['status'] = 'inactive'
+
+            self.update_case(case_obj, keep_date=False)
```
northwestwitch: You don't need to pass the keep_date here really, but maybe it's more clear this way?

scout/commands/update/case.py
```diff
@@ -107,7 +107,7 @@ def case(context, case_id, case_name, institute, collaborator, vcf, vcf_sv,
         case_changed = True
 
     if case_changed:
-        adapter.update_case(case_obj)
+        adapter.update_case(case_obj, keep_date=False)
```
northwestwitch: Here as well

scout/load/case.py
```diff
@@ -28,7 +28,7 @@ def load_case(adapter, case_obj, update=False):
 
     if existing_case:
         if update:
-            adapter.update_case(case_obj)
+            adapter.update_case(case_obj, keep_date=False)
```
northwestwitch: And here

scout/adapter/mongo/case.py
dnil: :blush: yes, I know - I just kept loosing track of which calls I had edited when it wasn't there..

scout/load/case.py
dnil: If we agree that's all, I can now safely remove them.

northwestwitch: If you want, otherwise it's fine.

scout/adapter/mongo/case.py
northwestwitch: ah ok

dnil:
>> @@ -28,7 +28,7 @@ def load_case(adapter, case_obj, update=False):
> 
>     if existing_case:
>         if update:
>-            adapter.update_case(case_obj, update)
>Why did you remove update?

And we agree on this one now, don't we? No update argument for the update_case function, only for load_case.

northwestwitch:
> And we agree on this one now, don't we? No update argument for the update_case function, only for load_case.

update will be True, it seems fine!

dnil:
Ok, ready for testing now methinks!

northwestwitch:
Thanks! I suggest we try it on stage with different scenarios, that way we could also simulate a re-upload from cg maybe

dnil:
Agreed - I've done some clicking around and cli uploading locally. Seems ok so far, but adding CG interaction is a good idea.

northwestwitch:
Tested to re-upload variants for an active case using (from stage) the command `cg upload case_id -r `

image

After reuploading the case becomes inactive.


image

northwestwitch:
The tissue and age form validation errors are fixed now

northwestwitch:
I think the PR is fine, but I found #1657 . I've added some code that should fix the error to this PR!

northwestwitch:
Tested the latest change locally (where only tissue type is specified for demo case). It works. and the following command: `scout --demo load case -u scout/demo/643594.config.yaml` doesn't change an already saved individuals age.

dnil:
Nice! Latest update to the branch is again on stage if you want to try it there!

dnil:
```cat current_version.txt 
fix_inactivation_by_form
661abbd17cbe24ddf73ad0454598c78a5cbaa022```

dnil:
I have not reinstalled, but a quick read of the changes suggest that a code update should be enough. I restarted now though.

northwestwitch:
Merging this PR!!

PR #1664, opened 2020-01-29
Fix panel udate bug

### LINKED ISSUE #1661: panel submission not working
I can not upload an update of my panel.
the file is not uploaded. even testet with your tenplate file.

:(

Michela

northwestwitch:
bug fix #1661

**How to test**:
1. Install it on scout stage and make sure you can update a gene panel (not the first one on the list) using the following options:
- [x] Add genes using the gene panel which was initially failing for the user
- [x] Replace genes using the gene panel which was initially failing for the user
- [x] Test add/replace genes in panel using the Scout demo panel

image

**Expected outcome**:

**Review:**
- [x] code approved by MM
- [x] tests executed by CR


northwestwitch:
Found a solution to fix the error when the panel file starts with BOM chars here: https://stackoverflow.com/questions/13590749/reading-unicode-file-data-with-bom-chars-in-python

Looks perfectly!

I've tested it on stage with the above scenarios and it works. Ready for review!

northwestwitch:
> 👍

Thanks!

PR #1681, opened 2020-02-04
Fix cancer variants HPO panel link

### LINKED ISSUE #1611: HPO clinical filter button needs to go to "cancer" variants..
From https://github.com/Clinical-Genomics/scout/issues/1460

northwestwitch:
This PR fixes the link that from a cancer case leads to variants hitting genes in the HPO panel. (fix #1611).

Current behavior:
1. When an HPO panel is created the link to `Clinical HPO SNV variants` :

image

1. This link opens the NON-CANCER variants page

**How to test**:
1. On a local implementation of Scout load a demo cancer case: 
`scout --demo load case scout/demo/cancer.load_config.yaml`

1. Generate an HPO panel on a case page using a random HPO term. 
1. Check that the button to `Clinical HPO SNV variants` opens the "Clinical Cancer Variants" page.

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -138,21 +138,17 @@
   </div>
 </div>
 {% if case.is_research%}
-  {% if case.track=="cancer" %}
-    <div class="form-group">
-      <div class="btn-group">
-        <a class="btn btn-outline-dark" href="{{ url_for('variants.cancer_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research SNV and INDELs</a>
-        <a class="btn btn-outline-dark" href="{{ url_for('variants.cancer_sv_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research structural variants</a>
-      </div>
+  <div class="form-group">
+    <div class="btn-group">
+    {% if case.track=="cancer" %}
+      <a class="btn btn-outline-dark" href="{{ url_for('variants.cancer_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research SNV and INDELs</a>
+      <a class="btn btn-outline-dark" href="{{ url_for('variants.cancer_sv_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research structural variants</a>
+    {% else %}
+      <a class="btn btn-outline-dark" href="{{ url_for('variants.variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research SNV and INDELs</a>
+      <a class="btn btn-outline-dark" href="{{ url_for('variants.sv_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research structural variants</a>
+    {% endif %}
     </div>
-  {% else %}
-    <div class="form-group">
-      <div class="btn-group">
-        <a class="btn btn-outline-dark" href="{{ url_for('variants.variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research SNV and INDELs</a>
-        <a class="btn btn-outline-dark" href="{{ url_for('variants.sv_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research structural variants</a>
-      </div>
-    </div>
-  {% endif %}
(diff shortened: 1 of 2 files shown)


northwestwitch:
Was it only about this link?

dnil:
I think so, but we may be missing some still.. If you are on it, perhaps click/read through other places where one should be able to reach a cancer variant / variants page. Otw we will find them eventually.

northwestwitch:
Didn't find any, only this:

image

But we don't have the clinical/research variant category separation for cancer, right?

dnil:
No, not at track yet, but we have sort of implemented most of the parsing etc to enable it. We should clearly link it cancer-variants and cancer-sv-variants - but why not leave it at research?

northwestwitch:
> No, not at track yet, but we have sort of implemented most of the parsing etc to enable it. We should clearly link it cancer-variants and cancer-sv-variants - but why not leave it at research?

So I'll add the check, without links to the cancer research variants for now, ok?

dnil:
Sure, thats the baseline. But why not just check, within that if case.is_research, if case.vcf_files.vcf_cancer_research and/or case.vcf_files.vcf_cancer_sv_research are loaded? If they are, then it should be great to link to cancer_variants and cancer_sv_variants with research variants? There is very little difference for the research/clinical views even for RD.

northwestwitch:
> Sure, thats the baseline. But why not just check, within that if case.is_research, if case.vcf_files.vcf_cancer_research and/or case.vcf_files.vcf_cancer_sv_research are loaded? If they are, then it should be great to link to cancer_variants and cancer_sv_variants with research variants? There is very little difference for the research/clinical views even for RD.

Ah OK, I didn't get that that these files where already potentially available, I'll fix it!

dnil:
Thanks, since we are already most of the way there. I don't think they will be used initially. There are none in the test cancer case, and I don't think Balsamic produces them. I have a feeling the ppl involved rather differentiate clinical / research based on gene panel designation or type of variant within the panel, but that is another story.

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -904,12 +904,21 @@ <h5 class="modal-title" id="exampleModalLabel">Edit synopsis</h5>
             Use HPO list for clinical filter</div>
         </div>
         <div>
-          <a class="btn btn-outline-secondary form-control"
-            href="{{ url_for('variants.variants', institute_id=institute._id,
-                             case_name=case.display_name, variant_type='clinical',
-                             gene_panels=['hpo']) }}">
-             Clinical HPO SNV variants
-           </a>
+          {% if case.track == 'cancer' %}
```
dnil: :+1:

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -137,13 +137,22 @@ <h4 class="mt-3">Case: {{case.display_name}}</h4><p class="text-muted {% if case
...
 </div>
-{% if case.is_research %}
-  <div class="form-group">
-    <div class="btn-group">
-      <a class="btn btn-outline-dark" href="{{ url_for('variants.variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research SNV and INDELs</a>
-      <a class="btn btn-outline-dark" href="{{ url_for('variants.sv_variants', institute_id=institute._id, case_name=case.display_name, variant_type='research') }}">Research structural variants</a>
+{% if case.is_research%}
+  {% if case.track=="cancer" %}
+    <div class="form-group">
+      <div class="btn-group">
```
dnil: This is also a little picky, but couldn't the divs here live outside the {%if%} clauses? :blush:

northwestwitch: Sure! I'll change it

northwestwitch:
If this is fine for you @dnil we can merge, it's only HTML code + flask anyway!

dnil:
Hold for the black discussion first I realise - otw you'll have to merge this on top?

dnil:
But, yeah, this is html. Should be fine, right?

northwestwitch:
> But, yeah, this is html. Should be fine, right?

Yes, that's why I chose it ! :D

northwestwitch:
3, 2, 1..

PR #1683, opened 2020-02-04
code format all files

patrikgrenfeldt:
This PR blacks all python files in scout

**How to test**:
1. look at the travis checks

**Expected outcome**:
- [ ] all checks should be green

**How to test ui**:
1. install on stage or local scout implementation
1. login

**Expected outcome**:
- [x] login successful

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


northwestwitch:
Which version of Black do you have @patrikgrenfeldt , just out of curiosity?

patrikgrenfeldt:
> Which version of Black do you have @patrikgrenfeldt , just out of curiosity?

`black, version 19.3b0`

northwestwitch:
> > Which version of Black do you have @patrikgrenfeldt , just out of curiosity?
> 
> `black, version 19.3b0`

Ok, I have version 19.10b0

dnil:
Hm, I'm on ```black, version 19.10b0```. Are we going to have issues with this? Like varying degrees of click arguments being shuffled around? What do they change between versions?

northwestwitch:
> Hm, I'm on `black, version 19.10b0`. Are we going to have issues with this? Like varying degrees of click arguments being shuffled around? What do they change between versions?

I have no idea, and I personally don't care, as long as it passes the tests..

patrikgrenfeldt:
I can't install `black, version 19.10b0` for some reason

dnil:
Ah, its one of those - "its portable, because its written in java" things. Overall, some things look nicer, some uglier. I also don't mind either way. I think it's a good enough baseline, as long as it can be easily overridden.

I guess I will be annoyed if there are large near-ws deviations in unrelated places from version differences, but we would get those anyway from e.g. @monso running his own black ops. What I would *really* like to have is a version of diff that is style agnostic, so that it treats as equal all style changes. And this is I suppose as close to that as we can get to that with todays available options.

requirements-dev.txt
```diff
@@ -21,7 +21,7 @@ mkdocs-material
 
 # PEP8 Compliance
 pre-commit
-black
+black==19.3b0
```
dnil: 👀

dnil:
Hey @patrikgrenfeldt aren't we going to have version conflicts and stuff over freezing the version? I appreciate the idea - it would solve my issue with the odd difference. What is the latest version of black anyway? Things should be incremental from there, and we can prod each other to just use it?

dnil:
How about using latest stable? They seem to change that only like once or twice a year?

dnil:
Which then happens to be the same commit as 19.10b0. What was your env issue with it anyway?

northwestwitch:
I agree, if we freeze to a specific Black version we could miss all those updates in the PEP8 definitions in Black, and also bugfixes

patrikgrenfeldt:
@dnil I think your version is the last one. 

When comparing my formatted code and Chiaras there are many files that differ. But the travis job only complain on two files all in all. 

Tried to push Chiaras changes for those two files but my automatic formatting changed them before commit 🤣  That's why I went for the freeze instead. I see your point about avoiding freezing, I'll fix it

dnil:
> Tried to push Chiaras changes for those two files but my automatic formatting changed them before commit 🤣 That's why I went for the freeze instead. I see your point about avoiding freezing, I'll fix it

🤣 Right! Sorry for the hassle - thanks!

patrikgrenfeldt:
running black with 19.3 changes 68 files of Chiaras branch.. Running 19.10 on my branch seems to change 2 files..

dnil:
Sounds like we will have black differences for quite some time. Anyhow, as long as we don't make that test mandatory, this should hopefully be a minor issue, especially when run only on the git diff subset of files.

dnil:
The earlier prolific "trailing commas" issue might stem from differences in the target version autodetect. The help paragraph on 36 sounds a lot like what happened. Anyhow.

patrikgrenfeldt:
> The earlier prolific "trailing commas" issue might stem from differences in the target version autodetect. The help paragraph on 36 sounds a lot like what happened. Anyhow.

Any suggestions?

patrikgrenfeldt:
The code formatting check on travis is green now, do we want to merge?

northwestwitch:
> > The earlier prolific "trailing commas" issue might stem from differences in the target version autodetect. The help paragraph on 36 sounds a lot like what happened. Anyhow.
> 
> Any suggestions?

I suggest we merge, and we don't care about differences in Black version, as long as PEP8 passes, do you agree?

northwestwitch:
> The code formatting check on travis is green now, do we want to merge?

Yes!

northwestwitch:
Merging!

patrikgrenfeldt:
🎆 🍾

PR #1685, opened 2020-02-04
Clear panel pending

### LINKED ISSUE #1665: panel edit botton not working
bild
 

I can not click in the edit, neither in the add ....:(

A question:
how do I remove an uploaded gene that i do not want to submitt any longer?

northwestwitch:
This PR allows to cancel the pending changes to a gene panel. It also improves the badges of the pending actions (fix #1665)

**How to test**:
1. Install the branch on a local implementation of Scout, or on stage.
1. Try to modify a gene panel on the panel page, by adding, editing and deleting a gene. The relative badge should now display deleteED, addED, editED  (instead of delete, add, edit) as the pending actions for those genes:

image

1. Try to cancel those pending actions by clicking on the `Cancel pending actions` button and the pending list of actions should be cleared, and gene panel version should stay the same.
image

**Expected outcome**:

**Review:**
- [x] code approved by MM
- [x] tests executed by CR


northwestwitch:
It works, using the pre-commit before pushing the new PR works, this is a new one and you only see the few changes I made to master! 👍

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 19.3b0
```
northwestwitch: Hi @patrikgrenfeldt, is there a way here to remove the specific version of Black to be used by pre-commit? I've tried to remove the line but it gave me error

northwestwitch:
@moonso this PR has the same oauth problem now. Time to fix it!

moonso:
Yes I fixed it!

scout/server/blueprints/panels/views.py
```diff
@@ -153,8 +153,16 @@ def panel(panel_id):
 def panel_update(panel_id):
     """Update panel to a new version."""
     panel_obj = store.panel(panel_id)
-    update_version = request.form.get("version", None)
-    new_panel_id = store.apply_pending(panel_obj, update_version)
+    if request.form.get("cancel_pending"):
+        updated_panel = store.reset_pending(panel_obj)
+        if updated_panel.get("pending") is None:
+            flash("Pending actions were correctly canceled!", "success")
+        elif updated_panel is None:
```
moonso: If `updated_panel` is `None` the first check will give an error. So these two needs to change place. Perhaps add a test ti that scenario as well?

northwestwitch: Thanks I'll fix it. I'd like to add some other fixes to this PR as well!

northwestwitch: I've fixed it. I think I'll leave this PR as it is and open another PR to fix more panel-related issues..  Could you check again @moonso? Thanks!

northwestwitch:
Merging this as soon as passes the tests!

PR #1689, opened 2020-02-06
Accepts  in load config to pass bam/cram files. Fix #1655

### LINKED ISSUE #1655: Add cram_path to scout load case
Since Scout now supports both bam and cram under the "bam_path" it seems inappropriate to add a anything that is not a BAM file under the "bam_path".

I would therefore like to add another key called "cram_path" to the scout config.

If it does not matter to Scout if the alignment file is in BAM or CRAM - then I think we should consider replacing the keys "bam_path" and "cram_path" with "alignment_file_path" (or something similiar) and deffer the need to check the format until it is actually required (for instance in IGV). But since that requires a major bump, I suggest we start with adding a "cram_path" key. What do you think? @northwestwitch @moonso @dnil

moonso:
**How to test**:
1. tests are testing it

**Expected outcome**:
It will now be possible to pass alignment files via the parameter `alignment_path` in load config.

**Review:**
- [x] code approved by dNil
- [x] tests executed by travis

### DIFF
--- a/tests/parse/test_parse_case.py
+++ b/tests/parse/test_parse_case.py
@@ -1,17 +1,12 @@
-import pytest
-
 from datetime import datetime
 from pprint import pprint as pp
-from scout.parse.case import (
-    parse_case,
-    parse_ped,
-    parse_individuals,
-    parse_individual,
-    parse_case_data,
-    removeNoneValues,
-)
-from scout.exceptions import PedigreeError
+
+import pytest
+
 from scout.constants import REV_SEX_MAP
+from scout.exceptions import PedigreeError
+from scout.parse.case import (parse_case, parse_case_data, parse_individual,
+                              parse_individuals, parse_ped, removeNoneValues)
 
 
 def test_parse_case_no_date(scout_config):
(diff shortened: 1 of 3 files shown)


tests/parse/test_parse_case.py
```diff
@@ -1,17 +1,12 @@
...
-    parse_case_data,
-    removeNoneValues,
-)
-from scout.exceptions import PedigreeError
+
+import pytest
+
 from scout.constants import REV_SEX_MAP
+from scout.exceptions import PedigreeError
+from scout.parse.case import (parse_case, parse_case_data, parse_individual,
```
dnil: It’s interesting to watch the different formatting styles evolve from commit to commit. 😆

tests/parse/test_parse_case.py
```diff
@@ -156,6 +151,45 @@ def test_parse_case_delivery_report(scout_config):
...
+        sample["bam_file"] = bam_path
+    # WHEN case is parsed
+    case_data = parse_case(scout_config)
+
+    # THEN assert that bam files are added correct
+    for ind in case_data["individuals"]:
+        assert ind["bam_file"] == bam_path
+
+
+def test_parse_case_alignment_path(scout_config):
```
dnil: 👍🏻

dnil:
☔️ ☂ 🌂

henrikstranneheim:
🌤 😄 
Thx!

PR #1667, opened 2020-01-30
Upload coverage on master an push CI checks

patrikgrenfeldt:
This PR fixes the issue with coverage reports missing on PRs (fix #1666)

**How to test**:
1. Open this PR on Github
1. Commit some changes that reduces the amount of code covered
1. Revert those changes

**Expected outcome**:
- [x] The build should fail due to decreased coverage
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by CR
- [x] tests executed by Travis CI

This is a patch/fix

PR #1684, opened 2020-02-04
Fix coverage check

patrikgrenfeldt:
This PR fixes the coverage check

**How to test**:
1. decrease the amount of tests significantly & commit
1. increase the amount of tests & commit
1. restore the amount of tests & commit

**Expected outcome**:
- [ ] First commit Red on travis
- [ ] Second commit Red on travis
- [ ] Third commit Green on travis
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by
- [ ] "merge and deploy" approved by

PR #1696, opened 2020-02-10
Coveralls good behavior - PLEASE DON'T UPDATE THE BRANCH

northwestwitch:
This PR is an attempt at fixing the coveralls behavior with I think is wrong. (fix #1667).

**How things should be**:
- [x] remove two files with tests, coverage goes down and coveralls test fails (red cross)
- [x] reintroduce one file with tests, coverage goes a bit up but it's still lower than master -> coveralls test fails (red cross)
- [x] Reintroduce the last missing file with tests so now PR coverage is the same as master -> coveralls test passes, green.
- [x] Every coverage change status should be associated to a coveralls report, updated.


moonso:
Should we not merge this and close the others?

northwestwitch:
> Should we not merge this and close the others?

I don't know if @patrikgrenfeldt is working on an alternative in #1684

northwestwitch:
Unless someone objects I'd like to merge this one as well.. @patrikgrenfeldt ??

PR #1699, opened 2020-02-11
Reuse variant events (pinned/classif/comments) in all variants pages

### LINKED ISSUE #1680: Show Pinned status etc on cancer variants list
Some obvious variant level events are missing from the cancer variants list. Thank you for pointing this out!

northwestwitch:
Fixes also missing info in cancer variants and cancer sv variants pages (fix #1680)
Fixes also: 
- a bug affecting case page for cases with pinned or causative variants in genes without hgvs_identifier. Example: https://scout-stage.scilifelab.se/cust000/620573fam30M
- a bug in case page due to missing parameters in hpo_group_item macro
- a bug caused by missing old_observation import in sv variant page

**How to test**:
1. Open a **snv variant from a non-cancer case** and pin it, add an ACMG classification, comment it globally and locally
1. Open a **structural variant from a non-cancer case** and pin it, add an ACMG classification, comment it globally and locally
1. Open a **str variant from a non-cancer case** and pin it, add an ACMG classification, comment it globally and locally
1. Open a **snv variant from a cancer case** and pin it, add an ACMG classification, comment it globally and locally
1. Open a **structural variant from a cancer case** and pin it, add an ACMG classification, comment it globally and locally

**Expected outcome**:
- [x] non-cancer variants page should show pinned, comments and classification info in the rank cell.
- [x] non-cancer SV variants page should show pinned, comments and classification info in the rank cell.
- [x] non-cancer strs variants page should show pinned, comments and classification info in the rank cell.
- [x] cancer variants page should show pinned, comments and classification info in the rank cell.
- [x] cancer SV variants page should show pinned, comments and classification info in the rank cell.

**Review:**
- [x] code approved by MM
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -20,7 +20,7 @@
                                       variant_id=variant._id) }}">
                     {% set display_genes = [] %}
                     {% for gene in variant.genes %}
-                      {{ display_genes.append(gene.hgnc_symbol + gene.hgvs_identifier or "") }}
+                      {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier or "") }}
                     {% endfor %}
                     {{ display_genes|join(", ") }}
 		             {% else %}
@@ -150,7 +150,7 @@
                                         variant_id=variant._id) }}">
                       {% set display_genes = [] %}
                       {% for gene in variant.genes %}
-                        {{ display_genes.append(gene.hgnc_symbol + gene.hgvs_identifier or "") }}
+                        {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
                       {% endfor %}
                       {{ display_genes|join(", ") }}
                   {% else %}
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -150,7 +150,7 @@
                                         variant_id=variant._id) }}">
                       {% set display_genes = [] %}
                       {% for gene in variant.genes %}
-                        {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
+                        {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier or "") }}
                       {% endfor %}
                       {{ display_genes|join(", ") }}
                   {% else %}
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -1,4 +1,6 @@
 
+{% from "variants/utils.html" import tier_cell %}
+
 {% macro causatives_list(causatives, institute, case) %}
   <div class="card panel-default">
     <div data-toggle='tooltip' class="panel-heading" title="Displays all variants that have been marked causative for this case">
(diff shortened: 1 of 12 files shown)


northwestwitch:
Testing non-cancer variants page using case https://scout-stage.scilifelab.se/cust002/F0000496.

snv variants:

image

sv variants:

image

str variants:

image


Testing cancer variants page using case https://scout-stage.scilifelab.se/cust002/F0000496.

snv variants:

image

I've tested the SV variants page on my local scout instance because there are not cases with cancer SVs on scout stage right now:

image

northwestwitch:
Ready for review!!

northwestwitch:
@dnil if you have some time this is also ready and tested

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -20,7 +20,11 @@
                                       variant_id=variant._id) }}">
                     {% set display_genes = [] %}
                     {% for gene in variant.genes %}
-                      {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
+                      {% if 'hgvs_identifier' in gene %}
```
moonso: 👍

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -150,7 +154,11 @@
                                         variant_id=variant._id) }}">
                       {% set display_genes = [] %}
                       {% for gene in variant.genes %}
-                        {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
+                        {% if 'hgvs_identifier' in gene %}
```
moonso: These two (this one and the `causatives_list` above) are almost identical. Could we break out these lines into its own macro?

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -1,6 +1,7 @@
 {% extends "layout_bs4.html" %}
 {% from "variants/components.html" import gene_cell, frequency_cell %}
 {% from "utils.html" import comments_table %}
+{% from "variants/utils.html" import cell_rank %}
```
moonso: 😄 nice to unify with macros!

scout/server/blueprints/cases/templates/cases/utils.html
northwestwitch: mmm I don't agree so much with having this code in a separate macro here. I agree with you that it's almost the same code as in the causative list, but it's a just a few lines. 
Macros are very good when it comes to perform elaborate and repetitive tasks, but the disadvantage is that from the main HTML page you don't see what happens in them, and you kind of lose the train of thoughts (The if/else conditions) and you need to look for the code in the utils. I don't think it's worth to move this code there.

northwestwitch: It's also good not to fragment too much the page into macros with very little code, because it will become more and more complicated to handle, this is what I mean.

scout/server/blueprints/variants/templates/variants/str-variants.html
dnil: Famous last words.. :wink:

dnil:
I think someone was not paying attention at step 2 of the tests here.. 😜

scout/server/blueprints/variants/templates/variants/str-variants.html
northwestwitch: 😅

moonso: What’s wrong?

dnil: 😊The SV version of that macro unification disabled the link to the SV variant page...

dnil: Cf #1724...

PR #1703, opened 2020-02-12
Fix #1702: activate cancer case on viewing variants.

### LINKED ISSUE #1702: Scout cancer variant view does not activate case
As a scout cancer user, I would like my cases to activate when I peruse variants much as for RD cases.

dnil:

**How to test**:
1. as a non-admin user, open a cancer case and visit the variants page
1. note that it does not activate
Screenshot 2020-02-12 at 10 40 02
1. apply this branch, and note that the same procedure now activates the case.
Screenshot 2020-02-12 at 10 40 38

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Cancer case HPO panel variants link
 - Fix so that some drop downs have correct size
 - First IGV button in str variants page
+- Cancer case activates on SNV variants
 
 ### Changed
 - Renamed `requests` file to `scout_requests`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fix so that some drop downs have correct size
 - First IGV button in str variants page
 - Cancer case activates on SNV variants
+- Cases activate when STR variants are viewed
 
 ### Changed
 - Renamed `requests` file to `scout_requests`
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -787,3 +787,25 @@ def verified_excel_file(store, institute_list, temp_excel_dir):
             written_files += 1
 
     return written_files
+
+def activate_case(store, institute_obj, case_obj, current_user):
+    """ Activate case when visited for the first time.
+
+        Args:
+            store(adapter.MongoAdapter)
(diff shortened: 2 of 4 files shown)


dnil:
PIng @christacosta: note that this will not work right away, but only after next release!

northwestwitch:
Do you need a review?

northwestwitch:
Works: 
image

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Cancer case HPO panel variants link
 - Fix so that some drop downs have correct size
 - First IGV button in str variants page
+- Cancer case activates on SNV variants
```
northwestwitch: What about SV variants? Because we have those as well now!

dnil:
We should unify the views as well - there should be a common base variant view class, and some children for different kinds of variants. But, mañana!

scout/server/blueprints/variants/views.py
```diff
@@ -277,6 +277,16 @@ def cancer_variants(institute_id, case_name):
         page = int(request.args.get("page", 1))
         form = CancerFiltersForm(request.args)
 
+    # update status of case if visited for the first time
```
northwestwitch: You could move this function to the controllers and re-use it from several other functions: 
- snv cancer variants
- sv cancer variants
- snv variants
- sv variants
- str variants? (current situation is that you don't activate the case when you look at these)

CHANGELOG.md
dnil: Already working.

northwestwitch: Ah ok!

scout/server/blueprints/variants/views.py
dnil: Good catch on the STR variants - I'll add that to this PR!

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
...
+- Cases activate when STR variants are viewed
```
northwestwitch: 👍

scout/server/blueprints/variants/controllers.py
```diff
@@ -787,3 +787,25 @@ def verified_excel_file(store, institute_list, temp_excel_dir):
             written_files += 1
 
     return written_files
+
+def activate_case(store, institute_obj, case_obj, current_user):
```
northwestwitch: 👍

Issue #1629, opened 2020-01-17
QoL changes to Variant list view for cancer cases


hassanfa:
During a brief discussion today with Chiara, Patrick, and Mikael, we come up with a small list of changes that we thought it would be useful to clear up some Variant list view for cancer cases. These items are:

- [x] Pop. Freq. column is named "Exac", either remove it or replace it with actual COSMIC variants that are matching. I think removing it with Cosmic ID or count of COSMIC variants is more reasonable. Since we already have gnomadAF in filters, removing it won't harm.
Column example of how it can be:
![image](https://user-images.githubusercontent.com/1086877/72622574-c9d51b00-3943-11ea-8af4-125bf59b26c6.png)

- [x] When there is no matching Normal sample, the value of Normal column is ambiguous. Showing "N/A" instead of "N/A /" is much better. Example of removing "/": 
![image](https://user-images.githubusercontent.com/1086877/72622365-56330e00-3943-11ea-879d-919cd5d34163.png)
~~- [ ] Add custom sorting per column. i.e click a column will rearrange the list.~~
- [x] Remove pedigree section from cancer cases: 
![image](https://user-images.githubusercontent.com/1086877/75860969-f8785980-5dfc-11ea-8d69-08c7e0fd0d95.png)
- [x] Genes names are not showing up in the Variant ID. This should be consistent, regardless of variant type or missing gene. Example of how it is right now:
![image](https://user-images.githubusercontent.com/1086877/72622622-e5402600-3943-11ea-9583-160ad7c710a0.png)

- [x] Header for second column is: `Variant:Transcript:Exon:HGVS`. It is not matching the actual value of the column which is: `Gene:Transcript:Exon:HGVS`. And related to previous case, sometimes gene names are not showing up. ![image](https://user-images.githubusercontent.com/1086877/72622306-3bf93000-3943-11ea-9849-882ad1497652.png)

- [x] Only show 2 decimals instead of 4-5 decimals in AF for Tumor and Normal. Example:
![image](https://user-images.githubusercontent.com/1086877/72623201-f8072a80-3944-11ea-8160-43b55b44b9fe.png)

~~- [ ] Ranks in Cancer view are not ranks, and it creates confusion sometimes. Column header shouldn't be called "Rank" if there is not rankscore used.~~

[...]

PR #1711, opened 2020-02-14
Round Allele and Normal Frequency to Two Decimals

mikaell:
Show two instead of four decimals for Allele and Normal frequency
in cancer view.

As from discussion here:
https://github.com/Clinical-Genomics/scout/issues/1629#event-3009077515

**How to test**:
1. how to install it

Clone and install Scout from `QoL-cancer_variant_view_200214`. Setup demo or preferred configuration showing cancer variants such as `cancer.load_config.yaml`

Go to `Clinical Cancer Variants Page`in Scout.

**Expected outcome**:
At the most two decimals are shown, instead of four, for Allele and Normal in Scout's cancer view. For example:

Screenshot 2020-02-14 at 15 52 17

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 - Renamed `requests` file to `scout_requests`
-
+- Cancer variant view shows two, instead of four, decimals for allele and normal
 
 ## [4.11.1]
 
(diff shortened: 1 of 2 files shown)


mikaell:
> 👍
> 
> Don't forget the changelog!

Thanks! Fixed!

dnil:
Nice @mikaell!
But eh, ENSMBL errors again? I thought we had those bypassed @northwestwitch @moonso?

moonso:
I think we should avoid testing other services, this should be mocked out right?

northwestwitch:
> I think we should avoid testing other services, this should be mocked out right?

I agree, I've tried to catch the single errors that they give but they give every time different ones. A good thing about testing them is that you know when they work, and if you can trust them, but if you don't want this type of problems you can mock them.

dnil:
Mock is good. We can also put them in a separate directory and run them allowing failure.

northwestwitch:
Merging this!

PR #1720, opened 2020-02-19
QoL: Cleaner <TD> in Cancer View When Missing Tumor or Normal

mikaell:
Remove a stand alone slash ' / ' from Scout's cancer variant view
when Tumor or Normal is n/a.

This PR is QoL. Ref: https://github.com/Clinical-Genomics/scout/issues/1629

**How to test**:
1. how to install it
git clone 'feature-cancer_variant_view-200219'.

Access cancer variant view page. 

**Expected outcome**:
If table data is "N/A", there will be no small-tag below (this previously contained a slash "/"). If table data exists, small-tag will display a ratio. 

Right click and select "Inspect Element" to confirm.

Compare Tumor and Normal fields in attached image:

Screenshot 2020-02-19 at 10 51 21

**Review:**
- [x] code approved by
- [ ] tests executed by


hassanfa:
Looks good to me.

Question: why is ref_depth `-1` in your example? Which VCF demo are you using?

mikaell:
> Question: why is ref_depth `-1` in your example? Which VCF demo are you using? Probably it is not relevant to this PR though...

I haven't got the faintest idea. Pulled a new 'master' and applied my two liner. Must be something hanging around in my mongoDB.

northwestwitch:
Is this ready to merge? We are creating a new release today and we could have it in it!

hassanfa:
I think so. I forgot to approve for code and functionality.

The only catch is that there is a `-1` in ref count number, which I don't understand why it is happening. It is most likely @mikaell 's local installation and remanent of old data.

dnil:
Test on stage perhaps? That would alleviate any concerns with -1 depth?

northwestwitch:
I'm going to test it on stage now!

northwestwitch:
Looking good:

image

dnil:
No its not! 3658 / 18 😊 Please invert!

dnil:
Also consider what it will look like for a single case?

northwestwitch:
> Also consider what it will look like for a single case? Like with tumor-only?

That is the case I tested, the tumor only

northwestwitch:
> No its not! 3658 / 18 😊 Please invert!

18/3658 ?

northwestwitch:
I couldn't find any case with tumor / paired normal with N/A info for the normal

dnil:
> > No its not! 3658 / 18 😊 Please invert!
> 
> 18/3658 ?

Yes, alt / ref.. That's sort of the natural way to show the ratio - even if I assume the percentage is actually alt / (ref + alt), so there could be some confusion. Perhaps just write the number of alt? That would solve both missing normal tissue and any confusion as to if the denominator is all counts or just ref counts?

dnil:
lovingtiger at least have normal N/A so one can see there is no mixup. Good enough?

northwestwitch:
> > > No its not! 3658 / 18 😊 Please invert!
> > 
> > 
> > 18/3658 ?
> 
> Yes, alt / ref.. That's sort of the natural way to show the ratio - even if I assume the percentage is actually alt / (ref + alt), so there could be some confusion. Perhaps just write the number of alt? That would solve both missing normal tissue and any confusion as to if the denominator is all counts or just ref counts?

Done in #1728

northwestwitch:
But I've just swapped them in that PR

PR #1722, opened 2020-02-19
Take care of empty bam paths values

northwestwitch:
This is an urgent bug fix

The current master branch doesn't handle well when both these keys are present in case config file:
1. An empty `bam_path`
1. A valid `alignment_path`

**How to test**:
1. Install it on hasta stage and upload a case with the above configuration
1. Make sure that the correct bam path is passed along and saved in the database.

**Expected outcome**:

**Review:**
- [x] code approved by MM
- [x] tests executed by HS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Discrepancy between the manual disease transcripts and those in database in gene-edit page
 - ACMG classification not showing for some causatives
 - Fix bug which caused IGV.js to use hg19 reference files for hg38 data
+- Bug when multiple bam files sources with non-null values are available
 
 ### Changed
 - Renamed `requests` file to `scout_requests`
--- a/scout/parse/case.py
+++ b/scout/parse/case.py
@@ -286,9 +286,12 @@ def parse_individual(sample):
     ind_info["predicted_ancestry"] = sample.get("predicted_ancestry")
 
     # IGV files these can be bam or cram format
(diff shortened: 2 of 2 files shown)


northwestwitch:
Waiting to see if the upload of a test case on hasta stage worked!

henrikstranneheim:
1. The current master branch doesn't handle well when both these keys are present in case config file:

    An empty bam_path
    A valid alignment_path

**How to test:**

Install it on hasta stage and upload a case with the above configuration
Screenshot 2020-02-19 at 14 44 28

Screenshot 2020-02-19 at 14 44 52

Works with vitalmouse on hasta

northwestwitch:
@henrikstranneheim and I would meed to deploy this bugfix ASAP.

My question is: do we want to merge to master and deploy everything we have done since the latest release of should we do a patch, based on the latest version + this PR?

I vote for the first option @moonso @dnil ?

dnil:
So far I think no release accepts alignment_path only current master, so a patch would be rather strange? If you are in a hurry to get that out I also suggest making a proper release.

northwestwitch:
> So far I think no release accepts alignment_path only current master, so a patch would be rather strange? If you are in a hurry to get that out I also suggest making a proper release.

One could pull the other changes from #1689 into last release branch, but since we have done several other things it would be nice to release them out, so they're used. And it's not even Friday this time (we always make new releases on Friday lately :D)

moonso:
I vote for releasing everything, it is about time anyway

northwestwitch:
We release everything then! 👍

henrikstranneheim:
🤞

PR #1726, opened 2020-02-20
Fix #1724 - SV cell rank

### LINKED ISSUE #1724: IGV view in Structural variant is not working properly
bild

WE used to have 3 IGV viewer:
1. entire variation
2. breakpoint1
3. breakpoint 2

now there in only one :(
and in not the one for the entire variation. could you check?
Can we have back the 3 opitions?

dnil:
Fix #1724 

**How to test**:
1. on sv variants lists (cancer and rd), click on a rank
1. note that you are taken to the corresponding sv variant page

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by DN

### DIFF
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -549,11 +549,19 @@
 {% endmacro %}
 
 {% macro cell_rank(variant, institute, case, form, manual_rank_options) %}
+{% if variant.category == "sv" %}
+  <a class="variants-row-item flex-small layout"
+      href="{{ url_for('variant.sv_variant', institute_id=institute._id,
+                       case_name=case.display_name, variant_id=variant._id) }}">
+     {{ variant.variant_rank }}&nbsp;
+   </a>
+{% else %}
   <a class="variants-row-item flex-small layout"
      href="{{ url_for('variant.variant', institute_id=institute._id, case_name=case.display_name,
                       variant_id=variant._id) }}">
     {{ variant.variant_rank }}&nbsp;
   </a>
+{% endif %}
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -549,11 +549,19 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
 {% endmacro %}
 
 {% macro cell_rank(variant, institute, case, form, manual_rank_options) %}
+{% if variant.category == "sv" %}
+  <a class="variants-row-item flex-small layout"
+      href="{{ url_for('variant.sv_variant', institute_id=institute._id,
+                       case_name=case.display_name, variant_id=variant._id) }}">
+     {{ variant.variant_rank }}&nbsp;
```
northwestwitch: Nice! If this line is in both if and else then it could be moved outside the conditions?

northwestwitch: And check if the link to the variant page is also in cell_rank macro

dnil: Pardon, what did you mean by 
>And check if the link to the variant page is also in cell_rank macro

northwestwitch:
The links to the variant page was not originally in the cell_rank macro? I don't remember

dnil:
It was indeed - that was sort of the issue: since there was a fixed link, but the macro was used for different kinds of variants.. Well, you get the picture. 😸

northwestwitch:
Yes, I see better now, I was in between the train stop and the metro stop when I made the review. Good to merge I think!

PR #1728, opened 2020-02-20
Swap ref/alt to alt/ref reads in cancer variants page

northwestwitch:
So it mirrors the actual alt allele frequency

**How to test**:
1. Install it and check that on variants page it is reported the right info

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Bug in variant.gene when gene has no HGVS description
 - SV variant links now take you to the SV variant page again
+- cancer variants show correct alt/ref reads mirroring alt frequency now 
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


dnil:
Ok, thanks that's what I asked for. Seems @hassanfas and @mikaell  issue with refs is also in my demo case:
Screenshot 2020-02-20 at 11 10 01

I'm leaning towards this being ok as things seemed ok on stage? What do you think?

northwestwitch:
> Ok, thanks that's what I asked for. Seems the issue of @hassanfa and @mikaell with refs is also in my demo case:
> Screenshot 2020-02-20 at 11 10 01
> 
> I'm leaning towards this being ok as things seemed ok on stage? What do you think?

It's OK! :)

northwestwitch:
I don't know about this -1 in the demo case but It's looking good on stage:

image

dnil:
Ok good. What about column width / nbsp for the cell? It looks a bit chaotic on that picture?

northwestwitch:
> Ok good. What about column width / nbsp for the cell? It looks a bit chaotic on that picture?

It's a lot of info in all the other cols, so I'm not surprised. What about leaving it like this for now. We can fix it later, there's a number of other changes that is going to happen in that page anyway

northwestwitch:
Shall I merge?

PR #1732, opened 2020-02-20
Fix #1729 - load all STRs also on region load.

### LINKED ISSUE #1729: Trying to load all variants for a gene removes all clinical variants
When running the command to load all variants for a gene (`scout load region`) an error occurs and all clinical variants are removed.

The command `scout load region --case-id hotlemur --hgnc-id 8582` gives the following error:
image

I've tried this command on three cases from cust000 with two different genes and get the same error.

dnil:
Attempt to load all STRs even if region is requested, as the STR VCF is small and index likely missing.

**How to test**:
1. attempt to upload a previously failing case in hasta scout stage with this branch.

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Bug in variant.gene when gene has no HGVS description
 - SV variant links now take you to the SV variant page again
+- Cancer variant view has cleaner table data entries for "N/A" data
 - Cancer variants show correct alt/ref reads mirroring alt frequency now
 - Always load all clinical STR variants even if a region load is attempted - index may be missing
 
@@ -51,7 +52,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - ACMG classification not showing for some causatives
 - Fix bug which caused IGV.js to use hg19 reference files for hg38 data
 - Bug when multiple bam files sources with non-null values are available
-- Cancer variant view has cleaner table data entries for "N/A" data
+
 
 ### Changed
 - Renamed `requests` file to `scout_requests`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,6 +51,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - ACMG classification not showing for some causatives
 - Fix bug which caused IGV.js to use hg19 reference files for hg38 data
 - Bug when multiple bam files sources with non-null values are available
+- Cancer variant view has cleaner table data entries for "N/A" data
```
northwestwitch: I think this should belong to version 4.12.1, since we merged today

scout/load/all.py
```diff
@@ -102,16 +102,12 @@ def load_region(adapter, case_id, hgnc_id=None, chrom=None, start=None, end=None
...
         LOG.info(
-            "Load clinical STR variants for case: {0} region: chr {1}, "
-            "start {2}, end {3}".format(case_obj["_id"], chrom, start, end)
+            "Load all clinical STR variants for case: {0}."
         )
         adapter.load_variants(
             case_obj=case_obj,
             variant_type="clinical",
             category="str",
-            chrom=chrom,
```
northwestwitch: Should work conceptually. But better test it first!

dnil: Now you want to test.. Tsssk.. 😉

CHANGELOG.md
dnil: Ah, sorry - please move it where you want it!

scout/load/all.py
northwestwitch: Ok, on hasta stage!

northwestwitch:
Tested on scout stage using the same gene for which the region upload failed in #1729
image

...

image

dnil:
Ok, lets get it over with then. We can discuss this (and making changes to CG yaml writing before making sure the downstream apps can handle it) with @henrikstranneheim later.. 😉

PR #1734, opened 2020-02-20
Fix observations bug in v4.12

### LINKED ISSUE #1730: Local observations displays only the current family
Local observations display only the family I'm looking in, and in one case the family name appears multiple times.

Example 1: the variant is said to have been seen 45 times before and the family I'm looking in is listed 16 times. This is a re-run so I'd expect to see my family once.
case 19173, variant in SLC7A2 with rank 45

Example 2: Helene looked at a variant yesterday with 4 local observations and different family names were listed. Today in the same variant only the ID of the family she's looking at is displayed (but it still says 4 local observations).

image

northwestwitch:
bug fix #1730 

**How to test**:
1. Install the branch on stage (requires having some actual observations in loqus stage)
1. Make sure that cases where variant was observed are NOT the same as variant's case, repeated all over, like this:

image

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Cancer variant view has cleaner table data entries for "N/A" data
 - Cancer variants show correct alt/ref reads mirroring alt frequency now
 - Always load all clinical STR variants even if a region load is attempted - index may be missing
+- Came case repetition in variant local observations
 
 
 ### Changed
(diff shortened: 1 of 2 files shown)


northwestwitch:
I think this might solve the problem, but I don't know how to test it!

moonso:
This should be fixed in the controller and not the view I think

dnil:
Please fix if you wish @moonso; otw I'll test and merge! I', a little too low on blood sugar right now to delve into it.

moonso:
Ok give me a minute

moonso:
Aha sorry now I understand this fix!

moonso:
Did you try this on stage? Just to be sure...

dnil:
Nope - feel free! I went for food in the meantime.. 😊

northwestwitch:
Ok, @moonso and I tested this PR on his computer, with a loqus db instance containing observations. Looks like it's displaying different cases names in observations. So it should probably work!

northwestwitch:
I'm going to merge as soon as Travis approves

PR #1735, opened 2020-02-20
Fix pin cancer var

dnil:
A patch over the cancer pinned variant hgvs view. Appears some details on the genes there (and for STRs) is missing. We should return to this, but this at least just doesn't leave an empty field for pinned variants.

Screenshot 2020-02-20 at 12 48 27

Screenshot 2020-02-20 at 12 48 17

**How to test**:
1. Pin a variant of each type (cancer, cancer_sv, str, snv. sv) and make sure they render reasonably on the case page.

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -22,10 +22,17 @@
                     {% for gene in variant.genes %}
                       {% if gene.hgvs_identifier %}
                         {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
-                      {% else %}
-                        {{ "" if display_genes.append(gene.hgnc_symbol) }}
+                      {% elif gene.hgnc_symbol %}
+                          {{ "" if display_genes.append(gene.hgnc_symbol) }}
+                      {% elif gene.hgvs_identifier and gene.hgnc_id %}
+                          {{ "" if display_genes.append( gene.hgnc_id|string + ' ' + gene.hgvs_identifier) }}
                       {% endif %}
                     {% endfor %}
+
+                    {% if not display_genes %}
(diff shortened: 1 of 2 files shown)


dnil:
Mental note to fix the CHANGELOG before release 4.12.2..

northwestwitch:
Is this on stage alerady?

northwestwitch:
Testing on stage now. Tried to pin and mark as causative a cancer variant (the same). I've got 2 different representations at the case level:

image

Same for strs:

image



This should be quite easy to fix.

NON-cancer snvs look fine:
image

dnil:
Right; I didn’t think about the causative part. I thought we did that earlier.. Just a moment..

dnil:
Like so: 
Screenshot 2020-02-20 at 14 10 06

It should also really be fixed in the controller, but..

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -13,19 +13,26 @@
             <div class="row align-items-between align-items-center">
               <div class="col-10">
                 <i class="fa fa-check-circle-o"></i>
-		            {% if variant.category == "snv" %}
+		            {% if variant.category in ("snv", "cancer", "str") %}
                   <a href="{{ url_for('variant.variant',
```
northwestwitch: Wait a moment' won't this link be wrong if the variant is not a NON-cancer snv? then?

dnil: Hm, no?? The same variant page for all of the three, and same sv page for sv and cancer_sv?

northwestwitch: Yes, right, sorry! 👍

PR #1746, opened 2020-02-21
Side panel fixes

### LINKED ISSUE #1745: Improving / fixing report links on the case side panel
Things to do:
- [x] Recent delivery reports are missing. This is the most urgent thing to do
- [x] MultiQC's link should look like the other report links (same icon and it should open a new browser tab)
- [x] Coverage PDF link generates an empty PDF (HTML link works). PDF report creation from the HTML one works anyway, so it's still a bug but not terribly urgent
- [x] Coverage HTML link opens 2 tabs instead of one

northwestwitch:
Fix #1745

**How to test**:
1. Install on stage and check that all the points in #1745 are OK for a recent case having those reports available

**Expected outcome**:
The functionality should be working:
- [x] Recent delivery reports are missing. Tested with https://scout-stage.scilifelab.se/cust000/643594-300M, before (master doesn't show report)
- [x] MultiQC's link should look like the other report links (same icon and it should open a new browser tab)
- [x] Tried to fix a solution to the blank download PDF coverage report, nothing worked so I decided to remove the option. PDF report creation from the HTML one works anyway, so users can use that button instead!
- [x] Coverage HTML link opens 2 tabs instead of one

**Review:**
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Bug when adding a new gene to a panel
+- Restored missing recent delivery reports
+- Fixed style and links to other reports in case side panel
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


northwestwitch:
Standardize MultiQC link OK

image

And link opens:

image

northwestwitch:
Missing delivery report test, before (master):

image


And after (links work as well!):

image

northwestwitch:
This is done and ready for review!

PR #1749, opened 2020-02-24
Show bam file in Alamut

### LINKED ISSUE #1460: Scout for cancer workshop (2019-11-06) notes
**Important features (MVP):**
- Variant Ranking is important
- Scout for cancer and Scout for rare diseases on difference instances and also different institutes so no mix-up of rare cases and cancer ones.
- Introduce a badge for cancer in cases table and search for cancer cases in the cases search
- Chanjo report for cancer cases: modify depth of coverage thresholds.
- Define relevant phenotype groups, HPO terms and cohorts (custom at the institute level)
- Define well how the Ranking is done
- Swegen frequency alongside ExAC or other databases frequencies. Use "Frequencies" instead of ExAC in table header.
- Dump Swegen  frequencies in our database
- Cancer ACMG/AMP different from ACMG for classification of variants
- Pre-fill ACMG/AMP if possible, when possible. Is it feasible.
- Another button for showing SVs
- Wig tracks in igv
- Gene models can be removed in cancer (variant view)
-  Add Swegen in the table of Frequency (variant view)
- Gene specific databases (Braca tp53) could be added
- Investigate what N/A means in local observations, is it 0? Make it clearer in the pages
- GT call : Genotype (GT) is always ./. for tumors, fix it in another way?
- Add filter variants by chromosome and position
- Show genotype calls by caller
- Add institutes and accounts for users
- Add balsamic report to the reports that can be exported

**Nice to have features:**
- Flag variants as germline when info is available
- Ranking of variants based on HPO terms or OMIM diagnosis provided by users. Ranking performed directly in Scout.

### LINKED ISSUE #1621: alamut to show bam file
From https://github.com/Clinical-Genomics/scout/issues/1460

northwestwitch:
fix #1621

**I HAVE NO IDEA IF THIS WILL WORK, BUT IT MIGHT!! :)**

Bonus fix: improved link buttons layout from this:
image

to this:
image

**How to test**:
1. Install the branch on computer running Alamut on port 10000 (http://localhost:10000)
1. If a bam file is available go to a variant and click on Alamut button
1. Case in scout demo with bam file: https://scout-stage.scilifelab.se/cust000/643594-300M/
1. If only a MT bam is available, select first a MT variant and then click on the Alamut button

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


moonso:
I guess @dnil will be best suited to test this, I have never launched Alamut 🥘

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -586,27 +586,38 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
...
           {% if case.mt_bams or case.bam_files %}
-            <a href="{{ variant.alamut_link }}" class="btn btn-outline-secondary" target="_blank">Alamut</a>
+            {% set alamut_position = [ "GRCh38:" if case.genome_build == 38 ,  variant.chromosome, ":", variant.position, variant.reference, ">", variant.alternative]  %}
+            <form>
+            {% if variant.chromosome == "MT" and case.mt_bams %}
+              <input type="button" class="btn btn-outline-secondary" value='Alamut' onclick='httpGetLoadBam({{ alamut_position|join() }} ,
+                {{ case.mt_bams[0] }});' />
+            {% elif case.bam_files %} <!-- not a MT variant and bam file available -->
+            <input type="button" class="btn btn-outline-secondary" value='Alamut' onclick='httpGetLoadBam({{ alamut_position|join() }} ,
+              {{ case.bam_files[0] }});' />
```
dnil: Ah, but that is the file name only? This should point to the endpoint for the scout bam file streaming url, right?

northwestwitch:
I think it's a path on the server, Ill check better

dnil:
It is /remote/static, methinks! Cheers!

dnil:
DF55BBB8-BDD5-4730-8540-E6EAA06D9EF2

A very naive version of that didn’t go all the way - perhaps needs some more fields, or be in the context of that JavaScript request. But at least it seems the url is forwarded ok, so so far so good!

dnil:
There are probably a couple of scout stage log entries now that may give a hint..

northwestwitch:
Something like this?

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -586,27 +586,38 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
...
           {% if case.mt_bams or case.bam_files %}
-            <a href="{{ variant.alamut_link }}" class="btn btn-outline-secondary" target="_blank">Alamut</a>
+            {% set alamut_position = [ "GRCh38:" if case.genome_build == 38 ,  variant.chromosome, ":", variant.position, variant.reference, ">", variant.alternative]  %}
+            <form>
+            {% if variant.chromosome == "MT" and case.mt_bams %}
+              <input type="button" class="btn btn-outline-secondary" value='Alamut' onclick='httpGetLoadBam({{ alamut_position|join() }} ,
+                {{url_for("alignviewers.unindexed_remote_static", file=case.mt_bams[0])}} );' />
+            {% elif case.bam_files %} <!-- not a MT variant and bam file available -->
+            <input type="button" class="btn btn-outline-secondary" value='Alamut' onclick='httpGetLoadBam({{ alamut_position|join() }} ,
+              {{url_for("alignviewers.unindexed_remote_static", file=case.mt_bams[0])}} );' />
```
dnil: Oops; not the mt_bam here, right?
Also probably not the unindexed url for the whole file.. I wonder how this would work with auth? Does it pass the token in the url?

northwestwitch: > Does it pass the token in the url?
Not that I'm aware of.

The problem with using the other method is that it will pass just a part of the file, see the code:
Screenshot from 2020-02-26 08-33-56

I'm not sure alamut likes it.

I've fixed the path to the right bam for the non-MT link now

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -595,7 +595,7 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
                 {{url_for("alignviewers.unindexed_remote_static", file=case.mt_bams[0])}} );' />
             {% elif case.bam_files %} <!-- not a MT variant and bam file available -->
             <input type="button" class="btn btn-outline-secondary" value='Alamut' onclick='httpGetLoadBam({{ alamut_position|join() }} ,
-              {{url_for("alignviewers.unindexed_remote_static", file=case.mt_bams[0])}} );' />
+              {{url_for("alignviewers.unindexed_remote_static", file=case.bam_files[0])}} );' />
```
dnil: :+1:

scout/server/blueprints/variant/templates/variant/variant.html
dnil: We'll have to test I suppose: I'm not so sure Alamut will be able to reach it without a token, but lets see. But sure, lets start with the full file for testing. That won't work for whole genomes though, so if Alamut can't do ranged/some sort of streamed queries, we will likely have to check for the sequencing type before adding that to the url. Trying to dl a full bam via the hospital network means waiting for a few h and likely failing.. 😊

northwestwitch: If it won' work with whole genome then there is no much point. We could try directly with passing ranges. I'll fix it as soon as I finish something else I'm doing

dnil: Well, I guess the question came from the cancer side, where we have the tiny panels. I guess those could be dl:ed in full if need be, much like the mito dito..

parlar:
Hello!

I'm definitely not an expert but after I started debug logging it seems that alamut requesting ranges. Might be a misinterpretation, though.

Saw that it things did not work when reading the comments on this pull request. You probably understand this much better than I do, but could it work if bam + bai files were served using apache2 and mod_auth_token?

The tokens are generated from a "secret" string as well as a timestamp as md5 hashes with python. These are then shared with the apache2 server and forwarded to the client via scout.

The path (according to the link below) will look like this:

/<path>/dee0ed6174a894113d5e8f6c98f0e92b/43eaf9c5/file_to_protect.txt

Since it works when files are served directly by an apache2 server, this might work as well, perhaps?

https://code.google.com/archive/p/mod-auth-token/

parlar:
I realized that I did not read the docs properly on mod_auth_token. Since tokens are based on checksums of individual files, the bam and the bam.bai would get different tokens. Alamut will not be able to handle that. Sorry.

northwestwitch:
> I realized that I did not read the docs properly on mod_auth_token. Since tokens are based on checksums of individual files, the bam and the bam.bai would get different tokens. Alamut will not be able to handle that. Sorry.

No problem, and actually it's great that you understood a bit more about it. So, what do you suggest? Have you tried to see if the PR works as it is? Unfortunately I don't have any ALAMUT implementation that I can use so I'm kind of coding blindly. It't great to have another tester, apart from @dnil !

dnil:
Perhaps not modauth, but a token feels like the only reasonable way to go. Is there any option to have flask-login or its partner libs generate a time limited token for the user clicking the Alamut link, ideally tied to the files so as not to open for dl of arbitrary files. Otherwise it should be doable using itsdangerous directly, which flask anyway uses.

moonso:
I've been spending some time on moving from flask-login to [Authlib](https://docs.authlib.org/en/latest/basic/oauth2.html#client-auth-methods), however it is quite difficult for me. Anyway there you can have tokens, even though I do not really understand how it works. Perhaps we could have a session where we sit down together and I can show what I have done so far.

dnil:
authlib does look nice, and the ability to custom grants to an endpoint with tokens more or less like what we would need. The devil is as always in the details, but promising! 👍

parlar:
Great stuff! I'll take a look at Authlib as well.

northwestwitch:
I've applied for a 30-days demo license for Alamut visual, so I'll be able to test this PR on my own!

dnil:
They should give us a commission!

northwestwitch:
I've just received the license!! 👩‍💻

northwestwitch:
I've been spending a lot of time on this PR but still couldn't find a solution to this.

What works so far (kind of, because it opens an intermediate window sometimes asking you which gene you mean) is zooming in a chromosomal region.

I still have problems loading the alignment file, see error message:

image

If anybody wants to give it a try any help would be super appreciated!

northwestwitch:
Closing this. It was one of the most frustrating PRs of my career. We need to have more info from Alamut in order to understand what we should send to the software. We can re-open it at a later stage if we feel like giving it another try.

moonso:
Oh no! So you had to throw away all the work 😞

northwestwitch:
> Oh no! So you had to throw away all the work 😞

I know! But it's complicated to interact with software that lacks documentation and is not open source.

PR #1762, opened 2020-02-27
Fix coordinates filter bug

bjhall:
This PR fixes a bug causing the coordinates filter to override all other filters.

In SVfilters, if you select these filters: SVtype=DEL, Chromosome 3. This will show all variants on chr3 reagardless of SVtype. The query looks like this:
```
mongo query: {'case_id': 'xxx', 'category': 'sv', 'variant_type': 'clinical', '$and': [{'$or': [{'chromosome': '3'}, {'end_chrom': '3'}]}]}
```

This will fix that and make the query look like this:
```
mongo query: {'case_id': 'xxx', 'category': 'sv', 'variant_type': 'clinical', '$and': [{'$or': [{'chromosome': '3'}, {'end_chrom': '3'}]}, {'sub_category': {'$in': ['del']}}]}
```

I think this was caused by a typo where `if mongo_query.get("query"):` should have been `if mongo_query.get("$and"):`, but I don't understand the query building well enough to remove the old check so I just added an `and`.

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -268,7 +268,7 @@ class QueryHandler(object):
 
         # if chromosome coordinates exist in query, add them as first element of the mongo_query['$and']
         if coordinate_query:
-            if mongo_query.get("query") or mongo_query.get("$and"):
+            if mongo_query.get("$and"):
                 mongo_query["$and"] = coordinate_query + mongo_query["$and"]
             else:
                 mongo_query["$and"] = coordinate_query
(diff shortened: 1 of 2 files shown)


scout/adapter/mongo/query.py
```diff
@@ -268,7 +268,7 @@ def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
 
         # if chromosome coordinates exist in query, add them as first element of the mongo_query['$and']
         if coordinate_query:
-            if mongo_query.get("query"):
+            if mongo_query.get("query") or mongo_query.get("$and"):
```
moonso:
```suggestion
            if mongo_query.get("$and"):
```
It think we can safely remove the first statement here as it is not used

moonso: Hi @bjhall , thanks for this. Yes the query building part of the code is kind of a nightmare. Not sure how to improve this part. If there are any suggestions from Lund please add them :)

PR #1765, opened 2020-02-27
Use optional flask config

moonso:
This PR adds the functionality to use an alternative flask config when starting the server with `scout serve`

**How to test**:
1. Install the branch local
1. Create a new flask config by copying the existing one into a local dir, e.g. : ```cp scout/server/config.py local/flask.conf.py```
1. Edit the new config, like adding loqusdb
1. Start the development server with ```scout --flask-config local/flask.conf.py serve```

**Expected outcome**:
The server should now use settings form the alternative config

**Review:**
- [x] code approved by CR


northwestwitch:
Testing this PR. I've copied the demo config file to a local folder (/Users/chiararasi/config_test.py).
When I run `scout --flask-config /Users/chiararasi/config_test.py serve` 
I get:
Error: no such option: --flask-config

moonso:
I had written `--flask_config` instead of `--flask-config`. It should work now and is ready for merging

moonso:
So I would really like to merge this as it's affects my local testing environment quite a lot. Could you please have a quick look @northwestwitch ? It's tiny

northwestwitch:
Passing the config file as a parameter is used only for testing? Because if you need to make it available on a real server then you should modify the auto.py

northwestwitch:
And the functionality is not compatible with the `--demo` parameter, so it will only run the "real" database. Wouldn't be better to make it work also with the demo server?

northwestwitch:
> And the functionality doesn't look compatible with the `--demo` parameter, so it will only run the "real" database. Wouldn't be better to make it work also with the demo server?

northwestwitch:
> And the functionality doesn't look compatible with the --demo parameter, so it will only run the "real" database. Wouldn't be better to make it work also with the demo server?

Never mind, it accepts the demo parameter!

northwestwitch:
Looks like that if you run `scout --demo --flask-config /Users/chiararasi/config_test.py serve`

Then it doesn't recognize the --demo parameter and connects to the "real" database. It would be nice to fix it to work also with the demo one.

moonso:
No this is only a functionality to use when running the cli, not the server. In that case you use a environmental variable to point to the flask config. I don't really see why one would use it with demo but I can try to add that as well

northwestwitch:
> No this is only a functionality to use when running the cli, not the server. In that case you use a environmental variable to point to the flask config. I don't really see why one would use it with demo but I can try to add that as well

Well for instance I almost never run a real server while developing. So it would be good to setup a demo and being able to connect to it, while specifying other app parameters in a local file which is not IN the demo folder.

moonso:
Ok but then I guess that the `--demo` flag will have to override some parameters like database name. The `--demo`-flag implies that the database is called `scout-demo`. If you create a config with this database name you have a demo config...

northwestwitch:
> If you create a config with this database name you have a demo config...

I guess that would be the easies way. Never mind then, I'll approve the PR!

CHANGELOG.md
```diff
@@ -43,6 +44,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Add SMA view with SMN Copy Number data
 - Easier to assign OMIM diagnoses from case page
 - OMIM terms and specific OMIM term page
+- Option to use alternative flask config when running `scout serve`
```
northwestwitch: Could it be that you didn't remember that you already modified the changelog in this PR?

moonso:
Thank you @northwestwitch

PR #1775, opened 2020-03-05
Scout version 4.13

northwestwitch:
### Added
- ClinVar SNVs track in IGV
- Add SMA view with SMN Copy Number data
- Easier to assign OMIM diagnoses from case page
- OMIM terms and specific OMIM term page

### Fixed
- Bug when adding a new gene to a panel
- Restored missing recent delivery reports
- Fixed style and links to other reports in case side panel
- Deleting cases using display_name and institute not deleting its variants
- Fixed bug that caused coordinates filter to override other filters
- Fixed a problem with finding some INS in loqusdb
- Layout on SV page when local observations without cases are present
- Make scout compatible with the new HPO definition files from `http://compbio.charite.de/jenkins/`

**Review:**
- [x] code approved by HF, DN 😸


northwestwitch:
It doesn't feel so good when even an change in the changelog doesn't pass the cheks. Could we do something about it when you have time @moonso ?

northwestwitch:
Thanks!!

dnil:
Still check with prod before deploying I suppose..

northwestwitch:
> still check with prod before deploying I suppose..

I'll ask when is a good moment for deploying on hasta, but I'm sending the email that we are updating the interface this afternoon!

PR #1780, opened 2020-03-06
Parse correctly MIP 8 clnsig keys from vep-annotated CSQ field

### LINKED ISSUE #1777: CLin Var link missing
https://scout.scilifelab.se/cust003/14027/a2aefd060f80b751b3d227e587ffb00f

For several variant the CLinVar info in not appearing. 
Se ex. 
Thought the gnomAD link I can see that the variant is in ClinVar.

I've check some more cases and this seam to be a general issue

northwestwitch:
Fix #1777

**How to test**:
1. Upload a recent sample and make sure that clnsig info is parsed correctly

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
Thanks @dnil, but talking with @moonso we reasoned that it's the csq that has to be parsed (since it's coming from VEP), so I think I'll revert and add the parsing in there instead. It is indeed confusing!

dnil:
Yes, it is for the CSQ fields! I was confused, and thought that was what you added.. 😆

northwestwitch:
It's also confusing because the clnsig now is found in 2 places (INFO field and CSQ field) and if we want backward compatibility we should leave them both there, but I guess the CSQ has priority now so it should be parsed before and then if the INFO field, only if the clnsig fields are none.. Does it make any sense?

dnil:
Yes, that makes sense! Isn't that sort of precisely what we now do for the REVEL field?

northwestwitch:
> sn't that sort of precisely what we now do for the REVEL field?



> Yes, that makes sense! Isn't that sort of precisely what we now do for the REVEL field?

I think so!

dnil:
Looking at the delivery reports, the switch to MIP8 must have happened around 20200220, so at least cases before this date should not have been affected.

scout/parse/variant/transcript.py
```diff
@@ -200,9 +200,17 @@ def parse_transcripts(raw_transcripts, allele=None):
             LOG.debug("Something went wrong when parsing frequencies")
             LOG.debug("Only splitted and normalised VEP v90+ is supported")
 
+        if entry.get("CLINVAR_CLNVID"):
+            transcript["clinvar_clnvid"] = entry["CLINVAR_CLNVID"].split("&")
```
dnil: Hm, I thought only CLNSIG was an &-separated list?

northwestwitch:
I wrote something that might work, but I think we should test it with some real case!

scout/parse/variant/transcript.py
northwestwitch: I don't know. In theory it should just have one value here, but I don't know

dnil:
Almost there! Tested uploading the index case to scout-stage using present branch. The end result looks tentatively like a step forward, but not quite there yet:
Screenshot 2020-03-06 at 17 08 11

Clinsig other in scout, Uncertain significance in Clinvar, as well as in the MIP8 VCF (Uncertain_significance). Plenty of warnings of the following kind when loading from cg:
```
Clinsig value _other does not have an internal representation
Clinsig value Benign does not have an internal representation
```

scout/constants/clnsig.py
```diff
@@ -28,6 +29,7 @@
     "benign": 2,
     "likely_benign": 3,
     "conflicting_interpretations_of_pathogenicity": 8,
+    "conflicting_interpretations": 8,
```
northwestwitch: This is possibly fixing a bug we never noticed before. Some variants have `conflicting_interpretations` instead of `conflicting_interpretations_of_pathogenicity` as clnsig value and since the option was not a valid key these accessions were not saved in the variant object

northwestwitch:
@dnil I made some fixes and possibly fixed the problem. Give it a try when you have time!

scout/parse/variant/clnsig.py
```diff
@@ -5,65 +5,71 @@
...
+            sig = transcripts[0].get("clinvar_clnsig")
+            revstat = transcripts[0].get("clinvar_revstat")
 
     # There are some versions where clinvar uses integers to represent terms
     if isinstance(acc, int) or acc.isdigit():
         revstat_groups = []
         if revstat:
-            revstat_groups = [rev.lstrip("_") for rev in revstat.split(",")]
+            revstat_groups = [
+                rev.lstrip("_") for rev in revstat.replace("&", ",").split(",")
```
dnil: :+1:

scout/parse/variant/clnsig.py
```diff
@@ -5,65 +5,71 @@
...
         clnsig_accsessions(list(dict)): A list with clnsig accessions
     """
+
     transcripts = transcripts or []
     acc = variant.INFO.get("CLNACC", variant.INFO.get("CLNVID", ""))
     sig = variant.INFO.get("CLNSIG", "").lower()
     revstat = variant.INFO.get("CLNREVSTAT", "").lower()
 
-    clnsig_accsessions = []
+    clnsig_accessions = []
```
dnil: :+1: 😄

scout/parse/variant/clnsig.py
```diff
@@ -5,65 +5,71 @@
...
 
-    if not acc:
-        if transcripts:
+    if not acc and transcripts:
```
dnil: I trust you, but that line is beyond my knowledge of python operator precedence. I would have had to throw in a parenthesis - or google it. I guess `not` binds harder / has higher precedence?

tests/parse/test_parse_clnsig.py
```diff
@@ -227,3 +228,42 @@ def test_is_pathogenic_no_annotation(cyvcf2_variant):
 
     ## THEN assert that The variant should be loaded
     assert pathogenic == False
+
+
+def test_parse_clinsig_vep97(
```
dnil: Perhaps some kind of test with both INFO and CSQ entries as well, since a lot of the logic above is focused on working around that?

scout/parse/variant/clnsig.py
northwestwitch: I'll add the parentheses! :)

tests/parse/test_parse_clnsig.py
northwestwitch: I could add a variant with a complicated clinvar situation to the test file and check that is parsed into something we expect!

northwestwitch: I've introduced the check that variant with vep97 is loaded and contains the right clinvar fields along with REVEL and the rest

northwestwitch:
I think the code now works but I'm not really sure that all the pathogenic are loaded. There was some code at this regard in scout/parse/variant/clnsig.py (function is_pathogenic), but since it calls the parse_clnsig without passing the transcripts I'm afraid that everything will be non pathogenic. I have to check!

scout/parse/variant/clnsig.py
```diff
@@ -76,38 +82,50 @@ def parse_clnsig(variant, transcripts=None):
...
+
     Returns:
         bool: If variant should be loaded based on clinvar or not
     """
+
     load_categories = set(
         [
             "pathogenic",
             "likely_pathogenic",
-            "conflicting_interpretations_of_pathogenecity",
```
northwestwitch: This typo made it so that all the conflicting interpretation were NOT loaded no matter what before.

northwestwitch:
I think I'm done with this PR now!

dnil:
Closer and closer... 
Screenshot 2020-03-09 at 16 43 21

dnil:
Should be CLINSIG Uncertain significance I guess; somewhere the evaluation criteria must have snuck in..

northwestwitch:
> Uncertain significance I guess; somewhere the evaluation criteria must have snuck in..

I'll check, thanks!

scout/parse/variant/clnsig.py
dnil: 😑 
Well found anyway!

scout/constants/clnsig.py
```diff
@@ -15,9 +15,11 @@
     13: "association not found",
     255: "other",
 }
+
 REV_CLINSIG_MAP = {
     "other": 255,
     "not_provided": 1,
+    "no_assertion_criteria_provided": 1,
```
dnil: Hold on, isn't that supposed to be a REVSTAT rather? I don't think you will find it in CLINSIG!

northwestwitch: Ok but then it should be safe to remove it from here!

dnil: Yes, I think so!

northwestwitch: You are right, I don't know why I added it here! Gone now!

dnil:
Looking good:

Screenshot 2020-03-10 at 13 57 52

tests/load/test_load_variant.py
```diff
@@ -58,12 +59,19 @@ def test_load_vep97_parsed_variant(
...
     assert isinstance(variant["revel_score"], float)
 
+    # conservation fields
     for key, value in variant["conservation"].items():
         assert value == ["NotConserved"]
 
+    # clinvar fields
+    assert isinstance(variant["clnsig"][0]["accession"], int)
+    assert variant["clnsig"][0]["value"] in REV_CLINSIG_MAP  # can be str or int
+    assert variant["clnsig"][0]["revstat"] in REV_CLINSIG_MAP  # str
```
northwestwitch: I don't like this, I'll change it

dnil: Yes, in particular the latter I would say is wrong? We dont really encode the revstats, do we? They go through unadjusted.

northwestwitch: Yes, I meant the latter!

northwestwitch:
Sooo... Merging this in 3, 2, ..

northwestwitch:
Thank you so much for the review and the assistance with the tests @dnil!!!

PR #1796, opened 2020-03-13
Feature remove pedigree 200313

mikaell:

**How to test**:
1. how to install it
Check out and install Scout as usual.

- load one cancer config
`scout --demo load case demo/cancer.load_config.yaml  -u`

- load one ordinary config
`scout --demo load case demo/643594.config.yaml  -u`

With two loaded cases, examine and compare them. Note that no Pedigree panel is displayed in the cancer case.

**Expected outcome**:

Screenshot 2020-03-13 at 11 50 03

Screenshot 2020-03-13 at 11 50 36

**Review:**
- [x] code approved by @hassanfa 
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -86,7 +86,7 @@
         </div>
         {% endif %}
 
-        <div class="col">
+        <div class="{% if case.track != 'cancer' %}col-8{%else%}col-12{%endif%}">
           {{ synopsis_panel() }}
           {{ comments_panel(institute, case, current_user, comments) }}
         </div>
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -76,6 +76,7 @@
       </div>
 
       <div class="row">
+        {% if case.track != 'cancer' %}
         <div class="col-md-4">
           {% if case.madeline_info and case.individuals|length > 1 %}
             {{ pedigree_panel(case) }}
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -76,19 +76,22 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
         <div class="col-md-4">
           {% if case.madeline_info and case.individuals|length > 1 %}
             {{ pedigree_panel(case) }}
           {% else %}
             <p>No pedigree picture available.</p>
           {% endif %}
         </div>
+        {% endif %}
 
-        <div class="col-md-8">
```
mikaell: Replacing `col-md-8` with `col` to make the lines expand the whole column. Are we using the other screen layouts? (small, medium, large, XL)

hassanfa: I don't know about this. Maybe @northwestwitch can chime in?

dnil: Agreed - @northwestwitch is the guru there! One way to think about it is that the sizes refer to the smallest size intended. And scout is generally not designed for the very small devices.

Some docs here https://getbootstrap.com/docs/4.4/layout/grid and more generally here https://getbootstrap.com/docs/4.4/layout/overview/#responsive.

northwestwitch: I'll take a look as soon as I finish reviewing another PR!

northwestwitch: Haha definitely not the guru, I just google all the time! 😄 
We generally use the layouts mentioned by @dnil. It works already with the code that you wrote, but if you want to be super sure you could write col-12 to force the column to stretch over the entire line!

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -76,19 +76,22 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
+        <div class="col">
```
northwestwitch: You could use a conditional here, like col-12 if it's cancer otherwise col-8

northwestwitch:
```suggestion
        <div class="{% if case.track != 'cancer' %}col-8{%else%}col-12{%endif%}"">
```

northwestwitch:
I take it as a yes!

PR #1800, opened 2020-03-16
Fix #1792 - assume GATK caller.

### LINKED ISSUE #1792: "GATK Pass" not visible in Scout
Below the coverage data for an SNV/indel i Scout, we could usually see "GATK Pass" etc. In the GMCK-RD set rules for reporting a variant without having to verify it with a secondary method, one of the criterias are that we have to see GATK Pass i Scout. Hence, we need you to add it to the Scout view again on the variant view. Thanks!!

dnil:
Hm, I thought I made a PR with this? Lost in translation somewhere? Anyway, this is a fallback fix for #1792, until MIP is back on track on the issue again, but also a potentially welcome addition for other users running GATK only on the SNV side.

In principle it is ready for review, but if you wait a test or two may appear. 😊 

**How to test**:
1. load SNV variants without INFO.FOUND_IN or INFO.set information about caller set and note that GATK pass/filer status is evident on the variants.

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by DN


dnil:
> Code wise looks good, if you have tested it and it works I think it's fine to merge! 👍

Not well.. Now it's better. Previous POT1:
Screenshot 2020-03-17 at 08 25 40
Now with GATK fallback:
Screenshot 2020-03-17 at 09 45 20

I also removed "INFO:set" for a demo variant with a failing filter:
Screenshot 2020-03-17 at 09 47 43
with this PR:
Screenshot 2020-03-17 at 09 50 28

dnil:
Since @henrikstranneheim described a nice workaround in #1792, I don't think we need to reload any cases, but we should remember to communicated this on the next (virtual!) Scout user meeting/workshop.

dnil:
Ok, merging this in 3..

PR #1805, opened 2020-03-18
Varsome

ninanorgren:
This PR adds a button with a link to Varsome specific for that transcript and variant

**How to test**:
1. click on varsome button for specific variant to get to a varsome url

**Expected outcome**:
If the variant is within varsomes search limits a description of the variant should be displayed on the varsome page

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/links.py
```diff
@@ -266,6 +267,12 @@ def smart(smart_domain):
     return link.format(smart_domain)
 
 
+def varsome(build, refseq_id, protein_sequence_name):
```
northwestwitch: It would be nice to handle the situation in which refseq_id and /or protein_sequence_name are missing. In that case the fuction should return None I think.

Another observation: default build passed is 37, which for the same variant as above, will convert the link in something like `https://varsome.com/variant/hg37/NM_015450%3Ac.903G>T`, that in turn is converted into `https://varsome.com/variant/hg19/NM_015450%3Ac.903G>T` by Varsome.

I guess it's OK.

Have you tested the **Varsome link created for variants from cases with genome build 38**? Because I don't think the right build is propagated to this function. It's not your fault of course because the error must have been there from before, but could you try to fix it?

Tip: You could simulate a case with genome build 38 by modifying one of the demo config files, for instance `scout/demo/643595.config.yaml`, writing `human_genome_build: 38`

northwestwitch: As soon as we merge #1808 this problem will be fixed, see that PR!

northwestwitch: Problem fixed. You might want to rebase this PR perhaps!

ninanorgren: I added so the function returns None if missing refseq id or protein sequence name now. I also had to add the genome build as input to the the function  add_gene_info  in the file server/blueprints/variant/utils.py to make the hg38 build work properly.

ninanorgren:
@northwestwitch I think I managed to fix the stuff that you wanted. Let me know if it works or if more is needed :)

northwestwitch:
Thanks @ninanorgren! Your branch history is a bit weird, would you mind rebase with our current master? I am afraid that if we merge it it will cause some unexpected consequence in our repo

ninanorgren:
I tried rebasing, but I think I failed miserably. Sorry about that @northwestwitch ... Did it make any difference?

northwestwitch:
> I tried rebasing, but I think I failed miserably. Sorry about that @northwestwitch ... Did it make any difference?

Haha I think it's much worse now!! 🤣 
Don't worry I did like this plenty of times!

Whenever I'm in this situation I just go back to my own master (your repo) and pull whatever there is new on the main repo (our repo), then I create a new branch (on your repo) and start afresh. The PR was not so big so maybe it's worth to do the same here?

ninanorgren:
That sounds like a good suggestion :) I'll try that tomorrow

ninanorgren:
I opened a new PR, [#1860](https://github.com/Clinical-Genomics/scout/pull/1860). Fingers crosssed that works better @northwestwitch

dnil:
The new one looks a lot better, though it is still a bit complicated. Waiting for @moonso to add you as a collaborator.. In the meantime, lets at least close this one since the functionality is now in #1860.

PR #1806, opened 2020-03-18
Avoid server error on variant page when VCF file is not available

bjhall:
This PR fixes a bug which causes a server error when trying to view a variant overview page (both for SNVs and SVs) while the VCF file is not currently accessible. 

I think this bug has been present since this commit https://github.com/Clinical-Genomics/scout/commit/2769b285354790077aa86e692e25b9cf9c41be29#diff-48cf522b95b5383c87a80b36ce46df9cL106 fixed the typo of "crom" to "chrom" in scout/server/utils.py, making the code work as intended, and thereby ironically causing this bug.

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -667,7 +667,12 @@ class VariantHandler(VariantLoader):
         if not variant_file:
             raise SyntaxError("Vcf file does not seem to exist")
 
-        vcf_obj = VCF(variant_file)
+        try:
+            vcf_obj = VCF(variant_file)
+        except:
+            LOG.error("Cannot read VCF: {0}".format(variant_file))
+            return None
+
         region = ""
 
         if gene_obj:


bjhall:
Obviously I don't know how to use exceptions in Python. Someone please tell the right way to do this...

moonso:
I'm happy to help. Could you explain a little more when this error happens and why?
You should catch the particular exception(s) that can be raised there, perhaps something like:

```python
try:
    vcf_obj = VCF(vcf_path)
except FileNotFoundError:
    do_something()
except cyvcf2.SomeException:
    perhaps_something(else)
```

bjhall:
The error happens when I click on any variant, SNV or SV, in the variant list to bring up the variant details page, if the VCF is not currently accessible on the file system. I get and 500 server error page. 

Normally the file is available but for instance if the storage server is down for some reason I think it is nice not to throw an error page when trying to view a variant.

The reason it fails is that it is trying to create the reduced VCF for IGV.js

moonso:
ok then I understand. So if you just specify what exception to catch we can merge it

moonso:
Hehe cyvcf2 raises `Exception`. Not so nice. Could we raise an exception there like we do a few lines up in the function. And then catch theese when this is called? I just saw that we got some problems with this as well. I can fix that if you like

moonso:
I hijacked this in #1815 , please have a look a thanks for fixing this

bjhall:
Thanks for fixing this properly @moonso!  I'll close and delete this PR.

PR #1807, opened 2020-03-18
Str fix 1803

dnil:

**How to test**:
- [x] go back and forth between variants and variant on all variant category pages (also cancer, cancer SV) and check that variant level nav bar navigation points back to the correct variants page
- [x] dito for suspects and causatives
- [x] note that a pinned or causative STR now shows locus name on the case page
Screenshot 2020-03-18 at 15 24 46

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -1,5 +1,41 @@
 {% from "variants/utils.html" import tier_cell %}
 
+{% macro pretty_link_variant(variant, case) %}
+{# Returns human readable links to the corresponding variant page #}
+
+  <a href="{{ url_for('variant.variant',
+                   institute_id=variant.institute,
+                   case_name=case.display_name,
+                   variant_id=variant._id) }}">
+  {% if variant.category in "str" %}
+  {{ variant.str_repid }} {{ variant.alternative }}
+  {% elif variant.category in ("snv", "cancer") %}
+  {% set display_genes = [] %}
+    {% for gene in variant.genes %}
+      {% if gene.hgvs_identifier and gene.hgnc_symbol  %}
+        {{ "" if display_genes.append(gene.hgnc_symbol + ' ' + gene.hgvs_identifier) }}
+      {% elif gene.hgnc_symbol %}
+          {{ "" if display_genes.append(gene.hgnc_symbol) }}
+      {% elif gene.hgvs_identifier and gene.hgnc_id %}
+          {{ "" if display_genes.append( gene.hgnc_id|string + ' ' + gene.hgvs_identifier) }}
+      {% endif %}
+    {% endfor %}
+
+    {% if not display_genes %}
+        {{ "" if display_genes.append( variant.simple_id ) }}
+    {% endif %}
+
+    {{ display_genes|join(", ") }}
+  {% else %}
+  <a href="{{ url_for('variant.sv_variant',
+                            institute_id=variant.institute,
+                            case_name=case.display_name,
+                            variant_id=variant._id) }}">
+  {{ variant.sub_category|upper }}({{ variant.chromosome }}{{ variant.cytoband_start }}-{{ variant.chromosome }}{{ variant.cytoband_end }})
+  {% endif %}
+  </a>  
+{% endmacro %}
+
 {% macro causatives_list(causatives, partial_causatives, institute, case) %}
   <div class="card panel-default">
     <div data-toggle='tooltip' class="panel-heading" title="Displays all variants that have been marked causative for this case">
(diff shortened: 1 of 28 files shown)


dnil:
Still an oops for cancer-SVs. Will fix!

northwestwitch:
> Still an oops for cancer-SVs. Will fix!

I can review when you feel ready!

dnil:
Thanks; I should have some time to finish this. Also adding a fix for #1809 to it.

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -13,12 +13,25 @@
...
+                                   variant_id=variant._id, str='yes') }}">
+                  {{ variant.str_repid }} {{ variant.alternative }}
+                {% elif variant.category in ("snv", "cancer") %}
+                  {% if variant.category in "cancer" %}
+                    <a href="{{ url_for('variant.variant',
                                       institute_id=variant.institute,
                                       case_name=case.display_name,
-                                      variant_id=variant._id) }}">
-                    {% set display_genes = [] %}
+                                      variant_id=variant._id, cancer='yes') }}">
```
northwestwitch: What about putting the conditional only in the last part, so you could avoid repetition of code. Something like:

```
{% if variant.category in ("snv", "cancer", "str") %}

<a href="{{ url_for('variant.variant',
                                   institute_id=variant.institute,
                                   case_name=case.display_name,
                                   variant_id=variant._id
{% if  variant.category ==  "str"%}
, str='yes'
{% elif variant.category ==  "cancer" %}
, cancer='yes'
{% endif %}
)}}">
{% set display_genes = [] %}
```

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -154,27 +167,40 @@
                 <div class="form-group row justify-content-between align-items-center">
                 <div class="ml-3">
                   <i class="fa fa-bookmark"></i>
-                  {% if variant.category in ("snv", "cancer", "str") %}
+                  {% if variant.category in "str" %}
```
northwestwitch: Same here?

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -51,10 +51,22 @@
           {% for overlapping_variant in overlapping_vars %}
             <tr>
               <td>
-                <a href="{{url_for('variant.sv_variant', institute_id=institute._id,
+                {% if overlapping_variant.category in ("sv", "cancer_sv") %}
```
northwestwitch: Same here?

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -553,6 +553,14 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
   <a class="variants-row-item flex-small layout"
       href="{{ url_for('variant.sv_variant', institute_id=institute._id,
                        case_name=case.display_name, variant_id=variant._id) }}">
+{% elif variant.category in "str" %}
```
northwestwitch: Same here

scout/server/blueprints/cases/templates/cases/utils.html
dnil: One would love to, right. Only the situation is a bit complex. :wink: The var passing is already in a jinja2 statement, and we don't do like two passes of jinja, so no nested use is allowed.

If it is hard to see at a glance, try thinking about if you could close the `{{}}` block early, say after `variant._id` above. But jinja evaluation would still be trying to execute the url_for, but needing all the arguments to be in place for that.

To complicate things, the offending variable parameter here becomes the function call argument name, so I have no use of the variable directly either in the jinja/python env. If the construction was like `url_for(variant_page_type=variant.category` that would be fine. I considered getting at it using e.g. `eval()`, but I don't think the linting will love it - very soon this will look like a good old perl program.

What I would suggest is to rewrite the case and variant pages, utils and views, and dynamically determine what page to display using `variant.category` instead of passing keywords like `cancer=yes`. This is obviously tricky for the variants page, and here I would suggest changing to passing the variant type as the argument rather than argument name.

dnil: I rant too much and write too little code; I'll check if it would work!

scout/server/blueprints/variant/templates/variant/utils.html
dnil: Yes! ;-)

scout/server/blueprints/variants/templates/variants/utils.html
dnil: Yes! 😄 :+1:

dnil:
Thank you: great review - a little bit of resistance is very useful. I think that turned out a lot better! 😄

scout/server/blueprints/variant/views.py
```diff
@@ -143,7 +138,7 @@ def variant_update(institute_id, case_name, variant_id):
     if manual_rank:
         try:
             new_manual_rank = int(manual_rank) if manual_rank != "-1" else None
-        except:
+        except ValueError:
```
northwestwitch: Thanks for fixing this as well!

scout/server/blueprints/variant/controllers.py
```diff
@@ -208,6 +215,7 @@ def variant(
         "institute": institute_obj,
         "case": case_obj,
         "variant": variant_obj,
+        variant_category: True,
```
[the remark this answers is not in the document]
dnil: Ah, that is to cater for the variants view. This boils down to setting data[“cancer”]=True that was previously set with multiple if args.cancer == ‘yes’ etc.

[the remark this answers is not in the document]
northwestwitch: Of couse, you need to pass those parameters somehow! 👍

PR #1812, opened 2020-03-19
Re-upload case variants with the option to keep custom actions (dismissed, manual rank, mosaic, acmg, comments)

### LINKED ISSUE #1649: Decide to keep or not the dismissed variants upon reruns
This was a topic of discussion but I had the impression we didn't reach an agreement during today's meeting. Any ideas?

northwestwitch:
This PR adds the functionality to keep the custom manual tags assigned by a user when the variants of a case are re-uploaded (fix #1649)

Custom actions that are kept upon variant re-upload are:
- manual_rank
- dismiss_variant
- mosaic_tags
- cancer_tier
- acmg_classification
- comments

**How to test**:
1. On a local instance of Scout choose one or more variants
1. Assign the following manual actions to the variant(s):
- [x] manual rank (even in two separate occasions, but remember the last rank assigned)
- [x] dismiss tag (one or more values, even in two separate occasions, but remember the assigned values)
- [x] mosaic tag (one or more values, even in two separate occasions, but remember the assigned values)
- [x] cancer_tier
- [x] acmg classification (manual)
- [x] acmg classification (via scout app)
1. From master branch use the following CLI command to re-upload the SNVs
`scout --demo load variants internal_id -f --snv`
and you'll see that the manual actions disappear
1. Switch to this branch and run the same command :
`scout --demo load variants internal_id -f --snv`
You should see that the actions from the old variants are propagated into the new ones.

Testing the propagation of comments requires reupload of a case on scout stage:
1. install the branch on hasta stage
1. Add a local and a global comment for a variant of a case on scout stage (clinical-db stage)
1. reupload the variants of a stage case
`scout --demo load variants internal_id -f --snv`
1. Make sure that the comments are present in the newly loaded variant
- [x] global comments if variant ID changes (Automatic tests)
- [x] local comments if variant ID changes (Automatic tests)

**Expected outcome**:

**Review:**
- [x] code approved by DNil
- [x] tests executed by  DNil, CR


northwestwitch:
This is ready for review now!

moonso:
Good I think we should assign everything:)

scout/commands/load/variants.py
```diff
@@ -42,11 +41,17 @@
     help="Specify the rank score treshold",
     show_default=True,
 )
+@click.option("-f", "--force", is_flag=True, help="upload without request")
+@click.option(
+    "-k",
+    "--keep-tags",
```
dnil: I think this should be the default behaviour!
```suggestion
    "--keep-tags/--no-keep-tags', default=True,",
```

dnil: Or make the flag `--kill-old-user-tags`.. 😸

scout/commands/load/variants.py
```diff
@@ -151,7 +163,7 @@ def variants(
                 start=start,
                 end=end,
                 gene_obj=gene_obj,
-                build=case_obj["genome_build"]
+                build=case_obj["genome_build"],
```
dnil: Hmm... ⬛️

tests/adapter/mongo/test_case_handling.py
```diff
@@ -509,7 +509,7 @@ def test_update_case_rerun_status(adapter, case_obj, institute_obj, user_obj):
     adapter.request_rerun(institute_obj, res, user_obj, "blank")
     res = adapter.case(case_obj["_id"])
     # THEN rerun_requested is flagged
-    assert res["rerun_requested"] == True
+    assert res["rerun_requested"] is True
```
dnil: :+1:

tests/adapter/mongo/test_case_handling.py
```diff
@@ -643,3 +643,141 @@ def test_get_cases_solved_since(
     # WHEN querying for cases solved within 1 day
     # THEN one case is found
     assert len([case for case in adapter.cases(finished=True, within_days=1)]) == 1
+
+
+def test_keep_manual_rank_tag_after_reupload(
```
dnil: :+1:

scout/adapter/mongo/case.py
```diff
@@ -826,6 +826,114 @@ def case_sanger_variants(self, case_id):
...
+
+            if new_var is None:  # same var is no more among case variants, skip it
+                continue
+
+            for tag in list(
+                updated_variants.keys()
+            ):  # manual_rank, dismiss_variant, mosaic_tags
+                if old_var.get(tag):  # tag new variant accordingly
+
+                    LOG.error("FOUND TAG {} in a variant!!!".format(tag))
```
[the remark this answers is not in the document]
northwestwitch: Oh YEs! Thanks!

scout/adapter/mongo/case.py
```diff
@@ -826,6 +826,114 @@ def case_sanger_variants(self, case_id):
...
+
+            # search for the same variant in newly uploaded vars for this case
+            display_name = old_var["display_name"]
+
+            new_var = self.variant_collection.find_one(
+                {"case_id": case_obj["_id"], "display_name": display_name}
+            )
+
+            if new_var is None:  # same var is no more among case variants, skip it
+                continue
```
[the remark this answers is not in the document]
northwestwitch: Sure!

scout/adapter/mongo/case.py
```diff
@@ -826,6 +826,114 @@ def case_sanger_variants(self, case_id):
...
+                    if user_obj is None:
+                        continue
+
+                    # create a link to the new variant for the events
+                    link = "/{0}/{1}/{2}".format(
+                        new_var["institute"], case_obj["display_name"], new_var["_id"]
+                    )
+
+                    if tag == "manual_rank":
+                        updated_variant = self.update_manual_rank(
```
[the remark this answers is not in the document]
northwestwitch: I agree!

scout/adapter/mongo/variant.py
```diff
@@ -267,10 +267,15 @@ def variant(
 
         if variant_obj:
             if case_obj:
-                variant_obj = self.add_gene_info(variant_obj=variant_obj, gene_panels=gene_panels,
-                    build=case_obj["genome_build"])
+                variant_obj = self.add_gene_info(
```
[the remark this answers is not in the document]
northwestwitch: Probably

scout/commands/load/variants.py
northwestwitch: > I think this should be the default behaviour!

I think that most users would agree. I'll make it default!

scout/commands/load/variants.py
northwestwitch: 🙄

northwestwitch:
I've included your suggestions in the code @dnil , thanks for the review!

dnil:
A random test thing:
- ACMG classifications don't show on the cell_rank after reload, though present on the variant.

dnil:
This is great btw - will be very useful if we would ever have a bug affecting things that need reloading! 😅

northwestwitch:
> * ns don't show on the cell_rank after reload, though present on the variant.



> This is great btw - will be very useful if we would ever have a bug affecting things that need reloading! 😅

If we don't re-upload the entire case

northwestwitch:
> This is great btw - will be very useful if we would ever have a bug affecting things that need reloading! 😅

Oh no, I'll check!

northwestwitch:
> > ACMG classifications don't show on the cell_rank after reload, though present on the variant.
> 
> Oh no, I'll check!

Ah but that is because this PR doesn't save the manual ACMG, just the tags:

image

Shall I include also the ACMG in this PR?

dnil:
Well, the comments and ACMG sort of work out of the box, since they don't store much on the variant collection, but apparently there is something in the variants view that doesn't work. Perhaps just check what it is and if it would be simple to fix?

northwestwitch:
> Well, the comments and ACMG sort of work out of the box, since they don't store much on the variant collection, but apparently there is something in the variants view that doesn't work. Perhaps just check what it is and if it would be simple to fix?

Sure, if we can have it fixed, why not!

dnil:
I just haven't checked; I imagine it is just one query. If it turns out to be complicated we can do it in another issue/PR!

northwestwitch:
@dnil that's because the acmg_classification disappears from the variant object when it's re-uploaded:
See last key value in this variant:

image

We could keep the manual acmg, it's not complicated, but what about those variants that get a classification via the app (acmg criteria), the info would be lost for those.

northwestwitch:
Ok, I'll modify this PR to propagate also ACMG classification, both manual and the one with the app criteria.

northwestwitch:
> Ok, I'll modify this PR to propagate also ACMG classification, both manual and the one with the app criteria.

But probably not today..

northwestwitch:
@dnil I think I'm done again with this PR.

Custom actions that are kept upon variant re-upload are:
- manual_rank
- dismiss_variant
- mosaic_tags
- cancer_tier
- acmg_classification
- comments

I've updated the instructions on how to test! Could you review again when you have time?

dnil:
Hi, I've tested again after some time:
- I first thought this was a bit broken: ACMG classifications were not re-displayed on cancer-variants page after reload.
Before:
Screenshot 2020-03-31 at 08 33 51
After reload:
Screenshot 2020-03-31 at 08 41 14
Event log looks unharmed, but the "acmg_classification" field on the variant appears to have been cleared. Seems ok for non-cancer snvs? But, alas, I had only forgotten the "--force" flag. When reloading with force, things look nice and dandy:
Screenshot 2020-03-31 at 09 19 47
Nothing to see here. 😆 

- [ ] An en passant discovery: deleted P/LP ACMG classifications, having been manually removed from the variant, are still shown on the variant view, but that is a problem with the current masterl. If there are no other issues with this PR, I'll make a new issue / PR to fix it.

scout/adapter/mongo/case.py
```diff
@@ -826,33 +827,41 @@ def case_sanger_variants(self, case_id):
...
                 'manual_rank' : [list of variant ids],
                 'dismiss_variant' : [list of variant ids],
-                'mosaic_tags' : [list of variant ids]
+                'mosaic_tags' : [list of variant ids],
+                'cancer_tier': [list of variant ids],
+                'acmg_classification': [list of variant ids]
+                'is_commented': [list of variant ids]
         """
-        updated_variants = {"manual_rank": [], "dismiss_variant": [], "mosaic_tags": []}
+        updated_variants = {
```
dnil: :+1:

scout/adapter/mongo/case.py
```diff
@@ -897,41 +911,76 @@ def update_manual_tagged_variants(
                         new_var["institute"], case_obj["display_name"], new_var["_id"]
                     )
 
-                    if tag == "manual_rank":
+                    updated_variant = None
+
+                    if action == "manual_rank":
```
dnil: This could have been expressed more elegantly in c++. 😄 Should have been sth like action.update() that in turn deals with the polymorphic subtypes. This is hardly the only place we do this though.

tests/adapter/mongo/test_case_handling.py
```diff
@@ -770,14 +766,156 @@ def test_keep_mosaic_tags_after_reupload(
...
+        institute_obj=institute_obj, case_obj=case_obj, old_eval_variants=[updated_old],
     )
     assert updated_new_vars["mosaic_tags"] == ["new_id"]
 
     # and the new variant should have a the same mosaic tags
     test_variant = adapter.variant_collection.find_one({"_id": "new_id"})
     assert test_variant["mosaic_tags"] == [1, 3]
+
+
+def test_keep_cancer_tier_after_reupload(
```
dnil: :+1:

scout/commands/load/variants.py
```diff
@@ -43,10 +43,9 @@
 )
 @click.option("-f", "--force", is_flag=True, help="upload without request")
 @click.option(
-    "-k",
-    "--keep-tags",
-    is_flag=True,
-    help="Export manual user tags from old variants to the new",
+    "--keep-actions/--no-keep-actions",
+    default=True,
```
dnil: :+1:

dnil:
> Event log looks unharmed, but the "acmg_classification" field on the variant appears to have been cleared. Seems ok for non-cancer snvs? But, alas, I had only forgotten the "--force" flag. When reloading with force, things look nice and dandy:

Actually, the force flag has nothing to do with it, upon rereading the code. It is only for forcing research if no request has been made. Perhaps the reloading of the exact same variant file, which I suppose may have changed slightly since my last setup then?

northwestwitch:
> * An en passant discovery: deleted P/LP ACMG classifications, having been manually removed from the variant, are still shown on the variant view, but that is a problem with the current masterl. If there are no other issues with this PR, I'll make a new issue / PR to fix it.

I'm looking at the code and thinking that this should be already be taken care of by the function that updates the new variant starting from the old one:

image

It should cover that case, right?

scout/adapter/mongo/case.py
northwestwitch: or in java (switch, case)

dnil:
Yeah, I tentatively think it behaves precisely like the rest of the system in that regard, meaning old classifications will be shown. 😅

northwestwitch:
> Yeah, I tentatively think it behaves precisely like the rest of the system in that regard, meaning old classifications will be shown.

You might be NOT wrong

dnil:
I'm sure this would be a small issue, but since you are at it: reloading doubles the comments. Could get interesting after a few rounds.. 😄 
Screenshot 2020-03-31 at 10 51 21

Also, manually classified P did survive but not a single clicked VUS..

northwestwitch:
> I'm sure this would be a small issue, but since you are at it: reloading doubles the comments. Could get interesting after a few rounds.. 😄

This is the expected behavior, the new variants get a comment as well. I don't know how to fix it

northwestwitch:
I think I've found and fixed the problem @dnil, same old bug. I never learn 😞

dnil:
Nice catch! I'm not convinced Python has found a good solution with this None-value: it reminds way too much of the situation with null pointers.

dnil:
Ok, tested ok on a reload with some different kinds of VUS:es in them! 

Let me just rinse and repeat once: that could have been most of the edge issues. I think I'll be back with a comment on the comments after a lunch break. 😸

northwestwitch:
> I think I'll be back with a comment on the comments after a lunch break

Any suggestion would be great!

dnil:
Yes, tested ok with some P and LP as well as deleted classifications as well. All are preserved well. A little too well in the latter case, but this PR is doing it right, its the variants view/utils that needs a little straightening out.

northwestwitch:
Thanks for the review and the accurate testing (I didn't catch the VUS error on my own). Do you think it is ready to be merged?

northwestwitch:
Ah no, the comments, sorry!

dnil:
Right, sorry, I was sidetracked. I'll have a look now. In principle I think it is good; the comments are a bit weird though. While possible as is, it will cause some confusion.

northwestwitch:
> Right, sorry, I was sidetracked. I'll have a look now. In principle I think it is good; the comments are a bit weird though. While possible as is, it will cause some confusion.

One thing we could do is not creating the comments for the new variants.

dnil:
I guess it all boils down to if the ids change. Since comments don't really store anything in the variant collection, but very neatly keep their payload in a content field, they should be good as long as we don't force clearing the event log. Unless they can get detached from their variant by some mechanism other than the variant actually changing coords or such.

scout/adapter/mongo/event.py
```diff
@@ -374,3 +374,50 @@ def comment(
...
+            old_var(Variant): the deleted variant
+            new_var(Variant): the new variant replacing old_var
+            institute_obj(dict): an institute object
+            case_obj(dict): a case object
+
+        Returns:
+            new_comments(int): the number of created comments
+        """
+        new_comments = 0
+
```
dnil:
```suggestion
if new_var[_id] == old_var[_id]:
    return new_comments
```

scout/adapter/mongo/case.py
```diff
@@ -826,6 +827,162 @@ def case_sanger_variants(self, case_id):
...
+                        updated_variant = self.update_acmg(
+                            institute_obj=institute_obj,
+                            case_obj=case_obj,
+                            user_obj=user_obj,
+                            link=link,
+                            variant_obj=new_var,
+                            acmg_str=str_classif,
+                        )
+
+                    if action == "is_commented":
```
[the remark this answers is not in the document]
dnil: Well, there will be events created, but not comments, right? I thought that was a reasonable compromise. While the new events may cause some confusion for our users ("I never did that - or, well, not this year"), updating them ensures any payload they leave on the variant collection gets written correctly. If we had it all in the events.. Another time!

scout/adapter/mongo/case.py
dnil: Yes, certainly! Actually it think it can be done in python as well, but requires some arcane syntax, eg like *kwargs for function arguments when the argument lists differ for the different function calls. This is ok.

dnil: The "problem" is that we (I? 😊) didn't make the tags one main type, with several subtypes. Then the list would have been one or two ifs and one updated_variant = self.update_tag().

scout/adapter/mongo/case.py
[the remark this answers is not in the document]
northwestwitch: Oh yes, events, not comments. Of course there will be questions but let't take one thing at the time, another time, I agree

dnil:
😆

northwestwitch:
> 😆

Damn, the test caught you immediately! 🤣

dnil:
> > 😆
> 
> Damn, the test caught you immediately! 🤣

🤣 Right! It was only meant as a suggestion.. I'll fire up the editor and do it properly..

northwestwitch:
> Right! It was only meant as a suggestion.. I'll fire up the editor and do it properly..

I liked it, computers are so inhuman sometimes..

dnil:
Or feel free to finish it - its your test after all! I have another PR to wrap up.. 😸

northwestwitch:
> Or feel free to finish it - its your test after all! I have another PR to wrap up.. 😸

I can fix it, thanks for spending time on this!

northwestwitch:
Testing the latest changes (create new comments only if variant id is changed):

Variant comments before re-upload:
image

Variant comments after re-upload are unchanged:

image

I'm going to merge it!

PR #1814, opened 2020-03-19
Calculate correct translocation lengths

moonso:
This PR adds the functionality to calculate end position for translocations when break points are on the same chromosome

**How to test**:
1. Tests are automated

**Expected outcome**:
Scout will now display the correct length for translocations

**Review:**
- [x] code approved by

### DIFF
--- a/scout/parse/variant/coordinates.py
+++ b/scout/parse/variant/coordinates.py
@@ -1,4 +1,6 @@
-from scout.constants import CYTOBANDS, BND_ALT_PATTERN, CHR_PATTERN
+"""Code to parse variant coordinates"""
+
+from scout.constants import BND_ALT_PATTERN, CHR_PATTERN, CYTOBANDS
 
 
 def get_cytoband_coordinates(chrom, pos):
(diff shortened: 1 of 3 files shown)


scout/parse/variant/coordinates.py
```diff
@@ -13,112 +15,95 @@ def get_cytoband_coordinates(chrom, pos):
...
-            subcategory = "snv"
-        else:
-            subcategory = "indel"
-    elif category in ("sv", "cancer_sv"):
-        subcategory = svtype
+    if end == pos:
+        return -1
 
-    return subcategory
+    return end - pos
```
dnil: Not +1?

scout/parse/variant/coordinates.py
```diff
@@ -13,112 +15,95 @@ def get_cytoband_coordinates(chrom, pos):
...
+
+    if svend == pos:
+        if svlen:
+            end = pos + svlen
 
     return end
 
 
+def get_end_chrom(alt, chrom):
+    """Return the end chromosome for a tranlocation
```
dnil:
```suggestion
    """Return the end chromosome for a translocation
```

tests/parse/conftest.py
```diff
@@ -1,6 +1,97 @@
+"""Fixtures for coordinates"""
+
 import pytest
 
 
+class CyvcfVariant:
```
dnil: :+1:

moonso:
No the problem with testing this on stage is that I need to upload a new case and that's always a hassle. Don't know how to make that more easy... But that should definitely be fixed

dnil:
No worries! I can look at it now that the missing-vcf pr is merged, but I’ll take a walk in the last rays of sun first! 🌞

dnil:
Memory is short when other priorities arise. Anyway, here is a screenshot from stage. Looks good! I suppose we could have added a little bit of logic to the view as well, to hide the issue from old cases. But, that will be a nice next issue now that this works. 

Screenshot 2020-03-23 at 09 46 36

PR #1816, opened 2020-03-22
Panel maintainers - fix #705

### LINKED ISSUE #705: Gene panel ownership
Gene panels should in effect have write permissions that can be given to users or admins.

dnil:
Gene panels can now have panel maintainers
Newly created genes are initially maintained by their creator. Admins can always maintain.

**How to test**:
1. modify, add, remove, create new, replace genes and what not as maintainer and non-maintainer. 
2. Old panels should work as before until maintainers are manually added to the db. 

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


moonso:
Is this ready @dnil ?

dnil:
Whenever! I was meaning to look at those codeFactor calls before long, but now is a probably a good time for feedback! It is not exactly urgent, but pretty fun. Review with a thought towards safety methinks: there are many adaptions and additional admin UI things that could go with this as well.

dnil:
Right, good points! I will give it a start, and shout when I need input!

moonso:
I wait until you tell it's ready 🐹

dnil:
Yep: that was only the first - a few more tests to go!

dnil:
Ok, this is probably a good time to check in again and see if anything is missing! Thanks!

scout/commands/update/panel.py
```diff
@@ -46,10 +60,46 @@ def panel(panel, version, update_date, update_version):
             LOG.warning(err)
             click.Abort()
 
+    # Any mintainer updates?
+
+    new_maintainer = None
+    if add_maintainer:
+        if add_maintainer not in [user["_id"] for user in adapter.users()]:
```
moonso:
```suggestion
        user_obj = adapter.user(user_id=add_maintainer)
        if not user_obj:
```
To avoid looping over all users

scout/commands/update/panel.py
```diff
@@ -46,10 +60,46 @@ def panel(panel, version, update_date, update_version):
...
+            # Check if maintainers exist in the user database
+            raise SyntaxError(
```
moonso: I would prefer a `LOG.warning` and `click.Abort` or just `return`

dnil: Sure; just following the nearest example.

scout/commands/update/panel.py
dnil: Hm, I considered that, but needed it for some reason. Perhaps that was earlier, or I was just imagining things. Will check!

scout/commands/update/panel.py
dnil: I made all of them click.Abort now. Hope it works!

scout/commands/update/panel.py
dnil: No, can't remember why; probably some older idea. Thanks!

dnil:
As for the remaining CodeFactor issue, it is right: update panel OMIM and update panel panel app should really be separate commands. But I will not alter the interface now just over that. I made a subroutine for the OMIM part to make it really obvious, and update panel is really already all contained in the adapter.

scout/commands/update/panel.py
```diff
@@ -46,10 +60,47 @@ def panel(panel, version, update_date, update_version):
...
+    if add_maintainer:
+        user_obj = adapter.user(user_id=add_maintainer)
+        if not user_obj:
+            # Check if maintainers exist in the user database
+            LOG.warning(
+                "Maintainer user id {0} does not exist in user database".format(
+                    add_maintainer
+                )
+            )
+            click.Abort()
```
moonso:
```suggestion
            raise click.Abort
```
Otherwise the function will continue, a test for that perhaps :)

scout/commands/update/panel.py
```diff
@@ -46,10 +60,47 @@ def panel(panel, version, update_date, update_version):
...
+
+        new_maintainer = panel_obj.get("maintainer", [])
+        if add_maintainer in new_maintainer:
+            LOG.warning("User {} already in maintainer list.".format(add_maintainer))
+            click.Abort()
```
moonso: Same here

scout/commands/update/panel.py
```diff
@@ -46,10 +60,47 @@ def panel(panel, version, update_date, update_version):
...
+
+    if revoke_maintainer:
+        current_maintainers = panel_obj.get("maintainer", [])
+        try:
+            current_maintainers.remove(revoke_maintainer)
+            new_maintainer = current_maintainers
+        except ValueError:
+            LOG.warning(
+                (
+                    "Maintainer user id {} is not a maintainer for panel {}.".format(
```
moonso: I think you should always use `%s` for LOG messages. Learned that from pylint

scout/commands/update/panel.py
```diff
@@ -46,10 +60,47 @@ def panel(panel, version, update_date, update_version):
...
+                        revoke_maintainer, panel
+                    )
+                    + " Current maintainers: {}.".format(current_maintainers)
```
moonso: Also this could be in one string like
```python
LOG.warning(
    "Maintainer %s is not a maintainer for panel %s. Current maintainers: %s.", 
    revoke_maintainer, panel, current_maintainers
)
```

scout/load/panel.py
```diff
@@ -77,13 +80,26 @@ def load_panel(panel_path, adapter, **kwargs):
         display_name = display_name or existing_panel["display_name"]
         institute = institute or existing_panel["institute"]
 
+    # Check if maintainers exist in the user database
+    maintainer = kwargs.get("maintainer")
+    if maintainer is not None:
+        scout_users = adapter.users()
```
moonso: Will not this return a iterable och user objects? If that's the case the check will fail since we ask if a string is in a list of dicts

scout/server/blueprints/panels/views.py
```diff
@@ -99,6 +109,14 @@ def panels():
     )
 
 
+def panel_write_granted(panel_obj, user):
```
moonso: 👍

tests/commands/update/test_update_panel_cmd.py
```diff
@@ -27,6 +27,30 @@ def test_update_panel(mock_app):
     )
     assert "panels.$.updated_at" in result.output
 
+    # Test adding a non-existing user
+    result = runner.invoke(
```
moonso: I think these should be split into multiple tests as it is much harder to follow what is wrong when these starts to fail

scout/adapter/mongo/panel.py
```diff
@@ -314,7 +314,7 @@ def gene_to_panels(self, case_obj):
 
         return gene_dict
 
-    def update_panel(self, panel_obj, version=None, date_obj=None):
+    def update_panel(self, panel_obj, version=None, date_obj=None, maintainer=None):
```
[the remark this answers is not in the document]
dnil: :+1:

scout/commands/load/panel.py
```diff
@@ -59,6 +63,7 @@ def panel(
     panel_type,
     panel_id,
     institute,
+    maintainer,
```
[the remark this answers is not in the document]
dnil: The idea was that we shouldn't be able to get there without having checked, and not impose too much in terms of the order things get bootstrapped. If we demand that, panel creation must e.g. come after user creation in setup etc.

scout/commands/update/panel.py
dnil: Hehe, I mistook click.Abort for click.abort() - not a brilliant naming scheme..

moonso: Yes I have done the same thing. I mixed it up with `context.abort()`

scout/commands/load/panel.py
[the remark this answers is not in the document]
moonso: Sure no need to check everything everywhere:)

scout/commands/update/panel.py
dnil: Ah, ok! I thought they did the same thing; cast the list into string and output, except that .format doesn't have the tuple issues that % has. Did you know what the advantage is? Hint: the formatting is may not have to happen! Thank you!

scout/commands/update/panel.py
dnil: 100 chars..

scout/load/panel.py
dnil: Yeah, pretty much! That was the reason for the list comprehension in the other function.. Fixed.

scout/commands/update/panel.py
moonso: Ok like that then

scout/load/panel.py
dnil: ..and added a test for it as well.

dnil:
I did not touch panel export. Should I?

moonso:
I think we can wait with panel export and do it in another PR

dnil:
Ok, good. That was my initial thought as well, but at some point it would make a nice full cycle to be able to write and read seamlessly.

PR #1817, opened 2020-03-23
hg38, Fix SV-variant links

bjhall:
This fixes the Ensembl link at the bottom of the variant details page for SVs when genome build is hg38. 

For DECIPHER I cannot find a way to link to hg38 coordinates, so I added the string "(GRCh37)" to the link as a warning to the user. Another option would be to remove the button completely for hg38. 

bild


moonso:
Thanks @bjhall , could you just move the line in conftest to the top of the file (x.xx). We did two releases today.

bjhall:
> Thanks @bjhall , could you just move the line in conftest to the top of the file (x.xx). We did two releases today.

@moonso If you meant to write "changelog" it is now fixed, if you really meant "conftest" I have no idea what you're talking about. :)

PR #1819, opened 2020-03-23
Release 4.14

dnil:
This minor update introduces mostly fixes, but also an additional optional option for the CLI.

### Added
- Use github actions to run tests
- Adds CLI command to update individual alignments path
- Update HPO terms using downloaded definitions files
- Option to use alternative flask config when running `scout serve`
- Requirement to use loqusdb >= 2.5 if integrated

### Fixed
- Do not display Pedigree panel in cancer view
- Do not rely on internet connection and services available when running CI tests
- Variant loading assumes GATK if no caller set given and GATK filter status is seen in FILTER
- Pass genome build param all the way in order to get the right gene mappings for cases with build 38
- Parse correctly variants with zero frequency values
- Continue even if there are problems to create a region vcf
- STR and cancer variant navigation back to variants pages could fail

### Changed
- Improved code that sends requests to the external APIs
- Updates ranges for user ranks to fit todays usage
- Run coveralls on github actions instead of travis
- Run pip checks on github actions instead of coveralls
- For hg38 cases, change gnomAD link to point to version 3.0 (which is hg38 based)
- Show pinned or causative STR variants a bit more human readable


dnil:
Oops, travis blocked again?

moonso:
It's still running

moonso:
Or is this happening, that it gets stuck?

dnil:
It did this weekend.. Hold on, I'll check..

dnil:
Yep, its the same again. See, it is fine over at travis:

Screenshot 2020-03-23 at 14 45 23

dnil:
Don't know - sth with GitHub actions, some mistake in the latest travis config yaml, the fact that none of the travis-ci tests are any longer required or a combination of the above?

dnil:
Anyway, 4.14 is now on hippo stage. Have you tried that cli command on hasta stage? Or did @northwestwitch?

moonso:
Mm maybe. Perhaps disable travis now? We are only doing lint checks there and now we have code factor for doing the same stuff + pep8 on GitHub Actions

PR #1820, opened 2020-03-23
Fix problem with fetching cases

moonso:
This PR fixes a bug when a variant found in loqusdb does not exist in another case

**Expected outcome**:
Should not fail when the above happens

**Review:**
- [x] code approved by DN
- [x] test by DN


dnil:
Test fake non-existing variant pinned on stage:
Screenshot 2020-03-23 at 16 03 24
Fails predictably on 4.14. Fixed on this branch:
Screenshot 2020-03-23 at 16 04 20

PR #1839, opened 2020-03-25
Arrange cards using card-columns

moonso:
This PR rearranges the cards on variant page

**Prepare for test**:
1. On stage
1. Open a variant, looks something like this:

Skärmavbild 2020-03-25 kl  13 34 01

Frequencies in one row, gene models in one etc.

**How to test**:
1. Refresh the variant from above

**Expected outcome**:
Models should be in one column, frequencies in one and severity/conservation in one

Skärmavbild 2020-03-26 kl  10 50 16

**Review:**
- [x] code approved by @northwestwitch 
- [X] tests executed by @moonso

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/variant_details.html
+++ b/scout/server/blueprints/variant/templates/variant/variant_details.html
@@ -229,7 +229,7 @@
           </li>
         {% else %}
           <li class="list-group-item">
-            <p class="card-text">    -</p>
+            <p class="card-text">-</p>
           </li>
         {% endif %}
       </ul>
--- a/scout/server/blueprints/variant/templates/variant/variant_details.html
+++ b/scout/server/blueprints/variant/templates/variant/variant_details.html
@@ -222,11 +222,17 @@
     {% endfor %}
     {% set superdups_fracmatches = superdups_fracmatches|sort %}
     <div class="panel-body">
-      {% if superdups_fracmatches %}
-        <span>mapping to {{superdups_fracmatches|length}} segm. dups. (min matching:{{ superdups_fracmatches|first|float|round(3) }}, max matching:{{ superdups_fracmatches|last|float|round(3) }})</span>
-      {% else %}
-        <p class="card-text">-</p>
-      {% endif %}
+      <ul class="list-group">
+        {% if superdups_fracmatches %}
+          <li class="list-group-item">
+            <span>mapping to {{superdups_fracmatches|length}} segm. dups. (min matching:{{ superdups_fracmatches|first|float|round(3) }}, max matching:{{ superdups_fracmatches|last|float|round(3) }})</span>
+          </li>
+        {% else %}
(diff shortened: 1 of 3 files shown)


moonso:
Would be nice with a review @northwestwitch or @dnil , it's tiny and tested

northwestwitch:
I can review tomorrow!

northwestwitch:
Code-wise it's fine I think and I also see that the page looks like the expected outcome you described. Now I don't remember exactly everything was said in the meeting but I understood that @4WGH suggested (via Nicole) to have all frequencies inside the same panel. Now frequency panels are on top of each other. But I might be wrong.

dnil:
I think this was roughly what was asked for, and it is easy to change back.

I didn't really mind the previous, as I am fairly used to reading left-right, but I suppose if you natively read a lot of languages with top-down orientation, this is better. I get sea-sick by the orientation of the boxes now though: couldn't we still do sth about the alignment so they stay in rows. 

But, importantly @NicoleCMMS; is this roughly how you visualised it?

moonso:
Just like you say it is easy to change back. And it is always hard to judge from first glance. Could we try this for a while and revise after a month or two?

4WGH:
After weeks of the new view I still think that the previous format for frequencies infornantion was better. 
Data easier to compare :) 
more similat to presenation/ tabler in article  :)

moonso:
Yes @4WGH that is why we propose a change in this pull request. Have you looked at the pictures above?

4WGH:
yes! confirming that freq looks better :)

I'm a bit concerned that the "OMIM" window has been moved lower (and also to the right).
I do look at the "OMIM" and "gene model" window muche more often compared to the  "gene annotation" and  "autozygous region" window. SO if the order can be swithed would be better for me.

Trying to be practical here :)
What we look at first should be higher up.
At the same time I know people work differently....

thankyou very much for the nice work :)

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -222,11 +222,17 @@
...
-        <p class="card-text">-</p>
-      {% endif %}
+      <ul class="list-group">
+        {% if superdups_fracmatches %}
+          <li class="list-group-item">
+            <span>mapping to {{superdups_fracmatches|length}} segm. dups. (min matching:{{ superdups_fracmatches|first|float|round(3) }}, max matching:{{ superdups_fracmatches|last|float|round(3) }})</span>
+          </li>
+        {% else %}
+          <li class="list-group-item">
+            <p class="card-text">    -</p>
```
northwestwitch: There is some space here before "-", is it a typo?

moonso: Forgot, thanks

moonso:
Ok I updated that according to your suggestion @4WGH

PR #1840, opened 2020-03-26
Add clinVar CNVs track in IGV

### LINKED ISSUE #1757: Display clinVar CNVs track in IGV
Similarly to SNVs. Check this PR: #1756. Loading the CNVs track actually worked but all the variants were overlapping on just one line!

### LINKED ISSUE #1793: Bigger clinvar variants hide the small ones in IGV track
Example interval: chr21:47,408,993-47,409,093

Looks like bigger things like this deletion hide small clinvar variants:

image

This is how the same track looks on UCSC browser:

image

northwestwitch:
fix #1757 and fix #1793

This works, but it's using the igv.js library that represents the patched version immediately before a new release (I don't know when the new release will happen). I suggest we merge and when they release the new version we update our code (I'm working on another PR involving the igv viewer anyway).  

This is what both clinvar tracks look like after the fix:
image

**How to test**:
1. Install this branch on scout-stage in clinical-db
1. Check some variants in IGV from a case which has alignments available (example: https://scout-stage.scilifelab.se/cust000/643594-300M) and make sure that clinVar CNV track is present
1. Go to the IGV browser of the case above and check this region: chr21:47,408,993-47,409,093. Bug #1793 should be fixed.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


northwestwitch:
This is ready for review!

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -0,0 +1,140 @@
+HG19REF_URL = (
```
northwestwitch: Moved a lot of track-related content to the controllers file, so the igv() method in the views is shorter and easier to read

scout/server/blueprints/alignviewers/views.py
```diff
@@ -103,66 +104,14 @@ def igv():
 
     display_obj = {}
 
-    # Add chromosome build info to the track object
```
northwestwitch: Moved to controllers

northwestwitch:
@moonso this one for instance?

northwestwitch:
> Looks nice!

Thanks!

dnil:
I think it has sth to do with the acmg collection (possibly acmg tag on variant) and ids.
Here's a load on a fresh setup:
Before:
Screenshot 2020-03-31 at 10 19 31
After reload:
Screenshot 2020-03-31 at 10 20 08

dnil:
Ah, but they are now still there in the db / on the variant! 👀 It might just be the variants view?

northwestwitch:
I'll check, I could reproduce this behavior by classifying a variant using the scout ACMG tool

dnil:
:blush: and sorry for getting this into the wrong thread. For reference, this is #1812.

northwestwitch:
> ead. For reference, this

Oh yes, I didn't even notice!

northwestwitch:
I'm going to merge this in 3, 2..

dnil:
Good fun! 👏

PR #1843, opened 2020-03-31
Add custom annotations back in for SVs

bjhall:
This commit https://github.com/Clinical-Genomics/scout/commit/e5468f439ee9c89fe68280ebb3cffe76611a53d7 accidentally (?) removed the "custom annotations" from the SV-variant template.

This PR adds them back in. I realize we're the only ones using these, but they don't affect anything if you're not using them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fix so that new cosmic identifier (COSV) is also acceptable #1304
 - Fixed COSMIC tag in INFO (outside of CSQ) to be parses as well with `&` splitter.
 - COSMIC stub URL changed to https://cancer.sanger.ac.uk/cosmic/search?q= instead.
+- Re-add "custom annotations" for SV variants
 
 ### Changed
 - Runs all CI checks in github actions
--- a/scout/server/blueprints/variant/templates/variant/sv-variant.html
+++ b/scout/server/blueprints/variant/templates/variant/sv-variant.html
@@ -225,6 +225,15 @@
           <div>Type</div><div class="float-right">{{ variant.sub_category|upper }}</div>
         </div>
       </li>
+      {% if variant.custom %}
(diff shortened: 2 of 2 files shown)


dnil:
:blush: Really sorry! Could you add tests and/or a test case file for this @bjhall otherwise we may forget again when we are not actively using it!

dnil:
😊 Sorry about that! We need a test and/or test case file for the display of this feature as well, otherwise we may forget it again!

bjhall:
> 😊 Really sorry! Could you add tests and/or a test case file for this @bjhall otherwise we may forget again when we are not actively using it!

No worries! There are already tests for loading the data. The problem here was that the code to show the data was removed from the template, the data was still properly loaded. 

I have no idea how to create tests for what is displayed in the web browser. Any pointers?

northwestwitch:
Sorry for the disappearance of the custom anno @bjhall. To test their presence in the template you could modify a test variant object and it it into the database. Then call the variant page in the test as in here: https://github.com/Clinical-Genomics/scout/blob/41e3f53b132ca2f9e91b260c5898085e633cb73a/tests/server/blueprints/variant/test_variant_views.py#L19

if you convert to string the response output into string (str(resp.data)) it should contain pair[0] and pair[1] substrings

bjhall:
Thanks for the pointer @northwestwitch. I made some half-hearted attempts at making a test for this, but it is obviously too hard for me (pytest makes me feel stupid). 

I'm a bit short on time at the moment, so could be merge this as is and I'll put it on my personal todo-list to make tests for this in the future?

northwestwitch:
> could be merge this as is

Agreed!! 👍

dnil:
I know the feeling; sometimes the tests seem so simple and straightforward, then when you want a slightly different thing.. 🤯 
We can help as well; perhaps make an issue and optionally assign yourself to it so we don't forget! Thanks!

dnil:
Ok, will do; testing it just now..

PR #1844, opened 2020-03-31
Name clinvar submission files according to their type

### LINKED ISSUE #1835: editing CLin VAr submission
Hi,
 we preprared some variants to submit, we wish to edit what we have written, how do we do it?

https://scout.scilifelab.se/cust003/clinvar_submissions
 from this page?

where?

northwestwitch:
Fix #1835.
In the current situation whenever a clinvar submission object is created, then the downloadable files have the same name. This PR fixes that and adds the type of file (case_data, variant_data) to the downloadable file names.

**How to test**:
1. Install the branch on a local instance of Scout.
1. Assign an HPO phenotype to the demo case
1. Pin a variant
1. Create a clinvar submission object and download the associated files

**Expected outcome**:
1. The 2 downloaded files should have different names. 

**Review:**
- [x] code approved by PG
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -230,21 +230,12 @@ def clinvar_submissions(institute_id):
                 csv_header = [
                     '"' + str(x) + '"' for x in csv_header
                 ]  # quote columns in header for csv rendering
-
+                download_day = str(datetime.datetime.now().strftime("%Y-%m-%d"))
                 headers = Headers()
                 headers.add(
                     "Content-Disposition",
                     "attachment",
-                    filename="".join(
-                        [
-                            clinvar_subm_id,
-                            "_",
-                            csv_type,
-                            "_",
-                            str(datetime.datetime.now().strftime("%Y-%m-%d")),
-                            ".csv",
-                        ]
-                    ),
+                    filename = f"{clinvar_subm_id}_{csv_type}_{download_day}.csv"
                 )
                 return Response(
                     generate_csv(",".join(csv_header), csv_lines),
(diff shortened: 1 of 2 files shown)


northwestwitch:
Tested, these are the saved files: 
image

northwestwitch:
Ready for review, one of the smallest PRs ever!

scout/server/blueprints/cases/views.py
```diff
@@ -236,10 +235,16 @@ def generate_csv(header, lines):
                 headers.add(
                     "Content-Disposition",
                     "attachment",
-                    filename=clinvar_subm_id
-                    + "_"
-                    + str(datetime.datetime.now().strftime("%Y-%m-%d"))
-                    + ".csv",
+                    filename="".join(
```
patrikgrenfeldt: Another possibility could be to make an f-string. I would go for the option that is easiest to read though. Which one do you prefer?

northwestwitch: I can make it an f string if you want!

northwestwitch:
fixed @patrikgrenfeldt !

scout/server/blueprints/cases/views.py
patrikgrenfeldt: I like it!

northwestwitch: Me too, thanks for the nice suggestion!

scout/server/blueprints/cases/views.py
```diff
@@ -230,21 +230,12 @@ def generate_csv(header, lines):
...
-                        [
-                            clinvar_subm_id,
-                            "_",
-                            csv_type,
-                            "_",
-                            str(datetime.datetime.now().strftime("%Y-%m-%d")),
-                            ".csv",
-                        ]
-                    ),
+                    filename = f"{clinvar_subm_id}_{csv_type}_{download_day}.csv"
```
dnil: 🥇 Me like too! Good both of you!

PR #1847, opened 2020-03-31
Bugfix other causatives case

bjhall:
This PR fixes a bug which caused causatives from other cases to show up only if there were at least **two** of them.

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -52,7 +52,7 @@
         <div class="col-sm-12">{{ variants_buttons() }}</div>
       </div>
 
-      {% if other_causatives|length > 1%}
+      {% if other_causatives|length > 0%}
         <div class="row ">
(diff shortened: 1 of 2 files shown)


dnil:
Indeed. 👀 

I confirm this does happen, and that the fix does indeed what is intended. Very valuable - thank you @bjhall!

Screenshot 2020-03-31 at 17 38 08

Screenshot 2020-03-31 at 17 37 41

- [x] tested on stage
- [x] code reviewed

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -52,7 +52,7 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
         <div class="col-sm-12">{{ variants_buttons() }}</div>
       </div>
 
-      {% if other_causatives|length > 1%}
+      {% if other_causatives|length > 0%}
```
dnil: 👀 🙄

PR #1858, opened 2020-04-02
New revstat filter strings

dnil:
OR

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR

### DIFF
--- a/tests/server/blueprints/variants/test_variants_views.py
+++ b/tests/server/blueprints/variants/test_variants_views.py
@@ -4,26 +4,24 @@ from flask_login import current_user
 from urllib.parse import urlencode
 from scout.server.extensions import store
 
+
 def test_variants_clinical_filter(app, institute_obj, case_obj):
 
     # GIVEN a variant non classified by clinvar
-    non_pathogenic_var = store.variant_collection.find_one({
-        'clnsig': { '$exists': False },
-        'variant_type' : 'clinical',
-        'category' : 'snv'
-    })
+    non_pathogenic_var = store.variant_collection.find_one(
+        {"clnsig": {"$exists": False}, "variant_type": "clinical", "category": "snv"}
+    )
     assert non_pathogenic_var
 
     # IF the variants receives a fake clisig compatible with the clinical filter
     clinsig_criteria = {
-        "value" : 5,
-        "accession" : 345986,
-        "revstat" : "criteria_provided,multiple_submitters,no_conflicts"
+        "value": 5,
+        "accession": 345986,
+        "revstat": "criteria_provided,multiple_submitters,no_conflicts",
     }
 
     store.variant_collection.find_one_and_update(
-        {"_id" : non_pathogenic_var},
-        {"$set" : {"clinsig_criteria": clinsig_criteria} }
+        {"_id": non_pathogenic_var}, {"$set": {"clinsig_criteria": clinsig_criteria}}
     )
 
     # GIVEN an initialized app
@@ -34,10 +32,9 @@ def test_variants_clinical_filter(app, institute_obj, case_obj):
         assert resp.status_code == 200
 
         # WHEN submitting form data to the page (POST method)
-        data = urlencode({
-            "clinical_filter": "clinical_filter",
-            "variant_type" : "clinical"
-            })  # clinical filter
+        data = urlencode(
+            {"clinical_filter": "clinical_filter", "variant_type": "clinical"}
+        )  # clinical filter
 
         resp = client.post(
             url_for(
--- a/tests/server/blueprints/variants/test_variants_views.py
+++ b/tests/server/blueprints/variants/test_variants_views.py
@@ -21,7 +21,7 @@ def test_variants_clinical_filter(app, institute_obj, case_obj):
     }
 
     store.variant_collection.find_one_and_update(
-        {"_id": non_pathogenic_var}, {"$set": {"clinsig_criteria": clinsig_criteria}}
+        {"_id": non_pathogenic_var["_id"]}, {"$set": {"clinsig_criteria": clinsig_criteria}}
     )
 
     # GIVEN an initialized app
@@ -50,7 +50,7 @@ def test_variants_clinical_filter(app, institute_obj, case_obj):
         assert resp.status_code == 200
 
         # containing the variant above (DOESN't WORK YET!!!)
-        assert "POT1" in str(resp.data)
+        assert "POT" in str(resp.data)
 
 
 def test_variants(app, institute_obj, case_obj):
--- a/tests/server/blueprints/variants/test_variants_views.py
+++ b/tests/server/blueprints/variants/test_variants_views.py
@@ -8,13 +8,13 @@ from scout.server.extensions import store
 
 def test_variants_clinical_filter(app, institute_obj, case_obj):
 
-    # GIVEN a variant non classified by clinvar
+    # GIVEN a variant without clinVar annotations
     test_var = store.variant_collection.find_one(
         {"clnsig": {"$exists": False}, "variant_type": "clinical", "category": "snv"}
     )
     assert test_var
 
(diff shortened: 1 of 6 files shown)


northwestwitch:
I'm going to write a specific test that uses the clinical filter to get variants which have new revstat criteria.

northwestwitch:
Possible testing for this PR: 
- Take this demo variant http://localhost:5000/cust000/643594/d57f0b1933560335fca26e6dd687e60c and add the following clnsig criteria:
```
clinsig_criteria = {
        "value" : 5,
        "accession" : 345986,
        "revstat" : "criteria_provided,multiple_submitters,no_conflicts"
    }
```
Variant should show up when clinical filter is called (because the other variant specifics for clinical filter are also met, I think)

northwestwitch:
I'm trying the clinical filter query with the tested suggested in the comment above, so far I didn't have much success. The clinical filter query looks quite complicated and I wonder if it can be simplified..

Now it is like this:
```
{
  'case_id': 'internal_id', 
  'category': 'snv', 
  'variant_type': 'clinical', 
  'panels': {'$in': ['panel1']}, 
  '$or': [
    {'$and': [
      {'$or': [
        {'gnomad_frequency': {'$lt': 0.01}}, 
        {'gnomad_frequency': {'$exists': False}}]}, 
        {'genes.functional_annotation': {'$in': ['transcript_ablation', 'splice_donor_variant', 'splice_acceptor_variant', 'stop_gained', 'frameshift_variant', 'stop_lost', 'start_lost', 'initiator_codon_variant', 'inframe_insertion', 'inframe_deletion', 'missense_variant', 'protein_altering_variant', 'transcript_amplification', 'splice_region_variant', 'incomplete_terminal_codon_variant', 'synonymous_variant', 'start_retained_variant', 'stop_retained_variant', 'coding_sequence_variant']}},
        {'genes.region_annotation': {'$in': ['exonic', 'splicing']}}]}, 
        {'clnsig': 
          {'$elemMatch': 
            {'$or': [
              {'$and': [
                {'value': {'$in': [4, 'Likely pathogenic', 5, 'Pathogenic']}}, 
                {'revstat': {'$in': ['mult', 'multiple_submitters', 'single', 'single_submitter', 'exp', 'reviewed_by_expert_panel', 'guideline', 'practice_guideline']}}]}, 
                {'$and': [
                  {'value': re.compile('Likely pathogenic|Pathogenic')}, 
                  {'revstat': re.compile('mult|multiple_submitters|single|single_submitter|exp|reviewed_by_expert_panel|guideline|practice_guideline')}
                  ]
                }
              ]
            }
          }
        }
    ]
}
```

dnil:
Sorry, just back from meeting. Right, it is a bit complex, and will probably remain so, but sure there could be simplifications! Lets look. It is supposed to be of the this - but also this type. There are two distinct populations that we want to let through: 1) those that fulfil a certain set of frequency and predicted severity criteria and 2) those that have previous classifications of clinical evidence, regardless of 1. And both while retaining the mask of the active (clinical) gene panel selection.

dnil:
As it is now it for instance has this extra `$and': [`from the query generation. I suppose that is from some features from the more general queries that we don't use in the clinical filter, but lets look!

dnil:
Oh, and it is doing those re:s and $in matching? Why is that?

dnil:
Right, the int valued ones.

northwestwitch:
I'm proposing a fix to the query.
I thing the clnsig part of the query should be like this:
```
    clnsig_query = {
                "clnsig": {
                    "$elemMatch": {
                        "$or": [
                            {
                                "$and": [
                                    {"value": {"$in": rank}},
                                    {'$or': [
                                        {"revstat": {"$in": trusted_revision_level}},
                                        {"revstat": re.compile("|".join(trusted_revision_level))}
                                    ]}
                                ]
                            },
                            {
                                "$and": [
                                    {"value": re.compile("|".join(str_rank))},
                                    {'$or': [
                                        {"revstat": {"$in": trusted_revision_level}},
                                        {"revstat": re.compile("|".join(trusted_revision_level))}
                                    ]}
                                ]
                            },
                        ]
                    }
                }
            }
```

northwestwitch:
Testing after my fix to the clinsig part of the query, The modified gene shows up!

Master branch:

image


This branch:
image

dnil:
Agreed; no use in having both $in and re for the same string. Since this is a re match, a few wildcards would shorten it a bit as well; e.g. mult*, but that would also require more careful checking of what strings are coming in from the CSQ..

dnil:
Also, the revstat part is the same for both string and rank types now and could be broken out!

northwestwitch:
> no use in having both $in and re for the same string.

So you would remove the $in?

dnil:
no, I'd keep the $in for the sake of the ints, unless you are very sure of the internal str - int conversion in python to mongo..

northwestwitch:
> unless you are very sure of the internal str - int conversion in python to mongo..

I'm not sure about anything when it comes to queries! 😆

northwestwitch:
Ok, I'll try to fix the tests

dnil:
I can try to break out the revstat-section so we only use it once!

dnil:
Like so: 
`clnsig_query = {
                "clnsig": {
                    "$elemMatch": {
                        "$and": [
                            {
                                "$or": [
                                    {"value": {"$in": rank}},
                                    {"value": re.compile("|".join(str_rank))},
                                ]
                            },
                            {"$or": [
                                {"revstat": {"$in": trusted_revision_level}},
                                {"revstat": re.compile("|".join(trusted_revision_level))}
                            ]}
                        ]
                    }
                }
            }
`

dnil:
Even better: 
```
            clnsig_query = {
                "clnsig": {
                    "$elemMatch": {
                        "$and": [
                            {"$or": [
                                    {"value": {"$in": rank}},
                                    {"value": re.compile("|".join(str_rank))},
                                ]
                            },
                            {"$or": [
                                {"revstat": re.compile("|".join(trusted_revision_level))}
                            ]}
                        ]
                    }
                }
            }
```

dnil:
The revstats don't need an in when they are being re:d; no int levels there.

northwestwitch:
Looks like a better solution, please push it to the branch!

dnil:
You were right! 😉

northwestwitch:
Testing your latest changes, it works:

image

northwestwitch:
> You were right! 😉

About what?

dnil:
> About what?

It could be simplified! 👍

dnil:
Ok, lets change the test then. Do you have a version already, or should I?

northwestwitch:
> Ok, lets change the test then. Do you have a version already, or should I?

You could fix the test_query ones and I can fix the test_variants_view with the clinical filter?

tests/server/blueprints/variants/test_variants_views.py
```diff
@@ -1,31 +1,66 @@
...
+            ),
+            data=data,
+            content_type="application/x-www-form-urlencoded",
         )
+
         # THEN it should return a page
         assert resp.status_code == 200
 
+        # containing the variant above
+        assert updated_var["_id"] in str(resp.data)
```
dnil, line 60: :+1:

tests/server/blueprints/variants/test_variants_views.py
```diff
@@ -58,22 +90,19 @@ def test_variants_post_data(app, institute_obj, case_obj):
         resp = client.get(url_for("auto_login"))
         assert resp.status_code == 200
 
-        # WHEN submitting form data to the page (POST method)
-        data = urlencode({"clinical_filter": True})  # clinical filter
-
-        # WHEN requesting a clinical filter to variants page
-        resp = client.post(
+        # WHEN accessing the variants page
+        resp = client.get(
```
dnil, line 94: 👍

PR #1859, opened 2020-04-03
Adds an extra case ID ("internal_id") which is independent of everything

bjhall:
This PR adds an extra field to cases, called "internal_id", which isn't used anywhere.

The use case for this is that we want to be able to keep track of our internal LIMS ID, but we don't want to use it as the family_id, since the LIMS ID is gibberish to most people. And family_id/family_name is also weirdly connected to the RankScore INFO-field (why does this feature exist, do you every have more than 1 family per vcf?)

If there is already an existing solution for this that I don't know of, I'm all ears :)


northwestwitch:
Regarding your question `"And family_id/family_name is also weirdly connected to the RankScore INFO-field (why does this feature exist, do you every have more than 1 family per vcf?)"` we have one VCF file per family/case.

We use Genmod (https://github.com/moonso/genmod) to rank variants in the family/case VCF.
RankScore INFO field is a custom field that we use to save the variant score coming from genmod. If for instance a variant has score -2, for a family named "XYZ" RankScore would be"XYZ:-2". 

I don't know why it was done this way, some decision taken in the past probably.

dnil:
Sure! Some of us as in klinisk genetik and klin immunologi in Sthlm) use our LIMS-id as display ID to make sure not to confuse cases. But eg CMMS have a separate display ID system - that HS alluded to in the GMS RD tech TC - that sort of goes back to their research projects. They might like having a lims id as well?

dnil:
Do you plan adding more things displaying/using it, or only to keep it in the DB? Perhaps consider injecting it into synopsis or as a case comment? You could search it anyway, and if it is in the synopsis users can automagically too..

bjhall:
@northwestwitch  I think that I've fixed the things you requested. Let me know if I misunderstood something! Loading a test case with lims_id works fine, and if the case does not have lims_id defined it ends up being set to "null" in the database. 

@dnil Currently we will use it for an external script that queries the Scout database and our LIMS API to generate a weird file that needs to go into some esoteric database at the hospital. Long story... In the future I plan to use it for fetching data from the LIMS and add to the PDF reports genrated from Scout.

tests/load/test_load_case.py
```diff
@@ -23,3 +23,19 @@ def test_load_case_rank_model_version(case_obj, adapter):
 
     assert loaded_case["rank_model_version"] == case_obj["rank_model_version"]
     assert loaded_case["sv_rank_model_version"] == case_obj["sv_rank_model_version"]
+
+
+def test_load_case_limsid(case_obj, adapter):
+
```
northwestwitch, line 29: Could you add a very short doctring for this function? One line would be sufficient

tests/parse/test_parse_case.py
```diff
@@ -62,6 +62,14 @@ def test_parse_case_owner(scout_config):
     assert case_data["owner"] == scout_config["owner"]
 
 
+def test_parse_case_limsid(scout_config):
```
northwestwitch, line 65: Could you add a very short doctring for this function? One line would be sufficient

northwestwitch:
I've created a commit to basically make your code pass the formatting test (Scout tests / Black) which was not passing previously. As you probably noticed, we introduced these more stringent code checks lately. Many times they helped us finding bugs or very weird code. This is obviously not the case with your PR because it was complaining for blank lines missing or extra.. Bah! Anyway, I used Black (https://github.com/psf/black) for the formatting of the modified files of this PR. Feel free to start using it as well, if you want! :)

bjhall:
@northwestwitch I've added docstrings and updated the changelog.

Regarding formatting, I blame emacs for frivolously inserting spaces on newlines. :) I guess I'll give black a try. Do you use it with the default settings?

northwestwitch:
Feel free to merge when you are ready!!

northwestwitch:
> Do you use it with the default settings?

yes, there's no many settings to choose

dnil:
> > Do you use it with the default settings?
> 
> yes, there's no many settings to choose

No, with `--line-length=100 / -l 100`. 😄

northwestwitch:
> No, with `--line-length=100 `

Ah ok, I never specified that

dnil:
You are not alone... But it does say so in the docs... 

To be clear @bjhall this is not a requirement, just a way to keep things tidy and avoid the more code aesthetic (@moonso 😉) of us completely freaking out, while the less so (@dnil) can direct their grumping at a program rather than a person changing whitespace and parenthesis indentation.

PR #1862, opened 2020-04-03
Adding varsome button

dnil:
This PR adds a functionality or fixes a bug.

From @ninanorgren, see #1805 and #1860!

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by
- [x] tests executed by

PR #1715, opened 2020-02-17
PR to fix issue #1304 with cosmic's new identifier and cosmic url

hassanfa:
This PR fixes handling new COSMIC id tag and updates the URL for COSMIC from ../id= (direct id) to ../q= (query/search).

This will help resolve four cases:
1. Older tags sometimes get merged or moved to new ids. This causes some tags with ../id= stub URL not work properly and end up not being found.
2. New tags won't work with ../id=, since COSMIC's stub URL seems to be always resolving to COSM.
3. New tags and old tags will be stored as string, NOT int.
4. COSMIC tag in INFO field is also parses the same way as CSQ with `&` separator.

**How to test**:
1. how to install it
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [x] tests executed by @hassanfa 
- [ ] "merge and deploy" approved by

PR #1867, opened 2020-04-06
Add Cosmic Badge

mikaell:
In Scout's cancer view, if cosmic IDs exists, add a badge [C].

OR

Ref: https://github.com/Clinical-Genomics/scout/issues/1629

**How to test**:

Install and run Scout in cancer view. 

**Expected outcome**:
If a variant has a Cosmic ID, a badge is shown in Exac-column.

Screenshot 2020-04-06 at 14 26 24

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -86,4 +86,8 @@
   ">
     {{ variant.gnomad_frequency|human_decimal if variant.gnomad_frequency else '~' }}
   </div>
+  {% if variant.cosmic_ids %}
+  <br>
+  <span style="border:1px; border-style:solid; border-color:#000000; padding: 0em;"> <strong> C </strong></span>
```
northwestwitch, line 91: Hi @mikaell, you could create a nice badge using the bootstrap 4 standards, and associate to it a link to cosmic too. Look at #1862, which does just this.

mikaell: Ok! I'll have a look

mikaell: Changed to use Bootstraps build in badges instead.

mikaell:
Handling several Cosmic links? @dnil @northwestwitch @hassanfa 

How to handle the case when we have several CosmicIDs and therefor could be several links? Now only the first in the list `cosmic_ids` will become a link, the rest are "forgotten"… (this is the current behaviour of Scout).

* In the detailed view create all links?
* be explicit about what cosmic IDs we have?



```
variant_obj: {‘_id’: ‘1d825c7f028fe1a1665364f618a41750’, ‘document_id’: ‘1d825c7f028fe1a1665364f618a41750’, ‘variant_id’: ‘9687b9706d14be2d4a98aa4518ddbe46’, ‘display_name’: ‘15_90631934_C_T_clinical’, ‘variant_type’: ‘clinical’, ‘case_id’: ‘cancer_test’, ‘chromosome’: ‘15’, ‘reference’: ‘C’, ‘alternative’: ‘T’, ‘institute’: ‘cust000’,
[...]
‘cosmic_ids’: [41875, 41590],
[...]
‘category’: ‘cancer’, ‘sub_category’: ‘snv’, ‘mate_id’: None,
[...]
‘hgnc_ids’: [5383],
[...]
}
```

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -82,8 +82,20 @@
       <div>
         <strong>Local (arch.)</strong>: {{ variant.local_obs_old or 0 }} obs.
       </div>
+      {% if variant.cosmic_ids %}
+      <strong>CosmicIDs</strong>:
+      <div>
+        {% for cosmicID in variant.cosmic_ids %}
+        cosm{{cosmicID}}
```
hassanfa, line 89: This might not work with new changes where we store COSMIC as a complete string. See this PR: https://github.com/Clinical-Genomics/scout/pull/1715

New COSMIC IDs might have COSV in them, so it will show the wrong ID here. It might bloat the code by adding conditions to check if cosmicID is int or str. But there might be other way to do more properly?

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -82,8 +82,20 @@
...
+        {% endfor %}
+      </div>
+      {% endif %}
     </div>
   ">
     {{ variant.gnomad_frequency|human_decimal if variant.gnomad_frequency else '~' }}
   </div>
+  {% if variant.cosmic_ids %}
+  <a href="{{ variant.cosmic_link }}" class="badge badge-light text=white visited =white" target="_blank">Cosmic</a>
```
hassanfa, line 98: Is it possible to post an image how this looks like?

dnil:
Yep, they would be much like the ClinVar entries, while they have made a large effort to condense. Compare how they are listed as a table on the variant view:
Screenshot 2020-04-07 at 11 12 52

For the variant*s* view, having a badge and popover with a list of the variants sounds about right? If they are rarely too many I suppose you could make several "cosmic" badges with links to one page each. Perhaps if there is only a couple, link them with separate badges, and if many make a popover or "..." between them to indicated there would be more on each var?

scout/server/blueprints/variants/templates/variants/components.html
mikaell: Screenshot 2020-04-08 at 08 49 52

scout/server/blueprints/variants/templates/variants/components.html
mikaell: Is this correctly understood?

* Cosmic IDs are stored as either 1. `1234` or 2. `COSV1234`.
* Cosmic IDs should be displayed as 1. `cosm1234` or 2. `COSV1234`

Could be done nicely in the code with a Jinja filter
```
    {% for cosmicID in variant.cosmic_ids %}
        cosmicID | cosmicPrefix


    @app.template_filter()
    def cosmicPrefix(cosmicID):
        if (cosmicID.startswith("COSV"):

       # etc etc


```

hassanfa: You're very close.

- COSMIC IDs start with "COS" all the time. they could be COSV, COSM, or COSN (legacy id, but they can still exist)
- If they are `int` type, then add `COSM` otherwise show them as they are.

PR #1876, opened 2020-04-15
Add verification comment box also for SVs

bjhall:
This small PR adds the verification email comment box that was available for small variants also to SVs. Only changes the template, since the backend functionality is shared with SNVs.

I have tested this, and it works as intented for me.


northwestwitch:
I'll review your PR now @bjhall !

bjhall:
> strange that we didn't include it before!!

That's probably my fault. I added that comment box back when we were  doing only exomes and didn't care about SVs... :)

PR #1886, opened 2020-04-27
Alamut link hg38

bjhall:
This PR fixes the Alamut link for hg38. 

When linking to a specific genome build in Alamut, the link must contain a query like this (GRCh37 is default if nothing is given):

```
3(GRCh38):g.37020386A>G
```

This has been tested and works properly on our setup (with Alamut 2.11), both for GRCh37 and 38.

### DIFF
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -312,7 +312,7 @@ def get_variant_links(variant_obj, build=None):
         cosmic_link=cosmic_link(variant_obj),
         beacon_link=beacon_link(variant_obj, build),
         ucsc_link=ucsc_link(variant_obj, build),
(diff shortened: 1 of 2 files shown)


scout/server/links.py
```diff
@@ -312,7 +312,7 @@ def get_variant_links(variant_obj, build=None):
         cosmic_link=cosmic_link(variant_obj),
         beacon_link=beacon_link(variant_obj, build),
         ucsc_link=ucsc_link(variant_obj, build),
-        alamut_link=alamut_link(variant_obj),
+        alamut_link=alamut_link(variant_obj, build),
```
dnil, line 315: :+1:

PR #1891, opened 2020-04-29
update igv.js to the newest version (2.5.4)

northwestwitch:
igv.js has released a bug fixed version containing the fix to our bigBed bug (I've done a nice assonance here!). The bug in question is the one in #1793

Our current master is using a patched version of igv.js at the moment (the version they use for developing in between official releases), so it would be nice to use the current version instead.

**How to test**:
1. Install this branch on clinical-db scout stage
1. make sure that bigBed tracks (for instance clinVar CNVs) are displayed correctly

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Better handling of delivery reports for rerun cases
 
 ### Changed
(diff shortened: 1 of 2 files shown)


northwestwitch:
Tested this on Scout stage right now and it's looking good:

image

PR #1892, opened 2020-04-29
Release 4.16

northwestwitch:
### Added
- Adds possibility to add "lims_id" to cases. Currently only stored in database, not shown anywhere
- Adds verification comment box to SVs (previously only available for small variants)
- Scrollable pedigree panel

### Fixed
- Error caused by changes in WTForm (new release 2.3.x)
- Bug in OMIM case page form, causing the page to crash when a string was provided instead of a numerical OMIM id
- Fix Alamut link to work properly on hg38
- Better handling of delivery reports for rerun cases

### Changed
- Updated the version of igv.js to 2.5.4

**Review:**
- [ ] code approved by


northwestwitch:
This codeowners thing must be pretty annoying, always asking you to be the reviewers of a PRs. So @dnil and @moonso don't blame me but @henrikstranneheim! 😄

dnil:
I don't particularly like the look of those padlocks. It gives the impression of the code being "locked" from changes for other collaborators, and not free. But if I got it right the settings are still as before, and the only change being that GitHub will send an email with review autoinvites to the ppl on the list? The reviewer suggestions based on recent changes seemed ok to me, so I don't quite understand why?

I don't mind particularly - I follow so get emails about PRs anyway - but I strongly suggest we do not do anything else to the settings. It would be totally counterproductive if our collaborators got the impression that their voices don't count as much as ours.

dnil:
Ok, it does. So, @moonso, I think it is imperative that you change the settings back to *NOT* require review by "codeowners" before this goes down the wrong alley. I can think of a few places in the code where other persons understand it better than the three of us do.

northwestwitch:
> Its not really in the update, but the owner must change Settings to not set "Require review by Code Owners".

Where is this part?

dnil:
On the Settings tab, but only the admin can change. Just a moment, I'm trying to understand why this feature is useful when there is already owners, admins and write permissions.

moonso:
I updated that

henrikstranneheim:
Had a chat with Daniel and if you do not like the code owners concept - we could achieve the same thing by adding the write permissions to the github repo at the same level as the code owner in the set the development guide.

northwestwitch:
> we could achieve the same thing by adding the write permissions to the github repo at the same level as the code owner in the development guide.

I think we'll opt for this other way. Thanks @henrikstranneheim !

dnil:
While bugged (delivery reports) we will merge this down and release a patch soonish.

PR #1900, opened 2020-04-30
MANE transcripts

bjhall:
This PR adds loading and parsing of MANE transcript information from VEP and marks it in the transcript overview:

bild

I know I should add tests for this, but I'm working my last day in 8 months today and I'm a bit stressed. I'll try to add tests some time next week.

MANE is hg38 only I think.


dnil:
Great! At least they said last year they had no intent of porting it to 37. Let's see what we can do  about testing then. I'm a little too tired right now. Have a wonderful 8 mo ahead!

northwestwitch:
I could review this today @dnil if you haven't started yet

scout/server/blueprints/variant/templates/variant/tx_overview.html
```diff
@@ -84,7 +84,13 @@
                     <div>
                       {{ refseq_id }}<span class="text-muted"> ({{ transcript.transcript_id }})</span>
                       {% if refseq_id in gene.common.primary_transcripts %}
-                      <span class="text-muted">(hgnc-primary)</span>
+                        <span class="text-muted">(hgnc-primary)</span>
+                      {% endif %}
+                      {% if transcript.mane_transcript %}
+                        {% set mane_transcript_parts = transcript.mane_transcript.split('.') %}
```
northwestwitch, line 90: How does it look like, the whole transcript.mane_transcript saved to the database?

bjhall: It's simply the refseq ID of the matched "MANE select" transcript. I'm removing the version number with the split() to be able to match it to the refseq variable (which doesn't have the version number).

I'll try to find time to add tests soon! Sorry about the delay...

northwestwitch: It's OK! 👍

moonso:
Shouldnt this be merged @northwestwitch ?

bjhall:
> Shouldnt this be merged @northwestwitch ?

I think maybe its waiting for me to write tests for this. However that won't happen before my paternity leave ends in january, so maybe this could be merged without test coverage for now and I'll put it on my january todo-list :)

dnil:
Or we can help you with a test, and you help someone else with a test in January? 😄

bjhall:
That also works. :) Do you need an example VCF, or do you already have some VEPed hg38 VCFs available?

dnil:
If you have one at hand, that is quicker. Otherwise I'm sure we can find some!

bjhall:
Here is an example vcf with a single variant. I just randomly selected the first variant I found with an annotated MANE transcript.

[hg38testcase.zip](https://github.com/Clinical-Genomics/scout/files/5128711/hg38testcase.zip)

northwestwitch:
I can work on this test if nobody started yet!

northwestwitch (approved):

northwestwitch:
Test written, I'm going to merge as soon as it passes the tests!

PR #1913, opened 2020-05-12
Fix remove causative variants

### LINKED ISSUE #1912: Remove causatives doesn't work any more
Mark a variant as causative and try to reset causative from case page. Page crashes

northwestwitch:
bug fix #1912

**How to test**:
1. Use a local instance of Scout
1. Mark a variant as causative
1. Assign an HPO phenotype to the case and mark another variant as partial causative for that specific phenotype
1. Make sure that removing causative works
1. Make sure that removing partial causative works

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/variant_events.py
+++ b/scout/adapter/mongo/variant_events.py
@@ -459,6 +459,7 @@ class VariantEventHandler(object):
             updated_case(dict)
 
         """
+        LOG.info("--------------------------------->{partial_causatives}")
         display_name = variant["display_name"]
         LOG.info(
             "Remove variant {0} as partially causative in case {1}".format(
@@ -468,6 +469,7 @@ class VariantEventHandler(object):
 
         # update partial_causative field of this case
         partial_causatives = case.get("partial_causatives") or {}
+        LOG.info(f"--------------------------------->{partial_causatives}")
         del partial_causatives[variant["_id"]]
 
         updated_case = self.case_collection.find_one_and_update(
--- a/scout/adapter/mongo/variant_events.py
+++ b/scout/adapter/mongo/variant_events.py
@@ -459,7 +459,6 @@ class VariantEventHandler(object):
             updated_case(dict)
 
         """
-        LOG.info("--------------------------------->{partial_causatives}")
         display_name = variant["display_name"]
         LOG.info(
             "Remove variant {0} as partially causative in case {1}".format(
@@ -469,7 +468,8 @@ class VariantEventHandler(object):
 
         # update partial_causative field of this case
         partial_causatives = case.get("partial_causatives") or {}
-        LOG.info(f"--------------------------------->{partial_causatives}")
+        if variant["_id"] not in partial_causatives:
+            return case
         del partial_causatives[variant["_id"]]
 
         updated_case = self.case_collection.find_one_and_update(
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/cases/views.py
```diff
@@ -1097,7 +1097,7 @@ def mark_causative(institute_id, case_name, variant_id, partial_causative=False)
 
     # send the user back to the case that was marked as solved
     case_url = url_for(".case", institute_id=institute_id, case_name=case_name)
-    return redirect(case_url)
+    return redirect(request.referrer)
```
northwestwitch, line 1100: If unmark causative is triggered from the variant page it's better to return that page instead of the case page

northwestwitch:
testing the branch:

1. Mark a causative and a partial causative:

image

1. Remove the causative (POT1):

image

1. Remove the partial causative:
image

1. Works also when the causative tag is removed from the variants page

northwestwitch:
This is also ready for review!

scout/server/blueprints/cases/views.py
```diff
@@ -1088,7 +1088,7 @@ def mark_causative(institute_id, case_name, variant_id, partial_causative=False)
         else:
             store.mark_causative(institute_obj, case_obj, user_obj, link, variant_obj)
     elif request.form["action"] == "DELETE":
-        if partial_causative:
+        if partial_causative == "True":
```
northwestwitch, line 1091: It's passed as an arg, so it's a string

scout/adapter/mongo/variant_events.py
```diff
@@ -468,6 +468,8 @@ def unmark_partial_causative(self, institute, case, user, link, variant):
 
         # update partial_causative field of this case
         partial_causatives = case.get("partial_causatives") or {}
+        if variant["_id"] not in partial_causatives:
```
dnil, line 471: Not that I disagree really, but any particular reason why this would happen in regular usage? Some warning, log or error might be in order? Silent fails can be a bit scary.

scout/server/blueprints/cases/views.py
dnil: Ouch..

dnil: Thanks, sorry - that was likely from my CodeFactor fix drive..

dnil: Ah, no, wasn't but anyway.

dnil: 😆 Was indeed - the removed eval! Thanks!

scout/adapter/mongo/variant_events.py
northwestwitch: This would not be useful on  a regular base, but I can imagine (and I have tested) this case:
having 2 tabs open :
1. Removing the partial causative from the case page
1. After point 1, trying to remove the partial causative also from the other tab (variant tab).

I could add a log, but I don't really think it will be very useful, because the effect of the cosde is that the page would refresh and the partial causative would disappear, as if the command works.
Does it make any sense?

dnil: You are probably right: sounds good, and I have no situation where I think it would be dangerous. Lets use this!

scout/server/blueprints/cases/views.py
northwestwitch: What?

northwestwitch: Ah Ok. Whatever! 😄

scout/adapter/mongo/variant_events.py
northwestwitch: Ok, but I'll merge then, thanks for the review!

PR #1919, opened 2020-05-15
Do not default to GATK caller for cancer SNVs when no caller info is provided

### LINKED ISSUE #1916: [BUG] GATK:Pass in report even though there is no GATK caller in cancer cases
It seems report picks up and reports a pinnged variant as `GATK:pass` even though there is no GATK caller used.

Steps to reproduce:
1. open a case in scout-stage with cancer track. e.g. https://scout-stage.scilifelab.se/cust000/KMP-00104_balsamic
2. Go to clinical variants and select a variant.
3. Pin the variant and generate a report: 
image

Variant caller that is used is VarDict, but for some reason is showing GATK:pass. It shouldn't show anything.

northwestwitch:
fix #1916

**How to test**:
1. Download this file https://github.com/Clinical-Genomics/scout/issues/1916
1. Decompress it and change this line:
`#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	TUMOR`
to:
`#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	999-99`

1. Edit scout/demo/scout/demo/cancer.load_config.yaml like this

- Remove these lines (normal sample):
```
- analysis_type: panel
    sample_id: 998-99
    capture_kit: Twist.V1
    sample_name: 998-99
    phenotype: unaffected
    sex: male
    expected_coverage: 90
    mt_bam: scout/demo/reduced_mt.bam
```
- Edit the link to the snv file to the downloaded (and edited) file of point 1
`vcf_cancer: path/to/scout_example_tumor_only.snv.indel.sorted.vcf`

1. Run the following command:
scout --demo load case scout/demo/cancer.load_config.yaml

**Expected outcome**:
Go to one snv variants of the cancer case and make sure that under the "GT call" it says vardict:Pass.

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/demo/cancer.load_config.yaml
+++ b/scout/demo/cancer.load_config.yaml
@@ -38,7 +38,3 @@ rank_score_threshold: -100
 analysis_date: 2018-10-12 14:00:46
 human_genome_build: 37
 track: cancer
-rank_score_threshold: -100
-analysis_date: 2018-10-12 14:00:46
-human_genome_build: 37
-track: cancer
--- a/scout/demo/cancer.load_config.yaml
+++ b/scout/demo/cancer.load_config.yaml
@@ -18,13 +18,26 @@ samples:
     msi: 15
     tumor_purity: 1
 
-vcf_cancer: /Users/chiararasi/downloads/scout_example_tumor_only.snv.indel.sorted.vcf
+  - analysis_type: panel
+    sample_id: 998-99
+    capture_kit: Twist.V1
+    sample_name: 998-99
+    phenotype: unaffected
+    sex: male
+    expected_coverage: 90
+    mt_bam: scout/demo/reduced_mt.bam
+
+vcf_cancer: scout/demo/cancer_test.vcf.gz
 vcf_cancer_sv: scout/demo/manta_vep_94_annotated_sv_cancer_file.vcf
 
 delivery_report: scout/demo/delivery_report.html
 
 # meta data
 rank_model_version: '1.1'
+rank_score_threshold: -100
+analysis_date: 2018-10-12 14:00:46
+human_genome_build: 37
+track: cancer
 rank_score_threshold: -100
 analysis_date: 2018-10-12 14:00:46
 human_genome_build: 37
--- a/scout/demo/cancer.load_config.yaml
+++ b/scout/demo/cancer.load_config.yaml
@@ -18,16 +18,7 @@ samples:
     msi: 15
     tumor_purity: 1
 
-  - analysis_type: panel
-    sample_id: 998-99
-    capture_kit: Twist.V1
-    sample_name: 998-99
-    phenotype: unaffected
-    sex: male
-    expected_coverage: 90
-    mt_bam: scout/demo/reduced_mt.bam
-
-vcf_cancer: scout/demo/cancer_test.vcf.gz
+vcf_cancer: /Users/chiararasi/downloads/scout_example_tumor_only.snv.indel.sorted.vcf
 vcf_cancer_sv: scout/demo/manta_vep_94_annotated_sv_cancer_file.vcf
 
 delivery_report: scout/demo/delivery_report.html
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -1,6 +1,6 @@
 name: Scout tests
 
-on: push
+on: ["push", "pull_request"]
 
 jobs:
   build:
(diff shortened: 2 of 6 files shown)


northwestwitch:
Works:

image

dnil:
Looks good to me! @hassanfa please confirm that varDict is what you would like for a cancer default. Note that you can control which callers get listed with their respective filter status by setting either the "set" (automatic from GATK merge) or "FOUND_IN" INFO tags when merging call sets from several callers - or a single one.

dnil:
Tests are great, aren't they! 😆

northwestwitch:
> Tests are great, aren't they! 😆

Yes they are! 😄

scout/demo/cancer.load_config.yaml
```diff
@@ -18,16 +18,7 @@ samples:
     msi: 15
     tumor_purity: 1
 
-  - analysis_type: panel
```
northwestwitch, line 21: Wops!

bjhall:
As a third party deployer of Scout I have to say that I find it a bit cumbersome that things like these are hardcoded in the source code. It requires us to maintain a fork for pure configurational things (which leads to occasional merge conflicts)

Would you consider moving this to either the global config.py or (even better I think) make it case-specific, ie set it in the yaml at load time. It is possible that the default caller changes over time or that you have different default callers for different cancer assay, for instance.

dnil:
Yes, good point. It evidently causes issues for us as well. I would prefer that VCF had a standard header option for the caller(s) that have edited FILTER. But, giving a default fallback caller name as an option to the CLI on load, or as you suggest yaml or conf, would work.

northwestwitch:
Hi @bjhall, I totally agree with you. Defining a case-specific default caller (or more, given the different types of pipelines used) would be the best solution for now. This is the way to go I think, but at our facility right now the config files that are fed to Scout come from different pipelines and are produced by other software. 
The change is very small but it would take quite some time before going into production so it's not a fast fix. But I'll propose it and try to move in that direction.

For the time being, if everybody agrees, we could call this default caller "custom", which is not very informative, and I'm sure many of our users won't be happy about it, but at least it's not conceptually wrong..

dnil:
We could then leave this "custom" part off the badge on the variant and only show "PASS" or "Filtered - whatnot"?

northwestwitch:
> We could then leave this "custom" part off the badge on the variant and only show "PASS" or "Filtered - whatnot"?

That would be smart, even if I'm sure we'll be caught 😆

dnil:
Yes, perhaps - there are a few rather sharp eyed and worried users out there - but it just might carry until we can get changes into CG. I like setting this to a general fallback name: it is somehow more honest than assuming a caller.

northwestwitch:
Ok, shall I go for custom then?

dnil:
:+1: As long as we hide it on display. 😆

hassanfa:
> Looks good to me! @hassanfa please confirm that varDict is what you would like for a cancer default. Note that you can control which callers get listed with their respective filter status by setting either the "set" (automatic from GATK merge) or "FOUND_IN" INFO tags when merging call sets from several callers - or a single one.

- Yes, we will use only VarDict for now. And all variants that go to Scout will have PASS filter. I can't fathom someone using a non-PASS filter for a clinical interpretation 😛 

- I like the idea of `INFO/FOUND_IN` and aware of it. It is what I am aiming for in BALSAMIC. I think all VCFs for Scout should conform to Scout defined tags instead of adding new ones

- Original issue was about Scout showing `GATK:pass` in report without having GATK. To be honest, prefer if it doesn't say anything. Then when we add other variant callers to final call set we can use `FOUND_IN` to its fullest :)

northwestwitch:
> Yes, we will use only VarDict **for now**

The devil is in the detail. :)

@hassanfa all this problem arose from @1792. Our clinicians specifically need to see the "GATK Pass"  in a variant GT call, otherwise they have to Sanger sequence it.

At this point I don't know what to do. The easiest would be add GATK pass only to the non-cancer SNVs.

hassanfa:
Well that's unfortunate. I don't like hardcoded stuff, cause I forget about them. You do whatever you think is best for now, and well adopt.

dnil:
Good choice @northwestwitch. Lets go with this for now - I think this should solve your issue @hassanfa without causing any issues for the RD track. @bjhall - I do agree this should not be hardcoded - it is an odd construction that is prone to error. Let's refactor at some point, but I can't help feeling that the issue is to a large part with the underlying pipelines/VCF format that does not represent call history consistently.

northwestwitch:
Merging this. I'm scared this PR will be stuck in a limbo of github actions changed otherwise @moonso 😆

PR #1928, opened 2020-05-20
Ucsc-hg38-sv-view

ViktorHy:
Check variant genome build and fix UCSC link accordingly in SV page.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR


scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -294,11 +294,23 @@
         {% endif %}
 
         {% if zoom == "length" and variant.chromosome == variant.end_chrom %}
-          - <a class="btn btn-outline-secondary btn-sm" href="http://genome.ucsc.edu/cgi-bin/hgTracks?db=hg19&position=chr{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}&dgv=pack&knownGene=pack&omimGene=pack" target="_blank">UCSC</a>
+          {% if case.genome_build == 37 %}
```
northwestwitch, line 297: Yet another occasion where the code is not chromosome build 38-friendly, thanks for the fix!!

The code works but it's a bit repetitive. You could avoid many lines of code if you define a variable "build" immediately before this block of code. You could assign it a value depending of chromosome build. Something like this:
```
{% set build = "hg19" %}
{% if case. genome_build == 38 %}
  {% set build = "hg38" %}
{% endif %}
```
And then use this variable inside the link to ucsc:
`href="http://genome.ucsc.edu/cgi-bin/hgTracks?db={{build}}&position=c....`

northwestwitch:
We are about to release a new Scout version and I'd like to include this one. @ViktorHy do you mind if I fix it and merge it?

ViktorHy:
Please do, sorry I was preoccupied and forgot.

PR #1936, opened 2020-05-27
Make variants cytoband genome-build aware

### LINKED ISSUE #1908: Make cytobands  build aware
All use of cytoband coordinates (e.g. variant names and PAR regions) should be build aware.

northwestwitch:
Fix #1908

**How to test**:
1. Modify the config file of one of the 2 test cases so the case genome build is "38"
1. Run `scout setup demo`
1. Make sure that the genome build of variants is saved correctly to database.
1. The best case scenario would be to load a variant that has different cytoband in genome build 37 and 38, but it's complicated because we have quite a few variants in the demo case. I've introduced  a specific test in tests anyway.

**Expected outcome**:

**Review:**
- [x] code approved by ML
- [x] tests executed by CR, github actions


scout/parse/cytoband.py
```diff
@@ -33,38 +33,3 @@ def parse_cytoband(lines):
             cytobands[chrom] = new_tree
 
     return cytobands
-
-
-import click
```
northwestwitch, line 38: Look what I found here!

northwestwitch:
Ready for review!

northwestwitch:
@mikaell this one is an easy review

mikaell:
> @mikaell this one is an easy review

I'll have a look!

northwestwitch:
> Looks good!

Thanks @mikaell!

PR #1937, opened 2020-05-28
Add VAF filter in cancer SNV variants filter form

### LINKED ISSUE #1922: [Feature reqeust] Filter variants based on their VAF
This is a customer request to filter variants in cancer view based on VAF.

A decision has to be made whether this VAF to be applied to both Tumor or Tumor&Normal.

northwestwitch:
Fix #1922

- Adds a field to filter variants with Tumor AF > than a certain floating point value
- Validates the whole filter on submission. So when some weird parameter is submitted, the search is not performed and an error will be flashed instead

image

**How to test**:
1. Install the branch on Scout stage
1. Go to a cancer case
1. Filter for variants with VAF > than a custom number (try with numbers like 0.25 or 0,25)
1. Try to fill in the form with wrong input for the fields

**Expected outcome**:
The functionality should be working:
1. VAF filter should work (compare with VAF for the returned variants)
1. Form validation errors should be returned when the occur

**Review:**
- [x] code approved by dNil
- [x] tests executed by @hassanfa

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -106,6 +106,7 @@ class QueryHandler(object):
                 'clingen_ngi': int,
                 'cadd_score': float,
                 'cadd_inclusive": boolean,
+                'vaf_frequency': float,
                 'genetic_models': list(str),
                 'hgnc_symbols': list,
                 'region_annotations': list,
@@ -188,7 +189,7 @@ class QueryHandler(object):
         primary_terms = False
 
         # gnomad_frequency, local_obs, clingen_ngi, swegen, spidex_human, cadd_score, genetic_models, mvl_tag
-        # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency
+        # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency, vaf_frequency
         secondary_terms = False
 
         # check if any of the primary criteria was specified in the query
@@ -209,7 +210,6 @@ class QueryHandler(object):
         # such as a Pathogenic ClinSig.
         if secondary_terms is True:
             secondary_filter = self.secondary_query(query, mongo_query)
-
             # If there are no primary criteria given, all secondary criteria are added as a
             # top level '$and' to the query.
             if primary_terms is False:
@@ -585,6 +585,13 @@ class QueryHandler(object):
                 mongo_secondary_query.append(
                     {"normal.alt_freq": {"$lt": float(query.get("control_frequency"))}}
                 )
+
+            if criterion == "vaf_frequency":
+                LOG.debug("add minimum VAF filter")
+                mongo_secondary_query.append(
+                    {"tumor.alt_freq": {"$gt": float(query.get("vaf_frequency"))}}
+                )
+
 
             if criterion == "mvl_tag":
                 LOG.debug("add managed variant list filter")
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -106,7 +106,7 @@ class QueryHandler(object):
                 'clingen_ngi': int,
                 'cadd_score': float,
                 'cadd_inclusive": boolean,
-                'vaf_frequency': float,
+                'tumor_frequency': float,
                 'genetic_models': list(str),
                 'hgnc_symbols': list,
                 'region_annotations': list,
@@ -189,7 +189,7 @@ class QueryHandler(object):
         primary_terms = False
 
         # gnomad_frequency, local_obs, clingen_ngi, swegen, spidex_human, cadd_score, genetic_models, mvl_tag
-        # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency, vaf_frequency
+        # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency, tumor_frequency
         secondary_terms = False
 
         # check if any of the primary criteria was specified in the query
@@ -586,10 +586,10 @@ class QueryHandler(object):
                     {"normal.alt_freq": {"$lt": float(query.get("control_frequency"))}}
                 )
 
-            if criterion == "vaf_frequency":
+            if criterion == "tumor_frequency":
                 LOG.debug("add minimum VAF filter")
                 mongo_secondary_query.append(
-                    {"tumor.alt_freq": {"$gt": float(query.get("vaf_frequency"))}}
+                    {"tumor.alt_freq": {"$gt": float(query.get("tumor_frequency"))}}
                 )
 
             if criterion == "mvl_tag":
--- a/scout/server/blueprints/variants/forms.py
+++ b/scout/server/blueprints/variants/forms.py
@@ -92,7 +92,7 @@ class VariantFiltersForm(FlaskForm):
 
     gnomad_frequency = BetterDecimalField("gnomadAF", places=2, validators=[validators.Optional()])
 
-    filters = SelectField(choices=[], validators=[validators.Optional()])
+    filters = NonValidatingSelectMultipleField(choices=[], validators=[validators.Optional()])
     filter_display_name = StringField(default="")
     save_filter = SubmitField(label="Save filter")
     load_filter = SubmitField(label="Load filter")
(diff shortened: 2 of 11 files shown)


scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -39,7 +39,7 @@
       <div class="card-header">
         <strong><a data-toggle="collapse" data-parent="#accordion" href="#collapseFilters">Filters</a></strong>
       </div>
-      <div class="card-body panel-collapse collapse" id="collapseFilters">
+      <div class="card-body panel-collapse collapse {{ "show" if request.args.get("expand_search") }}" id="collapseFilters" >
```
northwestwitch, line 42: If there is an error in the filter search, redirect and expand the filter search directly because it's likely that the user wants to continue using the filter form. It's annoying having to expand the form once again. This should be the default behavior in all types of variants pages I think

northwestwitch:
This is also ready for review

northwestwitch:
@hassanfa let me know if you want to test it (you can do it locally or on stage). If you want to test it on stage I can install the branch for you

hassanfa:
Nice! Let's talk tomorrow after townhall meeting about this. I still have issues with hasta-stage installation, but I don't want to install locally. Would be nice to test it in real environment.

hassanfa:
(selected only one gene NTRK3 to apply tumor AF filter):

test 1: without filter all variants are shown:
image

test 2: filter with 0.05 and filter worked:
image

test 2: tested with 0,05 (with , instead of .) and filter worked:
image

northwestwitch:
This PR is quite small and already testes, a quick review anyone?

scout/server/blueprints/variants/forms.py
```diff
@@ -115,9 +122,12 @@ class FiltersForm(VariantFiltersForm):
...
 
-    depth = IntegerField("Depth >")
-    alt_count = IntegerField("Min alt count >")
-    control_frequency = BetterDecimalField("Control freq. <", places=2)
+    depth = IntegerField("Depth >", validators=[validators.Optional()])
+    alt_count = IntegerField("Min alt count >", validators=[validators.Optional()])
+    control_frequency = BetterDecimalField(
+        "Control freq. <", places=2, validators=[validators.Optional()]
+    )
+    vaf_frequency = BetterDecimalField("Tumor AF >", places=2, validators=[validators.Optional()])
```
dnil, line 130: Tumor alt AF? Dunno, it does get long. Perhaps it is obvious, but it is easy to presuppose things in this context with allele pop freq, allele tumor/normal ratio and alt/ref ratio within tumor being on the picking list for confusion.

scout/server/blueprints/variants/templates/variants/cancer-variants.html
dnil: :+1: Good idea! We should change that soon so that the pages behave similarly. Open an issue perhaps. 
It is not entirely unlikely that the most common use case for the variants page would be to filter, so perhaps filters should also be open on first visit. I remember HT was a little frustrated over the filter folding away too much when he just started.

scout/server/blueprints/variants/views.py
```diff
@@ -54,7 +54,7 @@ def variants(institute_id, case_name):
     user_obj = store.user(current_user.email)
 
     if request.method == "POST":
-        # If special filter buttons were selected:
+        # If special filter were selected:
```
dnil, line 57: "special"? Simply if filter was selected perhaps? Is that really the only reason for POST nowadays?

dnil: There are no such comments elsewhere describing what entry points may have got there. Can you see any reason why we should not better just remove the comment?

tests/server/conftest.py
```diff
@@ -84,6 +84,7 @@ def app(real_database_name, real_variant_database, user_obj):
             MONGO_DBNAME=real_database_name,
             DEBUG_TB_ENABLED=False,
             LOGIN_DISABLED=True,
+            WTF_CSRF_ENABLED=False,
```
dnil, line 87: 🥇

scout/server/blueprints/variants/forms.py
dnil: Also, there is a slight risk of confusion to see % numbers in the list, but type fractions. But it is really the issue of the ones that wanted % in the table, not this input.

dnil:
To summarise, this essentially works. Except for one potential issue touching the persistent filters part of the form, everything else is just style and second guessing how users will interpret limited info.

scout/server/blueprints/variants/templates/variants/cancer-variants.html
northwestwitch: Issue added

scout/adapter/mongo/query.py
```diff
@@ -106,6 +106,7 @@ def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
                 'clingen_ngi': int,
                 'cadd_score': float,
                 'cadd_inclusive": boolean,
+                'vaf_frequency': float,
```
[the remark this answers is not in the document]
northwestwitch: tumor_frequency sounds better, I'll change it!

scout/server/blueprints/variants/forms.py
```diff
@@ -72,20 +79,20 @@ def process_formdata(self, valuelist):
...
+    cadd_score = BetterDecimalField("CADD", places=2, validators=[validators.Optional()])
     cadd_inclusive = BooleanField("CADD inclusive")
-    clinsig = SelectMultipleField("CLINSIG", choices=CLINSIG_OPTIONS)
+    clinsig = NonValidatingSelectMultipleField("CLINSIG", choices=CLINSIG_OPTIONS)
 
-    gnomad_frequency = BetterDecimalField("gnomadAF", places=2)
+    gnomad_frequency = BetterDecimalField("gnomadAF", places=2, validators=[validators.Optional()])
 
-    filters = SelectField(choices=[])
+    filters = NonValidatingSelectMultipleField(choices=[], validators=[validators.Optional()])
```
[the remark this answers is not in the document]
northwestwitch: Yes, i had to do it otherwise the whole form validation fails. We can change the class of the filter in a second moment of we need to, but it's a select, and its content should be already controlled, so I think it's fine to have it like this.

scout/server/blueprints/variants/forms.py
northwestwitch: This should be consistent with the control frequency I think, so a double. Tumor alt AF is more precise, I'll see if the label fits!

scout/server/blueprints/variants/views.py
northwestwitch: You are right, I'll remove it

scout/server/blueprints/variants/forms.py
northwestwitch: Damn! 😆 
image

scout/server/blueprints/variants/forms.py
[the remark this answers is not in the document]
dnil: Hm, I'm not convinced. The Select becomes a MultiSelect, allowing several filters to be checked. What happens then on load or delete?

scout/server/blueprints/variants/forms.py
dnil: Actually control freq (should probably be Normal freq for best disambiguation?) and tumor freq should be equal size?

scout/server/blueprints/variants/forms.py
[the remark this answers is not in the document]
northwestwitch: > Hm, I'm not convinced. The Select becomes a MultiSelect, allowing several filters to be checked. What happens then on load or delete?

It crashes! I'm don't remember much of how the filters select works. Good that you checked!

scout/server/blueprints/variants/forms.py
northwestwitch: Yes, I've removed one col space from `Min alt count >`, that I renamed to 'Min alt count' (no >). Now I'll rename the normal freq

scout/server/blueprints/variants/forms.py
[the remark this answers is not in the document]
northwestwitch: I've changed the class of filters to a non-validating single select class and loading or removing filters now works!

northwestwitch:
Ok, I'd merge in 3...2...

PR #1946, opened 2020-06-03
Patch release 4.17.1

northwestwitch:
This is a patch release because it's fixing the lack of data in the cytoband by chromosome objects populating the variants filter form.

The bug was observed on stage environment in a number of cases with case["genome_build"] either None or non standard ("37" or "38). It failed for instance on a case with build="grch37". 

We should be more strict in general when it comes to parsing the genome build value provided by the load config file.

**How to test**:
1. Install on stage and try to use the filter by cytobands form (SNVs, SVs variants page)

**Expected outcome**:

**Review:**
- [x] code approved by MM
- [x] tests executed by CR


northwestwitch:
Tested on stage and it works:

image

northwestwitch:
Ready for review!

northwestwitch:
> I have not followed the cytobandhandlerthing but I trust you!!

Thanks!!

dnil:
Well done! It’s in a way more about the handling of build version. I suspect we have a few more places where a defined but unexpected build value could cause a hiccup.

northwestwitch:
> Well done! It’s in a way more about the handling of build version. I suspect we have a few more places where a defined but unexpected build value could cause a hiccup.

Oh yes, I am certain we have plenty of those, unfortunately!

Issue #1942, opened 2020-06-02
Clicking on "cases" local observation table does nothing


hassanfa:
Clicking on "cases" in local observation table does nothing other than going to top of the page. Example: https://scout.scilifelab.se/cust087/bosssponge_rerun_20200113/09f36c6e4aa78d3b2a396af999d4059a?cancer=yes

dnil:
Hm, no. I think someone thought it more clear that there was more info hidden under it. Would you rather that it was not a link perhaps? We could presumably show the hover over the text without having it click-able.

hassanfa:
Hover text says "max 10 cases". I guess as starter, it'd be great to explain what does "max 10 cases" mean.  If it informative, we should keep it and show the text. But if it doesn't help interpreting "local observation table" then we might be better off removing it completely.

hassanfa:
One more thing, if we remove the link then it will be guess game of "is there a hover text" :-D one could hide easter eggs there 😁

dnil:
Max 10 cases is swenglish for ”At most 10 case ids out of all with matching variant shown”. I’d like to keep some mention of the number on there, but we can’t show all cases since that may be literally thousands e.g. for old ClinVar miss-annotations. Yeah, didn’t you know? Try hovering *everywhere*! 😜

dnil:
Remove it and introduce a symbol e.g. an ellipsis when and only when there are more than 10 matches?

hassanfa:
No I didn't know what does it mean. :)

I think ellipsis is a good idea. ✋

PR #1948, opened 2020-06-03
Dismiss badges on clinical. Fix #1939.

### LINKED ISSUE #1939: Mouse over for dismissed variants
On variant-pages "Clinical SNV and INDELS" and "Clinical structural variants" dismissed variants are marked with grey color. Is it possible to incorporate Mouse-over (like we have for Comments) so that it can easily be seen which frase/frases that have been used to dismiss each variant?
This would save some time in subsequent steps in our workflow in Lund as it would no longer be neccessary to open a variantpage to see which frases were used.

dnil:
Fix #1939. Simple badge with hover transfered from research variant display to clinical variant display. We now have at least four different approaches to storing labels for display via variants views and controllers. I would propose unifying them at some point. Possibly now? If so I would use the same style as with the clinical_assessments. Good?

**How to test**:
1. Dismiss SNV and SV variants. Hover over corresponding Dismiss labels and ensure consistency.

**Expected outcome**:

Screenshot 2020-06-03 at 12 08 37

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -111,18 +111,11 @@ def variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50
             )
 
             if clinical_var_obj is not None:
-                variant_obj["clinical_assessments"] = get_manual_assessments(
-                    clinical_var_obj
-                )
+                variant_obj["clinical_assessments"] = get_manual_assessments(clinical_var_obj)
 
         variants.append(
             parse_variant(
-                store,
-                institute_obj,
-                case_obj,
-                variant_obj,
-                update=True,
-                genome_build=genome_build,
+                store, institute_obj, case_obj, variant_obj, update=True, genome_build=genome_build,
             )
         )
 
@@ -150,9 +143,7 @@ def sv_variants(store, institute_obj, case_obj, variants_query, page=1, per_page
                 variant_type="clinical",
             )
             if clinical_var_obj is not None:
-                variant_obj["clinical_assessments"] = get_manual_assessments(
-                    clinical_var_obj
-                )
+                variant_obj["clinical_assessments"] = get_manual_assessments(clinical_var_obj)
 
         if variant_obj.get("dismiss_variant") is not None:
             assessment = {}
@@ -169,9 +160,7 @@ def sv_variants(store, institute_obj, case_obj, variants_query, page=1, per_page
             variant_obj["assessment"] = assessment
 
         variants.append(
-            parse_variant(
-                store, institute_obj, case_obj, variant_obj, genome_build=genome_build
-            )
+            parse_variant(store, institute_obj, case_obj, variant_obj, genome_build=genome_build)
         )
 
     return {"variants": variants, "more_variants": more_variants}
@@ -211,16 +200,12 @@ def get_manual_assessments(variant_obj):
 
             if assessment_type == "acmg_classification":
                 classification = variant_obj[assessment_type]
-                LOG.info(
-                    "Assessment type {}: {}".format(assessment_type, classification)
-                )
+                LOG.info("Assessment type {}: {}".format(assessment_type, classification))
                 if isinstance(classification, int):
                     acmg_code = ACMG_MAP[classification]
                     classification = ACMG_COMPLETE_MAP[acmg_code]
 
-                assessment["title"] = "Clinical ACMG: {}".format(
-                    classification["label"]
-                )
+                assessment["title"] = "Clinical ACMG: {}".format(classification["label"])
                 assessment["label"] = classification["short"]
 
             if assessment_type == "dismiss_variant":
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -88,17 +88,29 @@ def variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50
             evaluations.append(evaluation_obj)
         variant_obj["evaluations"] = evaluations
 
+        if variant_obj.get("dismiss_variant") is not None:
```
northwestwitch, line 91: It works! Was there any particular reason why you created this code in the variants controller? Because an alternative could have been to pass the DISMISS_VARIANT_OPTIONS along and to handle the situation in the template, as it's done for MANUAL_RANK_OPTIONS, CANCER_TIER_OPTIONS, SEVERE_SO_TERMS

dnil: 😄 Well, no, not more than that I followed my last version of this from the research side (compare https://github.com/Clinical-Genomics/scout/blob/ea2136e34094351d98b05c8e180850abb2c50839/scout/server/blueprints/variants/controllers.py#L146), where it is generated in a controller subroutine.

I'm all for plurality, but if I have to choose between the four-five different ways of doing essentially the same things we use, I do like the version where data for display is prepared in the controller over passing constants all the way to jinja. Don't get me wrong: jinja is cool, they all work, and repetitions tend to happen in all.

dnil:
True, that is a lot of text when you check a few. I doubt they will be read beyond "dismissed-for-reasons", but since that was the actual feature request.. How about this:
Screenshot 2020-06-03 at 16 09 02
Already my wonderful controller-template separation is crumbling since I'm now doing formatting in the controller that would arguably be better placed in the template.

scout/server/blueprints/variants/controllers.py
northwestwitch: I agree with you that stuff like this is better handled in the controllers but I also think that it would be best to have a consistent way to populate all these badges (with MANUAL_RANK_OPTIONS, CANCER_TIER_OPTIONS, SEVERE_SO_TERMS, DISMISS_VARIANT_OPTIONS). Would it be a lot of work to have a unique function (outside the variants views) that handles these variables and send stuff to every variantS template?

northwestwitch:
> True, that is a lot of text when you check a few. I doubt they will be read beyond "dismissed-for-reasons", but since that was the actual feature request.. How about this:
> Screenshot 2020-06-03 at 16 09 02
> Already my wonderful controller-template separation is crumbling since I'm now doing formatting in the controller that would arguably be better placed in the template.

Nice!!

scout/server/blueprints/variants/controllers.py
dnil: No, I don't think so, but they are used in several places, so I worried about missing something and introducing errors. But I do agree - and in its own way I also think CodeFactor agrees. Breaking out a few things of the controllers and templates would not hurt. I can give it a start and see if I run into any problems.

dnil: At least locally in cell_rank, it's just manual_rank that uses the lookup table directly in the template, the rest use controller (sic) based prepared strings. In the cancer case again  manual_rank as well as the analogous cancer_rank. I have a version now, but will sleep on it: in particular it touches the subtlety with how research classifications are different than clinical ones yet use the same display - but asymmetric since clinical should not be informed by research.

northwestwitch: ok, let me know when you are ready and I can review!

dnil:
Unifying also badge colors:
Screenshot 2020-06-04 at 12 33 49

.pre-commit-config.yaml
```diff
@@ -3,3 +3,4 @@ repos:
     rev: 19.3b0
     hooks:
       - id: black
+        args: [--line-length 100]
```
northwestwitch, line 6: 👏

northwestwitch: Standing ovation actually, and what about that rev? we don't have rev 19.3, but 19.10 something if I remember correctly

dnil: Right, well spotted!

dnil:
Ok, see what you think! Its difficult to say what is better between letting formatting slip into the controller and having too much data mangling out in the templates, but this at least makes the variantS badges behave similarly.

dnil:
Oh and the last was just sth I saw when testing the additional payload for #1942 that I managed to involve.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 19.3b0
+    rev: 19.10b0
```
northwestwitch, line 3: 👍 even if I'm not really a fan of freezing things but I welcome this change if it means we'll have less problems with formatting in our PRs

scout/constants/variant_tags.py
```diff
@@ -65,71 +65,91 @@
         "description": "Strong Clinical Significance. Biomarkers in FDA or guidlines that "
         "predict response, resistance to therapy, diagnosis or prognosis "
         "to specific tumor type.",
+        "label_class": "danger",
```
northwestwitch, line 68: 👍

.pre-commit-config.yaml
northwestwitch: Or we keep it for now and then we'll see?

PR #1953, opened 2020-06-05
Display number of filtered variants vs total number of variants in variants page

### LINKED ISSUE #1914: [Feature request] Option to show number of filtered variants and export filtered variants table
After applying filters to variants, there is no total count. It is hard to say even if filtered removed anything. It'd be great to show a total count of filtered variants. And then export them in excel format. There is option to export causative variants on institute level but not on case level. Of course one can pin these variants one by one and print out a report, but exporting before pinning can be quite useful.

northwestwitch:
Partial fix #1914

Creates a session object for each case that contains the variant count for all types of variants:
Example:
```
{
   "clinical": {
        "snv": 789, (or "cancer"),
        "str": 44,
        "sv": 63 (or "cancer-sv")
  },
  "research":{
      "snv": 789, (or "cancer")
      "sv": 63 (or "cancer-sv")
  }
}
```
And uses it on variants pages as the total number of variants by category.

This number is shown on variants page together with the number of variants retuned after filtering:
image

**How to test**:
1. Install the branch on stage and test filtering variants for clinical and research variants, SNVs and SVs, cancer cases.

**Expected outcome**:

**Review:**
- [x] code approved by ML
- [x] tests executed by CR


northwestwitch:
This is ready for review. There is one codefactor issue due to the complexity of variants controllers. I don't think I contributed to its complexity with this PR (I've just created a small function in there) so I'm going to ignore it!

scout/server/blueprints/variants/controllers.py
```diff
@@ -485,6 +489,34 @@ def cancer_variants(store, institute_id, case_name, form, page=1):
     return data
 
 
+def variant_count_session(store, institute_id, case_id, var_type, var_category):
```
northwestwitch, line 492: Al alternative to this would be to save these numbers in the database in the case object, so they'll be available all over the app. For instance in the institute page we could display the cases with their number of variants and so on.

northwestwitch:
> Nice work, looks good 👍

Thanks! Merging

PR #1959, opened 2020-06-09
Export (filtered) cancer variants

### LINKED ISSUE #1914: [Feature request] Option to show number of filtered variants and export filtered variants table

northwestwitch:
Fix #1914

Adds a button "Filter and export" to cancer variants page. When pressed the filtered variants are exported to CSV. 

This is how the page is now:
image

This is how the page becomes:
image

**How to test**:
1. Install the branch on Scout stage and go to a cancer case, for instance: https://scout-stage.scilifelab.se/cust000/strongfly
1. Test filtering and export of cancer variants

**Expected outcome**:

**Review:**
- [x] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -372,7 +372,7 @@ def variant_export_lines(store, case_obj, variants_query):
         hgvs_c = []
 
         # if variant is in genes
-        if gene_list is not None and len(gene_list)>0:
+        if gene_list is not None and len(gene_list) > 0:
             for gene_obj in gene_list:
                 hgnc_id = gene_obj["hgnc_id"]
                 gene_name = gene(store, hgnc_id)["symbol"]
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -420,10 +420,7 @@ def variant_export_genes_info(store, gene_list):
         # gather HGVS info from gene transcripts
         transcripts_list = gene_obj.get("transcripts")
         for transcript_obj in transcripts_list:
-            if (
-                transcript_obj.get("is_canonical")
-                and transcript_obj.get("is_canonical") is True
-            ):
+            if transcript_obj.get("is_canonical") and transcript_obj.get("is_canonical") is True:
                 hgvs_nucleotide = str(transcript_obj.get("coding_sequence_name"))
         hgvs_c.append(hgvs_nucleotide)
 
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -401,14 +401,14 @@ def variant_export_genes_info(store, gene_list):
             gene_list(list) A list of gene objects contained in the variant
 
         Returns:
-            variant_line(list)
+            gene_info(list) A list of gene-relates string info
     """
     gene_ids = []
     gene_names = []
     hgvs_c = []
 
     gene_info = []
-    
+
     for gene_obj in gene_list:
         hgnc_id = gene_obj["hgnc_id"]
         gene_name = gene(store, hgnc_id)["symbol"]
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -373,12 +373,12 @@ def variant_export_lines(store, case_obj, variants_query):
             gene_info = variant_export_genes_info(store, gene_list)
             variant_line += gene_info
         else:
-            i = 0
-            while i < 3:
+            empty_col = 0
+            while empty_col < 3:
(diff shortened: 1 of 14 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -416,6 +394,43 @@ def variant_export_lines(store, case_obj, variants_query):
     return export_variants
 
 
+def variant_export_genes_info(store, gene_list):
```
northwestwitch, line 397: Moved this code to another function to simplify the variant_export_lines function.

northwestwitch:
Ready for review.

scout/server/blueprints/variants/controllers.py
```diff
@@ -416,6 +394,43 @@ def variant_export_lines(store, case_obj, variants_query):
...
+        gene_name = gene(store, hgnc_id)["symbol"]
+
+        gene_ids.append(hgnc_id)
+        gene_names.append(gene_name)
+
+        hgvs_nucleotide = "-"
+        # gather HGVS info from gene transcripts
+        transcripts_list = gene_obj.get("transcripts")
+        for transcript_obj in transcripts_list:
+            if transcript_obj.get("is_canonical") and transcript_obj.get("is_canonical") is True:
```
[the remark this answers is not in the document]
northwestwitch: Better to be very sure? 😆

northwestwitch:
> Exporting all Clinical Structural variants creates an inconsistent csv-file. Checked with https://csvlint.io. I was running default demo cancer. Exported all.
> 
> Rows 6, 12, 13 and 16 have 14 columns. All other rows have 13. Example:
> 
> `0.0,1,2752092,A>[19:48324960[CA,2752092_A>[19:48324960[CA,-,-,-,-,121,17,-1,138,0,-1`
> 
> [cancer_case-filtered_sv-variants-3.csv.txt](https://github.com/Clinical-Genomics/scout/files/4769790/cancer_case-filtered_sv-variants-3.csv.txt)

Thanks, I'll check!

scout/server/blueprints/variants/controllers.py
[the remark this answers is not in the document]
northwestwitch: I'll improve the code here but it should be a double check here: first if the is_canonical key exists and then if it has a boolean value that is True

northwestwitch:
Fixed according to comments @mikaell

scout/server/blueprints/variants/controllers.py
```diff
@@ -374,9 +374,11 @@ def variant_export_lines(store, case_obj, variants_query):
             variant_line += gene_info
         else:
             i = 0
-            while i < 4:
-                variant_line.append("-")  # instead of gene ids
-                i = i + 1
+            while i < 3:
```
mikaell, line 377: Unnamed constant in while-loop (4).

northwestwitch: It's a good suggestion, I'll rename it to empty_col, thanks!

northwestwitch:
> 1. Linebreaks are '\n' and not '\r\n'. See [RFC](https://tools.ietf.org/html/rfc4180). Maybe no-one will notice.

I don't think I'm adding any \r in the exported text file, see code:

image

Could it be that the \r is added by the editor or the program you used to open  it?

northwestwitch:
@mikaell I've fixed the variant name to something more self-explanatory. I don't think the code is responsible to add any \r at the end of the lines, just \n.

mikaell:
> > 1. Linebreaks are '\n' and not '\r\n'. See [RFC](https://tools.ietf.org/html/rfc4180). Maybe no-one will notice.
> 
> I don't think I'm adding any \r in the exported text file, see code:
> 
> image
> 
> Could it be that the \r is added by the editor or the program you used to open it?

I meant '\r' is missing if we want to strictly follow the RFC:

>       Each record is located on a separate line, delimited by a line
>       break (CRLF).  For example:
>
>       aaa,bbb,ccc CRLF
>       zzz,yyy,xxx CRLF

This is no big deal, most editors and other software handles it just fine.

northwestwitch:
> I meant '\r' is missing if we want to strictly follow the RFC:

Ah no, we don't want to follow the RFC! These \r always cause a lot of trouble when parsing files, I'd gladly avoid them!

northwestwitch:
@mikaell if the PR is fine could you approve?

northwestwitch:
Thanks @mikaell !

PR #1964, opened 2020-06-11
Export dynamic HPO gene list

### LINKED ISSUE #1647: Export HPO gene list from the case page

northwestwitch:
fix #1647 
This PR allows to export the dynamic gene list generated from HPO phenotypes as a csv file.

**How to test**:
1. Install the branch on stage environment
1. Add one or more HPO phenotypes for a case and generate a gene list from these terms by clicking on `HPO panel`:

image

1. Click on the `Download HPO gene panel` button 

image

**Expected outcome**:
The functionality should be working and a gene list should be exported.

**Review:**
- [x] code approved by DN
- [x] tests executed by DN


northwestwitch:
This is ready for review.  There are some codefactor complains that I'm not going to fix, because they're not actually issues

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -374,6 +374,39 @@ def test_case_synopsis(app, institute_obj, case_obj):
         assert resp.status_code == 302
 
 
+def test_download_hpo_genes(app, case_obj, institute_obj):
```
dnil, line 377: :+1:

scout/server/blueprints/cases/views.py
```diff
@@ -1108,3 +1142,10 @@ def host_image_aux(institute_id, case_name, image, imgstr):
     img_path = abs_path + "/" + imgstr
     LOG.debug("Attempting to send {}/{}".format(img_path, image))
     return send_from_directory(img_path, image)
+
+
+def _generate_csv(header, lines):
```
dnil, line 1147: This works, and is readable to me, but if you feel like changing anything, consider being systematic about where you do the comma joining, quotation etc. Right now its a little different for header, lines and then end lines get added by this function.

dnil:
Tested ok locally. Don't forget CHANGELOG.. 😉

scout/server/blueprints/cases/views.py
northwestwitch: Thanks for the review. In general the only fields that need quotation are those which might contain commas (that are also the col separators). You have a point about consistency, but let's address next time we need this function to export something else!

PR #1976, opened 2020-06-25
text-break style when comments contain extremely long strings not separated by spaces

### LINKED ISSUE #1974: general report- pdf not working well
in fam 20101

the online page look like this (the table goes slightly outside the margin
bild

the pdf format is even worse. EDIT - try for yourself.

northwestwitch:
Fix #1974. Whenever there are long strings not containing space in comments at the case level or the variant level, the the general report HTML and PDF pages are not looking OK, see #1197 

**How to test**:
1. Make a very long comment of this type **this,is,a,impressively,long,comment,that,shoudnt,exist,in,the,first,place,but,sadly,it,does** at the case level and also at the variant (pinned, causative, dismissed, whichever type goes in the report) level. 
1. Make sure that the general report (HTML and PDF) is not OK using the master branch.
1. Switch to thus branch and see that the report looks fine

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case_report.html
+++ b/scout/server/blueprints/cases/templates/cases/case_report.html
@@ -196,7 +196,7 @@
 	                      <tr>
 	                        <td width="20%">{{comment.created_at.strftime('%Y-%m-%d')}}</td>
 	                        <td width="20%">{{comment.user_name}}:</td>
-	                        <td><strong>{{ comment.content | replace(",",", ") | replace(",  ",", ") }}</strong></td>
+	                        <td class="text-break"><strong>{{ comment.content }}</strong></td>
 	                      </tr>
 	                    {% endfor %}
 	                    </table>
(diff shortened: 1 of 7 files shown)


scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -196,7 +196,7 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 	                      <tr>
 	                        <td width="20%">{{comment.created_at.strftime('%Y-%m-%d')}}</td>
 	                        <td width="20%">{{comment.user_name}}:</td>
-	                        <td><strong>{{ comment.content }}</strong></td>
+	                        <td><strong>{{ comment.content | replace(",",", ") | replace(",  ",", ") }}</strong></td>
```
northwestwitch, line 199: Double replace is for taking into account the cases where there is already a space after the comma, so you don't get a double space

dnil:
I like that version!

dnil:
..only it didn't work with my version of the pdf libraries? Fingers crossed for rendering at stage!

northwestwitch:
Naa, the PDF doesn't work

northwestwitch:
I think I found a descent solution @dnil. I've created a custom Jinja filter that checks whenever there is no space after .,:;? and adds it. PDF looks decent now. It's on stage!

northwestwitch:
The latest changes work fine. This is the PDF from a case on stage:

image

i'm merging after the tests pass!

PR #1995, opened 2020-08-10
Existing rerun bug

northwestwitch:
Bug fix #1995 

**How to test**:
1. Very hard to reproduce since when a rerun is requested then the button disappears, but somehow someone managed to trigger the error

Install on stage and request a rerun for a case

**Expected outcome**:
The rerun function should work as expected

**Review:**
- [x] code approved by EM
- [x] tests executed by automatic tests and stage CR, EM


northwestwitch:
Tested on stage and the rerun functionality works:

image

Issue #1952, opened 2020-06-05
Phenotypes on individuals


dnil:
Currently, phenotypes (HPO terms) are logged at the case level. However sometimes the individuals in a trio or extended family do have different phenotypes. It would be more appropriate to store phenotypes on individuals in a case, rather than for the case as a whole.

In order not to break old data, phenotypes at the case level can perhaps be kept as a fallback.

PR #1999, opened 2020-08-11
Refactor individual dict 200811 -Draft for Discussion

mikaell:
Hi,
I'm looking for feedback on this idea to refactor how individual is handled in Scout. 

In the current situation we declare a softly typed dict named `individual` in `/models/case/case.py`. This soft type is -to my knowledge - never checked. However, that wouldn't matter. Because initalization of an individual is done like so (scout/build/individual.py):

` ind_obj = dict(individual_id=ind["individual_id"])`

Creating a new dict object. So we have unused code, types that are never checked and unreachable code for tests. I suggest we make individual into a class, directly inheriting dict. Running local tests I have yet to encounter problems.

Positives
+ A reasonable layer of abstraction 
+ Same operations and behaviour, `ind = Individual(); ind['phenotype']= 1` works
+ Mirrors other central and stored objects i.e. institute class
+ Possible to test -both functionality and code coverage
+ Type checking is not used -delete confusing and unused code
+ Possible to enforce our own type checking by overriding `__setitem__(self, key, value)` (see examples in the code)

Negatives
- Big change, might be hard to oversee consequences

What do you guys think?


northwestwitch:
I think that as long as the individual data is saved in the database in the same way as it is now then any change is fine. 

Specifically this would be an improvement so I'm all for it. Handling individuals as classes would probably help with some other things (I'm thinking about #1952 -Giving importance to phenotypes at the individual level and not at the general level of the family-).

Feel free to propose something and we can review!

northwestwitch:
Hi @mikaell, and how is it going with this other one? Because I realize this change needs a lot of refactoring all over the repo and if you don't plan to work on this in the near future we could perhaps close and open an issue about it instead?

mikaell:
I haven't worked on this for a while. The code change should be in very few places in the code. But the impact is large as this is an object sent through a case.

Making it an issue is ok by me.

@moonso Why not store `None` in the database? The disc usage and memory penalty would be minimal. You loose the ability to query for `None`. And -in my opinion worst -you need an extra layer. Is there an upside?

dnil:
Time for an overhaul of this after Chromograph is completely smooth, perhaps? We like, it just needs a mild update methinks?

mikaell:
I think I can start working on this now. My only caveat is @moonso's comment regarding null/None written to the database. What are the pros and cons of this?

moonso:
It is not a major dealbreaker. Storing None will affect disc space and ram, skip it and we save that. Queries asking if a field is None will work in the case that the attribute is missing.

Issue #1997, opened 2020-08-11
Find better alternatives to Flask-WeasyPrint


elevu:
Flask-WeasyPrint is used in Scout to export pdfs of reports. It doesn't support bootstrap styling.

elevu:
This will be done when https://github.com/Clinical-Genomics/scout-react/issues/51 and https://github.com/Clinical-Genomics/scout-react/issues/56 are done

PR #2002, opened 2020-08-13
Change pdf export library js solution #1997

elevu:
This pr changes the way pdfs are generated in the case reports. Instead of using Weasyprint, the javascript libraries jsPDF and html2canvas are used.

Pros of using this solution: 

- Weasyprint supports a limited amount of styling, in this way all the styling is applied and we don't have to worry about pdfs when changing the page.
- Weasyprint has a lot of dependencies that needs to be installed separately when setting up Scout. We don't need to install anything with this solution

Cons:
- It is not possible to copy text in the pdfs made with jsPDF, they behave like images.

This is a first iteration and if this solution is accepted I will remove Weasyprint from the project, add a loading animation to the page and change the naming of the generated pdfs.

**How to test**:
In Scout local, go to a case report page and click on the download button.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Nice! We want to get rid of WeasyPrint since a long time. This would also mean that we could uninstall packages that create a lot of annoying  warnings on the server!!! 🚀

dnil:
Hi there! Really cool to see more contributors @elevu!

Totally agree that we need to get rid of weasyprint, not super convinced a solution with bitmap only export is the way to go though? From a user perspective, it would be worth while trying to find something that can do proper pdfs, from which we can copy-paste if need be.

dnil:
I'll hopefully be back to devs next week; talk to you more then if not here!

elevu:
> I'll hopefully be back to devs next week; talk to you more then if not here!

Great! See you next week!

northwestwitch:
An alternative would be creating a version of the page without any bootstrap style which might might interfere with weasyprint, so that at least we don't have to freeze the library any more..

elevu:
> An alternative would be creating a version of the page without any bootstrap style which might interfere with weasyprint
Yes! Weasyprint supports a limited range of css specifications, the page can be written using those. We just need to be aware of those specifications when editing the styling in the future.

PR #2007, opened 2020-08-14
Filter variants by any gene panel from the case institute or the collaborators

### LINKED ISSUE #1637: Allow the use of arbitrary gene panels as gene lists for case filtering.
I created a new gene panel, and added to the institute: "trusight1". Now I want to show it in the filters of all those cases for that institute. It only shows HPO and The Panel, but not Trusight. Should I re-upload with that gene panel name specified, or is there an alternative way?

Here is the panel:
image

But in case variant view I can't find TruSight:
image

### LINKED ISSUE #2005: "The Panel" as default
Balsamic accessed 2020-08-01

POSSIBLE BUG?: When I use balsamic, "The Panel" is selected as default. It is very unclear to me what "the Panel" is. 

SOLUTION: As I guess that most times we use the GMCK solid panel, perhaps this could be default?. 
Even better if in addition, different cancer panels could be selected as a drop down list (GMCK panel and MSKI and then others as we add them perhaps leukemia SNV/SV, sarcoma...)

### LINKED ISSUE #2006: Allow for variants filtering using any panel in database
Do not restrict variant filtering to the case-specific gene panels.

northwestwitch:
fix #2006, fix #2005 and fix #1637.
This PR allows to filter variants of any type, from any case, by any existing gene panel in the database. 🕺 

**Current situation**:
- Variants filter accepts gene panels and gene symbols. You can specify both in the filter form.
- When you filter by gene panels, the only gene panels you can choose from are the default gene panels for the case.
- When the variants query is defined, if gene panels are provided, then the search returns variants with gene panel variant["gene_panels"]
- Variants page defaults to a default panel selected (panel1) or default panels for the case.

**Changes introduced by this PR:**
 - Variants filter still accepts gene panels and gene symbols, but all gene panels are translated into gene symbols before the query, and only gene symbols are used in variants query. (Simpler I think!)
- The initial variants filter contains the case-default panels + the eventual HPO panel.
- From the institute settings one can select any gene panel for the institute and the collaborator institutes (only the newest versions of each panel)
- Variants page still default to default panels for the case, but panel selection options will be broader whenever other institute panels are chosen.

**How to test -- Better on clinical-db stage**:
1. Install the branch on clinical-db stage
1. Try to go to the variants page of a case and check that the gene panels select contains case default panels + hpo panel.
1. Pre-select some different panels from the institute settings page.
1. Go back to the variants page and notice that all pre-selected panels show up, along with the case default panels and the HPO one 
1. Test the filtering, by gene panel(s), genes + gene panes, clinical filter, reset filter, whatever you come up with really.

Checklist when testing the functionality
- [x] Filter by empty OMIM panel should not return any results - same as now
- [x] Check and fix automatic filtering after loading of an existing filter
- [x] Check and fix clinical filter. This was fixed by #2044
- [x] Make sure that all default case panels are available in the panel select.
- [x] Automatic filtering by default panels when you click on the variants button
- [x] Only institute specific panels + panels of collaborators should be available to pre-select in the institute 
settings

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
This is ready for review now. What about a nice query-related PR to start over after the holidays @dnil ?? I'm suggesting you because you know well the monster (THE QUERY) 🤣

dnil:
Sounds like a fun plan! 👍 😜 
If there are any complexity issues with the code for the db query, I suppose I can just review and ask the author to simplify the code..
If you are in a rush, ping someone else. Most people are back today over att the clinic side, so give it a couple of days of convergence and I should be able to look at it.

scout/server/blueprints/institutes/controllers.py
```diff
@@ -64,7 +65,19 @@ def populate_institute_form(form, institute_obj):
...
+        store.gene_panels().sort(
+            [("panel_name", pymongo.ASCENDING), ("version", pymongo.ASCENDING)]
+        )
+    )
+    panel_choices = []
+    for panel in available_panels:
+        panel_id = str(panel["_id"])
+        n_genes = len(panel.get("genes", []))
+        panel_name = f"{panel['panel_name']} ({str(panel['version'])}) - {str(n_genes)} genes"
+        panel_choices.append((panel_id, panel_name))
```
northwestwitch, line 79: I'm using panel_id (ObjectID) here because I want all specific panel objects, not the latest version of a panel.

scout/adapter/mongo/panel.py
```diff
@@ -303,6 +304,27 @@ def gene_to_panels(self, case_obj):
...
+
+            Returns:
+                gene_list(list): a list of hgnc terms
+
+        """
+        gene_list = []
+        panel_obj = None
+        try:
+            panel_obj = self.panel(panel_id)
+        except InvalidId:  # Occurrs when clinical filter panel is used
```
northwestwitch, line 321: Because clinical filter is passing gene panels by name, not id. And the latest panel with that name is collected.

northwestwitch:
> I suppose I can just review and ask the author to simplify the code..

You won't believe it, but this PR is actually simplifying the code! But I don't know about the performance.

But don't worry, no rush, when you have time!

dnil:
This is great stuff, and has the potential to be very much appreciated both for the simplification of queries and from all the user side goodies. But I think we can still do this while keeping the really Good Thing from the old panel concept: carefully protecting the extent of the ordered clinical tests from simple user mistakes.

Some observations from local testing before diving into details:

- we need to think about/add a way to populate the Institutes (or cases) with at the very least the searchable panels that are default for the cases. Some users would find it deeply disconcerting - and possibly unethical - if they unintentionally ended up on a variant list with all loaded variants. E.g. variants in genes normally only shown on hereditary cancer or neurodegeneration uncovered for an immunology panel. This could perhaps be done at load (with default panels for cases automagically made searchable for the owner institute). Then there is collaborators as well..

- on the same line, the variants view must start with the default panel filtered variants shown. One way to do this might be to hijack the exception you have for clinical filter?

- speaking of which, clinical filter does seem broken, and returns an empty list,  even if the default panel is made searchable?

- an extra click is now required from loading filters to searching: on load filter, an empty variants list and a populated filter form is shown, then everything works when hitting Filter again. This is in contrast to 4.19 behaviour.

- I noticed the case page being very slow at times - perhaps just my local test db lacking some index or whatnot, but worth paying attention to on stage, e.g. on first visit with new panel or such?

northwestwitch:
> * we need to think about/add a way to populate the Institutes (or cases) with at the very least the searchable panels that are default for the cases. Some users would find it deeply disconcerting - and possibly unethical - if they unintentionally ended up on a variant list with all loaded variants. E.g. variants in genes normally only shown on hereditary cancer or neurodegeneration uncovered for an immunology panel. This could perhaps be done at load (with default panels for cases automagically made searchable for the owner institute). Then there is collaborators as well..

OK, I wasn't really considering this aspect, but you are right. We could make available for choice at the institute page only the panels of the institute + those of the collaborators perhaps?

> * on the same line, the variants view must start with the default panel filtered variants shown. One way to do this might be to hijack the exception you have for clinical filter?

Yes, this was a concert that I already had, and this is relatively easy to achieve.

> * speaking of which, clinical filter does seem broken, and returns an empty list,  even if the default panel is made searchable?

I'll check!

> * an extra click is now required from loading filters to searching: on load filter, an empty variants list and a populated filter form is shown, then everything works when hitting Filter again. This is in contrast to 4.19 behaviour.

I'll fix it.


> * I noticed the case page being very slow at times - perhaps just my local test db lacking some index or whatnot, but worth paying attention to on stage, e.g. on first visit with new panel or such?

Let's check before the new release!

northwestwitch:
Todo list:
- [ ] OMIM panel should not return any result when it doesn't have any genes.
- [ ] Check and fix automatic filtering after loading of an existing filter.
- [ ] Check and fix clinical filter
- [ ] Make sure that all case-specific panels are available in the panel select.
- [ ] Automatic filtering by default panels
- [ ] Only institute specific panels + panels of collaborators should be available to pre-select in the institute settings

northwestwitch:
> Todo list:
> 
> * [x]  Filter by empty OMIM panel should not return any results - same as now
> * [x]  Check and fix automatic filtering after loading of an existing filter. -this works for me -
> * [ ]  Check and fix clinical filter
> * [ ]  Make sure that all default panels are available in the panel select.
> * [ ]  Automatic filtering by default panels
> * [ ]  Only institute specific panels + panels of collaborators should be available to pre-select in the institute settings

northwestwitch:
> Todo list:
> 
> * [x]  Filter by empty OMIM panel should not return any results - same as now
> * [x]  Check and fix automatic filtering after loading of an existing filter.
> * [ ]  Check and fix clinical filter
> * [ ]  Make sure that all default panels are available in the panel select.
> * [ ]  Automatic filtering by default panels
> * [ ]  Only institute specific panels + panels of collaborators should be available to pre-select in the institute settings

dnil:
Very nice structured summary of action points! I'll ramble some more and let you translate! 😄 

> > * we need to think about/add a way to populate the Institutes (or cases) with at the very least the searchable panels that are default for the cases. Some users would find it deeply disconcerting - and possibly unethical - if they unintentionally ended up on a variant list with all loaded variants. E.g. variants in genes normally only shown on hereditary cancer or neurodegeneration uncovered for an immunology panel. This could perhaps be done at load (with default panels for cases automagically made searchable for the owner institute). Then there is collaborators as well..
> 
> OK, I wasn't really considering this aspect, but you are right. We could make available for choice at the institute page only the panels of the institute + those of the collaborators perhaps?

I was primarily thinking about the empty list issue, and the default/actually analysed panels not being selectable in the variants filter. But you are right too: there have historically been some guarding of the extent of panels with quite some enthusiasm, although the atmosphere has become much more open and collaborative.

Have you given any thought to warnings about missing content, eg when the panels listed as run for the case - where the variants are actually present in the selected vcf? This is generally not much of a problem when a recent OMIM panel is on the "selected" list, but can be a big issue if ppl believe they search an important disease gene, and the case has originally been run (in MIP, typically) with a set of small panels.

northwestwitch:
> here have historically been some guarding of the extent of panels with quite some enthusiasm, although the atmosphere has become much more open and collaborative.

That should be the spirit, right?

> Have you given any thought to warnings about missing content, eg when the panels listed as run for the case - where the variants are actually present in the selected vcf? 

No, not really. It's quite important but perhaps in another PR? This one is rather big already..

northwestwitch:
> speaking of which, clinical filter does seem broken, and returns an empty list, even if the default panel is made searchable?

@dnil I can't reproduce this. Clinical filter returns the same results as master (in demo)

dnil:
Interesting! Then we may have some incompatible db content issue? But it worked ok on stage?

northwestwitch:
> Interesting! Then we may have some incompatible db content issue? But it worked ok on stage?

Haven't checked. I'm going to install it there now

dnil:
Right, I tried it on a local with only the demo instance genes. Perhaps something to do with missing genes, or more likely a change in something related to genes from the demo data to current downloads?

dnil:
> > here have historically been some guarding of the extent of panels with quite some enthusiasm, although the atmosphere has become much more open and collaborative.
> 
> That should be the spirit, right?
> 
> > Have you given any thought to warnings about missing content, eg when the panels listed as run for the case - where the variants are actually present in the selected vcf?
> 
> No, not really. It's quite important but perhaps in another PR? This one is rather big already..

Absolutely on both accounts!

northwestwitch:
> Right, I tried it on a local with only the demo instance genes. Perhaps something to do with missing genes, or more likely a change in something related to genes from the demo data to current downloads? Still will need some fixing to get the demo to work - if I can reproduce it with a clean install say.

Doesn't work on stage, so must be my instance. I'll fix

northwestwitch:
I think I'm done with this PR for now. I've simplified it quite a bit and I made it in a way that only gene panels from collaborating institutes are selectable now. For the reviewer : re-read the description of this PR,  because I've changes it a bit. I've tested on stage all the items in the checkboxes and look like they work, but I'll uncheck everything now so you can re-test it!

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -46,7 +46,7 @@
       </div>
       <div class="card-footer text-center py-3">
         {% if "all_variants" in session %}
-          Filtered {{session['filtered_variants']}} / {{session["all_variants"]}} variants
+          Showing {{session['filtered_variants']}} / {{session["all_variants"]}} variants
```
dnil, line 49: :+1:

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -158,7 +158,7 @@ <h6>Phenotype terms (HPO):</h6>
 <div class="form-group">
   <div class="row">
     <div class="col-4">
-      {{ form.gene_panels.label}}
+      {{ form.gene_panels.label(class="control-label", data_toggle="tooltip", data_placement="left", title="This list can be modified from the institute settings page.") }}
```
dnil, line 161: :+1:

scout/server/blueprints/variants/views.py
```diff
@@ -67,7 +67,14 @@ def variants(institute_id, case_name):
         form = FiltersForm(request.args)
         # set form variant data type the first time around
         form.variant_type.data = variant_type
-        form.chrom.data = request.args.get("chrom", None)
+        # set chromosome to all chromosomes
+        form.chrom.data = request.args.get("chrom", "")
+        # set gene panels to case-specific gene panels
```
dnil, line 72: Hm, perhaps leave the comment out - or correct: the code implies case-specific default panels?

scout/server/blueprints/variants/views.py
```diff
@@ -76,11 +83,7 @@ def variants(institute_id, case_name):
     ]
 
     # populate available panel choices
-    available_panels = case_obj.get("panels", []) + [{"panel_name": "hpo", "display_name": "HPO"}]
-
-    panel_choices = [(panel["panel_name"], panel["display_name"]) for panel in available_panels]
-
-    form.gene_panels.choices = panel_choices
+    form.gene_panels.choices = controllers.gene_panel_choices(store, institute_obj, case_obj)
```
dnil, line 86: :+1:

scout/server/blueprints/variants/views.py
```diff
@@ -293,7 +296,14 @@ def cancer_variants(institute_id, case_name):
     else:
         page = int(request.args.get("page", 1))
         form = CancerFiltersForm(request.args)
-        form.chrom.data = request.args.get("chrom", None)
+        # set chromosome to all chromosomes
+        form.chrom.data = request.args.get("chrom", "")
+        # set gene panels to case-specific gene panels
```
dnil, line 301: Same here!

dnil: (And in general that implies we have a bit too much code duplication between the different variant type views, but that is for another PR!)

dnil (changes requested):
Almost there! This works really well, and the code is indeed simplified! My main misgiving is with panels of different versions. It appears this will generally call latest versions, even if say case default is 2.0 and institute latest is 3.0? While it is great to have the option to filter with 3.0, I'd say the default behaviour should still be filter with 2.0 unless explicitly chosen otherwise?

northwestwitch (commented):
> Almost there! This works really well, and the code is indeed simplified! My main misgiving is with panels of different versions. It appears this will generally call latest versions, even if say case default is 2.0 and institute latest is 3.0? While it is great to have the option to filter with 3.0, I'd say the default behaviour should still be filter with 2.0 unless explicitly chosen otherwise?

The default behavior in current master is that if even if a case as an associated panel of version 2.0, then the filter filters by the latest panel. This is because the query is looking for variants having the key/value:
`panels:["panel_name"] `

This PR is not changing the result, just the query: search for genes instead of panels.

scout/server/blueprints/institutes/controllers.py
```diff
@@ -54,7 +54,7 @@ def populate_institute_form(form, institute_obj):
     form.coverage_cutoff.default = institute_obj.get("coverage_cutoff")
     form.frequency_cutoff.default = institute_obj.get("frequency_cutoff")
 
-    # collect all available default HPO terms
+    # collect all available default HPO terms and populate
```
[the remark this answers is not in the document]
northwestwitch: Fixed

scout/adapter/mongo/query.py
```diff
@@ -166,8 +165,10 @@ def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
             # panel is used, addition of this is delayed until after the rest of
```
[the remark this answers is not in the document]
northwestwitch: Fixed

scout/adapter/mongo/query.py
```diff
@@ -409,24 +387,19 @@ def gene_filter(self, query, mongo_query):
...
-            mongo_query["hgnc_symbols"] = {"$in": hgnc_symbols}
-            LOG.debug("Adding hgnc_symbols: %s to query" % ", ".join(hgnc_symbols))
-        elif gene_panels:  # gene_panels
-            mongo_query["panels"] = {"$in": gene_panels}
-
-        return gene_query
+        hgnc_symbols = set(query.get("hgnc_symbols", []))
+
+        # Grab all genes from selected gene panels:
+        for panel in query.get("gene_panels", []):
```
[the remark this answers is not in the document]
northwestwitch: Yes, fixed!

scout/server/blueprints/variants/controllers.py
```diff
@@ -591,12 +591,38 @@ def upload_panel(store, institute_id, case_name, stream):
     return hgnc_symbols
 
 
+def gene_panel_choices(store, institute_obj, case_obj):
```
northwestwitch, line 594: This is the main function that populates the choices of the gene panel select for the variants filters. It is called every time you land on a variantS page. I was thinking that how it is now it's quite slim and fast. @dnil I like your suggestion to include also panel version and number of genes (as in a previous version of this PR) but to get this info it means that for each loop of the second for (line 612) you need to query the database for the panel object, so you know the version and number of genes. First loop (line 607) is fine because we know version and genes of case-specific panels.

I suggest a compromise here: specify version of panel and n. genes for case-specific panels and write instead 
`panel name (latest)` for the institute-specific panels. So no query is required. What do you say?

dnil: I think that is quite acceptable. It will actually be a slight improvement over - and a clarification or reminder of the current mode of operation.

northwestwitch:
Fixed. It looks like this now: 

image

I've also introduced a test of the function that populates this select

northwestwitch:
Do we want to reverse the order of the options in the select? Like first case panels, then HPO, then institute panels? Because it looks like that the first element in the list (case panels) end up in the bottom of the select

dnil:
> Do we want to reverse the order of the options in the select? Like first case panels, then HPO, then institute panels? Because it looks like that the first element in the list (case panels) end up in the bottom of the select

Sounds great! If we can get away with separators between the three in the list that would be even clearer!

northwestwitch:
> Sounds great! If we can get away with separators between the three in the list that would be even clearer!

So they can select the separators!! 😆

dnil:
> > Sounds great! If we can get away with separators between the three in the list that would be even clearer!
> 
> So they can select the separators!! 😆

Yes! And cause `== None` type errors all over!

northwestwitch:
> Do we want to reverse the order of the options in the select?

I've figured out that this phenomenon is due to the fact that I'm using sets to avoid duplicates, I'm going to switch to arrays instead. And perhaps there will be duplicates when you have `panelx v.7 (n genes)` and `panelx (latest)` but it doesn't matter, the query is still gene based!

northwestwitch:
> > Do we want to reverse the order of the options in the select?
> 
> I've figured out that this phenomenon is due to the fact that I'm using sets to avoid duplicates (and sets are not oredered), I'm going to switch to arrays instead. And perhaps there will be duplicates when you have `panelx v.7 (n genes)` and `panelx (latest)` but it doesn't matter, the query is still gene based!

northwestwitch:
Yes, now it's ordered right:

image

Do we really need these separators? I would skip them

dnil:
No, no need for separators - that is fine! Let’s just think once more about how this will work and be interpreted by the users. Need a ☕  for that - brb!

dnil:
No, this doesn't quite fly for me yet: sorry to be a nitpick! Let me try to show you where this is difficult for me. 

I made a new version of the Test panel, removing POT1 in v2.0. Here's what that will look like in the current branch, with latest panel fresh from the panels:
Screenshot 2020-09-03 at 10 21 28
This is practical and pragmatic, but the version shown gives the impression one is viewing the results from panel v1.0, which is not the case. One could just remove the version and say we always use latest, but that would change the present behaviour.

Swithing back to master we still use the case panels:
Screenshot 2020-09-03 at 10 22 15

I would love being able to just tick "latest" panel to be able to get a new version (hoping that that gene was already on the MIP Select list). The code will be more complicated but my feeling is that it is hard to get past the ability to fetch at least the version loaded with the case?

scout/server/blueprints/variants/views.py
dnil: 😝 I know the feeling.. 
Anyway, I think the code appears sound, just saying the comment says something different. The list of panels being set is filtered with `if panel["is_default"] is True` and the comment kind of ignores that. This makes me uneasy. 😨 😉

northwestwitch:
I totally agree with you, the select is showing the panel version of the case, but the query uses the last panel. Why don't we simplify and just remove the version? I'm pretty sure that even if a case was runned with panel version 1 then, if the panel was updated, users might be interested in seeing if any gene from the latest panel is found in the case.. It is possible to pass the panel version to the query function that gets the genes of a panel, but perhaps is an overkill?

dnil:
Assuming always using latest is a desired behaviour change (which I agree it overwhelmingly is), how about legacy panels and cases with panels deleted from db?

dnil:
Perhaps they are already partly broken enough anyway?

northwestwitch:
Its not broken because keep in mind how the query is created now --> look for variants that have` panels:[panel list]`, variants will be found even if the panels are removed.

Now variants will not be found because the query looks first for the panel obj and if it finds it, gets the genes: https://github.com/Clinical-Genomics/scout/blob/63ccd73a6890f4fb21529327e86d941df1219e74/scout/adapter/mongo/panel.py#L308

So now it's not broken either.

But isn't it impossible for common users to delete gene panels? And most importantly, why would they care to see genes for a panel that has been deleted (assuming it was deleted for a reason..)

🤔

northwestwitch:
So in the end what should I do with this one @dnil? I vote for simplifying it and just filter by the latest panels, both for cases-specific and institute-specific panels. I'd remove also versions and number of genes, because otherwise it may lead to a lot of hassle, (unclear why some panels have a version, some panels don't, request to include panels with some versions and so on). I'd say let's keep it simple and possibly bug-free

dnil:
Cleaning day at CG - will be back in a few h.. 😅

dnil:
Agreed!

I've had a discussion with @moonso: he had a sound suggestion for how to handle the missing genes. At some stage of the query (or possibly on the case page display) we should also compare to the case panel version genes, and list ones that are new in a flash - or such with the python set functions. That way we get both the latest set, and inform users about what genes in the latest that will not have any variants loaded for the case, and allow an informed decision about a rerun.

northwestwitch:
> I've had a discussion with @moonso: he had a sound suggestion for how to handle the missing genes. At some stage of the query (or possibly on the case page display) we should also compare to the case panel version genes, and list ones that are new in a flash - or such with the python set functions.

That's a very good idea! I'll fix it to show the flash message, and I'll revert to the simpler version with only gene panel name

northwestwitch:
I need a suggestion here
If users run queries using many panels, should we show:

- Panel x in the case is missing genes : [ list of genes ] that are present in latest panel
- Panel y in the case is missing genes : [ list of genes ] that are present in latest panel
- And so on?

Panels to genes conversions happens in the query (adapter), and the query returns variants, not the query itself. 
And gene panels at the case level does not contain the name of the genes, just the number. Example:

image


This means that every time a user makes a query, first we need to make another query to get all genes of each panel of a case, (then possibly another query to get the genes of the latest panels), and then we can compare them to flash the messages.

After this the variants query is runned. I think it's a bit heavy for the pages to have it like this. Any suggestion?

northwestwitch:
Wouldn't it be better to show on case page then (just once) that the gene panels used in the analysis might be outdated because version has changed (if it has changed)?

dnil:
> Wouldn't it be better to show on case page then (just once) that the gene panels used in the analysis might be outdated because version has changed (if it has changed)?

How about showing exactly that on the case page, with a link to something that will make the heavy querying once and determine what genes differ?

northwestwitch:
> How about showing exactly that on the case page, with a link to something that will make the heavy querying once and determine what genes differ?

YES!

northwestwitch:
I was thinking that the perfect place to have such alert would be this:

image

But can we take it in another pull request? I feel like this one is huge as it is already..

dnil:
Sure - and I can try to help out, even if I think you currently understand it best. But we need that before we can release this part. Then we leave this one hanging, or how should we do it? We can merge, if you are confident the other part is easy to do before next release?

northwestwitch:
> But we need that before we can release this part.

I wrote it, just on a different PR! ;)

dnil:
Very close.. This branch still has the issue with filtering for all genes if the search gene was not found, but that is exactly the same as master by the looks of it. We should presumably fix it, but does not have to be this PR - a new issue is a fine solution.

dnil (approved):
This is Good Stuff. The only remaining discussion is on the content of comments, so I approve!

northwestwitch:
I want to include this PR in the release of today @dnil, because it has also a couple of important fixes. Do you want other fixes here?

dnil:
Agreed: as per last comment, I only remember fixes to comments. I think it is ok to merge.

northwestwitch:
> I only remember fixes to comments

Didn't I fix those?

dnil:
> > I only remember fixes to comments
> 
> Didn't I fix those?

I'll check..

scout/server/blueprints/variants/views.py
dnil:
```suggestion
        # set gene panels to case-specific default gene panels
```

scout/server/blueprints/variants/views.py
dnil, line 301:
```suggestion
        # set gene panels to case-specific default gene panels
```

dnil: Or just remove the comments alltogether - or change them to discuss case level versus institute level or sth, but as they stand they only make me confused.. 😅

northwestwitch: Ok I'll remove them

PR #2010, opened 2020-08-17
Add pre-commit and black packages in dev requirements, edit README

elevu:
This PR adds Black and pre-commit packages to requirements-dev. Since Black is run in Github Actions
I think it's a good idea to have it installed by default when setting up the project.

**Review:**
- [x] code approved by

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,33 @@ A user-oriented guide describing how to share case and variant data to MatchMake
 be found [here][matchmaker-scout-sharing].
 
 
+## Development
+
+To keep the code base consistent, formatting with [Black](https://github.com/psf/black) is enforced.
+Black defaults to 88 characters per line, we use 100.
+
+To format all the files in the project run:
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -191,6 +191,33 @@ A user-oriented guide describing how to share case and variant data to MatchMake
 be found [here][matchmaker-scout-sharing].
 
 
+## Development
+
+To keep the code base consistent, formatting with [Black](https://github.com/psf/black) is enforced.
```
dnil, line 196: Enforce is a forceful word.😉 How about adding another sentence, like so:
```suggestion
To keep the code base consistent, formatting with [Black](https://github.com/psf/black) is always applied as part of the PR submission process via GitHub Actions. While not strictly required, to avoid confusion, it is suggested that developers apply Black locally.
```

elevu: Good suggestion, file edited!

PR #2013, opened 2020-08-18
Redirect to login page when user session expires and user requests a report

### LINKED ISSUE #2008: AttributeError: 'AnonymousUserMixin' object has no attribute 'is_admin'
I've been seeing this error more than twice now. Let's fix it when we have time. 

Full message:

2020-08-17 07:26:16,064 - scout.server.app - ERROR: Exception on /cust002/F0032947/delivery-report [GET] [in /home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py:1892]
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/hiseq.clinical/SERVER/apps/scout/git/scout/scout/server/blueprints/cases/views.py", line 780, in delivery_report
  File "/home/hiseq.clinical/SERVER/apps/scout/git/scout/scout/server/utils.py", line 55, in institute_and_case
    if not current_user.is_admin:
  File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: 'AnonymousUserMixin' object has no attribute 'is_admin'

northwestwitch:
fix #2008.

The error is caused by the fact that the session expires and the current_user is logged out. If the unlogged user is requesting one of the reports then the anonymous user object does not have the "admin" attribute any more. 

I'm fixing it in 2 ways:
- Simply redirect to login for all pages that are not static ("report" pages)
- Introduce an extra check in the method institute_and_case(that is used EVERYWHERE) to return HTTP error 401 page (unauthenticated) whenever the user is anonymous. (Might be an overkill but better safe than sorry)

**How to test**:
1. On a local instance of Scout, master branch, login user Clark Kent 
1. Go to a test case page and click on the case general report (HTML report is fine)
1. Go to the browser settings and remove the session cookies (or cookie)
1. Reload the report page and see that the page crashes.
1. Checkout to this branch and repeat the process.
1. The report page will redirect to the login page instead

**Expected outcome**:

**Review:**
- [x] code approved by DN


northwestwitch:
Done and ready for review

dnil:
Nice test description - thanks!
Screenshot 2020-08-20 at 09 24 20

scout/server/app.py
```diff
@@ -74,7 +74,7 @@ def create_app(config_file=None, config=None):
     def check_user():
         if not app.config.get("LOGIN_DISABLED") and request.endpoint:
             # check if the endpoint requires authentication
-            static_endpoint = "static" in request.endpoint or "report" in request.endpoint
+            static_endpoint = "static" in request.endpoint
```
dnil, line 77: 🙄

dnil: I suppose this is a leftover from some version of Chanjo- or other reports integration. Did you by any chance try this on stage and click through a case with all available report types? 😊

scout/server/utils.py
```diff
@@ -40,6 +40,11 @@ def public_endpoint(function):
 
 def institute_and_case(store, institute_id, case_name=None):
     """Fetch insitiute and case objects."""
+
+    # Check if user is authenticated. This should prevent the access to un-authenticated users and registered users with expired session
+    if current_user.is_authenticated is False:
```
dnil, line 45: But.. why would you need this? I can perhaps understand wanting to issue a 401 specifically, but the auth should be handled already for all endpoints under Flask-Login? Specifically by the server app before_request decorator that you edited? Please give a counterexample otherwise if I'm missing something! 

I'm a little worried constructs like this would cause later confusion about where auth is actually handled!

dnil (changes requested):

scout/server/utils.py
northwestwitch: No, I agree, that's why I wrote that it's probably an overkill. Shall we remove it?

scout/server/app.py
northwestwitch: Yes, probably from Chanjo. I'll install the branch on stage, better to check properly, I agree!

scout/server/utils.py
dnil: Aye, then we agree, and I see the reason. I missed the overkill caveat - better safe then sorry is not a bad principle. Personally, I think it better to keep it clean and correct. This is an area where worry and confusion often kicks in.

scout/server/app.py
northwestwitch: It's on stage now. Looks like all the reports redirect to the login page as expected.

dnil: Thanks!

dnil (approved):

dnil:
..but you already know what I will remind you to do..

northwestwitch:
> _says in Oracle voice_ ..but you already know what I will remind you to do..

The changelog?

PR #2014, opened 2020-08-18
Fix #2011. Add safety check for load config alignment_path.

### LINKED ISSUE #2011: Case does not display BAM
Seems like some (all) cases delivered lately lack the BAM-files or do not connect them to the cases? Analysis date 2020-08-16
I have problems with:
F0035793
F0035649
F0035665

My colleague could not see the BAM´s in IGV on these cases either.

dnil:
This PR adds a sanity check for file loading, changing the priority from bam to cram (alignment_path), warning the user if multiple files were added to config.

**How to test**:
1. add an extra non empty bam_file or bam_path entry to an individual that has an alignment_path entry - or vice versa.

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Bug occurring when rerun is requested twice
 - Peddy info fields in the demo config file
+- Added load config safety check for multiple alignment files for one individual
 
 ### Changed
 - Updated the documentation on how to create a new software release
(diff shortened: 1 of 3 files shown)


dnil:
Yes, this is an old Band-Aid that would have fared better pulling. I'm unsure which way to go though; removing either would not be backwards compatible - and potentially not compatible with other sites as well. But, I suppose "bam_file" and "bam_path" should go deprecated..

northwestwitch:
> But, I suppose "bam_file" and "bam_path" should go deprecated..

Yes. Regarding the backward compatibility we could only modify the upload process to accept "alignment_path" and keep the other 2 keys in the rest of the app. This way we'd know that at all new cases will have the right key.

Issue #2012, opened 2020-08-18
Highlight causative variant(s) on the variants lists


dnil:
With dismissed variants showing some color it would be appropriate with e.g. a green background for causatives on the variants lists. This would help help looking through old cases.

elevu:
I can start with this task!

PR #2015, opened 2020-08-18
Highlight causative variants in the variants list #2012

elevu:
This PR adds a functionality, in the variant list causative variants are now highlighted. The tables style was also improved.
Screen Shot 2020-08-18 at 10 40 15

**How to test**:
1. Run Scout locally, mark a variant causative.

**Expected outcome**:
In the list, the causative variant should be highlighted.

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Display number of filtered variants vs number of total variants in variants page
 - Search case by HPO terms
-- Dismiss variant column in the variant tables.
+- Dismiss variant column in the variants tables
 - Black and pre-commit packages to dev requirements
+- Highlight on causative variants in the variants tables
 
 ### Fixed
 - Bug occurring when rerun is requested twice
@@ -20,6 +21,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Genome build-aware cytobands coordinates
 - Styling update of the Matchmaker card
 - Select search type in case search form
+- Highlight color on normal STRs in the variants table from green to blue
 
 
 ## [4.19]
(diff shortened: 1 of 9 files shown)


scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -66,8 +66,10 @@
           {% set ns.allele0 = variant.chromosome + variant.position|string %}
           {% if variant.dismiss_variant %}
               <tr class="dismiss">
+          {% elif variant._id in case.causatives %}
+             <tr class="causative">
```
dnil, line 70: Ah, ouch! My bad! The STRs. They are sort of greeninsh when normal, so green will perhaps not be so great for causative, or what do you think? Unless green in some other way, like with outline or font color changed? Suggestions?

elevu: I changed the normal STRs color to primary blue. Since green is the color of solved cases in the UI I think it is consistent to use it for causative variants as well. Less consistent to have the same green in solved cases and normal STRs. The blue is a bit dark for my taste but I think it works. What do you think?
Screen Shot 2020-08-18 at 15 35 31

dnil: Thats excellent! Agreed, a lighter blue would have been more appealing here, but recurring color is also important.

CHANGELOG.md
```diff
@@ -8,8 +8,9 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Display number of filtered variants vs number of total variants in variants page
 - Search case by HPO terms
-- Dismiss variant column in the variant tables.
+- Dismiss variant column in the variants tables
```
dnil, line 11: 😄 👍

dnil (approved):
Nice! Thank you!

PR #2016, opened 2020-08-18
Version 4.20

northwestwitch:
I was preparing patch release 4.19.1 but then I realized that we actually have new features, so I made it a minor instead!

Contents of this release:

### Added
- Display number of filtered variants vs number of total variants in variants page
- Search case by HPO terms
- Dismiss variant column in the variant tables.
- Black and pre-commit packages to dev requirements

### Fixed
- Bug occurring when rerun is requested twice
- Peddy info fields in the demo config file
- Added load config safety check for multiple alignment files for one individual

### Changed
- Updated the documentation on how to create a new software release
- Genome build-aware cytobands coordinates
- Styling update of the Matchmaker card
- Select search type in case search form

I'm going to test all the above features before merging and deploying!

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
This branch is on stage now. Feel free to give it a try!

northwestwitch:
Woa, what happened to the case filter checkboxes?

image

dnil:
Filter boxes still look ok on my Safari, but I'm not 100% about cache status etc.

northwestwitch:
One thing that I've done and then immediately regretted is this:

image

I was going to fix this in #2007, by writing instead "showing n/n variants".

dnil:
Yeah, I was just thinking about how to mention that.. 😊 Thanks! The functionality is great though.

northwestwitch:
What if we install this version on hasta only then? We could always skip it and go for the next on clinical-db

dnil:
I also really like the functionality of dismissing from the variants view, but we may want to revisit the table element sizing again, as the first column can grow quite a bit:
Screenshot 2020-08-18 at 15 14 41

dnil:
> What if we install this version on hasta only then? We could always skip it and go for the next on clinical-db

Now that would be a first! I don't see why not. We could do a 4.21 fairly soon with some cosmetics and whatever we find?

northwestwitch:
Ok then, deployment on Hasta only

dnil:
Was there anything in that "Bug occurring when rerun is requested twice" that might have interaction issues?

northwestwitch:
> Was there anything in that "Bug occurring when rerun is requested twice" that might have interaction issues?

Naa, it's on the frontend side

PR #2028, opened 2020-08-20
Enable csrf protect module

### LINKED ISSUE #2021: Saved filters fail to load and throw an error.
Scout version: 4.19
Tested on Stage, Scout version: 4.20

Saved filters do not work and it throws a weird error:

image

test case (it is not a patient sample):
https://scout.scilifelab.se/cust087/calmpuma/cancer/variants?expand_search=True

northwestwitch:
Load csrf protection at the app level. Fix #2021. Documentation is here: https://flask-wtf.readthedocs.io/en/v0.12/csrf.html

**How to test**:
1. Install the branch either locally or on scout stage
1. If testing locally, load the demo cancer case with the following command:
`scout --demo load case scout/demo/cancer.load_config.yaml `
1. Go to the cancer variants and try to filter by some criteria (your choice really)
1. Make sure that variants are filtered correctly
1. Save the filters used using the "save filter" button
1. Reset filter and make sure you see again all variants
1. Load the saved filter.
1. Filter load and filtering itself should work
1. Remove the filter

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -6,6 +6,7 @@ from flask import Flask, current_app, redirect, request, url_for
 from flask_babel import Babel
 from flask_login import current_user
 from flaskext.markdown import Markdown
+from flask_wtf.csrf import CSRFProtect
 
 from scout.utils.matchmaker import mme_nodes
 
@@ -96,6 +97,7 @@ def configure_extensions(app):
     extensions.login_manager.init_app(app)
     extensions.mail.init_app(app)
 
+    CSRFProtect(app)
     Markdown(app)
 
     if app.config.get("SQLALCHEMY_DATABASE_URI"):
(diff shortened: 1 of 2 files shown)


northwestwitch:
I'm not super convinced that this is the right approach, since EVERY form in the app should have the token, otherwise it won't work. It's a bit scary because it might broken stuff that is not validated now.
Going back in time this would have been the best solution tho.

dnil:
Agree. I'm more than a little surprised to see another way to enable it - but I suppose that is for the validation mechanism?
I seem to recall several instances of debugging when it turned out the problem was a small form with missing CSRF token.

northwestwitch:
> but I suppose that is for the validation mechanism?

I guess

dnil:
Otherwise I will start blaming my missing filters on cross-site scripting attacks from StarLIMS instead of miss-clicks by colleagues. 😉

dnil:
We can check when reviewing.

northwestwitch:
> We can check when reviewing.

I'm confident that many forms won't work any more. I just don't know which ones

northwestwitch:
I'm going to close this, because I've opened another PR with safer code (#2030)

Issue #2032, opened 2020-08-21
Make sure newly created gene panels names do not contain leading and trailing spaces 


northwestwitch:
It's quite unfortunate that if creating a gene panel named "HELLO" one adds a space in the form and the panel name becomes "HELLO "

PR #2037, opened 2020-08-24
No trailing spaces in new gene panels #2032

elevu:
This PR filters remove leading and trailing whitespaces from the gene panel name
**How to test**:
Add a panel with a name that contains trailing whitespaces in your local or stage branch
**Expected outcome**:
The uploaded gene panel name should not have trailing whitespaces

**Review:**
- [x] code approved by CR
- [x] tests executed by CR, EM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Highlight causative variants in the variants list
 - Add tests. Mostly regarding building internal datatypes.
+- Remove leading and trailing whitespaces from panel_name and display_name
 ### Fixed
 - Report pages redirect to login instead of crashing when session expires
 - Variants filter loading in cancer variants page
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Highlight causative variants in the variants list
 - Add tests. Mostly regarding building internal datatypes.
-- Remove leading and trailing whitespaces from panel_name and display_name
+- Remove leading and trailing whitespaces from panel_name and display_name when panel is created
 ### Fixed
 - Report pages redirect to login instead of crashing when session expires
 - Variants filter loading in cancer variants page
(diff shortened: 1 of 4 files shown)


northwestwitch (changes requested):
Tested on a local instance of Scout and it works! 👍 

This fix can be handled at different levels:
- When you parse the request from the user form / or the params provided in the cli
- When you pass these params to the server controller
- When you build the panel object in scout/build/panel/panel.py
- When you save the new panel to the database (the solution you choose)

While your solution works, for consistency with the rest of the code, I would add the strip() to the build function instead, and ideally I would remove the spaces also from the other parameter that gets saved in the database (display_name). This way the mongo adapter function is only responsible for saving the panel, while all the fixing is done by the builder. Does it make any sense?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Highlight causative variants in the variants list
 - Add tests. Mostly regarding building internal datatypes.
+- Remove leading and trailing whitespaces from panel_name and display_name
```
northwestwitch, line 11:
```suggestion
- Remove leading and trailing whitespaces from panel_name and display_name when panel is created
```

tests/build/test_build_panel.py
```diff
@@ -28,10 +28,10 @@ def test_build_panel(institute_database, test_gene):
     adapter.load_hgnc_gene(test_gene)
 
     panel_info = {
-        "panel_id": "panel1",
+        "panel_id": " panel1",
```
northwestwitch, line 31: Perhaps add a small comment here on why you added a white space

tests/build/test_build_panel.py
```diff
@@ -41,6 +41,8 @@ def test_build_panel(institute_database, test_gene):
 
     ## THEN assert that the panel was given the right attributes
     assert panel_obj["institute"] == panel_info["institute"]
+    assert panel_obj["panel_name"] == "panel1"
```
northwestwitch, line 44: And here saying that there is no space in result panel

northwestwitch (approved):
Just small things!

PR #2042, opened 2020-08-25
User load docs refresh.

dnil:
Refreshing docs for user "loading". Also updated old login error message flash.

**How to test**:
1. on a demo instance/local instance without OAuth or LDAP, attempt logging in as a non-registered user and note the error message.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/docs/admin-guide/login-system.md
+++ b/docs/admin-guide/login-system.md
@@ -72,4 +72,4 @@ Please note that the while the `-id ldap_id` option is not a mandatory parameter
 
 ## Simple login with userid and password
 
-Basic login with userid and password is the login system available whenever no advanced login system (either Google or LDAP) is specified in the Scout config file. It is an un-secure system which is not recommended to use. This is also the login system available in the demo instance of Scout.
+Basic login with userid and password is the login system available whenever no advanced login system (either Google or LDAP) is specified in the Scout config file. To enable it, comment out or remove the LDAP and GOOGLE lines mentioned above from your config. It is an un-secure system which is not recommended to use. This is also the login system available in the demo instance of Scout.
--- a/scout/server/blueprints/login/views.py
+++ b/scout/server/blueprints/login/views.py
@@ -79,11 +79,11 @@ def login():
 
     if request.args.get("email"):  # log in against Scout database
         user_mail = request.args.get("email")
-        LOG.info("Validating user {} against Scout database".format(user_id))
+        LOG.info("Validating user email {} against Scout database".format(user_mail))
 
     user_obj = store.user(email=user_mail, user_id=user_id)
     if user_obj is None:
-        flash("User not whitelisted", "warning")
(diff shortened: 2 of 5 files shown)


dnil:
Screenshot 2020-08-25 at 07 41 37

scout/server/blueprints/login/views.py
```diff
@@ -79,11 +79,11 @@ def login():
 
     if request.args.get("email"):  # log in against Scout database
         user_mail = request.args.get("email")
-        LOG.info("Validating user {} against Scout database".format(user_id))
+        LOG.info("Validating user email {} against Scout database".format(user_mail))
 
     user_obj = store.user(email=user_mail, user_id=user_id)
     if user_obj is None:
-        flash("User not whitelisted", "warning")
+        flash("User not found", "warning")
```
northwestwitch, line 86: 👍

northwestwitch:
We could perhaps add somewhere that when loading a user for simple login or login using google oauth then the -id param should NOT be specified

northwestwitch:
> We could perhaps add somewhere that when loading a user for simple login or login using google oauth then the -id param should NOT be specified

Or the other way around: -id should be specified only if login is using LDAP

dnil:
Yeah, like the latter better: it doesn't exactly break if you give another id for pwd auth. Would it break OAuth? No?

northwestwitch:
> Yeah, like the latter better: it doesn't exactly break if you give another id for email only auth. Would it break OAuth? No?

Perhaps not, but it's not really required either

dnil:
> > Yeah, like the latter better: it doesn't exactly break if you give another id for email only auth. Would it break OAuth? No?
> 
> Perhaps not, but it's not really required either

I reread the login system page; I think it works as is, but feel free to suggest changes!

docs/admin-guide/login-system.md
```diff
@@ -72,4 +72,4 @@ Please note that the while the `-id ldap_id` option is not a mandatory parameter
 
 ## Simple login with userid and password
 
-Basic login with userid and password is the login system available whenever no advanced login system (either Google or LDAP) is specified in the Scout config file. It is an un-secure system which is not recommended to use. This is also the login system available in the demo instance of Scout.
+Basic login with userid and password is the login system available whenever no advanced login system (either Google or LDAP) is specified in the Scout config file. To enable it, comment out or remove the LDAP and GOOGLE lines mentioned above from your config. It is an un-secure system which is not recommended to use. This is also the login system available in the demo instance of Scout.
```
northwestwitch, line 75: 👍

northwestwitch (approved):
Ok!

dnil:
Aw, CHANGELOG. Oh well, docs.

northwestwitch:
Current master is based on an old branch, we should roll it back to this branch @mikaell

PR #2044, opened 2020-08-25
Fix #2033 -- clinical filter returning no variants

### LINKED ISSUE #2033: Check clinical filter performance on clinical-db stage
Clinical filter of release 4.19, 4.20 and current master doesn't work any more. (Well, it returns no variants, but if one clicks on "filter" immediately after, it works).

Thankfully (🙏) everything works on prod (version 4.19)

northwestwitch:
Fix #2033.

**How to test**:
1. Go to scout stage and install the current master branch.
1. Open a case and go to variants page
1. Click on the clinical filter button: 

image
1. No variants should be returned
1. Install this branch on scout stage
1. Repeat the search and variants should be shown

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/forms.py
+++ b/scout/server/blueprints/variants/forms.py
@@ -113,6 +113,7 @@ class VariantFiltersForm(FlaskForm):
 
     filter_variants = SubmitField(label="Filter variants")
     export = SubmitField(label="Filter and export")
+    clinical_filter = SubmitField(label="Clinical filter")
 
 
 class FiltersForm(VariantFiltersForm):
@@ -124,7 +125,6 @@ class FiltersForm(VariantFiltersForm):
     spidex_human = SelectMultipleField("SPIDEX", choices=SPIDEX_CHOICES)
 
     local_obs = IntegerField("Local obs. (archive)")
-    clinical_filter = SubmitField(label="Clinical filter")
 
 
 class CancerFiltersForm(VariantFiltersForm):
@@ -159,5 +159,3 @@ class SvFiltersForm(VariantFiltersForm):
     decipher = BooleanField("Decipher")
     clingen_ngi = IntegerField("ClinGen NGI obs")
     swegen = IntegerField("SweGen obs")
-
-    clinical_filter = SubmitField(label="Clinical filter")
--- a/scout/server/blueprints/variants/forms.py
+++ b/scout/server/blueprints/variants/forms.py
@@ -105,7 +105,9 @@ class VariantFiltersForm(FlaskForm):
     load_filter = SubmitField(label="Load filter")
     delete_filter = SubmitField(label="Delete filter")
 
-    chrom = SelectField("Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS)
+    chrom = SelectField(
+        "Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS, default=("", "All")
+    )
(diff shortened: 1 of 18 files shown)


scout/server/blueprints/variants/forms.py
```diff
@@ -105,14 +105,17 @@ class VariantFiltersForm(FlaskForm):
...
+        "Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS, default=""
+    )
     start = IntegerField("Start position", [validators.Optional()])
     end = IntegerField("End position", [validators.Optional()])
     cytoband_start = NonValidatingSelectField("Cytoband start", choices=[])
     cytoband_end = NonValidatingSelectField("Cytoband end", choices=[])
 
     filter_variants = SubmitField(label="Filter variants")
     export = SubmitField(label="Filter and export")
+    clinical_filter = SubmitField(label="Clinical filter")
```
dnil, line 118: Right, but this isn't really supposed to be instantiated, and Cancer does not have one (yet)?

scout/server/blueprints/variants/forms.py
```diff
@@ -105,14 +105,17 @@ class VariantFiltersForm(FlaskForm):
     load_filter = SubmitField(label="Load filter")
     delete_filter = SubmitField(label="Delete filter")
 
-    chrom = SelectField("Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS)
+    chrom = SelectField(
+        "Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS, default=""
```
dnil, line 109: Was that it?!

northwestwitch: It's the default option with All chromosomes, look at line 33 :
`CHROMOSOME_OPTIONS = [("", "All")] + [(chrom, chrom) for chrom in CHROMOSOMES]`

scout/server/blueprints/variants/forms.py
northwestwitch: This is not really affecting the query I guess, shall I revert this code?

dnil: I think it would be less confusing, as long as Cancer does not have one,  or what do you think?

northwestwitch: I can remove it!

scout/server/blueprints/variants/forms.py
dnil: > It's the default option with All chromosomes, look at line 33 :
> `CHROMOSOME_OPTIONS = [("", "All")] + [(chrom, chrom) for chrom in CHROMOSOMES]`

Yeah, agree, thats the way it should be - just dont quite see the connection to the clinical filter, which basically has its own set of parameters. But ah - perhaps not for chromosome?

northwestwitch: Nope, chrom was None

scout/server/blueprints/variants/forms.py
northwestwitch: Done!

scout/server/blueprints/variants/forms.py
dnil: One could also set that explicitly in the clinical filter dict. But this deals with multiple potential issues, so lets leave it at that! Well found!

dnil (approved):
:+1:

PR #2050, opened 2020-08-26
Pagination revert

dnil:

**How to test**:
1. check through all variants pages and make sure all pagination buttons (Next, First, final-First) work, and that the corresponding page layout looks ok.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -628,3 +628,34 @@
   {% endif %}
 
 {% endmacro %}
+
+{% macro pagination_hidden_div(page) %}
+  <div class="hidden">
+    <input type="submit" name="page" id="paginate-first" value=1 hidden="true">
+    <input type="submit" name="page" id="paginate-next" value={{page+1}} hidden="true">
+  </div>
+{% endmacro %}
+
+{% macro pagination_footer(more_variants) %}
+  <div class="container-fluid">
+      {% if more_variants %}
+        <div class="d-flex justify-content-around">
+	         <div>
+             <label for="paginate-first" class="btn btn-outline-secondary mx-auto d-block" tabindex="0">
+              First page
+            </label>
+          </div>
+          <div>
+            <label for="paginate-next" class="btn btn-outline-secondary mx-auto d-block" tabindex="0">
+              Next page
+            </label>
+          </div>
+	     </div>
+      {% else %}
+        <i class="text-muted">No more variants to display</i>
+        <label for="paginate-first" class="btn btn-outline-secondary mx-auto d-block" tabindex="0">
+          First page
+        </label>
+      {% endif %}
+  </div>
+{% endmacro %}
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -630,6 +630,7 @@
 {% endmacro %}
 
 {% macro pagination_hidden_div(page) %}
+{# Used inside filters form to introduce submit buttons for footer pagniation #}
   <div class="hidden">
     <input type="submit" name="page" id="paginate-first" value=1 hidden="true">
     <input type="submit" name="page" id="paginate-next" value={{page+1}} hidden="true">
@@ -637,6 +638,7 @@
 {% endmacro %}
 
 {% macro pagination_footer(more_variants) %}
+{# Used outside filters form to introduce footer pagniation labels #}
   <div class="container-fluid">
       {% if more_variants %}
         <div class="d-flex justify-content-around">
--- a/scout/server/blueprints/variants/templates/variants/variants.html
+++ b/scout/server/blueprints/variants/templates/variants/variants.html
@@ -159,15 +159,21 @@
       {% if more_variants %}
         <div class="d-flex justify-content-around">
 	         <div>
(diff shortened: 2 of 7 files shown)


dnil:
Yes, a new small form before closure of the main one. This will keep happening, so I went with the STR version, where pagination is handled as links. One could also do this with a small js, invoking the form submit from outside the form. That way, less code needs be reproduced, but I would have to test more.. 😝

dnil:
Already regretting not writing those js lines.

northwestwitch:
I can review!

northwestwitch:
Testing this branch on stage. 
**variants page**: on chrome, looks like the get request is adding the page number in the args, but pages don't go beyond page=2. Another thing is that the list of variants contains always the same variants (from page 1).

**structural variants**: whenever there are just a bunch of variants (for instance https://scout-stage.scilifelab.se/cust000/GIAluc14p19fam/sv/variants) there is this big "First page" button on the bottom of the page and if you click the app crashes..

dnil:
Thanks! Sorry, I got sidetracked before getting to testing. Can do later afternoon for an easier review..

northwestwitch:
> Thanks! Sorry, I got sidetracked before getting to testing. Can do later afternoon for an easier review..

No rush. The branch is on stage atm, if you want to check!

dnil:
🙄 Lint-action black on 147 files? What was that about? Did we manage to push code w/o black at some point or has it's config changed somehow?

dnil:
I think I'll try that again - some glitch in actions??

dnil:
🤦 Ok, lets solve that tomorrow. Food first. 😸

dnil:
ah, they released a new version of black yesterday. That explains it. Installing it locally and running agrees with the automatic linter.

dnil:
For the reviewer: the following files actually need checking:

northwestwitch:
I'd wait for the black one to be merged first, so the changes are more evident!

dnil:
Thats better. Back to 7 files changed.

northwestwitch:
> Thats better. Back to 7 files changed.

I'll review now!

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -628,3 +628,34 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
   {% endif %}
 
 {% endmacro %}
+
+{% macro pagination_hidden_div(page) %}
```
northwestwitch, line 632: I find it a bit confusing to have 2 macros to create a simple footer with just 2 buttons. Wouldn't it be more readable with just one? You could always put the submits in pagination_footer (where they'd be close to their labels), and pass it also the number of page..

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -154,24 +156,6 @@
   {% endif %}
 {% endmacro %}
 
-{% macro footer() %}
```
northwestwitch, line 157: But I really like the idea of having just one imported macro in all the variants page

northwestwitch (commented):
What about having just one macro for the footer?

scout/server/blueprints/variants/templates/variants/utils.html
dnil: Sorry, I don't quite follow you?
The beauty of this version - thanks to an anonymous forum participant - is that it keeps the inputs hidden within the form, while the labels act as proxy outside. And it still is in keeping with the html definition, and works with IE. If I were to, say, move the inputs to the footer without they would just stop working due to the introduction of new forms things like linking to IGV directly or dismissing variants. And if I make new submits outside the main filter form with `form="filters_form"` that will only work for some modern browsers.. 

I still think it may be possible to do this with java script, though I struggled to do it without importing some jQuery helpers, but then I would still need at least two macros - one for the buttons, and one for the js code?

dnil: Oh, and I perhaps forgot to mention: the STR page link solution will not work on the larger filters. As soon as you open for things like arbitrary gene lists, GET is not really feasible and one has to go with POST:

northwestwitch:
Ok but then you should have said that it was a design of beauty of some forum participant that stays anonymous to avoid notoriety (kidding 😆🌷). Anyway, I see yoit point, the form being much higher in the page, and separated from the footer. I'll take a better look at it tomorrow!

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: Your solution works and it's fine by me, but please add a couple of comments in the 2 macros referencing the other one, mostly for future maintenance. Took me a while to understand the situation there! :)

To use a GET request you what you could do is:
- Save the filter query in a session variable, whenever the user sends a POST request to the variants view.
- Use this variable whenever the variants view is accessed by GET request
- Have a single macro with the pagination, and making it cooler even: https://getbootstrap.com/docs/4.0/components/pagination/
- This pagination can send GET requests to the variants view with page= as argument.
- You can create a paginator with a custom number pages because you know already:
  - The number of variants returned by the query:
image
  - The number of variants in each page.

But I understand this is more complicated and our main concern is to fix the broken links now, so you choose!

northwestwitch (approved):
Works fine, but a suggestion: add comments in the macros explaining that the submit is in the hidden div

PR #2051, opened 2020-08-27
Automatic black 20.8b1, test compatibility with click>7

dnil:
Black updated to 20.8b1 yesterday. This causes some issues with the interaction of pre-commit and GitHub Actions linting. 
An updated black and lint agree well, but pre-commit is not quite there on the train yet. I would suggest disabling it for now?

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by github actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Variants filter loading in cancer variants page
 - User, Causative and Cases tables not scaling to full page
 - Improved docs for an initial production setup
+- Compatibility with latest version of Black
+- Fixed tests for Click>7
+- Clinical filter required an extra click to Filter to return variants
 ### Changed
 - Highlight color on normal STRs in the variants table from green to blue
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -19,6 +19,3 @@ pre-commit
 # docs
 mkdocs
 mkdocs-material==4.6.1
-
-#compatibility
-click<7.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -20,3 +20,5 @@ pre-commit
 mkdocs
 mkdocs-material==4.6.1
 
+#compatibility
+click<7.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ pymongo<3.7
 pathlib
 WeasyPrint==0.42.3
 xlsxwriter
-click<7.0
+click
 
 # webapp login
 authlib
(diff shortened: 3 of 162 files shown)


dnil:
Ok, I think that would be it from the black part. Please check, but from my point of view this works. 

Black does not show up X:ed, and Lint Action does not find anything to fuss about or actually change. Depending on ones local config it may be necessary to go with manually 
```bash
black -l 100 --exclude "/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|\.svn|buck-out|dist)/"
git commit --no-verify
```
or similar to avoid having to see pre-commit formatting and lint-actions until pre-commit catches up.

The new version of black also has newer deps, so we will have problems in October. I suggest we resolve the Click situation soon. In principle it may have been resolved by switching the OAuth system?

Here's the pip summary:
```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

mkdocs-material 4.6.1 requires markdown>=3.2, but you'll have markdown 3.0.1 which is incompatible.
black 20.8b1 requires click>=7.1.2, but you'll have click 6.7 which is incompatible.
```

northwestwitch:
Why did we freeze click to <7 ??

dnil:
I think the old OAuth lib had to do with it, but there may be more. Let's see!

northwestwitch:
> I think the old OAuth lib had to do with it, but there may be more. Let's see!

🤞

dnil:
Doesn't look like much, but help me demarcate here: should I just back out black out of the dev req for now, and solve the click thing in a separate PR, or do it all at once?

northwestwitch:
I think click >=7 has modified the error messages slightly

dnil:
Agree, so not much to change, IF thats the only thing. Maybe I give it a couple of iterations then.

northwestwitch:
> Doesn't look like much, but help me demarcate here: should I just back out black out of the dev req for now, and solve the click thing in a separate PR, or do it all at once?

I have no idea how many changes would that be honestly, let's pytest locally and see how many of them fail perhaps?

dnil:
I'm missing an "update brach" option - will it appear after approval? Or is that because master fails?

northwestwitch:
We're good @dnil! All tests pass now! 👯

northwestwitch (approved):
Looks fine!

northwestwitch:
> "update brach" option - will it appear after approval? Or is

Haven't seen that option in a long while

northwestwitch:
Merging in 3, 2, 1..

dnil:
No, you are right, its keeping track of mergeability now instead.

PR #2057, opened 2020-09-01
Format cancer variants table

elevu:
This PR formats the cancer variant tables

**How to test**:
1. Deployed this branch to staging
2. Checked out the table of this variant that contains a lot of text.
https://scout-stage.scilifelab.se/cust000/20190327_cancer_T_N/cancer/variants?variant_type=clinical

**Expected outcome**:
It should look fine.
Screen Shot 2020-09-01 at 09 25 00
The content is still a bit compressed but it has to be with so much text. The other solution is to decrease the text size but then it wouldn't be so readable.
Edit: with shrinked tags and tooltip:
Screen Shot 2020-09-01 at 12 03 10
When not hovering:
Screen Shot 2020-09-01 at 12 07 11

**Review:**
- [x] code approved by CR
- [x] tests executed by EM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Compatibility with latest version of Black
 - Fixed tests for Click>7
 - Clinical filter required an extra click to Filter to return variants
-- Restore pagination to variants pages
+- Restore pagination and shrink badges in the variants page tables
 
 ### Changed
 - Highlight color on normal STRs in the variants table from green to blue
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Bug occurring when rerun is requested twice
 - Peddy info fields in the demo config file
 - Added load config safety check for multiple alignment files for one individual
+- Formatting of cancer variants table
 
 ### Changed
 - Updated the documentation on how to create a new software release
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -105,43 +105,43 @@ CANCER_TIER_OPTIONS = {
 
 MANUAL_RANK_OPTIONS = {
     8: {
-        "label": "Known pathogenic",
-        "description": "Previously known pathogenic in ClinVar, HGMD, literature, etc",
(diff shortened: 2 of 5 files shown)


hassanfa:
👍🏽  Nice changes.

One thing that can be done in future is to trim away "Gene" from "Gene:transcript:..." column. 

How will this work with a longer tag: 'likely pathogenic'?

northwestwitch:
> How will this work with a longer tag: 'likely pathogenic'?

One could always use a red LP badge that will show "Likely pathogenic" on mouseover

hassanfa:
Badge'em all! Good idea with badges. Let me know if you need feedback. Great work.

northwestwitch:
Let me know when this is ready so I can review!

elevu:
Updated the PR with shrinked badges (screenshots in first comment), now it's ready for review!

northwestwitch (changes requested):
I couldn't test the PR on stage because the variant page doesn't work at the moment, but locally I've observed that now manual rank options tag on variant pages (snv variant, cancer variant etc) looks like this:

image

Instead of like this:

image

I think you should keep the select labels as they were before because it's more clear that way.

What about adding a new key/value to the variant_tags dictionary with the short name for these labels instead? You could use this other key in the select options of the variants, without messing up the pages where the labels should be more self-explanatory..

elevu:
> What about adding a new key/value to the variant_tags dictionary with the short name for these labels instead? You could use this other key in the select options of the variants, without messing up the pages where the labels should be more self-explanatory..

ohh I had looked at it and didn't see it, now I realise the page was not refreshed :D I will apply your solution!

elevu:
Added the property "name" to the variant_tag object. Didn't check on staging because of the temporary variant issue.

Screen Shot 2020-09-01 at 13 38 49

northwestwitch (approved):
Tested again and it works nicely! 👍

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Compatibility with latest version of Black
 - Fixed tests for Click>7
 - Clinical filter required an extra click to Filter to return variants
-- Restore pagination to variants pages
+- Restore pagination and shrink badges in the variants page tables
```
northwestwitch, line 23: 👍

PR #2070, opened 2020-09-03
Fix #2067 - previous button

### LINKED ISSUE #2067: Pagination: previous
Especially for cancer, but hey, why not generally. See https://github.com/Clinical-Genomics/scout/issues/2047

dnil:
This PR adds a functionality: previous buttons.

It also reverts the STR variants page to previous behaviour (e.g. v4.19 and v4.20 - prior to current master), without the general pagination footer. This can migrate when filters are turned on for STRs.

Screenshot 2020-09-03 at 17 20 27

Feel free to suggest an alternative row layout, e.g. button spacing and alignment. 😸 

**How to test**:
1. access snv, sv, str, cancer and cancer_sv variants pages and click all pagination buttons, with and without applied filters to ensure behaviour is as per the button labels

**Expected outcome**:

**Review:**
- [x] code approved by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Remove leading and trailing whitespaces from panel_name and display_name when panel is created
 - Mark MANE transcript in list of transcripts in "Transcript overview" on variant page
 - Show default panel name in case sidebar
+- Previous buttons for variants pagination
 
 ### Fixed
 - Report pages redirect to login instead of crashing when session expires
--- a/scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
+++ b/scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
@@ -88,7 +88,7 @@
     </table>
   </div>
   <div class="container-fluid">
-    {{ pagination_footer() }}
+    {{ pagination_footer(more_variants, page) }}
   </div>
 {% endblock %}
 
--- a/scout/server/blueprints/variants/templates/variants/cancer-variants.html
+++ b/scout/server/blueprints/variants/templates/variants/cancer-variants.html
@@ -110,7 +110,7 @@
   <br>
 </div><!--end of container-float -->
 <div class="container-fluid">
-  {{ pagination_footer() }}
+  {{ pagination_footer(more_variants, page) }}
 </div>
 
 {% endblock %}
--- a/scout/server/blueprints/variants/templates/variants/str-variants.html
+++ b/scout/server/blueprints/variants/templates/variants/str-variants.html
@@ -128,46 +128,36 @@
 {% macro footer() %}
 <form>
   <div class="container-fluid">
-    <div class="form-group text-center">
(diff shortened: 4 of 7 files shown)


dnil:
This is about ready for review. I think I went through most of the testing - may have lacked a case with many cancer_svs on stage - but this is a function that somehow often breaks. At least I think the current macros are much more malleable!

dnil:
Btw note the line about STR pagination: I think it looks broken in the current master - I guess I was working under the assumption we already had filters up for it.

northwestwitch:
I can review!

northwestwitch:
For the button layout, what about a button group? -> https://getbootstrap.com/docs/4.0/components/button-group/

But not grey!

northwestwitch (changes requested):
Tested on snvs, svs, cancer snvs. Looks like it's working as it should.

Some observations:

- when you are on page 2, then it shown both previous page and first page,  but since it's the same page we could perhaps remove "previous page"? example: https://scout-stage.scilifelab.se/cust000/lovedmarlin/cancer/variants

This behavior is not showing for the strs, that have a different button style from the other variants pages by the way (https://scout-stage.scilifelab.se/cust003/20001/str/variants?page=2&variant_type=clinical&chrom=&csrf_token=)

-  When we are on first page we could perhaps remove the "fist page" button, or just change its color (this is totally optional and up to you)

- Couldn't find any case with many cancer SVs to test the page, so I'm assuming it's working as the SNVs page..


- As in my comment, I'd use the bootstrap button group to style it a bit better, but it works! 👍

dnil:
> Tested on snvs, svs, cancer snvs. Looks like it's working as it should.
> 
> Some observations:
> 
> * when you are on page 2, then it shown both previous page and first page,  but since it's the same page we could perhaps remove "previous page"? example: https://scout-stage.scilifelab.se/cust000/lovedmarlin/cancer/variants
> 
> This behavior is not showing for the strs, that have a different button style from the other variants pages by the way (https://scout-stage.scilifelab.se/cust003/20001/str/variants?page=2&variant_type=clinical&chrom=&csrf_token=)
> 
> * When we are on first page we could perhaps remove the "fist page" button, or just change its color (this is totally optional and up to you)
> * Couldn't find any case with many cancer SVs to test the page, so I'm assuming it's working as the SNVs page..
> * As in my comment, I'd use the bootstrap button group to style it a bit better, but it works! 👍

I kind of agree, just to make clear that I left the unnecessary First (and by analogy Previous) since that was current behaviour. It is easy to do - you don't think there is any risk of confusing anyone? I'll try your style suggestion anyway - thanks!

northwestwitch:
Testing it on stage. I've noticed that you removed "first page" and you kept just previous and next:

image

Don't you think that users might still find useful to have a first page?

northwestwitch:
strs page: "no more variants to display" overlaps to the button:

https://scout-stage.scilifelab.se/cust003/20001/str/variants?page=2&variant_type=clinical&chrom=&csrf_token=

image

dnil:
Thanks; those two are fixed now and retested on stage. Merging!

PR #2071, opened 2020-09-04
Fix sv variant header labels#2064

elevu:
This PR adds fixes a bug

**How to test**:
1. Deploy to staging and click on a SV variant tables. Tested locally but I still cannot deploy to clinical-db.
Screen Shot 2020-09-04 at 08 07 28

**Expected outcome**:
The table header should include Score

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/templates/variants/sv-variants.html
+++ b/scout/server/blueprints/variants/templates/variants/sv-variants.html
@@ -73,9 +73,9 @@
           <th style="width:5%">Stop loc</th>
           <th style="width:8%">Length</th>
           <th style="width:8%">Region</th>
-          <th style="width:8%">Function</th>
+          <th style="width:9%">Function</th>
           <th style="width:8%">Frequency</th>
-          <th style="width:10%">Gene(s)</th>
+          <th style="width:9%">Gene(s)</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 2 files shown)


northwestwitch:
Current master SVs page:

image

This branch: 🆗 
image

scout/server/blueprints/variants/templates/variants/sv-variants.html
```diff
@@ -65,16 +65,17 @@
       <thead>
         <tr>
           <th style="width:8%">Rank</th>
```
northwestwitch, line 67: The sum of this is 93, why not 100?

northwestwitch (approved):
tested and it works, I've just a comment.

Don't forget the changelog!

scout/server/blueprints/variants/templates/variants/sv-variants.html
elevu: The table is fitting the screen so it doesn't have to get to 100, the percentage column ratios will be calculated anyway as if the total was 100. But you are right, it might be confusing to not have a percentage reaching 100, I changed it now!
Still looking ok:
Screen Shot 2020-09-04 at 08 31 57

northwestwitch: 💯 !!

northwestwitch (approved):

PR #2077, opened 2020-09-04
Warn users if case default panels are outdated

northwestwitch:
Display a warning badge if case default panels' version is older than panel version in database. This PR goes together with #2007

image

**How to test**:
1. Can be done both on stage or a local implementation of scout.
1. Update a gene panel that is the default gene panel of a case
1. Go to the case and see the warning badge
1. On mouseover it should tell you the gene differences between case panel and updated panel.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -131,7 +131,7 @@ def case(store, institute_obj, case_obj):
             missing_genes = [
                 gene["symbol"] for gene in hgnc_ids_latest_panel if gene not in hgnc_ids_case_panel
             ]
-            if len(extra_genes) or len(missing_genes):
+            if extra_genes or missing_genes:
                 case_obj["outdated_panels"][panel_name] = {
                     "missing_genes": missing_genes,
                     "extra_genes": extra_genes,
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -223,25 +223,19 @@ def _check_outdated_gene_panel(panel_obj, latest_panel):
         missing_genes, extra_genes
     """
     # Create a list of minified gene object for the case panel {hgnc_id, gene_symbol}
-    hgnc_ids_case_panel = [
+    case_panel_genes = [
         {"hgnc_id": gene["hgnc_id"], "symbol": gene.get("symbol", gene["hgnc_id"])}
         for gene in panel_obj["genes"]
     ]
     # And for the latest panel
-
-    hgnc_ids_latest_panel = [
+    latest_panel_genes = [
         {"hgnc_id": gene["hgnc_id"], "symbol": gene.get("symbol", gene["hgnc_id"])}
         for gene in latest_panel["genes"]
     ]
     # Extract the genes unique to case panel
-    extra_genes = [
-        gene["symbol"] for gene in hgnc_ids_case_panel if gene not in hgnc_ids_latest_panel
-    ]
+    extra_genes = [gene["symbol"] for gene in case_panel_genes if gene not in latest_panel_genes]
     # Extract the genes unique to latest panel
-    missing_genes = [
-        gene["symbol"] for gene in hgnc_ids_latest_panel if gene not in hgnc_ids_case_panel
-    ]
-
+    missing_genes = [gene["symbol"] for gene in latest_panel_genes if gene not in case_panel_genes]
     return extra_genes, missing_genes
 
 
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -96,23 +96,39 @@ def case(store, institute_obj, case_obj):
     # Set of all unique genes in the default gene panels
     distinct_genes = set()
     case_obj["panel_names"] = []
+    case_obj["outdated_panels"] = {}
     for panel_info in case_obj.get("panels", []):
         if not panel_info.get("is_default"):
             continue
         panel_name = panel_info["panel_name"]
         panel_version = panel_info.get("version")
         panel_obj = store.gene_panel(panel_name, version=panel_version)
+        latest_panel = store.gene_panel(panel_name)
         if not panel_obj:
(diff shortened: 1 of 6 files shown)


dnil:
Tremendously good! Now have a very good and deserved weekend!

northwestwitch:
You too! 😊

dnil:
Now this is certainly an edge case, but still:
Screenshot 2020-09-07 at 11 20 35
Version 2 w/o POT1, version 3 with it again..

scout/server/blueprints/cases/controllers.py
```diff
@@ -96,23 +96,45 @@ def case(store, institute_obj, case_obj):
...
+                f"Case default panel '{panel_name}' version {panel_version} could not be found, using latest existing version",
+                "warning",
+            )
+
+        # Check if case-specific panel is up-to-date with latest version of the panel
+        if panel_obj["version"] < latest_panel["version"]:
+            extra_genes = [
+                gene.get("symbol", gene["hgnc_id"])
+                for gene in panel_obj["genes"]
+                if gene not in latest_panel["genes"]
```
dnil, line 122: So I guess the issue is here, right? The "genes" are fairly complex objects, and can easily differ over time? Should be something like comparing the actual gene["hgnc_id"] for each of those?

dnil (changes requested):

scout/server/blueprints/cases/controllers.py
northwestwitch: Do you want to compare only the HGNC id or the symbol and the HGNC ID?

dnil: ID only: that is not supposed to change, whereas the gene symbol is quite likely to have changed for a few from the first version!

northwestwitch:
image


> Version 2 w/o POT1, version 3 with it again..

Weird, it's like if it doesn't update, nice find!

northwestwitch:
@dnil should be fixed now. It started complaining for the code complexity, so I moved the gene check to a separate function. It didn't work. I think it keeps complaining because the case  controller is still very long but it was like this from before..

dnil:
Function tests ok! Really nice one!

How about adding a text "The case panel is used in variants filter" to the tooltip. This can then be changed for #2007, to also state that "The latest panel version is used in variants filter".

scout/server/blueprints/cases/controllers.py
```diff
@@ -96,23 +96,37 @@ def case(store, institute_obj, case_obj):
...
                 continue
-            LOG.info("Using panel %s, version %s", panel_name, panel_obj["version"])
+            flash(
+                f"Case default panel '{panel_name}' version {panel_version} could not be found, using latest existing version",
+                "warning",
+            )
+
+        # Check if case-specific panel is up-to-date with latest version of the panel
+        if panel_obj["version"] < latest_panel["version"]:
+            extra_genes, missing_genes = _check_outdated_gene_panel(panel_obj, latest_panel)
```
dnil, line 119: :+1:

scout/server/blueprints/cases/controllers.py
```diff
@@ -198,6 +212,39 @@ def case(store, institute_obj, case_obj):
...
+    ]
+    # And for the latest panel
+
+    hgnc_ids_latest_panel = [
+        {"hgnc_id": gene["hgnc_id"], "symbol": gene.get("symbol", gene["hgnc_id"])}
+        for gene in latest_panel["genes"]
+    ]
+    # Extract the genes unique to case panel
+    extra_genes = [
+        gene["symbol"] for gene in hgnc_ids_case_panel if gene not in hgnc_ids_latest_panel
```
dnil, line 238: Sets might have been even faster, but this doesn't seem bad either. Nicely done!

dnil (approved):

scout/server/blueprints/cases/controllers.py
dnil: > @dnil should be fixed now. It started complaining for the code complexity, so I moved the gene check to a separate function. It didn't work. I think it keeps complaining because the case controller is still very long but it was like this from before..

Well, yes, I agree it is complex, and I suppose what pushed it over the edge is the new nested if statement. Its possible you would get away with it if you just un-nest them - the vars will not be more or less set if you check them outside the outer if-statement I suppose.

dnil: That is to say, I don't mind much either way, being c-schooled, but I suppose flat if-statements is more the python way.

northwestwitch: > Sets might have been even faster, but this doesn't seem bad either. Nicely done!

The problem with sets here is that I'm comparing the hgnc ids, but I need to keep also the symbols, and to do so I store the values in dictionaries. Dictionaries can't be stored in sets (not hashable) so I'd need to have two loops instead, and in the end it would be slower I think. I'll keep it like this instead

northwestwitch: > That is to say, I don't mind much either way, being c-schooled, but I suppose flat if-statements is more the python way.

I wouldn't like to add even more code to a function that by itself is simple. It's just that it's part of a nested check and the complexity is also given by the levels above. Let's keep it this way. We can fix it in a second moment if we feel like it.

northwestwitch:
> Function tests ok! Really nice one!
> 
> How about adding a text "The case panel is used in variants filter" to the tooltip. This can then be changed for #2007, to also state that "The latest panel version is used in variants filter".

I think it would be good to have a tooltip, but I'd rather add it to the variants search select instead. Otherwise this tooltip will become super long in the case there are many gene differences between the case panel and the latest panel

northwestwitch:
I think I'm going to merge this, and I'll add the tooltip that the filtering is performed using the latest panel on the variants filtering page, do you agree @dnil?

northwestwitch:
I changed my mind and added "Case panel version is used in variants filtering" to the case tooltip instead. Have you noticed how voluble I am in my decisions sometimes? 😆

dnil:
I'd call it a great flexibility of mind! One should not needlessly get stuck on one of many almost equally good solutions.. 😉

PR #2079, opened 2020-09-08
Move outdated panel info from tooltip to popover

northwestwitch:
While testing #2077 on stage using older cases, I've realized that there are so many differences in genes that the info no longer fits in a tooltip, so a popover looks like a better solution.

**How to test**:
1. On stage, master branch, go to a case runned with an older panel, for instance: https://scout-stage.scilifelab.se/cust000/WNB127N
1. Notice that the gene panel differences from the latest panels are a lot:

image

1. Install this branch and see that the visualization of the message improves

image

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
@elevu feel free to push improvements directly to the branch!

elevu (approved):

PR #2080, opened 2020-09-08
Scout release 4.21

northwestwitch:
A minor release because includes new features as well as bugfixes

### Added
- Support to configure LoqusDB per institute
- Highlight causative variants in the variants list
- Add tests. Mostly regarding building internal datatypes.
- Remove leading and trailing whitespaces from panel_name and display_name when panel is created
- Mark MANE transcript in list of transcripts in "Transcript overview" on variant page
- Show default panel name in case sidebar
- Previous buttons for variants pagination
- Adds a gh action that checks that the changelog is updated
- Adds a gh action that deploys new releases automatically to pypi
- Warn users if case default panels are outdated
- Define institute-specific gene panels for filtering in institute settings
- Use institute-specific gene panels in variants filtering
- Show somatic VAF for pinned and causative variants on case page

### Fixed
- Report pages redirect to login instead of crashing when session expires
- Variants filter loading in cancer variants page
- User, Causative and Cases tables not scaling to full page
- Improved docs for an initial production setup
- Compatibility with latest version of Black
- Fixed tests for Click>7
- Clinical filter required an extra click to Filter to return variants
- Restore pagination and shrink badges in the variants page tables
- Removing a user from the command line now inactivates the case only if user is last assignee and case is active
- Bugfix, LoqusDB per institute feature crashed when institute id was empty string
- Bugfix, LoqusDB calls where missing case count
- filter removal and upload for filters deleted from another page/other user
- Visualize outdated gene panels info in a popover instead of a tooltip in case page side panel

### Changed
- Highlight color on normal STRs in the variants table from green to blue
- Display breakpoints coordinates in verification emails only for structural variants

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil (approved):

dnil:
Here's something related to the latest panel PR. This is likely different on local, and due to the old panels on stage, but presumably it will look much the same on prod?

Screenshot 2020-09-08 at 13 13 33

northwestwitch:
I noticed that oldish panels contain panel_name values that look like mongodb Object IDs (hexadecimal strings like the one you found). Then something must have changed in how these names are saved, and the recent ones are those chosen by the users. It the one in the test a very old one?

dnil:
Probably - fairly randomly chosen from the list of lists. Let's roll with it for now and rather clean up in the db if need be?

northwestwitch:
It's from 2018. And the one you see is the real panel name. So it's not really a bug, the panel is named like that.

image

northwestwitch:
Ok then, merging in 3, 2, 1..

PR #2085, opened 2020-09-09
Restore HPO panel use in clinical filtering for SNVs and cancer SNVs

### LINKED ISSUE #2082: Bug "Use HPO list for clinical filter "??
Using the function "Use HPO list for clinical filter " does not seem to work - I get all variants from default panel.....

Not good   :=(

northwestwitch:
fix #2082
The dynamic gene list generated by the selected HPO terms in a case is not currently recognized as the list of genes to be used by clinical filter, even if it is marked so:

image

This PR fixes this behavior and introduces the same behavior in cancer SNVs filtering, that currently don't support this feature either

**How to test**:
1. On stage, master branch, assign one of more HPO phenotypes to a case, then click "HPO panel":

image

1. Then select "Use HPO list for clinical filter" option at the bottom of that panel, and click the `Clinical HPO snv variants button`:
image

1. The variants page, filters field named "HGNC symbols", should be empty and the variants search should return variants from the case default gene panels:

image

1. Switch to this branch and repeat the search from the case page. Notice that:
 - "HGNC symbols" field will be populated with gene names
 - Variants search returns variants in those genes

1. The functionality should work also for cancer SNVs

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/views.py
+++ b/scout/server/blueprints/variants/views.py
@@ -299,9 +299,7 @@ def cancer_variants(institute_id, case_name):
         form = CancerFiltersForm(request.args)
         # set chromosome to all chromosomes
         form.chrom.data = request.args.get("chrom", "")
-        if request.args.get("gene_panels"):
-            form.gene_panels.data = [request.args.get("gene_panels")]
-        else:
+        if form.gene_panels.data == []:
             form.gene_panels.data = [
                 panel["panel_name"]
                 for panel in case_obj.get("panels", [])
--- a/scout/server/blueprints/variants/views.py
+++ b/scout/server/blueprints/variants/views.py
@@ -56,6 +56,7 @@ def variants(institute_id, case_name):
 
     if request.form.get("hpo_clinical_filter"):
         case_obj["hpo_clinical_filter"] = True
+        form.gene_panels.data = ["hpo"]
 
     user_obj = store.user(current_user.email)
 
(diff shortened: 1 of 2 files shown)


northwestwitch:
This is ready for review now!

scout/server/blueprints/variants/views.py
```diff
@@ -69,11 +69,13 @@ def variants(institute_id, case_name):
         form.variant_type.data = variant_type
         # set chromosome to all chromosomes
         form.chrom.data = request.args.get("chrom", "")
-        form.gene_panels.data = [
-            panel["panel_name"]
-            for panel in case_obj.get("panels", [])
-            if panel["is_default"] is True
-        ]
+
+        if form.gene_panels.data == []:
```
dnil, line 73: It somehow doesn't strike me as standard python, but it does away with the length and is not operators I suppose. I will not fail working code on style, especially if the linters and what not give it a pass!

scout/server/blueprints/variants/views.py
```diff
@@ -297,11 +299,12 @@ def cancer_variants(institute_id, case_name):
         form = CancerFiltersForm(request.args)
         # set chromosome to all chromosomes
         form.chrom.data = request.args.get("chrom", "")
-        form.gene_panels.data = [
-            panel["panel_name"]
-            for panel in case_obj.get("panels", [])
-            if panel["is_default"] is True
-        ]
+        if form.gene_panels.data == []:
```
dnil, line 302: Well found! :+1:

dnil (approved):

PR #2088, opened 2020-09-10
Fix coverage report PDF download

### LINKED ISSUE #2086: Link to pdf of coverage report not working
It is not possible to download pdf of coverage reports. The link open some strange login site, see screen dump. I have tried both CMMS case 20132 and 20145.

This need to be fixed as soon as possible since we can not report our results without the coverage reports.

bild

northwestwitch:
fix #2086 

This is a fix to the problem of the PDF coverage report redirecting to the login page. This problem is affecting 2 pages:
- Coverage report, HTML page, when you click on the "PDF" button
- PDF version of the general report. The last section of the report, which is containing the coverage data.

Perhaps there is a better solution to fix this problem, but I haven't found it. Any other suggestion is highly appreciated!

**How to test**:
1. Test with one case on stage, for instance this: https://scout-stage.scilifelab.se/cust002/F0015031
1. Make sure that with master branch the 2 errors described above show up.
1. Install the branch and make sure that the 2 errors don't occur any more

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -79,6 +79,7 @@ def create_app(config_file=None, config=None):
             relevant_endpoint = not (static_endpoint or public_endpoint)
             # if endpoint requires auth, check if user is authenticated
             if relevant_endpoint and not current_user.is_authenticated:
+                LOG.error("HERE")
                 # combine visited URL (convert byte string query string to unicode!)
                 next_url = "{}?{}".format(request.path, request.query_string.decode())
                 login_url = url_for("public.index", next=next_url)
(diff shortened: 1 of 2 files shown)


northwestwitch:
This is also ready for review and open for discussions as well

scout/server/app.py
```diff
@@ -74,7 +74,7 @@ def create_app(config_file=None, config=None):
     def check_user():
         if not app.config.get("LOGIN_DISABLED") and request.endpoint:
             # check if the endpoint requires authentication
-            static_endpoint = "static" in request.endpoint
+            static_endpoint = "static" in request.endpoint or request.endpoint == "report.report"
```
dnil, line 77: 🤣
Sure, why not! Then we fix chanjo-report in due time? I'm in if you also file an issue there!

dnil (approved):

scout/server/app.py
northwestwitch: > Sure, why not!

It's actually different from before. Before it was `or "report" in request.endpoint`

PR #2093, opened 2020-09-11
Omim disease missing

dnil:
This PR fixes a bug.

If variant-gene objects were missing disease terms, these were not shown on the variants page tooltip.

**How to test**:
1. Try hovering over the gene names over a few genes with associated OMIM disorders. In all likelihood you will see no OMIM disorders on current master.
2. Do the same on current branch, and note that you now get the data fresh from hgnc_genes.

Screenshot 2020-09-11 at 17 07 15

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
So this is now fixed and better, but I do worry it will not be quite future proof. The design with freezing cases to the state everything was in on load (or first view as in this case) has its advantages, but we are slipping into showing latest information and having users expect seeing the latest updates to external information even on older cases.
 
We should have a little chat with @moonso and other interested about a design that can cater for both the archive and stability, as well as concurrency and continuity between different views.

scout/parse/omim.py
```diff
@@ -387,70 +387,3 @@ def get_mim_phenotypes(genemap_lines):
                 phenotypes_found[mim_nr] = phenotype
 
     return phenotypes_found
-
-
-@click.command()
```
northwestwitch, line 392: 👍

northwestwitch (approved):
Neat! Works perfectly:

image

northwestwitch:
> we are slipping into showing latest information and having users expect seeing the latest updates to external information even on older cases.

Should be this way. Whenever we have more info, it should be available even on older cases. Of course populating these variables on the fly on variants loading might slow down the pages. Then we should try to write smarter code. 
I suggest we make a list of the functionalities that need maintenance (refactoring for better performance) and whenever we have time we fix them.

dnil:
Agree to that! I also still worry that it is sometimes complex to describe what information is available at what point, and what its source is - but one can argue that that starts as soon as the mixing of vcf-info and later sources, so perhaps it does just fall on maintenance to keep it as clear as possible.

moonso:
I agree on that it is much better to populate this information on the fly to get updated information. However I didn't dare to do that since the loading of pages was already slow. You do whatever you feel like of course! Perhaps this kan be fixed with a new index on the phenotypes collection.

On a side note when discussing performance, we tried to look at the summary page under "Dashboard" but when choosing "All institutes" we could not get any information since it was too slow. Perhaps something to look into.

dnil:
The variants page is fairly ok again, and this will only trigger ever so often. It is not really a fetch-new-always, rather populate with new first time variants are viewed. Feel free to submit an issue.. 😉

PR #2096, opened 2020-09-14
Edit case and variant's comments

### LINKED ISSUE #1918: Make it possible to edit one's own comments in Scout
It would be great if it could be made possible to edit your own comments in Scout as it would save a lot of time, at least for me. Today the only way to correct for errors (typos etc) is to copy paste the old comment, correct and delete the old one.  What do you think - can this be done?

northwestwitch:
Fix #1918.

This PR makes it possible for users to edit their own comments. Editable comments:
- **Case comments**: comment text is editable
- **Variant comments** : comment text and comment's level (global -> specific and specific -> global) is editable

**How to test**:
1. Install on stage
1. Go to a case with comments from some other user
1. Make sure you can't edit those comments from other users
1. Insert a comment for a case and then change it
1. Insert a comment for a variant and then change it (modify also the comment level)

**Expected outcome**:

**Review:**
- [x] code approved by EM
- [x] tests executed by EM


northwestwitch:
Ready for review

elevu:
I can take this review!

northwestwitch:
> I can take this review!

Thanks! let me know if you don't find the comments!

elevu (approved):

northwestwitch:
Before merging this one I have a question. Do you think that it is sufficient to say that the comment is edited (because there is an updated_at field in the event), or should we register another event for the user: namely that the user has updated a comment?

elevu:
Good point, I don't think it it essential but it would be good to have the "comment edited" event in the Activity panel, to track changes more accurately.

northwestwitch:
Ok, I'll add the variant edited event then!

northwestwitch:
Inserted comment update event for case:

image

The updated variants update event is not showing in the variant's activity panel because of a bug I think, see #2097

northwestwitch:
Checked the variants page after fixing the empty activity panel bug. Works! :

image

Works also from the list of variants on variants page:

image

northwestwitch:
@elevu if the latest fix is OK for you we could perhaps merge this one as well

scout/server/templates/utils.html
```diff
@@ -51,17 +51,19 @@
                     {% if comment.level == 'global' %}
                       <span class="badge badge-info">GLOBAL</span>
                     {% endif %}
-                    {% if comment.created_at < case.updated_at %}
-                      <span class="badge badge-warning">OLD</span>
+                    {% if comment.created_at != case.updated_at %}
```
dnil, line 54: Ehrm; do you really want to remove the OLD label? And should everything that is not created at update be edited?! Didn't you just invent an edited event that could be checked?

northwestwitch: > do you really want to remove the OLD label?

Yes, why not? If a comment is updated at a different time than when it's created I think it's fine to say that it was edited.

Keep in mind that the `comment.created_at < case.updated_at` is not working and the OLD label is never shown in current master..

> Didn't you just invent an edited event that could be checked?

Yes, but the edited event is not a comment, so it's not available here in this macro.

dnil: > Yes, why not? If a comment is updated at a different time than when it's created I think it's fine to say that it was edited.

I'm probably missing something here, but isn't at a different time than when *the case* was updated something else? It looks like a copy-paste issue to me? Didn't you mean `comment.updated_at` or such?

> Keep in mind that the comment.created_at < case.updated_at is not working and the OLD label is never shown in current master..

Eh, what is wrong with the OLD label now? It worked like last week?

northwestwitch: No you are right. I've read  {% if comment.created_at != comment.updated_at %}
but it's case.updated_at, of course it didn't work. I'll fix it. Thanks!

northwestwitch: But why do we have the OLD badge, it is referred to the case or the comment then?

northwestwitch: I'm a bit confused here..

dnil: My understanding is that the OLD label should accompany entries that were made by some user when the case looked different, and so indicate that that user may not have seen the data then that you are looking at now.

dnil: As such, if the comment is older than the latest case update, that kind of makes sense to me?

dnil: We did have some discussion about this regarding minor data changes that still updated the case updated_at, such as delivery report.

northwestwitch: Ok but then if you agree I would leave the condition:
```
{% if comment.created_at < case.updated_at %}
    <span class="badge badge-warning">OLD</span>
{% endif %}
```

and I'd add another condition immediately after:

```
{% if comment.created_at < comment.updated_at %}
    <span class="badge badge-secondary">edited</span>
{% endif %}
```

dnil: Sounds great!

scout/server/templates/utils.html
```diff
@@ -16,10 +16,13 @@
...
-                  {% endif %}
-      	    {% if comment.created_at < case.updated_at %}
-                    <span class='badge badge-warning'>OLD</span>
-                  {% endif %}
+            {% endif %}
+            {% if comment.created_at < case.updated_at %}
+              <span class="badge badge-warning">OLD</span>
+            {% endif %}
+      	    {% if comment.created_at < comment.updated_at %}
+              <span class='badge badge-secondary'>edited</span>
```
dnil, line 24: 💯

northwestwitch:
I'm going to merge as soon as the tests pass. Stop me now or never..

dnil:
I'm very much looking forward to be able to edit variants! No stopping on my account! 😄

PR #2099, opened 2020-09-15
Transform event cursor to list to avoid having an empty variants event panel

### LINKED ISSUE #2097: Variant's activity not updating?
I've noticed that the variant activity panel (at the bottom of a variant's page) doesn't get populated when the user triggers events (ACMG classification, comments, pinned..) A bug perhaps?

northwestwitch:
fix #2097 --> Empty activity panel at the end of  a variant page
The bug is caused by the events being a cursor that becomes empty after a loop in variants controller. Looping over the events is possible if the cursor is transformed into a list.

**How to test**:
1. On a local branch, or stage go to a variant and do something on the page, for instance: comment, tag it, pin it, etc.
1. Note that no activity shows on master branch
1. Install this branch and you should see that activity panel is populated

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -116,7 +116,7 @@ def variant(
         variant_case(store, case_obj, variant_obj)
 
     # Collect all the events for the variant
-    events = store.events(institute_obj, case=case_obj, variant_id=variant_id)
+    events = list(store.events(institute_obj, case=case_obj, variant_id=variant_id))
     for event in events:
         event["verb"] = VERBS_MAP[event["verb"]]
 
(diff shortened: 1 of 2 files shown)


northwestwitch:
Before:

image


And after:

image

northwestwitch:
Ready for review!

scout/server/blueprints/variant/controllers.py
```diff
@@ -116,7 +116,7 @@ def variant(
         variant_case(store, case_obj, variant_obj)
 
     # Collect all the events for the variant
-    events = store.events(institute_obj, case=case_obj, variant_id=variant_id)
+    events = list(store.events(institute_obj, case=case_obj, variant_id=variant_id))
```
dnil, line 119: Yes, absolutely! 👍

dnil (approved):

PR #2102, opened 2020-09-17
*DRAFT* Feature pymongo3.11 200915

mikaell:
This PR prepares Scout for moving from PyMongo 3.7 to 3.11

**How to test**:

Edit requirements.txt: `pymongo==3.11`

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

*This is a draft. Please have a look and comment.*


scout/adapter/mongo/hgnc.py
```diff
@@ -113,7 +113,7 @@ def hgnc_genes(self, hgnc_symbol, build="37", search=False):
             search(bool): if partial searching should be used
 
         Returns:
-            result()
+            pymongo.cursor
```
moonso, line 116: 👍 Nice

scout/load/setup.py
```diff
@@ -65,7 +65,7 @@ def setup_scout(
 
     LOG.info("Check if there was a database, delete if existing")
     existing_database = False
-    for collection_name in adapter.db.collection_names():
+    for collection_name in adapter.db.list_collection_names():
```
moonso, line 68: Why not use the `adapter.collections` function that you changed above? Next time we would only need to update in one place :)

scout/server/blueprints/institutes/controllers.py
```diff
@@ -279,18 +279,16 @@ def get_sanger_unevaluated(store, institute_id, user_id):
     return unevaluated
 
 
-def gene_variants(store, variants_query, institute_id, page=1, per_page=50):
+def gene_variants(store, pymongo_cursor, institute_id, page=1, per_page=50):
```
moonso, line 282: I would prefer type hints so we get both information that it is a cursor and that it includes variants, so something like:
```suggestion
def gene_variants(store: Adapter, variants: pymongo.Cursor, institute_id: str, page: int = 1, per_page: int = 50):
```

scout/server/blueprints/institutes/controllers.py
```diff
@@ -279,18 +279,16 @@ def get_sanger_unevaluated(store, institute_id, user_id):
...
     """Pre-process list of variants."""
 
-    # We need to call variants_collection.count_documents here
-    variant_count = variants_query.count()
+    variant_count = count_cursor(pymongo_cursor)
```
moonso, line 285: Clever solution! Hope you have checked that the cursor can be traversed multiple times, just so it does not work as a generator which gets exhausted after one loop-through

scout/server/blueprints/variants/controllers.py
```diff
@@ -496,24 +467,26 @@ def cancer_variants(store, institute_id, case_name, variants_query, form, page=1
...
     # Setup variant count session with variant count by category
     variant_count_session(store, institute_id, case_obj["_id"], "clinical", "cancer")
     session["filtered_variants"] = variant_count
-
-    more_variants = True if variant_count > (skip_count + per_page) else False
     variant_res = variants_query.skip(skip_count).limit(per_page)
+
+    variants_list = []
+
+    for variant in variant_res:
```
moonso, line 480: 👍 😄

scout/server/utils.py
```diff
@@ -190,3 +190,11 @@ def find_index(align_file):
         if not os.path.exists(index_file):
             index_file = "{}.bai".format(align_file)
     return index_file
+
+
+def count_cursor(cursor):
+    """Count objects at cursor."""
+    # Replaces depricated call to Pymongo.cursor.count()
+    # Clone cursor to not interfer with other queries
+    new_cursor = cursor.clone()  # Clone cursor, otherwise original will move
```
moonso, line 199: Aha you fixed it like that, nice!

moonso (commented):
Looks like a great start! Should be fairly straight forward to deploy and test clicking around on stage

dnil:
Great stuff - will feel very good to be able to lift that freeze! Just hit "Ready for review" - and perhaps edit away the "*DRAFT*" otherwise its easy to overlook. Did it end up fast enough for the big operations with countDocuments rather than estimatedDocumentCount?

mikaell:
> Great stuff - will feel very good to be able to lift that freeze! Just hit "Ready for review" - and perhaps edit away the "_DRAFT_" otherwise its easy to overlook. Did it end up fast enough for the big operations with countDocuments rather than estimatedDocumentCount?

Ran into discrepancies while testing. Not ready for merge yet. Will resume this work next week.

mikaell:
Bug fixed. Will test and create pull request.

scout/server/blueprints/institutes/controllers.py
mikaell: This would be a separate refactoring pull request.

mikaell:
The naive implementation does not perform well on large collections. Investigating further in the depths of Mongodb and Pymongo.

mikaell:
|Method | Result | Time|
| --------| ------ | ------ |
|incorrect replacement method estimated() | 2172 | 0.00059|
|depricated count() | 680 |                                           0.00291|
|hack cursor object() | 680 |                                         0.00369|
|convert to list, then len | 680 |                                    0.09869|
|iterate pointer | 680 |                                                   0.10848|


Timed a simple query with different methods. Only run once. Results may vary a bit. Have not tested memory usage.

mikaell:
I made a complete rewrite of the migration from 3.7 to 3.11. 

It should be read for review as soon as the tests are passed.

mikaell:
Soon ready for review!

I removed DRAFT status -mainly because tests weren't running- do execute all checks on drafts?

mikaell:
PR moved to **_#2152_**

Issue #2103, opened 2020-09-17
KeyError in event["verb"] = VERBS_MAP[event["verb"]]


dnil:
When we introduce new events in feature branches, the stage database becomes populated with events that will crash a trunk version. I suggest we catch the exception and possibly make a visible note of the event rather than crash. This would also work to more quickly stabilise a situation where we would have to revert a version introducing a new event in production.

northwestwitch:
Good suggestion!

dnil:
Closed by #2109

PR #2109, opened 2020-09-17
Fix #2072 - small group assessment sharing

### LINKED ISSUE #2072: Persistent information between reruns
In case of re-run, it is either customer request, resequencing, bioinformatic workflow update, or just good old re-upload of data.

To begin with, I think it'd be useful to preserve some information from previous upload instead of completely wiping it. Here is the list of wishlist I can think about:

1. Are pinned/causative variants still there? Do they still have the same:
   1. AF
   1. AD/DP
   1. GT/GQ

2. What is the aggregated difference between past case and this one? i.e. compare previous run to this run:
   1. Total VAF
   1. Total DP or AD
   1. Tally of pinned/causative variants. Are they same with regards to point 1 above.
   1. Difference in difference of values for VAF, DP, AD, or some other metrics.

(I'll more to this if I can think about more)

Overall I think it'll be a great value to do this. I'm planning this in accordance with Vogue package, where validating each of our TGA or WES runs for cancer samples.

This is probably not so fun for WGS run and RD in general, there are too many variants, score, etc to take care of. But maybe it is more fun :-)

Yay or nay?

dnil:
This PR adds a functionality: shadow manual assessments on variants view from a small case group, intended as a simple workaround e.g. for relapses, additional remote family members, exome+genome or possibly genome + RNA - or indeed small cohorts with similar phenotype

**How to test**:
1. on local, load another test case with different internal id and perhaps for simplicity display id (643595)
1. add a group tag to this case using the case group card under the case pedigree and link to the mirror case (643594)
Screenshot 2021-01-19 at 10 50 59
1. make a few different manual evaluations on the original case and make sure these are now visible on the new case:
Screenshot 2021-01-19 at 10 52 43
1. and vice versa
Screenshot 2021-01-19 at 10 53 28

For now, this is manual to try out the potential, but it could also be extended to on load via config and via CLI. Good ways to do this may include Orderportalen, CG and/or via Balsamic and MIP. Basically anytime the case group relation is made clear and it is obvious the user wants them linked.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Is this one ready for review?

dnil:
> Is this one ready for review?

Feedback is always welcome, but I think it also needs a way to mark cases as grouped first. Otw its db admin privilege only..

scout/constants/case_tags.py
```diff
@@ -57,12 +57,11 @@
     "update_clinical_filter_hpo": "updated clinical filter HPO status for",
     "mme_add": "Exported to MatchMaker patient",
     "mme_remove": "Removed from MatchMaker patient",
-    "filter_stash": "stored a filter for ",
+    "filter_stash": "stored a filter for",
     "update_sample": "updated sample data for",
+    "update_case_group_ids": "updated case group ids for",
 }
 
-VERBS = list(VERBS_MAP.keys())
```
dnil, line 64: Seems unused.

scout/server/blueprints/cases/controllers.py
```diff
@@ -177,7 +182,7 @@ def case(store, institute_obj, case_obj):
 
     events = list(store.events(institute_obj, case=case_obj))
     for event in events:
-        event["verb"] = VERBS_MAP[event["verb"]]
+        event["verb"] = VERBS_MAP.get(event["verb"], "did %s for".format(event["verb"]))
```
dnil, line 185: Fixes #2103 - and avoids later crashes when updating the event verb set.

dnil:
Almost there now.. 
Screenshot 2021-01-18 at 16 36 14

It needs tests and an h or two of typo debug before review. Case name auto completion for the case add modal would not be bad either. Mañana, mañana!

dnil:
Ready for review! Thanks!

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -217,7 +218,7 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
 
 {% macro synopsis_panel() %}
   <div class="card panel-default">
-    <div data-toggle='tooltip' class="panel-heading" title="Free text field. Write a summary about the case! Markdown format">Synopsis</div>
+    <div data-toggle='tooltip' class="panel-heading" title="Free text field. Write a summary about the case! Markdown format"><i class="fa fa-id-card-o"></i>&nbsp;Synopsis</div>
```
northwestwitch, line 221: Nice!

scout/server/blueprints/cases/templates/cases/phenotype.html
```diff
@@ -2,7 +2,7 @@
 
 {% macro cohort_panel(case, institute, cohort_tags) %}
   <div class="card panel-default">
-    <div class="panel-heading">Cohort tag</div>
+    <div class="panel-heading"><span class="fa fa-tag"></span>&nbsp;Cohort tag</div>
```
northwestwitch, line 5: 👍

scout/server/blueprints/cases/templates/cases/phenotype.html
```diff
@@ -120,7 +120,7 @@
 
 {% macro phenotype_groups_panel(case, institute, hpo_groups) %}
   <div class="card panel-default">
-    <div class="panel-heading">Phenotype groups</div>
+    <div class="panel-heading"><span class="fa fa-stethoscope"></span>&nbsp;Phenotype groups</div>
```
northwestwitch, line 123: 👍

scout/server/blueprints/cases/templates/cases/phenotype.html
```diff
@@ -156,7 +156,7 @@
 {% macro phenotype_terms_panel(case, institute) %}
   {% set url = 'http://hpo.jax.org/app/browse/term/HP:0000018' %}
   <div class="card panel-default">
-    <div class="panel-heading">Phenotype terms (<a target="_blank" href="{{ url }}" class="text-white">HPO web</a>)</div>
+    <div class="panel-heading"><span class="fa fa-stethoscope"></span>&nbsp;Phenotype terms (<a target="_blank" href="{{ url }}" class="text-white">HPO web</a>)</div>
```
northwestwitch, line 159: 👍

scout/server/blueprints/cases/templates/cases/phenotype.html
```diff
@@ -192,7 +192,7 @@
 {% macro phenotypes_panel(case, institute, config) %}
   <form action="{{ url_for('cases.phenotypes_actions', institute_id=institute._id, case_name=case.display_name)+'#phenotypes_panel' }}" method="POST">
     <div id="phenotypes_panel" class="card panel-default">
-      <div class="panel-heading">Added phenotypes</div>
+      <div class="panel-heading"><span class="fa fa-stethoscope"></span>&nbsp;Added phenotypes</div>
```
northwestwitch, line 195: 👍

scout/server/blueprints/cases/views.py
```diff
@@ -94,7 +94,7 @@ def matchmaker_matches(institute_id, case_name):
     # check that only authorized users can access MME patients matches
     panel = 1
     if request.method == "POST":
-        panel = panel = request.form.get("pane_id")
```
northwestwitch, line 97: Right!

scout/server/blueprints/variant/controllers.py
```diff
@@ -140,7 +140,7 @@ def variant(
     # Collect all the events for the variant
     events = list(store.events(institute_obj, case=case_obj, variant_id=variant_id))
     for event in events:
-        event["verb"] = VERBS_MAP[event["verb"]]
+        event["verb"] = VERBS_MAP.get(event["verb"], "did %s for".format(event["verb"]))
```
northwestwitch, line 143: 🥇

scout/server/blueprints/variants/controllers.py
```diff
@@ -84,6 +84,25 @@ def variants(store, institute_obj, case_obj, variants_query, variant_count, page
             )
 
         variant_obj["clinical_assessments"] = get_manual_assessments(clinical_var_obj)
+
+        if case_obj.get("group"):
```
northwestwitch, line 88: I got a bit lost here. What is this code doing? Could we add a comment?

scout/server/blueprints/variants/controllers.py
```diff
@@ -84,6 +84,25 @@ def variants(store, institute_obj, case_obj, variants_query, variant_count, page
...
+            variant_obj["group_assessments"] = []
+
+            group_case_ids = set()
+            for group_id in case_obj.get("group"):
+                group_case_ids.update(store.case_group_ids(group_id))
+
+            for group_case_id in group_case_ids:
```
northwestwitch, line 95: A comment here as well would be nice to have

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -229,6 +230,67 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
+      <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="case_pick_dialog">Add case to existing case group - case {{case.display_name}}</h5>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+        </div>
+        <div class="modal-body">
+          <input type="text" name="other_case_name" class="typeahead_cases form-control" data-provide="typeahead" autocomplete="off" placeholder="Search..." cols="30" rows="10">Case name</input>
```
northwestwitch, line 283: Perhaps the placeholder for the input could be "search case name", so you can remove Case name under it

image

northwestwitch (changes requested):
Nice work, this is going to be a super appreciated feature!

Overall the feature works fine and it's amazing that you fixed also the recurrent crash whenever we introduce new action-related verbs that didn't exist in the database before.

I have comments for several few fixes but I think the interface to create and assign cases to groups could be improved. Here are some suggestions:

- Wouldn't it be easiest (for the users) if we allow to define groups with custom names (I'm sure that someone is going to ask for this feature in the future) and then assign cases to the groups?
- The 2 buttons thing is a bit confusing to me, especially if you can create a lot of groups with the same and only case in them
- It feels like you don't have 100% control on the groups using the interface. For example I've tested using the 3 demo samples: cancer_case, 643595 and 643594. If you create multiple groups with only one case (for instance cancer case).

image

Let's say we want to add 643595 in the same group of cancer_case: and we do it from the page of case: 643595.

This creates yet another group on cancer_case:

image

What if I want to add 643594 to the same group of cancer_case and 643595 and I do it from the case page of 643594? It creates yet another group in cancer_case:

image

So you have to manually go to cancer_case and remove the previous group. I think all this would be fixed by creating custom group names, but perhaps I'm missing something?

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -229,6 +230,67 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
+                <span class="pull-right">
+                  <a href="{{ url_for('cases.remove_case_group', institute_id=institute._id, case_name=grouped_case.display_name, case_group=group_id) }}" class="btn btn-link btn-sm">
+                    <span class="fa fa-remove"></span></a>
+                </span>
+            </li>
+          {% endfor %}
+          </ul>
+        {% endfor %}
+      {% endif %}
+      <a href="{{ url_for('cases.add_case_group',institute_id=institute._id, case_name=case.display_name) }}" class="btn btn-primary btn-xs text-white"><span class="fa fa-user-plus text-white mr-3"></span>Add group</a>
```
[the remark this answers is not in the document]
dnil: > Why there are 2 distinct buttons to add a case to a group? If you click several times on the one on the bottom you find yourself in this situation, which is a bit weird

Thanks a bunch for the review! Super to see what is not obvious to others.

The underlying reason is that a case should be able to join multiple groups, to allow for different investigations, like one with a related family, one with cancer remission sample, one with another tumor or sth. Also, groups don't get made on load (yet?) so they have to be made through the interface at some point.

There are three kinds of buttons, and in my simple mind they form a nice minimum, though perhaps we can come up with better names for them:
- "Add group" adds a new group (unnamed, can add names to see if that helps make it more clear) - and it is assumed the current case is always going to be in that new group, so it is automatically bound into the group. Groups are after all only shown on the case pages if the case belongs to the group. Perhaps it could be named "Create group"?
- "Add to group" adds other cases to the selected group. We agree on this one. 😸 Perhaps it could be named "Add other case" or such?
- "X" on group list item cases removes the case from the pertinent group. If no cases are left in the group, the group is deleted. If there are still other cases in the group, but present case A is removed, it just leaves the group, and the group is no longer visible on the case page of A. It will still be visible from the other group pages.

dnil:
> * Wouldn't it be easiest (for the users) if we allow to define groups with custom names (I'm sure that someone is going to ask for this feature in the future) and then assign cases to the groups?

Sure, I'll add a name to them - the ObjectId are good as id but don't do very well on display.. 👍 Experience from filters said people are not super eager to name things themselves, but the option doesn't hurt here.

> * The 2 buttons thing is a bit confusing to me, especially if you can create a lot of groups with the same and only case in them

😆 Yeah, you are not supposed to do that. It doesn't help much.

> * It feels like you don't have 100% control on the groups using the interface. For example I've tested using the 3 demo samples: cancer_case, 643595 and 643594. If you create multiple groups with only one case (for instance cancer case).

> Let's say we want to add 643595 in the same group of cancer_case: and we do it from the page of case: 643595.

Same, same really? I wanted to avoid adding a search-for-groups feature, to keep them a bit more lightweight. Maybe it is unavoidable in the end, but I bet most users will have a clear set of cases they want to group, like from a referral or the hospital LIMS, and can just add them from any case. 
 
> What if I want to add 643594 to the same group of cancer_case and 643595 and I do it from the case page of 643594? It creates yet another group in cancer_case:

To me, it would be fine with either just making this new group, and perhaps then rather skip making the previous one with only two cases, or adding the straggling case 643594 from either of the two other case pages that already show the group.

> So you have to manually go to cancer_case and remove the previous group. I think all this would be fixed by creating custom group names, but perhaps I'm missing something?

I was afraid it would get more complicated if I add a "Bind to existing group" button, but with searchable names one would not have to display a huge list of all possible groups. Let me start with renaming buttons, and add group names and we see.

scout/server/blueprints/cases/templates/cases/case.html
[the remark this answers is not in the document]
northwestwitch: I did understand that one case might belong to different groups, but to me it is a bit confusing that if you want to assign a case (case x) to a group with other 2 cases then you should do it from the case page of the other 2 cases, because if you do it from the page of case x it's only assigned to one case, in a new group, and not where you want.. But probably this will be more clear if one assigns a name to the groups..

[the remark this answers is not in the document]
dnil: Well, you can do it from that third case page, but you then need to mention both the two other previously added cases that you wanted in that group. More action economy would be gained by adding the single case from either of the two already in a group. Fair enough to me, and nothing that can't be fixed, but let's see what happens with just a rename.

scout/server/blueprints/variants/controllers.py
dnil: Good to know - then we need a new function and a docstring! See if the new version reads easier?

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -229,6 +230,67 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
+{% endmacro %}
+
+{% macro modal_add_to_group(group_id) %}
+  <form action="{{ url_for('cases.add_case_group', institute_id=institute._id, case_name="unused") }}" method="POST">
+    <div class="modal fade" id="add-case-group-{{group_id}}" role="dialog">
+      <input type="hidden" name="group_id" value="{{ group_id }}">
+      <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="case_pick_dialog">Add case to existing case group - case {{case.display_name}}</h5>
```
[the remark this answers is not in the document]
dnil: Thanks for noticing! "The add case" heading is ambiguous and really distracts from the purpose. 
To make it make sense quicker, let me fix that dialog, showing what is already in the group!
Screenshot 2021-01-20 at 15 31 27

scout/server/blueprints/variants/controllers.py
northwestwitch: Yes, I like it!

scout/server/blueprints/cases/templates/cases/case.html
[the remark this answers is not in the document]
dnil: Ok, see what you think - I just added a placeholder name based off the ObjectId, much like git commits. We *can* make an editable name, but it will add some complexity. I feel just having some more separator beyond an icon and a button between the different groups might have done the trick for now. If you still think the name request is inevitable, I'll add it anyway.

Screenshot 2021-01-20 at 16 17 49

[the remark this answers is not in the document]
northwestwitch: Mmm I don't think displaying the _id of the groups is much of help here. I think it would be best not to have it if it can't be chosen by the user, sorry. I guess we can go with this system and fix it later if they want to specify a custom name for the group?

I thinking for instance at different types of analyses for the same case, or different sequenced tissues for a cancer case and so on. It would be useful to have a way to name these groups, but perhaps I'm overthinking and it's enough to be able to assign them together!

northwestwitch:
> I was afraid it would get more complicated if I add a "Bind to existing group" button, but with searchable names one would not have to display a huge list of all possible groups. Let me start with renaming buttons, and add group names and we see.

This looks already better because doesn't have repeated "create another group" buttons and "bind other case" is more self-explanatory. If you make it possible to search for group name you don't have to display a huge list of groups, but then the group name should be custom.. We can introduce it later I guess?

dnil:
> > I was afraid it would get more complicated if I add a "Bind to existing group" button, but with searchable names one would not have to display a huge list of all possible groups. Let me start with renaming buttons, and add group names and we see.
> 
> This looks already better because doesn't have repeated "create another group" buttons and "bind other case" is more self-explanatory. If you make it possible to search for group name you don't have to display a huge list of groups, but then the group name should be custom.. We can introduce it later I guess?

And if the names should be searchable, they need a practical level of uniqueness - it wont help much if there are like 3000 "tumor_vs_first_remission" groups. And just adding the case name to the mix doesnt really help terribly, as cases can be added or removed from the group. I actually dont think they will ever be searchable by group name, but one could imagine searching for a case name, and getting hits also for two groups that that case is already in.

northwestwitch:
> > > I was afraid it would get more complicated if I add a "Bind to existing group" button, but with searchable names one would not have to display a huge list of all possible groups. Let me start with renaming buttons, and add group names and we see.
> > 
> > 
> > This looks already better because doesn't have repeated "create another group" buttons and "bind other case" is more self-explanatory. If you make it possible to search for group name you don't have to display a huge list of groups, but then the group name should be custom.. We can introduce it later I guess?
> 
> And if the names should be searchable, they need a practical level of uniqueness - it wont help much if there are like 3000 "tumor_vs_first_remission" groups. And just adding the case name to the mix doesnt really help terribly, as cases can be added or removed from the group. I actually dont think they will ever be searchable by group name, but one could imagine searching for a case name, and getting hits also for two groups that that case is already in.

As an unexperienced user I would be tempted to name a group with the name of the case, and then assign all different analyses of that case to the group, In that situation searching for a group name would make sense, But perhaps this is not the intended use of the feature?

northwestwitch (approved):
I think this is already working so I'll approve. what I would do before merging is removing group _id and perhaps try to separate a bit the groups in the card? The title of the group below is a bit more close to the <ul> element of the group above
image

dnil:
> As an unexperienced user I would be tempted to name a group with the name of the case, and then assign all different analyses of that case to the group, In that situation searching for a group name would make sense, But perhaps this is not the intended use of the feature?

Gotcha. Well, see it this way perhaps: the current case *is* already in the group where it was created, so no need to have it in the group name/label. Also, the same can be said for any other case which is added to the group. And we don't have free floating analyses in the system (yet?) - they all exist as cases, possibly with very similar names.
The thing is I think your group name in this example would end up being like F00010101_with_F00010102_and_F00010101-2. 

That said, I am quickly looking at labels now - will see how that pans out before merging anything.

dnil:
Ok, so there is a risk for it getting kind of bulky, especially for a small screen. And the group labels will need to be constrained in order not to take too much space at any rate?
Screenshot 2021-01-20 at 19 09 37
I dunno atm: I think I would just go with the previous one, and just stash the label-stuff for now? I'll eat, sleep and possibly play a game of SC2 before deciding. And feel free to weigh in!

northwestwitch:
> Ok, so there is a risk for it getting kind of bulky, especially for a small screen

Indeed. A way to fix this could be to change the list of cases into comma separated items, each with the delete button of course, but I don't know how messy it would look.

dnil:
Meh, since I already wrote a version of it, I'll finish the label thing off. This is what it looks like:
Screenshot 2021-01-21 at 13 08 31
Its code-wise a rather obvious add-on, rather than by design, and I didn't give it an event, but the functionality is there, and it can be more streamlined and eventified without any changes to the database. Lets see if coverage complains, since a couple of the label things will not get full test coverage, otherwise I may just merge it as is.

northwestwitch:
> Meh, since I already wrote a version of it, I'll finish the label thing off. This is what it looks like:
> Screenshot 2021-01-21 at 13 08 31
> Its code-wise a rather obvious add-on, rather than by design, and I didn't give it an event, but the functionality is there, and it can be more streamlined and eventified without any changes to the database. Lets see if coverage complains, since a couple of the label things will not get full test coverage, otherwise I may just merge it as is.

It looks very nice!

PR #2111, opened 2020-09-18
Move current igv tracks to constants and let users choose which IGV tracks to display

### LINKED ISSUE #1872: IGV track ClinVar CNV
We would like to have a discussion on the necessity to include ClinVar CNV track in the IGV view. All (majority) of users at Clin Genetics would like to have it removed. It does not provide any useful information. Can it be "collapsed" as default, or removed?
Discuss at user-meet.

northwestwitch:
fix #1872.

This PR does 4 things:
- Moves the IGV track info from alignviewers/controllers.py to the scout/constants (makes sense to me since they're all constants)
- Refactors the alignviewers.views so that most of its code is now in the controllers
- Introduces the possibility for users to customize which tracks they want to see from the tracks common to all cases (Genes, ClinVar and ClinVar CNV)
- Introduces tests for the alignviewers, which was largely untested before.

Note that **STRs variants page doesn't change** and the possibility to customize tracks is present only on variants and SV-variants page:
image

**How to test**:
1. Install the branch on stage and go to a recently uploaded case containing alignment files.
1. Make sure that alignments of variants/cancer variants/ SV variants and STRs still work.
1. Update your users settings to visualize custom igv tracks and make sure the right tracks show up in igv.js browser

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by DN


northwestwitch:
This is ready for review

dnil:
Hm, has at least some minor issue to fix (with builds?): naive testing on stage returns

```
-22 10:26:12,576 - scout.server.app - ERROR: Exception on /igv [POST] [in /home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py:1891]
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
   reraise(exc_type, exc_value, tb)
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
   raise value
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
   rv = self.dispatch_request()
 File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
   return self.view_functions[rule.endpoint](**req.view_args)
 File "/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/server/blueprints/alignviewers/views.py", line 72, in igv
   chromosome_build
KeyError: 'grch37'
```

dnil:
I doubt it will be hard to fix, but perhaps we leave this one off the current release not to push things too much?

northwestwitch:
> I doubt it will be hard to fix, but perhaps we leave this one off the current release not to push things too much?

Yes, let's leave it out for now!

northwestwitch:
> Hm, has at least some minor issue to fix (with builds?): naive testing on stage returns
> 
> ```
> -22 10:26:12,576 - scout.server.app - ERROR: Exception on /igv [POST] [in /home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py:1891]
> Traceback (most recent call last):
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
>    response = self.full_dispatch_request()
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
>    rv = self.handle_user_exception(e)
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
>    reraise(exc_type, exc_value, tb)
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
>    raise value
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
>    rv = self.dispatch_request()
>  File "/home/hiseq.clinical/miniconda/envs/prod-stage/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
>    return self.view_functions[rule.endpoint](**req.view_args)
>  File "/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/server/blueprints/alignviewers/views.py", line 72, in igv
>    chromosome_build
> KeyError: 'grch37'
> ```

Oh no, I forgot about the genome build naming inferno we have in real life! Will fix!

northwestwitch:
This should be fixed now. Tested on stage on a recent case (vitalmouse) and looks OK

dnil:
Again, great stuff! That now works just fine on stage. A couple of minor things, and a pre-codereview worry:
- hitting save from the variant page opens a new window; guessing redirect target got to be the same as the old alignviewer one. 
- settings could be with a capital S for conformity
- additional tracks from load conf, UPD, rhocall, global (TIDDIT) coverage are not on the dropdown. Could absolutely be another PR, I just worry they may be off.. Will check.

northwestwitch:
> * additional tracks from load conf, UPD, rhocall, global (TIDDIT) coverage are not on the dropdown. Could absolutely be another PR, I just worry they may be off.. Will check.

I have included in the dropdown only the items that are ALWAYS available for all cases. I was thinking that since those tracks are not in all cases (example cancer cases don't have them), and perhaps cases runned not at cg Solna, we could handle in a slightly different way. Since I didn't know which way I've left them out for a different PR.

northwestwitch:
But you are absolutely right, I think that in the code I wrote, if a user has track preferences and if they're not in the list of these preference, they won't show

northwestwitch:
So perhaps I should add them (showing in the select that the track depends on the case type)

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -1,3 +1,26 @@
+{% macro igv_track_selection(igv_tracks, current_user) %}
+<div id="accordion">
+  <a data-toggle="collapse" data-parent="#accordion" href="#track_settings">settings</a>
+  <div id="track_settings" class="collapse">
+    <form action="{{url_for('variant.update_tracks_settings')}}" target="_blank" method="POST">
```
dnil, line 5:
```suggestion
    <form action="{{url_for('variant.update_tracks_settings')}}" method="POST">
```

scout/server/blueprints/variant/views.py
```diff
@@ -21,14 +21,24 @@
...
 variant_bp = Blueprint("variant", __name__, static_folder="static", template_folder="templates")
 
 
+@variant_bp.route("/update_tracks", methods=["POST"])
+def update_tracks_settings():
+    """Update custom track settings for a user according to form choices"""
+    user_obj = store.user(email=current_user.email)
+    selected_tracks = request.form.getlist("user_tracks") or []
+    # update user in database with custom tracks info
+    user_obj["igv_tracks"] = selected_tracks
```
dnil, line 37: Perhaps add to User object model as well, and init with something sensible for new users?

northwestwitch:
No, never mind, they will show, because they're included in a function: `def set_case_specific_tracks():` that uploads these tracks independently from the tracks selectable by the user. So it should work

scout/server/blueprints/alignviewers/views.py
```diff
@@ -50,117 +51,39 @@ def unindexed_remote_static():
...
+    # General tracks (Genes, Clinvar and ClinVar SNVs are shown according to user preferences)
+    controllers.set_common_tracks(display_obj, chromosome_build, request.form)
 
+    # Set up bam/cram alignments for case samples:
+    controllers.set_sample_tracks(display_obj, request.form)
 
-def make_igv_tracks(name, file_list):
-    """ Return a dict according to IGV track format. """
+    # Set up case-specific tracks (might be present according to the pipeline)
+    if request.form.get("bam"):
```
dnil, line 84: Hm, will that really always be set? E.g. UPDsites or TIDDIT_coverage still left on file system with an old archived bam?

scout/constants/igv_tracks.py
```diff
@@ -0,0 +1,131 @@
...
+    "type": "annotation",
+    "sourceType": "file",
+    "displayMode": "EXPANDED",
+    "format": HG38GENES_FORMAT,
+    "url": HG38GENES_URL,
+    "indexURL": HG38GENES_INDEX_URL,
+}
+
+CASE_SPECIFIC_TRACKS = {
+    "rhocall_bed": "Rhocall Zygosity",
```
dnil, line 118: 👍

scout/server/blueprints/alignviewers/views.py
```diff
@@ -50,117 +51,39 @@ def unindexed_remote_static():
...
-    if request.form.get("align") == "mt_bam":
-        bam_files = request.form.get("mt_bam").split(",")
-        bai_files = request.form.get("mt_bai").split(",")
-    else:
-        if request.form.get("bam"):
-            bam_files = request.form.get("bam").split(",")
-            LOG.debug("loading the following BAM tracks: %s", bam_files)
-        if request.form.get("bai"):
-            bai_files = request.form.get("bai").split(",")
-        if request.form.get("rhocall_bed"):
```
dnil, line 86: But, let's see, this actually is not connected now, is it? While new and nice functions to add the tracks act if files are noted, no splitting of files from config occurs?

dnil (changes requested):
Looks very nice! Lets just make sure the UPD/RHO etc are not broken by this, and get automatically added if configured, then we can add manual config for them later in another PR. We should have some cases with from last testing bout.. Checking..

scout/server/blueprints/variant/views.py
northwestwitch: But if I add all available tracks to the user object when a user is created, then if we add one or more selectable tracks later, these will not be selected for those users, and perhaps they won't know that they even exist if they don't go into the settings and change them manually. Isn't it better if the key doesn't exist of is None so that all available tracks are always loaded if a user doesn't save any track preference?

dnil: Wouldn’t the same be true as soon as they save settings once, with the defaults being implicit in this function rather than on user init? How about making new tracks show, eg only hiding ones that have explicitly been marked hidden - not ones that didn’t exist before?

scout/server/blueprints/alignviewers/views.py
northwestwitch: Isn't this a problem we also have now in current master then? I don't think we ever reach this function (alignviewers) if there is no bam or MT_bam assigned for the case individual..

dnil: 😆 True enough! Makes one one wonder why the check is there?

scout/server/blueprints/variant/views.py
northwestwitch: Absolutely, but looks quite complicates to handle

scout/server/blueprints/alignviewers/views.py
northwestwitch: But you raised an important issue here, we should fix it for when there is no alignment file but other tracks are still available

northwestwitch: But in another PR perhaps?

northwestwitch: But this if might be perhaps removed in this PR, I'll fix it

northwestwitch: I think that code was like this originally to prepare those tracks only if the chromosome is not MT, you know? That's why I left the condition, now I remember. Is this the case? -- > do you want these tracks loaded also when igv is showing MT alignments?

dnil: A yes, right. I remember I used that as a workaround sometimes when looking at the other tracks locally, but not having a full bam, only an MT one. No, agreed, that is good. But perhaps make that clear so we don't loose it at a later stage!

scout/server/blueprints/alignviewers/views.py
northwestwitch: You are probably right. I wanted to update this repetitive code here but I couldn't find any case with these tracks to test. Did you find one btw?

scout/server/blueprints/alignviewers/views.py
northwestwitch: I'll change the condition to check if chrom != MT instead, so it's more clear

scout/server/blueprints/alignviewers/views.py
northwestwitch: This should be fixed now, but not tested

dnil: Was just looking.. We had them on stage vitalmouse for the pre-MIP9 test, but there was some issue with the track formats that caused them to look dotty rather than like bar plots, so we turned them off in cg for now. Vitalmouse has probably been re-uploaded a few times since then. If stage looks calm I can try and make a test case, otw test locally.

scout/server/blueprints/alignviewers/views.py
northwestwitch: Done

scout/server/blueprints/alignviewers/views.py
dnil: Agreed, the code is needlessly repetitive. I'd suggest we do all of that, including track selection in a separate PR? But lets just not disable it for now..

scout/server/blueprints/variant/views.py
dnil: Generally I'd say we want to have all stored elements on objects in the class / object model. There may be exceptions. Not documenting them there seems like a recipe for confusion, e.g. when serialising objects?

In this case that would ensure new users get sensible defaults, but old users must choose once. Doesn't that make sense?

northwestwitch: Sure, I'll add all general tracks to the newly created users.

scout/server/blueprints/alignviewers/views.py
northwestwitch: I've added the splitting of the form input in https://github.com/Clinical-Genomics/scout/pull/2111/commits/5b6694d2b4e277f42488e8d0bf23b986cc8b5b5d

scout/server/blueprints/variant/views.py
northwestwitch: Done, and I've updated the "build_user" test to check that the default tracks are created when a new user is created.

scout/server/blueprints/alignviewers/views.py
dnil: Nice! Almost there, then, but the info still doesn't really propagate does it?
Setting locally
Screenshot 2020-10-05 at 11 45 58
and on a master-ish branch:
Screenshot 2020-10-05 at 11 43 58
but on this branch:
Screenshot 2020-10-05 at 11 45 02

northwestwitch: Shouldn't the upd_regions_bed file end in .bed? Or is is just a test you're doing with no real file?

dnil: Since when do we run DOS on the machines? ;-) But yeah, that would be the common situation. Still, works with master, but not branch?

northwestwitch: .txt, whatever! 😄 I'm trying!

dnil: Thanks - don’t want to give you too much headache over this as it will anyway have to change soon again, but on the gist of not mending what ain’t broken - or not breaking what soon will need mending?

I’ll be out for lunch, but can arrange a case with some other tracks on stage later, unless you beat me to it.

northwestwitch: It's OK, it should work. I think I'll wait for your test case, you know better than me what it should look like! Have a nice lunch!

northwestwitch: I'll try to fix the code in the meanwhile

northwestwitch:
I've attempted a fix with some debugging HTML code on the top of the alignments, to make sure the right stuff is passed, but I'm not sure it works

scout/server/blueprints/alignviewers/views.py
dnil: Thanks! I'll add back the TIDDIT coverage bw that is currently on HK from MIP9. That should work on any number of cases, but I guess I'll try it on a clone of vitalmouse. I'll test your latest locally while waiting for hasta..

dnil: Looks like it now loads locally (Still some DEBUG out left in there ofcourse.. 😄)! I didn't realise there were that many low scoring variants on the case so still loading for the clone-vital-mouse. I'll perhaps drop it and load a new one with only a few variants.

Screenshot 2020-10-05 at 13 33 56

dnil (approved):
:+1: But lets test on stage once more..

northwestwitch:
Fingers crossed 🤞

dnil:
Meh.. Almost..

northwestwitch:
> Meh.. Almost..

It still says Uploading screenshot

dnil:
One test too early to remove debug statements.. 😆 
Screenshot 2020-10-05 at 14 24 14
IGV view from here: https://scout-stage.scilifelab.se/cust000/643594-300M/4384c7e710935a483418dae8ad901b2f

dnil:
Ooo wait a sec - this time its a bw. The local one I tested was a bed.. Could it be just an odd warning for a track format mismatch?

northwestwitch:
> Ooo wait a sec - this time its a bw. The local one I tested was a bed.. Could it be just an odd warning for a track format mismatch?

I guess it could

northwestwitch:
Looks like it's missing the file extension

dnil:
File extension? As in DOS file extension? 😆

northwestwitch:
> File extension? As in DOS file extension?

Exaxctly

dnil:
And with missing you mean does not recognise? Bc the filename matches *.bw?

dnil:
But who on earth would make a system that depends on file extensions? That was tried with DOS and failed miserably..

northwestwitch:
> But who on earth would make a system that depends on file extensions? That was tried with DOS and failed miserably..

A new issue for igv.js perhaps? 🤣

dnil:
But they have examples with bigwigs ending with .bw?! Still, I do have the feeling that if I would just explicitly give it `format: bigwig`...

northwestwitch:
Yes, but in that case I think we have to specify on the track type that is a bigwig, perhaps we set it to another type?

northwestwitch:
which track is the one not working?

dnil:
Its this one:
```
                     {% for ttrack in tiddit_wig_tracks %}
                       {
                         type: "wig",
+                        format: "bigwig",
                         name: '{{ ttrack.name }}',
```

northwestwitch:
In the example they don't specify the format:

https://github.com/igvteam/igv.js/wiki/Wig-Track

dnil:
Do you want me to try changing it?

dnil:
Or perhaps just remove the format: wig according to the above example?

northwestwitch:
Done, format removed

northwestwitch:
😆 

image

dnil:
Ok, now we get the right track displaying at least. Its still not showing the data, but one can argue it wasn't before either. This will need to be fixed eventually anyway, e.g. we have an obvious PR with selecting among specific case tracks, adding individual names to the additional tracks, fixing their formats etc.

This is a good start that appears to be working to me! Let's merge?

northwestwitch:
> Let's merge?

YES!

northwestwitch:
Thanks for your review @dnil, must have been a painful one, especially testing it! 🙏

northwestwitch:
Merging now! ...

dnil:
> Thanks for your review @dnil, must have been a painful one, especially testing it! 🙏

Yeah, well, thank you! We should have had examples for those extra tracks ready on stage. The local part was fine.. 😄

PR #2113, opened 2020-09-21
use single quotes in inner html badge class

### LINKED ISSUE #2110: Strs variants page is looking really weird for a case on stage
image
 The case is this one: https://scout-stage.scilifelab.se/cust000/643594-300M/str/variants?variant_type=clinical

Perhaps there is something wrong just with this case, because I haven't seen the behavior in another one, but I'll just drop it here so we can check!

northwestwitch:
Fix #2110 
Double quotes in inner HTML probably closed the string content of the popover.

 
**How to test**:
1. Install master branch on Stage and check that this page doesn't render properly: https://scout-stage.scilifelab.se/cust000/643594-300M/str/variants?variant_type=clinical
1. Install the branch and check that same page is looking normal

**Expected outcome**:

**Review:**
- [x] code approved by EM
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Editable variants comments
 ### Fixed
 - Empty variant activity panel
+- STRs variants popover
 ### Changed
 - Updated RELEASE docs.
 - Pinned variants card style on the case page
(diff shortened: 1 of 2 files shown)


elevu (approved):

northwestwitch:
Before:

image


After:

image

PR #2121, opened 2020-09-22
SpliceAI annotations

dnil:
This PR adds a functionality: SpliceAI annotations from VCF info if provided - as by e.g. MIPv9.

**How to test**:
1. (re)load a case with SpliceAI variants
1. note that these are being parsed correctly, and displayed on the variant page

Screenshot 2020-09-23 at 14 57 44

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [ ] tests executed by


scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -182,6 +182,10 @@
           SPIDEX
           <span class="float-right">{{ variant.spidex_human }}</span>
         </li>
+        <li class="list-group-item">
+          SpliceAI DS Max
+          <span class="float-right">{{ variant.spliceai_ds_max }}</span>
```
northwestwitch, line 187: We're saving the variant spliceai string (bla|bla|bla|..) but we're not showing it on the variant page. Wouldn't users be interested in those values?

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -182,6 +182,10 @@
           SPIDEX
           <span class="float-right">{{ variant.spidex_human }}</span>
         </li>
+        <li class="list-group-item">
```
northwestwitch, line 185: Perhaps we could have a condition here, like {% if variant.spliceai_ds_max %} so the li is not shown if there's no value in the field? But perhaps it's informative knowing that there is no spliceai. In that case you could write later in the span:
<span class="float-right">{{ variant.spliceai_ds_max or "-"}} </span>

northwestwitch (approved):
Looks perfect! I've got a couple of suggestion, see comments!

scout/server/blueprints/variant/templates/variant/variant_details.html
dnil: Sure! I'll do the same for spidex!

scout/server/blueprints/variant/templates/variant/variant_details.html
dnil: Do you know what values in particular? But, presumably I would be interested in all. Perhaps a hover? I'll try something!

northwestwitch: A hover would be perfect!

scout/server/blueprints/variant/templates/variant/variant_details.html
dnil: It turns out spidex has a solution where the value is always looked up, and spidex_human set to "not_reported" when not found. To each his own I guess.. Chalk one for refactor. I currently like your solution better so I'll use it.

scout/server/blueprints/variant/templates/variant/variant_details.html
dnil: Hm, I wonder if they will get that? I would probably have to look up the VCF-info key and then read the SpliceAI-paper to understand, but at least now it is displayed in some form.

dnil:
Testing with real data is king. Turns out SpliceAI and SpliceAI_DS_max are in fact lists, not just floats: one value per gene, but not per transcript. 👀 

Anyway, will fix!

dnil:
@jemten will have a look at the MIP code for producing and scoring: it is not necessarily fully aware of the list nature either. He will test with VEP first, and see how the CSQ annotation looks. That means this PR will go on hold until further notice..

dnil:
In fact, I think I will close it and we can resume from it or start anew whenever.

northwestwitch:
I guess it's always good to test with real data. The format of this spliceAI looks complicated!

PR #2114, opened 2020-09-21
Fix splitting multiple clnsig terms in variant parsing

northwestwitch:
fix #2112.

**How to test**:
1. Install this branch on Hasta stage.
1. Load a case that before showed warnings (vitalmouse?)
1. Make sure that there are no warnings on reupload
1. Go to a variant that previously caused warnings on upload (significance: "Conflicting_interpretations_of_pathogenicity&_other"). For instance: chr6, start:26091179. When clicking on the variant page, it should show the correct significance (conflicting_interpretations and other)

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by MM
- [x] tests executed by CR

PR #2127, opened 2020-09-28
Fix parsing multi-cnlsig vars when clnsig acc is None

### LINKED ISSUE #2123: Warnings about CLINSIG while uploading a dummy case
While testing this PR: https://github.com/Clinical-Genomics/scout/pull/2117 I noticed some strange warnings about clinsig. Please see the screenshot below:

image

Case name on stage is: actualgecko and cust000

northwestwitch:
fix #2123 

I've been testing with some variants and I think the error occurs when the variant has clnsig values with no accession. In #2114 I've been kind of fixing the same situation when the clnsig has accession values and the't why this problem is still there in this case.

Clinvar parsing in Scout has been changing over the years and for backward compatibility we have kept old code and added new lines without really planning what we expect from the pipelines. And that's why we get unexpected situations and different keys and values from MIP and BALSAMIC. We often don't know if something is wrong unless a user complains about warnings. In the future we should simplify this code, otherwise it's too prone to bugs.

Anyway, this should fix the error in the issue.

**How to test**:
1. Keep in mind a variant with values that has VCF annotation causing warnings, for instance this one: https://scout-stage.scilifelab.se/cust000/actualgecko/afacfdf4b9e4ba11b6fa45d86d9fea22?cancer=yes
At the VCF level this variant has the following annotation: `pathogenic/likely_pathogenic&likely_pathogenic&pathogenic`
1. Variant on current master looks like this:
image
1. Install this branch on hasta stage and reupload the case (with -u option to force-update).
1. Make sure that there are no warnings during the variants upload
- [x] Make sure that the clnsig values of the variant on the variant page changes, and `other` annotation should disappear.
- [x] `Clinsig value _other does not have an internal representation `warning should not show any more

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - CNV report link in cancer case side navigation
 ### Fixed
 - missing `vcf_cancer_sv` and `vcf_cancer_sv_research` to manual.
-- Split ClinVar multiple clnsig values (slash-separated) for annotations without accession number
+- Split ClinVar multiple clnsig values (slash-separated) and strip them of underscore for annotations without accession number
 ### Changed
 - Do not freeze mkdocs-material to version 4.6.1
 - Remove pre-commit dependency
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,10 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [x.x.x]
 ### Added
+- CNV report link in cancer case side navigation
 ### Fixed
 - missing `vcf_cancer_sv` and `vcf_cancer_sv_research` to manual.
-
+- Split ClinVar multiple clnsig values (slash-separated) for annotations without accession number
 ### Changed
 - Do not freeze mkdocs-material to version 4.6.1
 - Remove pre-commit dependency
@@ -18,7 +19,6 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Editable cases comments
 - Editable variants comments
-- CNV report link in cancer case side navigation
 ### Fixed
 - Empty variant activity panel
 - STRs variants popover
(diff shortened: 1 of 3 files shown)


northwestwitch:
@hassanfa if you point me to the config file of this case I can test the re-upload myself. Thanks!

hassanfa:
> @hassanfa if you point me to the config file of this case I can test the re-upload myself. Thanks!

I sent you a DM on Slack.

northwestwitch:
Ok the fix works, but while testing I've observed also this other warning: Clinsig value _other does not have an internal representation.

I'll fix this one as well in this PR

northwestwitch:
The variant causing the warning described in the comment above is this: https://scout-stage.scilifelab.se/cust000/actualgecko/12e95fc1314c880ac2a5631263ce9428?cancer=yes

And the clnsig values created in master branch look like this:
image

At the VCF level the values are the following: `drug_response&likely_benign&_other`

northwestwitch:
- [x] No errors on variant upload (you have to trust me on this because it's more then 4000 snvs and it would take me forever for screenshots.
- [x] Uploaded variant https://scout-stage.scilifelab.se/cust000/actualgecko/afacfdf4b9e4ba11b6fa45d86d9fea22?cancer=yes. The `other` due to un-recognized clnsig during parsing disappears and the correct values are shown:

image

- [x] Uploaded variant: https://scout-stage.scilifelab.se/cust000/actualgecko/12e95fc1314c880ac2a5631263ce9428?cancer=yes.  The same values of before are displayed (they were correct also before, but just because _other was not recognized and whenever something was not recognized it was assigned to `other`). Now the value is collected for what it is:

image

northwestwitch:
Ready for review!

moonso (approved):
👍 A test maybe to highlight the fix?

northwestwitch:
> A test maybe to highlight the fix?

Good idea, I'll add it!

northwestwitch:
Test included, merging. Thanks @moonso for the review!

PR #2137, opened 2020-10-01
MT vs autosome (14) mean coverage stats

northwestwitch:
This PR adds 3 following fields to the MT report of each sample of a case:

image

These new lines are added at the bottom of each sheet containing the MT variant info for the given sample.

**How to test**:
1. Install the branch in Scout stage (clinical-db)
1. Go to a case page and download the MT report

**Expected outcome**:
- [x] MT report of samples downloaded from Scout stage should contain the additional lines
- [x] A local instance of Scout (demo) shouldn't contain this info because the app is not connected to a chanjo instance

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
This is also ready for review

northwestwitch:
Since cust003 has been waiting for a long time for this feature I'd really like to include this one in the next release could someone review?

dnil:
Sure!

northwestwitch:
> Sure!

Keep in mind that with cust003 we're still deciding which autosome to use (I guess we'll decide today). Any suggestion btw?

dnil:
hehe, this is partly why I talked about those other files.. Your best candidate is a gc% matched set of bins across multiple autosomes. This may or may not be simple with the current solution though. One of the larger ones is a fairly good bet if you must choose one and only one. For e.g. cfDNA, a set is chosen, but then the need for exact calls is naturally higher as one strives to detect small percent changes to the totals of [13,18,21].

northwestwitch:
A paper describing a complete different method then the transcript coverage from Chanjo, but has a nice insight on pros and cons of choosing specific chromosomes

dnil:
Right: a typical approach for SMN or other specific CNV region calls is to establish a baseline of bins that are CN-stable, and typically GC-matched to the region of interest. Anyway, let me test now.. 😛

northwestwitch:
Woa, I'm reading that paper and look what I've found:

image

They multiply the ratio X2 to calculate the copy number, why?

dnil:
😄 Autosome baseline CN=2.. (1/2)^-1..

northwestwitch:
> 😄 Autosome baseline CN=2.. (1/2)^-1..

Right! Mitochondria are bacteria after all. I need to fix the formula in this PR

scout/constants/variants_export.py
```diff
@@ -19,6 +19,9 @@
     "AD Alternative",
 ]
 
+# Used only if app is connected to chanjo report
```
dnil, line 22: I'm not sure that the comment is warranted? If you feel it is, perhaps describe rather that it contains human readable descriptions of computed coverage stats for the MT reports?

dnil: But good that it is a constant!

scout/server/app.py
```diff
@@ -75,7 +75,10 @@ def create_app(config_file=None, config=None):
     def check_user():
         if not app.config.get("LOGIN_DISABLED") and request.endpoint:
             # check if the endpoint requires authentication
-            static_endpoint = "static" in request.endpoint or request.endpoint == "report.report"
+            static_endpoint = "static" in request.endpoint or request.endpoint in [
+                "report.report",
+                "report.json_chrom_coverage",
```
dnil, line 80: We should fix that at some point.. But not today.

scout/constants/variants_export.py
```diff
@@ -19,6 +19,9 @@
...
+MT_COV_STATS_HEADER = ["Mean MT coverage", "Mean autosome coverage", "MT/autosome ratio"]
```
dnil, line 23: Eh, this is semantics but it is rather mean autosome N coverage currently? 😊

scout/server/blueprints/cases/controllers.py
```diff
@@ -388,6 +390,48 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
...
+    Returns:
+        coverage_stats(dict): a dictionary with mean MT and autosome transcript coverage stats
+    """
+    coverage_stats = {}
+    ind_ids = []
+    for ind in individuals:
+        ind_ids.append(ind["individual_id"])
+
+    # Prepare complete url to Chanjo report chromosome mean coverage calculation endpoint
+    cov_calc_url = url_for("report.json_chrom_coverage", _external=True)
```
dnil, line 409: Oh, is external needed? Nothing wrong, just a little surprised. Well found then!

scout/server/blueprints/cases/controllers.py
```diff
@@ -388,6 +390,48 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
...
+    # Prepare request data to calculate mean MT coverage
+    data = dict(sample_ids=",".join(ind_ids), chrom="MT")
+    # Send POST request with data to chanjo endpoint
+    resp = requests.post(cov_calc_url, json=data)
```
dnil, line 413: 👍

scout/server/blueprints/cases/controllers.py
```diff
@@ -388,6 +390,48 @@ def coverage_report_contents(store, institute_obj, case_obj, base_url):
...
+    ref_cov_data = json.loads(resp.text)  # mean coverage over the transcripts of ref chrom
+
+    for ind in ind_ids:
+        if not (mt_cov_data.get(ind) and ref_cov_data.get(ind)):
+            continue
+        coverage_info = dict(
+            mt_coverage=round(mt_cov_data[ind], 2),
+            autosome_cov=round(ref_cov_data[ind], 2),
+            mt_autosome_ratio=round(mt_cov_data[ind] / ref_cov_data[ind], 2),
```
dnil, line 428:
```suggestion
            mt_autosome_ratio=round(mt_cov_data[ind] / ref_cov_data[ind] * 2, 2),
```
But you knew that already! 😊

scout/server/blueprints/cases/controllers.py
```diff
@@ -422,13 +471,26 @@ def mt_excel_files(store, case_obj, temp_excel_dir):
...
+
         for col, field in enumerate(MT_EXPORT_HEADER):
             Report_Sheet.write(row, col, field)
 
         # Write variant lines, after header (start at line 1)
         for row, line in enumerate(sample_lines, 1):  # each line becomes a row in the document
             for col, field in enumerate(line):  # each field in line becomes a cell
                 Report_Sheet.write(row, col, field)
+
+        if bool(
```
dnil, line 483: Works, thought the bool() should not be needed, but better safe I guess. 🤷‍♂️

tests/server/blueprints/cases/test_cases_controllers.py
```diff
@@ -1,8 +1,47 @@
...
+
+    def mock_post(*args, **kwargs):
+        return MockResponse()
+
+    # AND a mock response from chanjo API with coverage stats for some samples
+    class MockResponse:
+        def __init__(self):
+            self.text = '{"sample1":36, "sample2": 32, "sample3": 34}'
+
+    monkeypatch.setattr(requests, "post", mock_post)
```
dnil, line 28: 🐒 🔧

dnil (approved):

northwestwitch:
If you looks at the supplementary table of that paper: 

image

It's not the same approach and we use only transcripts over gene regions, so perhaps it's fine, but anyway looks like a good choice to exclude chroms: 19, 22, 17, 16, 20.

Chromosome 21, the one used in the current PR looks fine, but it has only about 450 transcripts on Chanjo, so perhaps a bigger chrosome with low CG content and hopefully more transcripts? What about 18? Or even 13 if the computation doesn't take that long. I'll try with 13!

tests/server/blueprints/cases/test_cases_controllers.py
northwestwitch: Works so nicely!

scout/constants/variants_export.py
northwestwitch: Yes, I mean to add the chrom, but I still don't know which one to use!

dnil:
Yes, never 19 or 17!
I would personally feel a tad irrationally uncomfortable with 13, since trisomy of is viable, though very rare. 11 or 15 aren't that different I guess, and have a little higher / closer GC% to MT?

scout/server/blueprints/cases/controllers.py
northwestwitch: Yes, otherwise you get an error when you create the url (it tells you that the url can't be created, did you mean blabla ? *blabla = some other random view)

scout/server/blueprints/cases/controllers.py
northwestwitch: > Yes, never 19 or 17!
> I would personally feel a tad irrationally uncomfortable with 13, since trisomy of is viable, though very rare. 11 or 15 aren't that different I guess, and have a little higher / closer GC% to MT?

I've seen many genomics profiles with somatic trisomy 15 in elderly in my previous job, I would exclude it as well.

northwestwitch:
What about 18? It's bigger then 21 and has more genes and still low CG content

dnil:
> What about 18? It's bigger then 21 and has more genes and still low CG content

Yeah, but isn't MT like 44%? My intuition would be to pick a similar autosome to get a fair ratio?

northwestwitch:
> Yeah, but isn't MT like 44%? My intuition would be to pick a similar autosome to get a fair ratio?

14? Otherwise I'm afraid we're going up on transcripts and the computation might take a long time?

northwestwitch:
Updated with the requested changes I think, but let me know if I forgot something!

northwestwitch:
@dnil did you try it again on stage or should I?

dnil:
I didn't; I think the comment edits are safe enough, but some performance evaluation of the chr change is probably in order! 😸

northwestwitch:
Works fine, also the fix in the copy number estimate formula: 

image

I'm going to merge. Thanks @dnil !!

PR #2138, opened 2020-10-01
Missing default panel

dnil:

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -713,7 +713,9 @@ def case_default_panels(case_obj):
         case_panels(list): a list of panels (panel_name)
     """
     case_panels = [
-        panel["panel_name"] for panel in case_obj.get("panels", []) if panel["is_default"] is True
+        panel["panel_name"]
+        for panel in case_obj.get("panels", [])
+        if panel.get("is_default", None) is True
     ]
     return case_panels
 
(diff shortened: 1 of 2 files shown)


northwestwitch (approved):
It works fine! 👍 

I've reproduced the error by removing the "is_default" key from the panel in test case. Master branch crashes. When I switch to this branch the variants are shown.

dnil:
> It works fine! 👍
> 
> I've reproduced the error by removing the "is_default" key from the panel in test case. Master branch crashes. When I switch to this branch the variants are shown.

Brilliant! Thanks!

scout/server/blueprints/variants/controllers.py
```diff
@@ -713,7 +713,9 @@ def case_default_panels(case_obj):
         case_panels(list): a list of panels (panel_name)
     """
     case_panels = [
-        panel["panel_name"] for panel in case_obj.get("panels", []) if panel["is_default"] is True
+        panel["panel_name"]
+        for panel in case_obj.get("panels", [])
+        if panel.get("is_default", None) is True
```
northwestwitch, line 718: You could also write if panel.get("is_default",False) is True, but the result would be the same

dnil: Yes, I was torn.. 😸

PR #2141, opened 2020-10-02
Fix empty other causatives lines

### LINKED ISSUE #2139: Blank matching causatives?
Screenshot 2020-10-01 at 15 57 20

Looks like we might be making some cards / rows even if we are not supposed to show the variants for various reasons?

northwestwitch:
fix #2139 --> Whenever the other causative variant has no hgnc symbol, display hgnc id instead!

**How to test on a local scout instance (scout demo)**:
1. load another case into cust000: case  643595 that is the exact replica of test case 643594.
1. Go to case 643595, variants page, and in variants filter UNcheck the test panel, so that all variants are shown:

image

1. Mark as causative a couple of variants that have no gene symbol, for instance the one ranked 1 (gene 14931) and 12 (gene 12015).

1. Go to the other case page and see two empty lines under the other causatives.

1. Switch to this branch and make sure that the other causatives under the case 643594 are now shown (their hgnc id).

1. Bonus test: mark POT1 as causative on case 643595 and see that "POT1" is shown under the other cases other causatives

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -243,7 +243,7 @@
         <li class="list-group-item">
           <a href="{{ url_for('variant.variant', institute_id=institute._id,
                               case_name=case.display_name, variant_id=variant._id) }}">
-            {{ variant.hgnc_symbols|join(', ') }}
+            {{ variant.hgnc_symbols|join(', ') or variant.hgnc_ids|join(', ')}}
           </a>
         </li>
       {% else %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Timeout of `All SNVs and INDELs` page when no valid gene is provided in the search
 - Round CADD (MIPv9)
 - Missing default panel value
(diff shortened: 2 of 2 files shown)


northwestwitch:
Ready for review!

dnil:
Works like a charm:
Screenshot 2020-10-02 at 09 39 20
Screenshot 2020-10-02 at 09 39 37

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -243,7 +243,7 @@
         <li class="list-group-item">
           <a href="{{ url_for('variant.variant', institute_id=institute._id,
                               case_name=case.display_name, variant_id=variant._id) }}">
-            {{ variant.hgnc_symbols|join(', ') }}
+            {{ variant.hgnc_symbols|join(', ') or variant.hgnc_ids|join(', ')}}
```
dnil, line 246: :+1:

dnil: And causative STRs that may not yet have hgnc_ids never get here anyway, so should be quite safe.

dnil (approved):

PR #2152, opened 2020-10-07
Feature pymongo3.11 201007

mikaell:
Re-PR of #2102. Which did not invoke Coveralls.

This PR adds a functionality or fixes a bug: Pymongo 3.7 --> 3.11

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


mikaell:
This PR is ready for review (finally). 

The original (#2102) stopped calling the GitHub action to run coverage and pytests for some  unknown reason. Will close that PR and delete the branch.

moonso:
I guess this will need quite extensive manual testing, like clicking around and stress the system before deployment. Perhaps we could do a common effort on this with @henrikstranneheim @northwestwitch @dnil and @elevu ?

dnil:
If it rains, we can have a session tomorrow with very rapid clicking instead of outdoor sports.. 😉

northwestwitch:
I'll start testing it locally today!

northwestwitch:
Tests performed on a fresh conda env, python=3.8 and mongo=3.6

- [x] scout setup demo
image

scout/server/blueprints/variants/views.py
```diff
@@ -256,13 +257,17 @@ def sv_variants(institute_id, case_name):
 
     # Setup variant count session with variant count by category
     controllers.variant_count_session(store, institute_id, case_obj["_id"], variant_type, category)
-    session["filtered_variants"] = variants_query.count()
+    query = store.get_query(case_obj["_id"], form.data, None, category)
+    result_size = variants_query.collection.count_documents(query)
```
northwestwitch, line 261: Why don't you use the store.variants_count that you created in scout/adapter/mongo/variant.py to do that instead?

scout/server/blueprints/variants/views.py
```diff
@@ -343,12 +348,14 @@ def cancer_variants(institute_id, case_name):
 
     variant_type = request.args.get("variant_type", "clinical")
     variants_query = store.variants(case_obj["_id"], category="cancer", query=form.data)
+    query = store.get_query(case_obj["_id"], form.data, None, category)
```
northwestwitch, line 351: Why don't you use the store.variants_count that you created in scout/adapter/mongo/variant.py to do that instead?

scout/server/blueprints/variants/views.py
```diff
@@ -379,22 +386,23 @@ def cancer_sv_variants(institute_id, case_name):
...
     form = controllers.populate_sv_filters_form(store, institute_obj, case_obj, category, request)
-
     cytobands = store.cytoband_by_chrom(case_obj.get("genome_build"))
-
     variants_query = store.variants(case_obj["_id"], category=category, query=form.data)
 
     # Setup variant count session with variant count by category
     controllers.variant_count_session(store, institute_id, case_obj["_id"], variant_type, category)
-    session["filtered_variants"] = variants_query.count()
+    query = store.get_query(case_obj["_id"], form.data, None, category)
```
northwestwitch, line 395: Same as above: the store.variants_count does just that.

scout/server/blueprints/variants/views.py
```diff
@@ -161,15 +160,17 @@ def variants(institute_id, case_name):
     cytobands = store.cytoband_by_chrom(case_obj.get("genome_build"))
 
     variants_query = store.variants(case_obj["_id"], query=form.data, category=category)
+    query = store.get_query(case_obj["_id"], form.data, None, category)
```
northwestwitch, line 163: Why don't you use the store.variants_count that you created in scout/adapter/mongo/variant.py to do that instead?

scout/adapter/mongo/variant.py
```diff
@@ -204,6 +204,19 @@ def variants(
 
         return result
 
+    def variants_count(self, case_id, query=None, variant_ids=None, category="snv"):
```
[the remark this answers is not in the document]
northwestwitch: Nice, but I see that it's called only in the tests. You could use it also from the variants views, I've added a few suggestion later in the comments

scout/adapter/mongo/variant.py
```diff
@@ -204,6 +204,19 @@ def variants(
...
+        """Count number of variants
```
northwestwitch, line 208: Could you elaborate this docstring a bit more? You could follow the general convention that we use in Scout. It contains a short description, then Args and  Returns

scout/adapter/mongo/variant.py
```diff
@@ -326,6 +339,32 @@ def gene_variants(
 
         return result
 
+    def count_gene_variants(
+        self, query=None, category="snv", variant_type=["clinical"], institute_id=None
+    ):
+        """Count all variants seen in a given gene.
+
+        Arguments:
```
northwestwitch, line 347: It's OK but to be consistent with the other docstrings, could you rename this one to Args and add a Returns?

scout/server/blueprints/cases/views.py
```diff
@@ -755,42 +730,29 @@ def mark_validation(institute_id, case_name, variant_id):
...
+        "variant.variant", institute_id=institute_id, case_name=case_name, variant_id=variant_id
     )
     store.validate(institute_obj, case_obj, user_obj, link, variant_obj, validate_type)
     return redirect(request.referrer or link)
 
 
 @cases_bp.route(
-    "/<institute_id>/<case_name>/<variant_id>/<partial_causative>/causative",
-    methods=["POST"],
+    "/<institute_id>/<case_name>/<variant_id>/<partial_causative>/causative", methods=["POST"]
```
northwestwitch, line 740: That's weird, a lot of formatting differences due to Black I guess?

scout/server/blueprints/variants/controllers.py
```diff
@@ -39,20 +39,14 @@
...
-)
+from .forms import CancerFiltersForm, FiltersForm, StrFiltersForm, SvFiltersForm, VariantFiltersForm
 
 LOG = logging.getLogger(__name__)
 
 
-def variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50):
+def variants(store, institute_obj, case_obj, variants_query, variant_count, page=1, per_page=50):
     """Pre-process list of variants."""
-    variant_count = variants_query.count()
```
northwestwitch, line 55: 👍

tests/server/blueprints/institutes/test_institute_views.py
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 from flask import url_for
 from scout.server.extensions import store
+import logging
```
northwestwitch, line 4: I think you can remove the import and the debug LOG message from this test. It won't be seen anyway

scout/adapter/mongo/variant.py
```diff
@@ -326,6 +339,32 @@ def gene_variants(
 
         return result
 
+    def count_gene_variants(
```
northwestwitch, line 342: Nice! Could you add a simple test for this one?

northwestwitch (changes requested):
I've gone through the code and wow you did a great job here! 🥇

I've got some suggestions on how to improve the code. It might seem a lot, but it's not entirely my fault, you picked a difficult task with this PR.. ;)

CHANGELOG.md
```diff
@@ -28,7 +28,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Do not freeze mkdocs-material to version 4.6.1
 - Remove pre-commit dependency
-
+- Update PyMongo to 3.11 (from 3.7)
```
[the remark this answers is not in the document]
mikaell: Done!

scout/adapter/mongo/variant.py
mikaell: This is inconsistent all over the code base. "Args:" win 381 to 58 for "Arguments". I could to a separate refactoring PR and change it everywhere. 

Though, personally I think "Arguments" is nicer.

ps. Looking into Python source code the ratio is about same. Args: 1129, Arguments: 211.

tests/server/blueprints/institutes/test_institute_views.py
mikaell: Done!

mikaell: > I think you can remove the import and the debug LOG message from this test. It won't be seen anyway

`pytest --log-cli-level=10` 🤓

scout/adapter/mongo/variant.py
northwestwitch: Bah I've always used Args, and as you counted it's mostly Args, so I'd change it to Args. If you want to rename it all over the place is fine, but I don't think it's super important at the moment

moonso: [Type hints!!!](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) then we can remove all of those

northwestwitch: I agree, but sometimes is also good to have some more info from the arguments instead of only the expected type..

scout/adapter/mongo/variant.py
mikaell: Done!

mikaell: Done!

scout/adapter/mongo/variant.py
mikaell: Done!

scout/adapter/mongo/variant.py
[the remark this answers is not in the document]
mikaell: Removed. Was only used for testing. Replaced with `len(list(cursor_to_variantDB.clone()))`.

scout/adapter/mongo/variant.py
```diff
@@ -204,6 +204,19 @@ def variants(
...
+        """Count number of variants
+        Replace deprecated pymongo.cursor.count()"""
+        mongo_query = self.build_query(
+            case_id, query=query, variant_ids=variant_ids, category=category
+        )
+        result = self.variant_collection.count_documents(mongo_query)
+
+        return result
+
+    def get_query(self, case_id, query, variant_ids, category):
```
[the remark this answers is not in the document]
mikaell: `get_query` was written at a time I thought a universal solution for replacing `cursor.count()` was possible…

scout/adapter/mongo/variant.py
mikaell: Function deleted.

scout/adapter/mongo/variant.py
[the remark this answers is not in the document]
mikaell: The design choice was made to be semantic and ideologically consistent with how the database interface is written. Direkt results are never returned from the API. Instead you get "handels" to be used for further actions. Usually a cursor (disguised by the naming convention "xxx_query" leading my thoughts to 1. a list of results or 2. a query dict used for further interactions). 

I'll rewrite it as:
```
    def count_query(self, case_id, query, variant_ids, category):
        """Returns number of variants

        Arguments:
            case_id(str): A string that represents the case
            query(dict): A dictionary with querys for the database
            variant_ids(List[str])
            category(str): 'sv', 'str', 'snv', 'cancer' or 'cancer_sv'

        Returns:
             integer
        """

        query = self.build_query(case_id, query=query, variant_ids=variant_ids, category=category)
        self.variant_collection.count_documents(query)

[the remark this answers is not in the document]
Comment (northwestwitch) [reply scout/adapter/mongo/variant.py:207]: Perhaps move it to the scout/adapter/mongo/query.py file?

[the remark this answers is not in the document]
Comment (mikaell) [reply scout/adapter/mongo/variant.py:207]: > Perhaps move it to the scout/adapter/mongo/query.py file?

I don't think that is possible. As `self.variant_collection` need to be referred.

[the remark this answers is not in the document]
Comment (northwestwitch) [reply scout/adapter/mongo/variant.py:207]: I'm quite sure it will work, since variant collection is defined in the setup of the adapter class: https://github.com/Clinical-Genomics/scout/blob/c97a4b1f977fd50e7e4dd92bdf9228beeeff526d/scout/adapter/mongo/base.py#L101

Try to invoke it in the same way (self. variant_collection)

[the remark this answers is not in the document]
Comment (northwestwitch) [reply scout/adapter/mongo/variant.py:207]: Or perhaps import the other class, dunno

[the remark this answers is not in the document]
Comment (mikaell) [reply scout/adapter/mongo/variant.py:207]: I don't agree that moving the functionality to 'query.py' is desired. 
1. 'query.py' purpose seems to be acting as a library module, functions are called to return dicts later used for querying.
2. 'query.py' is a mess! Adding more lines of code will only make the much needed refactoring harder.
3. the purpose is counting number of variants -so I'll rename it `count_variants()`

**Comment by northwestwitch** [comment]
Hi @mikaell, let me know when you are done with this so I can re-review!

**Comment by mikaell** [comment]
> Hi @mikaell, let me know when you are done with this so I can re-review!

Hi,
I think its ready to a second go!

File: scout/adapter/mongo/variant.py (hunk 13 above)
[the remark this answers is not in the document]
Comment (northwestwitch) [reply scout/adapter/mongo/variant.py]: Ok!

File: scout/adapter/mongo/hgnc.py [hunk 14]
```diff
@@ -138,6 +138,43 @@ def hgnc_genes(self, hgnc_symbol, build="37", search=False):
 
         return self.hgnc_collection.find({"build": build, "aliases": hgnc_symbol})
 
+    def hgnc_genes_count(self, hgnc_symbol, build="37", search=False):
```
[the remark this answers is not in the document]
northwestwitch: You marked as solved this one but you haven't really considered my comment. You are creating 2 new functions here (hgnc_genes_count, get_query_alias_or_id) to do in a quite complicated way I think what you could solve in literally 1 line of code. I think the code would be much more easy to maintain if you avoid adding these functions here.

tests/server/blueprints/institutes/test_institute_views.py
northwestwitch: Nice!

northwestwitch (changes requested):
Hi, I've got just one comment left. Don't kill me but I've done a grep for ".count()" over the code of this PR and found the following:

```
./scout/server/blueprints/variants/templates/variants/utils.html:  {% set comment_count = variant.comments.count() %}
./scout/server/blueprints/cases/templates/cases/case_report.html:	      {% if comments.count() %}
./scout/server/blueprints/cases/templates/cases/case_report.html:      {% if variant.comments.count() %}
./scout/server/blueprints/cases/templates/cases/case_report.html:    {% if variant.comments.count() %}
./scout/server/blueprints/institutes/templates/overview/clinvar_submissions.html:        {% if subm_obj.variant_data and subm_obj.variant_data.count() > 0 %}
./scout/server/blueprints/institutes/templates/overview/clinvar_submissions.html:        {% if subm_obj.case_data and subm_obj.case_data.count() > 0 %}
./scout/server/blueprints/managed_variants/controllers.py:    variant_count = managed_variants_query.count()
./scripts/merge_users.py:        if oi_user_events.count() > 0:
./scripts/merge_users.py:        if oi_user_acmg.count() > 0:
./scripts/merge_users.py:        if oi_user_clinvar.count() > 0:
./scripts/merge_users.py:        if oi_user_clinvars.count() > 0:
./scripts/merge_users.py:        if oi_user_cases.count() > 0:
```

These ones would be deprecated as well..

I've tried to install the branch on Hasta stage and got the following error:

image

When on the other hand one installs cg gets this:

image

The problem is that they're installed on the same environment. How to fix this? Any idea @moonso and @dnil ?

moonso:
> The problem is that they're installed on the same environment. How to fix this? Any idea @moonso and @dnil ?

Yes the requirement is set to `pymongo<3.7` in CG to not mess up the scout dependency since they live in the same env. Whenever this requirement is gone from Scout we can modify `requirements.txt` in CG and remove `pymongo` from that file.

northwestwitch:
> Yes the requirement is set to `pymongo<3.7` in CG to not mess up the scout dependency since they live in the same env. Whenever this requirement is gone from Scout we can modify `requirements.txt` in CG and remove `pymongo` from that file.

Ah OK! But then you could remove `pymongo<3.7` from requirements.txt in cg also now, since pymongo is already installed in the env and when you install cg the requirements would be already fulfilled. Unless you need to install an upgrade of pymongo also in cg. At this regard I have a question to you all, why do we want freeze pymongo to 3.11 in this PR?

mikaell:
> Don't kill me but I've done a grep for ".count()" over the code of this PR and found the following:

I'll have a look at these. But can we go forward with this PR and do the new findings in a separate PR?

northwestwitch:
> I'll have a look at these. But can we go forward with this PR and do the new findings in a separate PR?

Sure, they ca be addressed in another one, this is already big as it is!

northwestwitch:
I'm going to install the branch on clinical-db, so we can test it there before merging!

northwestwitch:
It's on clinical-db stage now and looks like it's working fine!

requirements.txt
```diff
@@ -15,7 +15,7 @@ query_phenomizer
 Flask-Babel
 livereload
 python-dateutil
-pymongo==3.11
+pymongo>=3.7
```
northwestwitch, line 18: 👍

northwestwitch (approved):
Approving this, but please consider my comment regarding hgnc_genes_count before merging

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Breakpoint two IGV button now shows correct chromosome when different from bp1
 ### Changed
 - Make matching causative and managed variants foldable on case page
+- Update PyMongo to 3.11 (from 3.7)
```
northwestwitch, line 14:
```suggestion
- Update PyMongo to >=3.7
```

CHANGELOG.md
```diff
@@ -30,7 +31,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Do not freeze mkdocs-material to version 4.6.1
 - Remove pre-commit dependency
-
+- Update PyMongo to 3.11 (from 3.7)
```
northwestwitch, line 34: This line should be removed

northwestwitch, line 34:
```suggestion
```

scout/adapter/mongo/hgnc.py
```diff
@@ -131,34 +131,19 @@ def hgnc_genes(self, hgnc_symbol, build="37", search=False):
             )
         return self.hgnc_collection.find({"build": build, "aliases": hgnc_symbol})
 
-    def hgnc_genes_count(self, hgnc_symbol, build="37", search=False):
-        """Count all hgnc genes that match a hgnc symbol. Replaces depricated
+    def hgnc_genes_find_one(self, hgnc_symbol):
```
northwestwitch, line 134: Hi, I meant that this function is not necessary. You can call the store.hgnc_collection.find_one directly from the functions that now call this one.

mikaell: I think we should have an interface layer towards MongoDB. This makes the code more likely to be maintainable next time the Pymongo library is changed.

moonso: I agree with @mikaell . This is the reason why we got stuck on an old version of pymongo in the first place, it's because we used pymongo methods directly instead of defining an api

northwestwitch: I agree as well on the principle, but in this case it's a function of one line that calls a direct pymongo method that is called only 2 times in the whole repo. But if you want to keep it this way it's fine by me!

CHANGELOG.md
northwestwitch: I took the liberty to fix it ;)

mikaell: Reworded to: "Remove calls to PyMongo functions marked as deprecated (as of version 3.7)."

northwestwitch: I saw! the line I removed was a duplicated entry in the changelog, under the last scout version (4.23)

PR #2155, opened 2020-10-07
Fix #2153 - hide causatives subtly

### LINKED ISSUE #2153: Subtly hide matching causatives
At the moment the pendulum locally swings more towards hiding things that may potentially cause ethical concern and trigger corresponding review than towards increasing diagnostic yield. Without destroying functionality we would like to make it possible to slightly hide matching causative (and managed variants) from the case page view. 

A suggestion may be to introduce a collapsible fold out for the two cards, and an indicator stating if and how many such are available when folded away. This way, the operator can decide to e.g. after finished initial clinical test, they also want to see those variants. Or if they so choose, elect to see them before going through said analysis.

dnil:

**How to test**:
1. open a case with matching causative and / or matching managed variants
1. note that managed variants and causative variants are shown folded with a number of variants until clicked

Screenshot 2020-10-07 at 15 18 17

Screenshot 2020-10-07 at 15 18 28

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Update dismissed variant status when variant dismissed key is missing
 ### Changed
+- Make matching causative and managed variants foldable on case page
 
 ## [4.23]
 ### Added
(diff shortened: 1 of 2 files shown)


northwestwitch:
I can review this one!

northwestwitch (approved):
Code-wise looks good and it seems like that you tested and it works, so it's fine to merge I think!

northwestwitch:
Merging this!

PR #2156, opened 2020-10-08
Improve "update individual" cli command

### LINKED ISSUE #2133: Make the update individuals command more user-friendly
Sometimes it is necessary to update the alignment path for an individual of a case.

Scout has a dedicated command to do that: `scout update individual`, that accepts the following inputs:
 -c, --case-id TEXT     [required]
 -i, --ind-id TEXT     [required]
 -a, --alignment-path PATH Replace alignment file

case id is already displayed on the case web page so it's fine,  but the --ind-id is not the same as the individual display name that can be seen on the case page. It's the `individual_id` values instead. It would be nice to change it into individual display name so that it would be much easier for prod team to run the command independently.

northwestwitch:
A user-friendly `update individual` cli command with more update options. fix #2133

The same command in current master allows users to update only path to bam/cram alignments (bam_file key). 
This branch will make it possible to update the following keys:
 - "bam_file",
 - "mt_bam",
 - "vcf2cytosure",
 - "rhocall_bed",
 - "rhocall_wig",
 - "tiddit_coverage_wig",
 - "upd_regions_bed",
 - "upd_sites_bed"

**How to test**:
1. Install branch locally or on hasta stage.
1. Run the command `scout update individual ` and follow the instructions.
1. Try providing non-valid individual name, keys or non-existent path to files and see how the command behaves

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Ready for review. An easy one!

mikaell (approved):
Looks good.

northwestwitch:
> Looks good.

Thanks!

Issue #1022, opened 2018-10-18
gnomAD PLi


dnil:
New pLI scores based on the entire gnomAD 2.1 are now available! https://macarthurlab.org/2018/10/17/gnomad-v2-1/
Should replace old exac pLI table!

moonso:
Could you point to that file?

dnil:
Here, I guess:
https://console.cloud.google.com/storage/browser/gnomad-public/release/2.1/ht/constraint?pli=1

(from gnomAD Downloads page).

moonso:
Hmm in this file we have different pLi scores for different transcripts. Not really sure how to handle that in Scout. Any suggestions?

Also I think pLi scores are included in the rank model so MIP has to update and score per transcript, not sure if it is MIP or VEP who does that, do you know @henrikstranneheim ?

henrikstranneheim:
Not by heart. I will take a look soon, I am about to update VEP in the near future

moonso:
Ping @henrikstranneheim

henrikstranneheim:
There is a gnomad pLI per gene file: Download it: https://storage.googleapis.com/gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_gene.txt.bgz

which Scout can use. MIP 7.0 is using this one (so we are still on gene level as the VEP plugin currently cannot handle transcript level)

henrikstranneheim:
For SV we are taking the most severe pli score for genes affected by the variant as a proxy:
```
##INFO=<ID=most_severe_pli,Number=1,Type=Float,Description="Most severe pli score.">
most_severe_pli=0.02
```
This part of the SV rank model in MIP 7.1. Display on SV view if relevant

Shicheng-Guo:
Do we have v3.1.1 download link? 

https://storage.googleapis.com/gnomad-public/release/3.1.1/constraint/gnomad.v3.1.1.lof_metrics.by_gene.txt.bgz

Shicheng-Guo:
I downloaded this gnomad.v2.1.1.lof_metrics.by_gene.txt.bgz and find these columns. Anyone can share some information about this p_amr, p_asj? What are these columns for? 

       p_amr      p_asj      p_eas     p_fin      p_nfe p_oth      p_sas

Thanks

![image](https://user-images.githubusercontent.com/6641283/134595433-8b1be669-ee90-4a14-ac98-e18a6ba5d5af.png)

[...]

PR #2189, opened 2020-11-10
Fix pLI link

dnil:

The link to ExAC was broken - or actually going over a slow redirect to their new server - 
Screenshot 2020-11-10 at 10 52 20

Screenshot 2020-11-10 at 11 26 14

Redirected link directly to new page, and also added a link to the corresponding gnomAD page, in waiting for a solution to #1022.

Screenshot 2020-11-10 at 11 26 22

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Breakpoint two IGV button now shows correct chromosome when different from bp1
 - Missing font lib in Docker image causing the PDF report download page to crash
 - Sentieon Manta calls lack Somaticscore - load anyway
-- clinVar submissions crashing due to pinned variants that are not loaded
+- ClinVar submissions crashing due to pinned variants that are not loaded
+- Point ExAC pLI score to new gnomad server address
 ### Changed
 - Make matching causative and managed variants foldable on case page
 - Remove calls to PyMongo functions marked as deprecated in backend and frontend(as of version 3.7).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Breakpoint two IGV button now shows correct chromosome when different from bp1
 - Missing font lib in Docker image causing the PDF report download page to crash
 - Sentieon Manta calls lack Somaticscore - load anyway
-- clinVar submissions crashing due to pinned variants that are not loaded
+- ClinVar submissions crashing due to pinned variants that are not loaded
```
northwestwitch, line 17: Right! :)

scout/server/blueprints/genes/templates/genes/gene.html
```diff
@@ -82,9 +82,9 @@ <h1>
               </span>
             </li>
             <li class="list-group-item">
-              pLi Score(ExAC)
+              pLi Score(ExAC) - (compare to <a href={{ record.gnomad_link }} target="_blank">GnomAD here</a>)
```
northwestwitch, line 85: How often would it happen that genes don't have Ensembl Ids? Because if it's extremely rare then it's fine, otherwise the record.gnomad_link will be none

northwestwitch (approved):
Nice! 👍

scout/server/blueprints/genes/templates/genes/gene.html
dnil: Hm, that shouldn't really happen, but we may - as the tests suggest - not have pLI for all genes. Making it conditional on that, which should hopefully take care of all of the rare missing ENSEMBL ids as well.

PR #2191, opened 2020-11-10
IGV compatible filtering of variants

mhkc:
This PR adds a new input field when filtering variants. The filter takes an IGV position string (<chr>:<start>-<end>) as input which is a feature frequently requested by the analysts at Clinical Genomics Lund. The intent is to enable copy-pasting in position information from genome browsers such as IGV directly into scout.

**How to test**
1. Load test case 643594 into scout.
2. Go to http://localhost:5000/cust000/643594/sv/variants
3. Try searches such as in Chromosome position field 9:1,8500,001-2,800,0001

**Expected outcome**:
1. The same type of filtering as when using the ordinary filters.

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -1,21 +1,15 @@
 import logging
 import re
-from scout.constants import (
-    FUNDAMENTAL_CRITERIA,
-    PRIMARY_CRITERIA,
-    SECONDARY_CRITERIA,
-    TRUSTED_REVSTAT_LEVEL,
-)
+from scout.constants import FUNDAMENTAL_CRITERIA, PRIMARY_CRITERIA, SECONDARY_CRITERIA
 
 LOG = logging.getLogger(__name__)
 
-from scout.constants import SPIDEX_HUMAN, CLINSIG_MAP
-
+from scout.constants import (SPIDEX_HUMAN, CLINSIG_MAP)
 
 class QueryHandler(object):
-    def build_variant_query(
-        self, query=None, institute_id=None, category="snv", variant_type=["clinical"]
-    ):
+
+    def build_variant_query(self, query=None, institute_id=None, category='snv',
+                            variant_type=['clinical']):
         """Build a mongo query across multiple cases.
         Translate query options from a form into a complete mongo query dictionary.
 
@@ -30,7 +24,7 @@ class QueryHandler(object):
 
         Args:
             query(dict): A query dictionary for the database, from a query form.
-            category(str): 'snv', 'sv', 'str' 'cancer_sv' or 'cancer'
+            category(str): 'snv', 'sv', 'str' or 'cancer'
             variant_type(str): 'clinical' or 'research'
 
         Possible query dict keys:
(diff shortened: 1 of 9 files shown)


dnil:
Nice; interoperability is a Good Thing, and I personally like free text fields much better than divided boxes. Have you considered adding a conditional "chr" to the regexps at the start of the chromosome name, for pasting from builds that default to names like that?

Screenshot 2020-11-10 at 17 06 27

mhkc:
I can add the optional chr in the regex.

Btw I kept the separate chr, start and end fields since I dont know if some analyst prefer to have separate fields compared to one unified. Should we keep both input options for now?

dnil:
I agree it is best to leave the old way in until current users have had a chance to try the new. Perhaps they will all realise it is easier and at least as safe just to type than click dropdowns, but I am not convinced they will see it that way. 😊

mhkc:
Sounds good! What can I do to help with the merging of the PR?

dnil:
We (@northwestwitch perhaps for the js, otw I can be a fallback) review to give you some feedback, but perhaps you wanted to extend the regexp to catch "chr". Also to forestall a review, have a look at the lint, test and code factor output. I have a feeling you may disagree style-wise with the js lint, but just note that if so.

northwestwitch:
I can definitely give a look at this PR this week! :)

mhkc:
Ill update the regex tomorrow :)

northwestwitch:
I'm going to review this one!

scout/adapter/mongo/query.py
```diff
@@ -179,6 +180,34 @@ def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
                 else:  # sv
                     coordinate_query = [self.sv_coordinate_query(query)]
 
+            elif criterion == "chrom_pos" and query.get("chrom_pos"):  # filter by coordinates
```
northwestwitch, line 183: I've been testing the branch on a local instance and I see that every time a user inserts coordinates in the chrom_pos fields -> chromosome, start, and stop fields are also updated (not applying the eventual padding). If you apply the padding in the web page (at the same time you update these fields) then you don't have to touch this file at all, because you can perform the search using those values instead (only chromosome, start and stop). It would avoid a lot of additional code in a function that is already exploding (not your fault, it's already a monster function as it is)

scout/server/blueprints/variants/forms.py
```diff
@@ -105,6 +105,12 @@ class VariantFiltersForm(FlaskForm):
     load_filter = SubmitField(label="Load filter")
     delete_filter = SubmitField(label="Delete filter")
 
+    chrom_pos = StringField(
```
northwestwitch, line 108: 👍

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,12 +1,23 @@
 function populateCytobands(cytobands){
+  var chromPosPattern = new RegExp("^([0-9]{1,2}|[X|T|MT]{1,2})");
```
northwestwitch, line 2: Testing with coordinates I've noticed that the regex allows for searching for chromosomes ranging 23-99, which don't exist of course. For instance if you enter chromosome 99 the form is submitted and the chromosome on top is assigned to "9"

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,12 +1,23 @@
...
+    chromosome = chromPos.match(chromPosPattern)[0];
+  } else if (chrom==="" && chromPos!==""){
+    chromosome = chromPos.match(chromPosPattern)[0];
+  } else {
+    chromosome = chrom;
   }
-
-  var chrom_cytobands = cytobands[chrom]["cytobands"]; // chromosome-specific cytobands
+  
+  var chrom_cytobands = cytobands[chromosome]["cytobands"]; // chromosome-specific cytobands
```
northwestwitch, line 20: Updating cytobands according to the values of chrom_pos is nice, but it's not really necessary for the subsequent search, so you can leave it out if you want.

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -66,6 +78,27 @@ function validateForm(){
             return false;
         }
     }
+    // Validate Chromosome position form
+    //Expected format: <chr number>:<start>-<end>[+-]?<padding>
+    var chrom_pos = document.forms["filters_form"].elements["chrom_pos"].value.replaceAll(',', '')
+    const chromPosPattern = new RegExp("^([0-9]{1,2}|[X|T|MT]{1,2})(:([0-9]+)-([0-9]+)([+-]{1}[0-9]+)?)?$");
```
northwestwitch, line 84: As I mentioned before, the pattern could be improved by checking that chromosome range is [1-22] and X,Y,MT.
Also the suggestion from @dnil (allowing "chr" as prefix) could be implemented

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -66,6 +78,27 @@ function validateForm(){
...
+    if(chrom_pos) {
+        if (!chromPosPattern.test(chrom_pos)) {
+            alert("Invalid format of chromosome position, expected format <chr number>:<start>-<end>[+-]?<padding>");
+            return false;
+        }
+        var _, chr, start, end, sign, padding;
```
northwestwitch, line 90: I think it's safe to ignore codefactor here.

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -113,3 +147,53 @@ function enableDismiss(){
...
 }
+
+
+// Update chromosome position
+function updateCoordinateFields(element) {
+  const chrom = document.forms["filters_form"].elements["chrom"];
+  const chromPos = document.forms["filters_form"].elements["chrom_pos"];
+  const cytoStart = document.forms["filters_form"].elements["cytoband_start"];
+  const cytoEnd = document.forms["filters_form"].elements["cytoband_end"];
+  const chromPosPattern = "^([0-9]{1,2}|[X|T|MT]{1,2})(?::([0-9]+)?(?:-([0-9]+)?)?)?";
```
northwestwitch, line 158: Fix pattern, see above.

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -113,3 +147,53 @@ function enableDismiss(){
     btnElem.disabled = true;
   }
 }
+
+
+// Update chromosome position
```
northwestwitch, line 152: Perhaps elaborate a bit more on this comment?

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -113,3 +147,53 @@ function enableDismiss(){
...
+function updateCoordinateFields(element) {
```
northwestwitch, line 153: What does this function do? I think it's very useful if:

- if the user enters values in the chrom_pos:
  - The chromosome select is updated
  - start is updated (with eventual padding)
  - end is updated (with eventual padding)

What I don't think is useful (at least for the subsequent filtering):
- Update chrom_pos  if a user modifies chrom and cytobands, also because chrom_pos doesn't update if you modify chromosome, start and stop

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -256,6 +256,25 @@ <h6>Phenotype terms (HPO):</h6>
       {{ wtf.form_field(form.cytoband_end) }}
     </div>
   </div>
+  <div class="form-group">
```
northwestwitch, line 259: This would look much nicer code-wise if you use a grid system like this: https://getbootstrap.com/docs/4.0/layout/grid/#variable-width-content

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -373,6 +392,25 @@ <h6>Phenotype terms (HPO):</h6>
     </div>
   </div>
 </div>
+<div class="form-group">
```
northwestwitch, line 395: Same here

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -176,7 +176,17 @@
     var chromSel = document.getElementById("chrom");
     chromSel.addEventListener("change", function() {
       populateCytobands({{cytobands|safe}});
+      updateCoordinateFields(event.target)
```
northwestwitch, line 179: I've noticed that while the search field is present in the SNVs and SVs pages, it's missing in the search form of cancer variants and SV cancer variants. It would be very useful if you could add it there too

northwestwitch (changes requested):
Nice feature and job! I think it would be very much appreciated by the users!

I have some suggestion for improvement, as you can see in the comments. 
I think the biggest change that I'd do is not touching the adapter/mongo/query function at all. Instead , given a chrom_pos string provided in the form, I would update properly the chrom, start and stop coordinates so that they will continue to be used in the search as it is now.

northwestwitch:
Hi @mhkc let me know when you are ready for another review. Have a good day! :)

mhkc:
@northwestwitch I have updated the things you mentioned so feel free to have a second review

northwestwitch:
> @northwestwitch I have updated the things you mentioned so feel free to have a second review

Great, I'll review today!

northwestwitch:
Hi @mhkc since this PR is changing the code and the variant filtering I think that it would be best if you could move it to a branch started from our repo, so we see if it passes the tests and the effect on the coverage. Could you do it? Otherwise I could do it for you!

mhkc:
@northwestwitch if you could do it it would be great. I have a bunch of meetings today and would not have the time until tomorrow.

northwestwitch:
I've moved this PR to COPY_filter-gene-by-c, closing this one!

PR #2197, opened 2020-11-19
 COPY of IGV compatible filtering of variants

northwestwitch:
This PR adds a new input field when filtering variants. The filter takes an IGV position string (<chr>:<start>-<end>) as input which is a feature frequently requested by the analysts at Clinical Genomics Lund. The intent is to enable copy-pasting in position information from genome browsers such as IGV directly into scout.

**How to test**
1. Load test case 643594 into scout.
2. Go to http://localhost:5000/cust000/643594/sv/variants
3. Try searches such as in Chromosome position field 9:1,8500,001-2,800,0001

**Expected outcome**:
1. The same type of filtering as when using the ordinary filters.

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch (commented):
Hi @mhkc 

Thanks for the changes, it's definitely getting there:
- Filtering by position with chrom as "chr:n" now works
- Filtering by chromosome chr99 now doesn't work and raises a warning
- The start and stop coordinates change in real time when you add the +- padding (great job!) 💯 
- chrom_pos filter appears on the SV cancer page and filtering works (nice!) 🏅 

Needs fixing:
- You reverted scout/adapter/mongo/query.py to how it was before the PR but I realize that this file in your fork is very different from our file! 😮 One more reason to keep filtering using only chromosome, start and end coordinates I think! I'm sure you guys have your reasons to have modified how the query is built but we wouldn't like to introduce these changes in our repo. In order to merge this PR the file should be kept as we have it in our current master.

- SV cancer variants filtering using chrom_pos is working fine, but filtering of cancer SNV variants doesn't. You can test by loading the demo cancer sample with this command:
scout --demo load case scout/demo/cancer.load_config.yaml

I could try to fix the above things for you in this branch and see how it goes if you want!

northwestwitch:
I'd like to include this functionality in the next release. Today I have time to work on it so I'll come up with a query page fix and hopefully a PR that passes the tests!

northwestwitch:
I think this is ready for another test now. I'll install it on stage. Ideally should be tested with all types of varians types except strs.

@dnil could you test it a bit and see if's OK?
@mhkc does it still work with your version of the query.py?

dnil:
Ahem, something seems to have broken along the road @northwestwitch? I installed on stage, and neither of the two position filter options seems to be working atm? Could you have a look perhaps, and holler if it is still working for you and you want me to see where I went wrong!

northwestwitch:
> Ahem, something seems to have broken along the road @northwestwitch? I installed on stage, and neither of the two position filter options seems to be working atm? Could you have a look perhaps, and holler if it is still working for you and you want me to see where I went wrong!

Oh no, I'll check!

northwestwitch:
- [ ] replace 'let' and 'const' with 'var', hoping that will improve portability to browsers like IE
- [ ] Include the chromosome position search in a modal (to make it look nicer)
- [ ] Parse the value contained in the field in the controllers and set chrom and start and stop there if they don't have a value of their own
- [ ] Entering values in start and/or stop fields should erase content in Chromosome position

dnil:
Yes! Thats more like it - me like! The js is now working well for Safari as well, and the foldout makes the layout clean. How about some arrow icon in the button as well indicating a foldout? We don't have that everywhere, but it would be nice if we did. Did you want to fix parsing as well? Im not sure its really needed anymore? This works for most browsers, and anyway is hidden so that it is rather clear that it is a bonus?

I think my roomies will want an IGV button on there as well, but I can do that in another PR so we can get this merged.

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,12 +1,26 @@
...
+  var chromPos = document.forms["filters_form"].elements["chrom_pos"].value;
+  var chromosome = "";
+  console.log("Populate cytobands")
+  var matchedChrName = chromPos.match(chromPosPattern)
+  if(chrom==="" && chromPos===""){
     startElem.value = "";
     endElem.value = "";
     return //only reset cytoband select element
+  // Set the selected chromosome name
+  } else if (chrom !== "" && matchedChrName !== null){
```
dnil, line 13: I'm so glad we didn't write the entire app in js.. === && !==, !== && === and == && ==..

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -469,27 +493,35 @@ <h6>Phenotype terms (HPO):</h6>
...
           {{ wtf.form_field(form.cytoband_start) }}
         </div>
-        <div class="col-md-3 col-lg-3 col-xl-2">
+        <div class="col-2">
           {{ wtf.form_field(form.cytoband_end) }}
         </div>
-        <div class="col-md-1 col-lg-1 col-xl-5" id="empty_space"></div>
+      </div>
+      <div class="row">
+        <div class="col-2">
```
dnil, line 513: Any chance of having this on the row before for cancer, sort of next to the chromosome field again? The overall layout isn't quite as elegant as it was in the previous version anymore. 😊

dnil (approved):
Very well done! I haven't tested the latest on stage, and in particular not on IE or such, but which maybe we should before merging, but unless it totally crashes I think its quite ok if this bonus feature would not work on all browsers, especially when it is folded away.

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: Yes, I agree, I'll revert to how it was before!

northwestwitch:
Merging!

PR #2199, opened 2020-11-19
Bug fix --> uploading of cases where no phenotype is specified in config file

### LINKED ISSUE #2198: Bug when loading cases in current master
I must have introduced it with #2165. 
To reproduce upload the demo cancer sample with the command: `scout --demo load case scout/demo/cancer.load_config.yaml`

image

I'll fix it!

northwestwitch:
A scary bug in current master that would have made it impossible to upload any sample without pre-assigned phenotypes in the config file. Good that I've found it before deploying! :sweat_smile:

fix #2198 

**How to test**:
Try to upload the cancer demo file (doesn't have synopsis, cohort and phenotype info on the config file): `scout --demo load case scout/demo/cancer.load_config.yaml`

**Expected outcome**:
When using master branch the case won't upload
When using this branch the case will be loaded into Scout

**Review:**
- [x] code approved by MM
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Sentieon Manta calls lack Somaticscore - load anyway
 - ClinVar submissions crashing due to pinned variants that are not loaded
 - Point ExAC pLI score to new gnomad server address
+- Bug uploading cases missing phenotype terms in config file
 ### Changed
 - Make matching causative and managed variants foldable on case page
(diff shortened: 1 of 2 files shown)


northwestwitch:
Ready for review, an easy one!

moonso (approved):
Nice! I actually saw this today as well when I was playing around with scout. I think you should treat the phenotype_groups in the same way just below in the code.

northwestwitch:
> Nice! I actually saw this today as well when I was playing around with scout. I think you should treat the phenotype_groups in the same way just below in the code.

Good idea! Phenotype groups are not saved automatically using the config file (I think), but you never know!

PR #2208, opened 2020-11-25
Str filters

dnil:
This PR adds a functionality: STR variants filter, use new Stranger 0.6.1 fields
This PR fixes a bug: STR variant pagination broken again..

Screenshot 2020-11-25 at 10 55 26

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
I can review when you feel ready with this one!

northwestwitch:
@dnil it would be nice to have this one in the next Scout release, do you want help fixing and testing it?

dnil:
Any day now.. I have a couple more pressing things, then I'll get back to it. Speaking of things that would be good to have in release, how about #2259? At least lets flag down in the report to not claim refseq where there is actually ENSEMBL..

dnil:
Hm, not quite ready for review yet; I need a quiet day or two to work through esp handling old cases.

dnil:
Now with link out to source.
Screenshot 2021-02-01 at 21 10 14

Still a couple of more things to fix before this is ready for prime time.

dnil:
Sorry for adding several things at once again, but en passant things aside, at its core it is just two things in not too many lines:
- update of the keys read from the Stranger STR VCF annotation (HGNC id, source, disease, Sweden_STR_mean/std etc)
- addition of filters for the STR variantS page, enabled by the addition of approximate HGNC-ids from annotation

I dropped the idea of making this retroactive, with queries for repids. That would basically require new indexes, in addition to quite a bit of new translation functionality for repids to HGNC ids to cope with gene panels, and obviously some changes to the query functions on top. Instead I added some demo files, and tried not to break anything for viewing old cases. Additional ideas for how to indicate to the users that the current case is outdated on the HGNCid front are very welcome! I added a few hints in the tooltips, but more - ideally a way to selectively disable the search boxes - would not hurt.

Oh, Coveralls failed again while I was typing this essay. Will add another test too then!

northwestwitch:
> Sorry for adding several things at once again

It's good to introduce many things together if they're interconnected.. But 19 files changed might also be enough 😄 
I can review (probably tomorrow) but get ready because I might have one question or 10!

northwestwitch:
Tested locally and it works fine. Tried filtering, saving, resetting and using saving filters and works. Perhaps one thing that I would do is avoid filtering by gene panels whenever you land on strs page. I think if you decide strs you might not be interested in filtering them at all in the beginning

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Added a page about migrating potentially breaking changes to the documentation
 - markdown_include in development requirements file
+- STR variants filter
+- Display source, Z-score, inheritance pattern for STR annotations from Stranger (>0.6.1) if available
```
northwestwitch, line 12: Do we have a case with strs annotations from Stranger on stage that I could check?

scout/build/variant/variant.py
```diff
@@ -115,6 +115,11 @@ def build_variant(
         str_status = str, this indicates the severity of the expansion level
         str_normal_max = int, max number of repeats to call an STR variant normal
         str_pathologic_min = int, min number of repeats to call an STR variant pathologic
+        str_disease = str, Associated disease name
```
northwestwitch, line 118: 👍

scout/build/variant/variant.py
```diff
@@ -210,6 +215,25 @@ def build_variant(
...
+            float(variant["str_swegen_std"]) if variant["str_swegen_std"] else None
+        )
+
+    if "str_inheritance_mode" in variant:
+        variant_obj["str_inheritance_mode"] = variant["str_inheritance_mode"]
+
+    if "str_disease" in variant:
+        variant_obj["str_disease"] = variant["str_disease"]
+
+    if "str_source" in variant:
```
northwestwitch, line 234: Wasn't this one called "str_source_display" when you initialize it before, or is it another one?

dnil:
Yeah, thanks for the support - I really feel your way. The logic behind this is similar to why there are panels for SNVs as well. You may not want to deal with seeing a late onset neuromuscular disorder if you were looking for a present metabolic disorder with a kid, so hence filter on from the start. That said, there is considerable phenotypic overlap with many of these, so this way places a larger burden on designing the panels. But, one can always disable the list if that was intended.

northwestwitch (commented):
> one can always disable the list if that was intended.

Fair enough!

CHANGELOG.md
dnil: I don't think so; the demo case sort of covers it, but we should get that going with the MIP folks as well. The most important thing on stage is making sure old cases dont break - and decide if it is too confusing that the filter is there but not really working for the old ones.

scout/build/variant/variant.py
dnil: str_source is the dict, but yeah, I got the docstring wrong! 👍

scout/models/variant/variant.py
```diff
@@ -46,8 +46,16 @@
     dbsnp_id=str,
     # str variant
     str_ru=str,
+    str_display_ru=str,
```
northwestwitch, line 49: what does _ru stand for?

scout/models/variant/variant.py
```diff
@@ -46,8 +46,16 @@
...
     str_repid=str,
     str_ref=str,
+    str_disease=str,
+    str_inheritance_mode=str,  # STR disease mode of inheritance "AD", "XR", "AR", "-"
+    str_source=dict,  # STR source dict with keys {"display": str, "type": str ("PubMed", "GeneReviews"), "id": str}
```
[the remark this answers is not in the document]
northwestwitch: Ah no, sorry, this is correct

scout/parse/variant/variant.py
```diff
@@ -264,6 +295,11 @@ def parse_variant(
     for gene in parsed_variant["genes"]:
         hgnc_ids.add(gene["hgnc_id"])
 
+    # STR HGNCIds are annotated by Stranger
+    str_hgnc_id = variant.INFO.get("HGNCId")
+    if str_hgnc_id:
+        hgnc_ids.add(str_hgnc_id)
```
northwestwitch, line 301: 👍

scout/server/blueprints/variants/controllers.py
```diff
@@ -877,7 +884,7 @@ def dismiss_variant_list(store, institute_obj, case_obj, link_page, variants_lis
         if variant_obj is None:
             continue
         # create variant link
-        link = link = url_for(
```
northwestwitch, line 880: 👍

scout/models/variant/variant.py
dnil: Repeat Unit - those are not very self explanatory but same as in Stranger and ExHu.

scout/server/blueprints/variants/forms.py
```diff
@@ -153,9 +154,30 @@ class StrFiltersForm(FlaskForm):
...
+
+    hgnc_symbols = TagListField("HGNC Symbols/Ids (case sensitive)")
     gene_panels = SelectMultipleField(choices=[])
-    repids = TagListField()
+
+    chrom = SelectField(
+        "Chromosome", [validators.Optional()], choices=CHROMOSOME_OPTIONS, default=""
+    )
+
+    start = IntegerField("Start position", [validators.Optional()])
```
northwestwitch, line 165: You could skip adding many (if not all) of these fields if you make StrFiltersForm a subclass of VariantFiltersForm instead of FlaskForm. This way it can inherit:
- chrom
- start
- end
- cytoband_start
- cytoband_end
- All the filters block
- hide dismissed
-  export

scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
```diff
@@ -48,7 +48,7 @@
         <div id="collapseFilters" class="card-body panel-collapse collapse {{ 'show' if expand_search is sameas true }}">
           <div class="card-body">{{ sv_filters(form, institute, case) }}</div>
         </div>
-        {{ filter_form_footer(form, result_size, total_variants) }}
+        {{ filter_form_footer(form, result_size, total_variants, more_variants, page, variants|length) }}
```
northwestwitch, line 51: Was this broken?

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -117,53 +128,43 @@
       </tbody>
     </table>
   </div>
-  <br>
-  {{ footer() }}
+</div> <!-- end of class="container-float"> -->
+<div class="container-fluid">
+  {{ pagination_footer(more_variants, page) }}
```
northwestwitch, line 133: So much better!

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -176,7 +177,57 @@
...
+  target="_blank"
+  data-toggle="popover"
+  data-html="true"
+  data-trigger="hover"
+  data-content="
+  <div class='text-left'>
+    {% if variant.str_disease %}
+      <div><strong>Disease:</strong> {{ variant.str_disease }} ({{ variant.str_inheritance_mode }})</div>
+    {% else %}
+      <div>Disease information not provided with locus. Consider updating to a new version of Stranger.</div>
```
northwestwitch, line 224: Doesn't break the demo case at least:

image

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -176,7 +177,57 @@
...
+  href="{{ variant.str_source_link }}"
+  class="link text-black"
+  target="_blank"
+  data-toggle="popover"
+  data-html="true"
+  data-trigger="hover"
+  data-content="
+  <div class='text-left'>
+    {% if variant.str_disease %}
+      <div><strong>Disease:</strong> {{ variant.str_disease }} ({{ variant.str_inheritance_mode }})</div>
```
northwestwitch, line 222: Really nice!

scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
northwestwitch: Ah ok now I get it, you changed it in this PR

scout/server/blueprints/variants/views.py
```diff
@@ -208,18 +235,24 @@ def str_variants(institute_id, case_name):
             ("position", pymongo.ASCENDING),
         ]
     )
-    result_size = store.count_variants(case_obj["_id"], {}, None, category)
+
+    result_size = store.count_variants(case_obj["_id"], query, None, category)
+
     data = controllers.str_variants(
-        store, institute_obj, case_obj, variants_query, result_size, page
+        store, institute_obj, case_obj, variants_query, result_size, page=page
```
northwestwitch, line 242: This change is a bit weird, i'd remove =page

scout/server/links.py
```diff
@@ -369,10 +369,33 @@ def get_variant_links(variant_obj, build=None):
...
+        return None
+
+    if source["type"] is None:
+        url_template = "{}"
+    if source["type"] == "GeneReviews":
+        url_template = "https://www.ncbi.nlm.nih.gov/books/{}/"
+    if source["type"] == "PubMed":
+        url_template = "https://pubmed.ncbi.nlm.nih.gov/{}/"
+
+    pp("returns ok!")
```
northwestwitch, line 395:
```suggestion
```

northwestwitch (approved):
Really nice work! It will be super appreciated! I have some small comments but nothing major, so I'll approve! 🏅

scout/server/blueprints/variants/forms.py
dnil: Hm, it has some extra stuff that I would rather not include but I could make another least common denominator class and have them all inherit again.

scout/server/blueprints/variants/views.py
dnil: It is named, but sure, it is also in the right order so will work anyway. It is presumably a leftover from testing something else.. 😄

scout/server/blueprints/variants/templates/variants/str-variants.html
dnil: I suppose that will have to be the sign to the users for now that filters are not working for this case. As soon as it is updated in MIP we can also let ppl know from what version/date etc.

scout/server/blueprints/variants/forms.py
northwestwitch: but then you just don't include those in the html page and they won't be a problem

northwestwitch: I think

dnil: Meh, I give in - changing and going for another LCD will require loads of extra testing. I agree and will just inherit - those extra fields should not hurt anyone, and ideally unique ones should only be on the child forms eventually.

northwestwitch: One easy test that you could do is checking the request form after it arrives to the view, I'm pretty sure that it won't contain the extra keys if you don't include them in the html page

Issue #2166, opened 2020-10-27
STR pagination broken on master..


dnil:
I think a new positional argument was introduced for the variants controller recently, but not on str variants, which regrettably does not formally inherit it? Just a reminder - will fix in STR filters update.

northwestwitch:
Is it also on prod?

dnil:
Nope, should be fine; no rush - we should just have the fix in before releasing/deploying current master. I just noticed during testing of said other fix with str filters.

northwestwitch:
Hi @dnil have you started working on this? Otherwise I can take it.

dnil:
Yeah, I think I solved it as part of the STR filters PR. Two other prio tasks in queue, then I'll get back to this. End of the week approx..

dnil:
Fixed by #2209!

PR #2209, opened 2020-11-25
Fixed missing str vars from browser page

### LINKED ISSUE #2207: STRs variants don't show up in current master
Even if they are loaded and in database

northwestwitch:
Bug fix #2207 

**How to test**:
1. On a local instance of scout, load the demo case

**Expected outcome**:
1. Strs should show in their respective page

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -203,15 +203,16 @@ def get_manual_assessments(variant_obj):
     return assessments
 
 
-def str_variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50):
+def str_variants(
+    store, institute_obj, case_obj, variants_query, variant_count, page=1, per_page=50
+):
     """Pre-process list of STR variants."""
 
     # Nothing unique to STRs on this level. Inheritance? yep, you will want it.
 
     # case bam_files for quick access to alignment view.
     case_append_alignments(case_obj)
-
-    return variants(store, institute_obj, case_obj, variants_query, page, per_page)
+    return variants(store, institute_obj, case_obj, variants_query, variant_count, page, per_page)
 
 
 def parse_variant(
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/variants/views.py
```diff
@@ -208,8 +208,10 @@ def str_variants(institute_id, case_name):
             ("position", pymongo.ASCENDING),
         ]
     )
-
-    data = controllers.str_variants(store, institute_obj, case_obj, variants_query, page)
+    result_size = store.count_variants(case_obj["_id"], {}, None, category)
```
northwestwitch, line 211: To make this compliant with the str filters, just substitute {} with the request.form.data, I guess

scout/server/blueprints/variants/controllers.py
```diff
@@ -203,15 +203,16 @@ def get_manual_assessments(variant_obj):
     return assessments
 
 
-def str_variants(store, institute_obj, case_obj, variants_query, page=1, per_page=50):
+def str_variants(
+    store, institute_obj, case_obj, variants_query, variant_count, page=1, per_page=50
```
dnil, line 207: :+1: Yep, exactly the same as #2166.

scout/server/blueprints/variants/views.py
dnil, line 211: Yeah, this should basically the only collision, the rest will be identical, so no worries.

dnil (approved):

dnil:
I'll just merge this! Objections? 5.. 4..

scout/server/blueprints/variants/controllers.py
northwestwitch: I see. The "pagination" in the title was a bit of an understatement 😆

PR #2214, opened 2020-11-26
SMN None new - fix #2140

### LINKED ISSUE #2140: Parsing None values from SMN file
Classic python: None and False values are mixed up somewhere early in parsing SMN files for load / display:
`test.smn.tsv`
Screenshot 2020-10-02 at 09 25 35

SMN view
Screenshot 2020-10-02 at 09 25 41

DB
Screenshot 2020-10-02 at 09 27 51

Perhaps it can simply not be a boolean - a string would be safer.. 😊

dnil:
This PR fixes a bug.

When SMNCaller is uncertain about carriership or affected status, it calls None rather than True or False. This was previously parsed as False for python-esque reasons. 

Before:
Screenshot 2020-11-26 at 11 39 50

After:
Screenshot 2020-11-26 at 16 52 19

**How to test**:
1. The demo test case now has a None carrier status.

**Expected outcome**:

**Review:**
- [x] code approved by


northwestwitch:
I'll review now!

northwestwitch:
Tested locally. Master:

image

This branch:
image

scout/utils/convert.py
```diff
@@ -17,6 +17,14 @@ def isint(x):
         return a == b
 
 
+def make_bool_pass_none(string):
+    if string is None:
```
northwestwitch, line 21: It works but perhaps if you are sure that the returned value if "None" or else then you could just write:
if string is None or str == 'None:

northwestwitch (approved):
👍

scout/utils/convert.py
dnil: Now that you mention it, the str()-cast is None safe, so the first isn't needed at all! Thanks!

PR #2221, opened 2020-11-30
Delete variants by analysis type

northwestwitch:
- Allow to delete variants for cases with one or more analysis type in "wes", "wgs" and "panel" with "wgs" as a default.
- Print also case status and if it's research in the log resuly

**How to test**:
1. Have cancer case and demo case on a local instance of scout
1. See the difference in results when you remove variants with the 2 commands
  - `scout --loglevel WARNING --demo delete variants -u clark.kent@mail.com --analysis-type wes  --analysis-type panel --status inactive --rank-threshold 5 --dry-run`
  - `scout --loglevel WARNING --demo delete variants -u clark.kent@mail.com --analysis-type wes   --status inactive --rank-threshold 5 --dry-run`

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


scout/commands/delete/delete_command.py
```diff
@@ -8,7 +8,7 @@
 LOG = logging.getLogger(__name__)
 
 BYTES_IN_ONE_GIGABYTE = 1073741824  # (1024*1024*1024)
-DELETE_VARIANTS_HEADER = "Case n.\tNcases\tInstitute\tCase name\tCase ID\tCase track\tAnalysis date\tTotal variants\tRemoved variants"
+DELETE_VARIANTS_HEADER = "Case n.\tNcases\tInstitute\tCase name\tCase ID\tCase track\tAnalysis date\tStatus\tResearch\tTotal variants\tRemoved variants"
```
dnil, line 11: See, thats why the join is so good.. 😄

scout/commands/delete/delete_command.py
```diff
@@ -22,6 +22,13 @@
     help="Restrict to cases with specified status",
 )
 @click.option("-older-than", type=click.INT, default=0, help="Older than (months)")
+@click.option(
+    "-analysis-type",
```
dnil, line 26: Shouldn't they be "--" btw?

dnil (approved):
👍

scout/commands/delete/delete_command.py
northwestwitch: it works.. But I'll add it!

scout/commands/delete/delete_command.py
northwestwitch: Right I'll fix also the header like that!

northwestwitch:
image

image

northwestwitch:
@dnil do we want to default the case status to any status?

northwestwitch:
Merging in 3, 2, ...

northwestwitch:
Starting the simulation on hasta now!

PR #2226, opened 2020-12-01
Fixed error in delete variants counting when removing vars

northwestwitch:
There is currently a bug in printing the deleted variants for each case. This PR is fixing it.

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil (approved):
:+1:

PR #2232, opened 2020-12-02
Relax three letter conversion

dnil:
This PR fixes a bug.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/constants/__init__.py
+++ b/scout/constants/__init__.py
@@ -173,4 +173,6 @@ AMINO_ACID_RESIDUE_3_TO_1 = {
     "Trp": "W",
     "Tyr": "Y",
     "Val": "V",
+    "Ter": "*",
+    "del": "del",
 }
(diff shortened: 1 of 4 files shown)


dnil:
I guess the alternative is explicitly casting each group element as str - or writing a slightly more general conversion function. 😅 This ought to do the trick, but it is possible the end point dbs may support a few more kinds of variations than plain missense, stop or del,

scout/constants/__init__.py
```diff
@@ -173,4 +173,6 @@
     "Trp": "W",
     "Tyr": "Y",
     "Val": "V",
+    "Ter": "*",
```
northwestwitch, line 176: What's Ter, the third letter?

northwestwitch (approved):
Perhaps it can be directly converted into a hotfix format?

scout/constants/__init__.py
dnil: Stop (Ter-minationcodon)

dnil:
Sure, but we don't really have anything else in master right now, do we? So may as well just patch release in good order.

northwestwitch:
> Sure, but we don't really have anything else in master right now, do we? So may as well just patch release in good order.

like bugs?

PR #2234, opened 2020-12-04
Exclude one or more variant categories in variant delete command

northwestwitch:
**How to test**:
1. Run the delete variants command with and without the exclude-ctg param and see the difference

**Expected outcome**:
1. If you exclude a variant category from deletion all variants of that category will be left, regardless of their rank score

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -41,13 +41,16 @@ class QueryHandler(object):
             case_query["status"] = {"$in": list(status)}
         return case_query
 
-    def delete_variants_query(self, case_id, variants_to_keep=[], min_rank_threshold=None) -> dict:
+    def delete_variants_query(
+        self, case_id, variants_to_keep=[], min_rank_threshold=None, exclude_ctg=[]
+    ) -> dict:
         """Build a query to delete variants from a case
 
         Args:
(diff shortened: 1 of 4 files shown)


northwestwitch:
Test without the `exclude-ctg` param:
image

image

Test excluding strs from delete command:
image

image

northwestwitch:
Ready for review

scout/adapter/mongo/query.py
```diff
@@ -41,13 +41,16 @@ def build_case_query(
             case_query["status"] = {"$in": list(status)}
         return case_query
 
-    def delete_variants_query(self, case_id, variants_to_keep=[], min_rank_threshold=None) -> dict:
+    def delete_variants_query(
+        self, case_id, variants_to_keep=[], min_rank_threshold=None, exclude_ctg=[]
```
dnil, line 45: This is nitpicking, but could it be perhaps keep_category - ctg makes me think config.. 😊

dnil (approved):

scout/adapter/mongo/query.py
northwestwitch: Sure, better to be clear

PR #2238, opened 2020-12-07
Fixed link to Marrvel

### LINKED ISSUE #2220: MARRVEL gene link out broken
The MARRVEL link out is currently broken. Presumably they switched from a POST to a GET API.

Screenshot 2020-11-30 at 16 26 02

northwestwitch:
Marrvel has recently (last summer I think) updated the website to v2 (beta).
I've tried to update the input fields to query Marrvel v2 but without success. The thing is I don't understand the name of the new input fields, check here: https://github.com/LiuzLab/MARRVEL2/blob/master/src/app/components/search-box/search-box.component.html

But I managed to fix the link to the v1 of the portal, so at least the links are not broken. Fix #2220

**How to test**:
1. Install this branch and go to a variant, check that the Marrvel link works.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/variant.html
+++ b/scout/server/blueprints/variant/templates/variant/variant.html
@@ -634,16 +634,7 @@
             <a href="{{ gene.oncokb_link }}" class="btn btn-outline-secondary" target="_blank">OncoKB</a>
             <a href="{{ gene.civic_link }}" class="btn btn-outline-secondary" target="_blank">CIViC</a>
           {% endif %}
-          <form action="http://v1.marrvel.org/search" method="POST" target="_blank" style="display:inline;">
-            <button type="submit" class="btn btn-outline-secondary">MARRVEL</button>
-            <input type="hidden" name="variantGeneId" value="{{ variant.chromosome }}:{{ variant.position }} {{ variant.reference }}>{{ variant.alternative }}">
-            <input type="hidden" name="inGeneSymbol" value="{{ gene.common.hgnc_symbol if gene.common }}">
-            <input type="hidden" name="OMIMdbCheck" value="on">
-            <input type="hidden" name="ExACdbCheck" value="on">
-            <input type="hidden" name="Geno2MPdbCheck" value="on">
-            <input type="hidden" name="DGVdbCheck" value="on">
-            <input type="hidden" name="DECIPHERdbCheck" value="on">
-          </form>
+            <a href="http://marrvel.org/human/variant/{{ variant.chromosome }}:{{ variant.position }} {{ variant.reference }}>{{variant.alternative}}" class="btn btn-outline-secondary" target="_blank">MARRVEL</a>
       </div>
     {% endfor %}
   </div>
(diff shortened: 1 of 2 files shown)


northwestwitch:
@dnil I've asked them how to fix our code to query MARRVEL 2: https://github.com/LiuzLab/MARRVEL2/issues/43

dnil:
The v1 link now works just fine, thanks!

The only thing I see on the v2 interface is the NCBI gene ID, rather than HGNC. That is what we (Scout and HGNC genes) call "entrez_id"), and should work well I guess? See if it doesn't work if you just url like this:
`"http://marrvel.org/human/gene/{}".format(entrez_id)` They also allow linking directly to the variant, but I haven't really tried that for new variants? Anything good or just not found?

dnil (approved):
I'll approve this: it is working. 
However if you wish you could remove this and go to the v2 beta. That is likely to be more future safe, also looks a lot simpler with an easy REST GET url rather than this POST form.

dnil:
Yes, the variant version (that we did use before) looks very good. So just GET link to 
`http://marrvel.org/human/variant/6:99365567T%3EC` (chr, pos, ref, alt) instead?

northwestwitch:
> The v1 link now works just fine, thanks!
> 
> The only thing I see on the v2 interface is the NCBI gene ID, rather than HGNC. That is what we (Scout and HGNC genes) call "entrez_id"), and should work well I guess? See if it doesn't work if you just url like this:
> `"http://marrvel.org/human/gene/{}".format(entrez_id)` They also allow linking directly to the variant, but I haven't really tried that for new variants? Anything good or just not found?

I've tried to compose the link this way. The problem is entrez_id is present only in the gene documents and not in the gene annotations inside the variant documents of the database. So, in order to compose these links we'd need an additional passage to retrieve entrez ids, add it to the annotated variant and then compose the link. A query would be easier at this point I think. Let's see what they reply!

northwestwitch:
> Yes, the variant version (that we did use before) looks very good. So just GET link to
> `http://marrvel.org/human/variant/6:99365567T%3EC` (chr, pos, ref, alt) instead?

Easier this way, I'll go for it!

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -634,16 +634,7 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
...
-            <button type="submit" class="btn btn-outline-secondary">MARRVEL</button>
-            <input type="hidden" name="variantGeneId" value="{{ variant.chromosome }}:{{ variant.position }} {{ variant.reference }}>{{ variant.alternative }}">
-            <input type="hidden" name="inGeneSymbol" value="{{ gene.common.hgnc_symbol if gene.common }}">
-            <input type="hidden" name="OMIMdbCheck" value="on">
-            <input type="hidden" name="ExACdbCheck" value="on">
-            <input type="hidden" name="Geno2MPdbCheck" value="on">
-            <input type="hidden" name="DGVdbCheck" value="on">
-            <input type="hidden" name="DECIPHERdbCheck" value="on">
-          </form>
+            <a href="http://marrvel.org/human/variant/{{ variant.chromosome }}:{{ variant.position }} {{ variant.reference }}>{{variant.alternative}}" class="btn btn-outline-secondary" target="_blank">MARRVEL</a>
```
northwestwitch, line 637: Tested and it works:

image

PR #2239, opened 2020-12-07
Allow users to add any numerical value in case page, OMIM genes panel

### LINKED ISSUE #2225: Cannot add OMIM gene ID to cases
I have been adding OMIM phenootype and gene IDs' prepparing for a Clinvar submission. During this process i could not add the gene ID in the Case page and just get an error message, see below. 

image

I tried to repeat this for both families (11 and 213) but also for a newer family (20193) but it does not work for any so far.

image

Is there a general error regarding this function?

northwestwitch:
Fix #2225, read discussion there to have an idea of what this is fixing

**How to test**:
1. Install this branch locally and add/remove OMIM gene IDs in the OMIM Genes panel on case page:
image
1. Make sure that the functionality of adding OMIM phenotypes is not disrupted.

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 ### Fixed
 - Marrvel link
+- OMIM gene fields accept any custom number as OMIM gene
 ### Changed
 
 ## [4.27]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 ### Fixed
 - Marrvel link
-- OMIM gene fields accept any custom number as OMIM gene
+- OMIM gene field accept any custom number as OMIM gene
 ### Changed
 
 ## [4.27]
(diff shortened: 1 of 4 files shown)


northwestwitch:
Ready for review

mikaell:
Hi, 
I started looking at this PR. Seem alright so far. But I'm unsure how to test phenotypes?

> 2. Make sure that the functionality of adding OMIM phenotypes is not disrupted.

northwestwitch:
> Hi,
> I started looking at this PR. Seem alright so far. But I'm unsure how to test phenotypes?
> 
> >

You might not have OMIM terms present in your local database, so perhaps you could test it on stage. 
- You should check that adding/removing OMIM terms using this panel (case page) works:

image

- You should check that if you add/remove a numerical ID using the panel on the side (OMIM Genes) works. Whenever you add a valid OMIM gene, for instance ID **150320** from this disease (check last column on the right)
image

Then after you add it the numerical ID is shown and if you click on it, it will open the page of the 150320 gene on OMIM

mikaell (approved):
Looks good!

northwestwitch:
> Looks good!

Thanks @mikaell. I don't know if the users want to completely remove the OMIM genes now. Let's see, otherwise we'll merge

northwestwitch:
I'm going to merge this, if we need to remove that panel we can do it in a second moment. At least this bug if fixed

PR #2241, opened 2020-12-08
Hide and lessen dependence on variant_rank.

dnil:
This PR is for keeping book of potential changes to shrink and harmonise indices.
May or may not become mergeable.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Is this ready for review?

dnil:
No, it is 5 mins of fixes intended as food for thought! I'll context switch back to it eventually, but feel very free if you want to elaborate on it before that! I would likely revert part of the report but hunt more for the variant-id and rank queries.

scout/adapter/mongo/variant.py
```diff
@@ -191,8 +191,6 @@ def variants(
             case_id, query=query, variant_ids=variant_ids, category=category
         )
         sorting = []
-        if sort_key == "variant_rank":
-            sorting = [("variant_rank", pymongo.ASCENDING)]
```
northwestwitch, line 195: Perhaps pymongo.DESCENDING instead, as we want the ones with higher numbers before?

northwestwitch:
What about the variant_id variant index? Should we remove it as well? Is Sept 14 the last date it was used?

image

northwestwitch:
> What about the variant_id variant index? Should we remove it as well? Used 92 times since September.

Couldn't even find it in the constant index models. Did we decide to remove it before and removed it just from the constants instead?

dnil:
I think I'm going to close this not to have to look at it. We will revisit when its time for a major update, or server move, and we have a good opportunity to re-index. Part of this is already in event indices and in the cancer variantS view with no rank. When merging that part to RD it will be rather evident what to do.

PR #2249, opened 2020-12-09
Store variants count in case document

### LINKED ISSUE #2247: Speed up variants page loading
By saving number of variants of any type in the case document. 

- Whenever a case is uploaded
- Whenever a case is re-uploaded
- Whenever variants are removed
- Whenever a case is opened and doesn't have any of these numbers saved (old cases)
?

This will allow to skip counting the number of variants of any type when one lands on the variants page with a GET request. 

Saving number of variants by type in the case session will become redundant and should be removed.

northwestwitch:
Speed up slightly queries by storing variants stats in the case document. (fixes #2247 only partially)

- when a case is uploaded the case document doesn't have variants_stats key in the case document (also all existent cases)
- Variants stats gets saved when a user clicks on variants pages (of any type) and gets used as a reference for total number of variants of any category
- Variants stats is updated if variants get removed for a case using the command line
- Variants stats are set to none every time that a case is re-uploaded

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


mikaell:
> This is ready for review. I don't know honestly how much faster will be with this in place. A bit perhaps, but I think the main reason why the variants pages are much slower now is the count_documents, and this PR is not touching that.

I'll have a look.

scout/server/blueprints/variants/controllers.py
```diff
@@ -494,34 +491,6 @@ def cancer_variants(store, institute_id, case_name, variants_query, variant_coun
     return data
 
 
-def variant_count_session(store, institute_id, case_id, var_type, var_category):
```
mikaell, line 497: 👍

scout/server/blueprints/variants/views.py
```diff
@@ -53,6 +52,9 @@ def variants(institute_id, case_name):
     category = "snv"
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
     variant_type = request.args.get("variant_type", "clinical")
+    variants_stats = case_obj.get("variants_stats") or store.case_variants_count(
+        case_obj["_id"], institute_id, True
+    )
```
mikaell, line 57: Maybe refactor into a function call? 


```
variants_stats = get_variant_count(case_obj, institute_id)

```

scout/server/blueprints/variants/views.py
```diff
@@ -232,9 +233,12 @@ def sv_variants(institute_id, case_name):
     page = int(request.form.get("page", 1))
     variant_type = request.args.get("variant_type", "clinical")
     category = "sv"
-
     # Define case and institute objects
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
+    variants_stats = case_obj.get("variants_stats") or store.case_variants_count(
```
mikaell, line 238: >Maybe refactor into a function call?

The pattern seems to appear twice -a separate function.

```
variants_stats = get_variant_count(case_obj, institute_id)
```

scout/server/blueprints/variants/views.py
```diff
@@ -288,9 +291,11 @@ def sv_variants(institute_id, case_name):
 def cancer_variants(institute_id, case_name):
     """Show cancer variants overview."""
     category = "cancer"
-
+    variant_type = request.args.get("variant_type", "clinical")
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
-
+    variants_stats = case_obj.get("variants_stats") or store.case_variants_count(
+        case_obj["_id"], institute_id, True
+    )
```
mikaell, line 298: 3

scout/server/blueprints/variants/views.py
```diff
@@ -376,9 +382,12 @@ def cancer_sv_variants(institute_id, case_name):
     page = int(request.form.get("page", 1))
     variant_type = request.args.get("variant_type", "clinical")
     category = "cancer_sv"
-
     # Define case and institute objects
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
+    variants_stats = case_obj.get("variants_stats") or store.case_variants_count(
+        case_obj["_id"], institute_id, True
+    )
+
```
mikaell, line 390: 4

🙂

mikaell (changes requested):
Looks good. But some minor refactoring should be done.

scout/server/blueprints/variants/views.py
northwestwitch: Good idea, I'll fix as you suggested, thanks for the review!

northwestwitch: Done, I have fixed store.case_variants_count so it does everything in one place. I'm goin to install the branch on stage and give it a try there!

northwestwitch:
It's on stage now, and looks like it's working fine

mikaell (approved):
Looks good!

PR #2251, opened 2020-12-10
Bugfix: ' vs " in Test

mikaell:
This PR adds a functionality or fixes a bug. `'` vs `"` in string assert on error message caused an error. At least locally.

**How to test**:

`$ pytest tests/commands/load/test_load_cnv_report_cmd.py`

**Expected outcome**:
image

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/tests/commands/load/test_load_cnv_report_cmd.py
+++ b/tests/commands/load/test_load_cnv_report_cmd.py
@@ -35,7 +35,5 @@ def test_invalid_path_load_cnv_report(mock_app, case_obj):
         cli,
         ["load", "cnv-report", case_obj["_id"], "invalid-path", "-u"],
     )
-    error_string1 = "Path 'invalid-path' does not exist."
-    error_string2 = 'Path "invalid-path" does not exist.'
-    assert error_string1 or error_string2 in result.output
+    assert 'does not exist' in result.output
     assert result.exit_code == 2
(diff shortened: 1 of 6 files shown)


northwestwitch:
It's not really a bug I think, but might be due to an oldish Click version you have installed on your local env. For instance tests pass on my env, with click==7.1.2

mikaell:
> It's not really a bug I think, but might be due to an oldish Click version you have installed on your local env. For instance tests pass on my env, with click==7.1.2

But this failed on Github. So strange.

tests/commands/load/test_load_cnv_report_cmd.py
```diff
@@ -35,6 +35,7 @@ def test_invalid_path_load_cnv_report(mock_app, case_obj):
         cli,
         ["load", "cnv-report", case_obj["_id"], "invalid-path", "-u"],
     )
-
-    assert "Path 'invalid-path' does not exist." in result.output
+    error_string1 = "Path 'invalid-path' does not exist."
```
northwestwitch, line 38: What about omitting the part of the statement with the quotes and leave just `does not exist`. But I'm afraid that more tests will fail for you after this because I remember clearly that some time ago when we unblocked the Click dependency we modified many tests to be up to date with a more recent version of Click, so I think the easiest would be for you to uninstall and reinstall Click, or just install Scout on a clean environment

northwestwitch (changes requested):

tests/commands/load/test_load_cnv_report_cmd.py
mikaell: Fixed!

northwestwitch:
This might be not your local settings after all, see #2252

northwestwitch (approved):
👍

PR #2257, opened 2020-12-10
Fix #2108 - VAF in cancer case general report

### LINKED ISSUE #2108: Scout report improvements for Cancer view
Besides being able to know which filter was used to generate the report, I'd like to also be able to see a table of variants in the report. For rare diseases this doesn't matter much, right? But for cancer view, it'd be great so the report contains the following information:

- [x] Discard inheritance model reporting
- [x] Discard GQ (?)
- [x] Add AF of reported variant
- [ ] Squash variant table into a more compact form. e.g. we can remove allele depth and add AD to reference and alternative
image

- [x] It still says "Affected": Yes/No even for cancer variants. I'd like to have it as Tumor or Normal. Or just use "Phenotype", Daniel: possibly with another name
- [x] Discard `Single sample family. No pedigree available.` for cancer samples. There is no pedigree.
- [ ] Discard 1000G, always show gnomad? pop_freq_max?
Daniel: Population frequencies are shown as annotated (and parsed on import). We can certainly ditch or deprecate 1000G from parsing, but if it is not annotated it will also not be shown. This is a general issue for Scout really, that could so slowly perhaps be resolved by only showing GnomAD assuming that essentially all cases should have it.
- [ ] Add bait set name to the report instead of "PANEL". HFA: Could this be a value in Scout config? Also ultimately the information about panel according to https://clincial-genomics-hybrid-capture.readthedocs.io/en/latest/ in a nice format as BALSAMIC report.
Daniel: this has been extensively discussed in other issues, and conforming to Scout panels will do much good for the overall experience. But if you want not only the gene panel for use with Scout display filtering, but also a capture panel linked, add it as a key to the load config and we'll link to it from both the case page and show it on the case report. Compare e.g. the rank model links that we have now.
- [x] Discard ancestry if empty (i.e. N/A)
- [x] Discard parenthood if empty (i.e. N/A)

What do you think? @vwirta any more items I missed?

dnil:

Fix the suggestions for the cancer general report from #2108. A few more general points remain, but this should cover most of the things already loaded. While a couple of the removed entries are a little Panel / BALSAMIC specific, we can add them back later or differentiate when we get more whole genome cancer cases and functionality.
 
- [x] Discard inheritance model reporting
- [x] Discard GQ (?)
- [x] Add AF of reported variant
- [x] It still says "Affected": Yes/No even for cancer variants. I'd like to have it as Tumor or Normal. Or just use "Phenotype", Daniel: possibly with another name
- [x] Discard `Single sample family. No pedigree available.` for cancer samples. There is no pedigree.
- [x] Discard ancestry if empty (i.e. N/A)
- [x] Discard parenthood if empty (i.e. N/A)

I note that both the reports as well as controllers and views for the reports contain some code duplication. Fixing without working local report pdf generation to ensure nothing goes wrong with weasyprinting seems difficult. Lets do that in a later PR! 😊 

Screenshot 2020-12-10 at 23 05 40

Screenshot 2020-12-10 at 22 47 27

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by


northwestwitch:
If this one is ready I can review!

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -125,22 +133,32 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
...
-	                    <span class="badge badge-pill badge-danger" title="Affected">YES</span>
-	                  {% else %}
-	                    <span class="badge badge-pill badge-secondary" title="Not affected">NO</span>
+                      {% if cancer %}
+                        <span class="badge badge-pill badge-danger" title="{{ ind.phenotype_human }}">{{ ind.phenotype_human }}</span>
+                      {% else %}
+                        <span class="badge badge-pill badge-danger" title="Affected">YES</span>
+                      {% endif %}
+                  {% else %}
+                      {% if cancer %}
```
northwestwitch, line 142: Here also, I suggest to treat RD and cancer in the same way by using ind.phenotype_human

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -389,6 +389,16 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 </div>
 {% endmacro %}
 
+{% macro variant_content(variant, index) %}
```
northwestwitch, line 392: Very nice! 👍

northwestwitch (approved):
This is great @dnil!

Perhaps I would have used another variant, and not TP53 that has coverage 0 in the demo data to show the example, but it works well. Perhaps it was to show that N/A is printed when variant is not covered?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [x.x.x]
 ### Added
+- Add VAF to cancer case general report, and remove some of its unused fields
```
northwestwitch, line 9: I've just realized that the current changelog contains two recent features in the wrong position:

- Chromograph integration for displaying PNGs in case-page
- Variants filter compatible with genome browser location strings

Could you move them up here instead? I guess that is what happens if you keep a PR open for quite some time! 😄

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [x.x.x]
 ### Added
+- Chromograph integration for displaying PNGs in case-page
 - Add VAF to cancer case general report, and remove some of its unused fields
+- Variants filter compatible with genome browser location strings
```
northwestwitch, line 11: Thanks!

dnil:
I'll see if I can't add a suitable test as well..

dnil:
Also tested on stage:
Screenshot 2020-12-14 at 15 23 11

Merging..

PR #2278, opened 2020-12-23
Add canonical transcript and protein change of canonical transcript to variants export excel file

### LINKED ISSUE #2193: Export variants - transcript and protein change
As a researcher, the customer would like to find transcript names and protein changes on exported variant lists.

northwestwitch:
fix #2193 

Introduce 2 more columns to the exported variant excel sheet: `Canonical_transcript` and `Canonical_transcript_protein_change`

**How to test**:
1. Install this branch and export variants from the variants page with or without filtering.
2. Note that two additional fields are present in the excel sheet: `Canonical_transcript` and `Canonical_transcript_protein_change`

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Ready for revie

scout/server/blueprints/variants/controllers.py
```diff
@@ -434,20 +436,23 @@ def variant_export_genes_info(store, gene_list):
...
         # gather HGVS info from gene transcripts
+
         transcripts_list = gene_obj.get("transcripts")
         for transcript_obj in transcripts_list:
-            if (
-                transcript_obj.get("is_canonical") is not None
-                and transcript_obj.get("is_canonical") is True
-            ):
-                hgvs_nucleotide = str(transcript_obj.get("coding_sequence_name"))
+            if transcript_obj["transcript_id"] == gene_obj.get("canonical_transcript"):
```
dnil, line 448: `<rant>`
We should take a couple of h and go over all of these at some point.
- `canonical_transcript` is annotated by VEP, and corresponds to the Canonical transcript status in ENSEMBL: typically the longes transcript around. In many cases this overlaps with, but can absolutely be distinct from primary_transcript - is also typically expressed as ENSEBL transcript.
- `primary_transcript` is the HGNC gene preferred transcript, which should be used for clinical reporting nomenclature unless there are pressing exceptions. Is expressed as a REFSEQ transcript. Now VEP "HGVS" are afaik based on ENSEMBL anyway. For hg38 MANE select unifies RefSeq and ENSEMBL but for 37 we are left with what we have for mapping RefSeq to ENSEML.

In general, when we export only one thing, I think this should actually be `primary_transcript`, and may have to be fetched off the `hgnc_gene` at more points, as it is not currently on the variant-stored gene_obj we use here. 
`</rant>`
In this case, the file very clearly states canonical transcript etc so fine, so I guess it is ok. 😅

scout/server/blueprints/variants/controllers.py
```diff
@@ -434,20 +436,23 @@ def variant_export_genes_info(store, gene_list):
...
+                hgvs_nucleotide = transcript_obj.get("coding_sequence_name") or "-"
+                protein_change = transcript_obj.get("protein_sequence_name") or "-"
         hgvs_c.append(hgvs_nucleotide)
+        pt_c.append(protein_change)
 
-    gene_info.append(";".join(str(x) for x in gene_ids))
-    gene_info.append(";".join(str(x) for x in gene_names))
-    gene_info.append(";".join(str(x) for x in hgvs_c))
+    for item in [gene_ids, gene_names, canonical_txs, hgvs_c, pt_c]:
+        gene_info.append(";".join(str(x) for x in item))
```
dnil, line 455: :+1:

dnil (approved):
Good stuff.

Exporting primary transcript changes would arguably be more useful to clinical users, but this appears to have been a research request, and canonical is clearly stated on the header. I approve after some soul searching. 😄

scout/server/blueprints/variants/controllers.py
northwestwitch: Yes. For me this is (and will always be) a mess. But I start to grasp the nuances

northwestwitch: So this is fine to merge or do we want to revisit what should be exported (primary instead of canonical?)

dnil: I think it is fine. It is pretty clear as is, and was requested mainly from the research side if I recall correctly from the meeting? That makes me feel it's even more ok: it is not as important to stick to HGNC there. For most genes they will be the same anyway, and for a subset of the remainder it will be wrong even with our current implementation of primary. Go forward! We can fix later if called for, and it will make better more sense to fix it once we have 38 for all.

PR #2275, opened 2020-12-21
Fix wrong transcript cdna and protein change on variant overview and case report

northwestwitch:
I'm very very confused about how variants and transcripts are parsed and used, so this is an attempt. Fix #2259 

What I've changed:
- I've removed the list of gene.primary_transcripts in the variant object
- In variant page (transcript overview) and general report I'm looping through gene.transcripts instead of gene.primary_transcripts and then I print the transcript only if it has a refseq (which was the condition for the transcript to be added to the gene.primary_transcripts in the first place)
- I've modified which transcripts are actually printed in the variants transcript overview. From this: 
image to this:
image
- I've modified in the same way the transcripts shown in the case report:
image

**How to test on a local instance of Scout**:
1. Setup the demo instance of scout:
```scout --demo update genes```
1. load all genes with the command:
```scout --demo update genes```
1. Save this test VCF file in the scout/demo folder, as testfile.vcf:
```
##fileformat=VCFv4.2
##ALT=<ID=NON_REF,Description="Represents any possible alternative allele not already represented at this location by REF and ALT">
##ExACpLI=ExACpLI value for gene
##FILTER=<ID=LowQual,Description="Low quality">
##FILTER=<ID=PASS,Description="All filters passed">
##FILTER=<ID=VQSRTrancheINDEL99.99to100.00+,Description="Truth sensitivity tranche level for INDEL model at VQS Lod < -28864.1591">
##FILTER=<ID=VQSRTrancheINDEL99.99to100.00,Description="Truth sensitivity tranche level for INDEL model at VQS Lod: -28864.1591 <= x < -17.0651">
##FILTER=<ID=VQSRTrancheSNP99.99to100.00+,Description="Truth sensitivity tranche level for SNP model at VQS Lod < -113980.0266">
##FILTER=<ID=VQSRTrancheSNP99.99to100.00,Description="Truth sensitivity tranche level for SNP model at VQS Lod: -113980.0266 <= x < -30.6328">
##FORMAT=<ID=AD,Number=R,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">
##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth (reads with MQ=255 or with bad mates are filtered)">
##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
[...]

PR #2279, opened 2020-12-23
Improve display of HGNC primary transcript on case report

### LINKED ISSUE #2259: wrong transcript/connection on case report
https://scout.scilifelab.se/cust002/F0037125/cce4d3ad01f80405137e886f6b26f519

The position in transcript HGNC NM_022089 should be c.2978 not as below, which is shown in Scout.

  | ATP13A2 | NM_022089 (ENST00000452699) (hgnc-primary) | c.2963C>T

northwestwitch:
Today with a fresh mind I've realized that at the gene level it is already saved which the canonical transcript is, and probably overlaps with a refseq with version. To be tested on stage, but perhaps works as it should!

fix #2259

For instance changes the output in the case report with from this:

image

To this: ✌️ 

image

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -19,8 +19,8 @@
           {% if transcript.is_disease_associated %}
             <span class="badge badge-pill badge-danger" title="disease_associated">D</span>
           {% endif %}
-          {% if transcript.is_primary %}
-            <span class="badge badge-pill badge-primary" title="primary_ref_transcript">hgnc primary:{{transcript.refseq_id}}</span>
+          {% if gene.canonical_transcript == transcript.transcript_id %}
```
northwestwitch, line 22: or perhaps {% if gene.canonical_transcript in transcript.refseq_ids %} ??

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Save case variants count in case document and not in sessions
 - Style of gene panels multiselect on case page
 - Collapse/expand main HPO checkboxes in phenomodel preview
+- Show canonical transcript as hgnc primary on case report
```
dnil, line 42: Hm, rather show RefSeq transcript as HGNC primary, right?

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -117,15 +120,30 @@ <h5 class="modal-title" id="exampleModalLabel">Beacon submission</h5>
         <!-- show transcript info only if it's canonical, disease-associated or primary (print max 5 primary transcripts) -->
       {% if transcript.refseq_identifiers and n_primary_txs.count <= 5 or transcript.is_canonical or transcript.is_disease_associated or transcript.is_primary %}
         <li>{{transcript.transcript_id}}, RefSeq:[{{ transcript.refseq_identifiers|join(", ") or "-" }}], {{ (transcript.coding_sequence_name or '')|truncate(20, True) }},  {{ (transcript.protein_sequence_name or '')|url_decode }}
+
+          <!-- Capture HGNC transcript primary transcript from canonical or primary transcript id or refseq id-->
+          {% if hgnc_primary.refseq is none and transcript.is_canonical or transcript.is_primary %}
+            {% if hgnc_primary.refseq is none and transcript.is_canonical and transcript.transcript_id.startswith('NM_') %}
```
dnil, line 126: Here too: while often the same, it seems there is no guarantee `is_canonical`will find the right transcript

Starts with `NM_` is a sufficient but not necessary criteria: there may still be some `XM_` in HGNC.

Dunno, it is clearly better, but still not crystal clear to me. Unless you have checked very carefully, I feel we should not just guess on correcting this one? We will have to answer those "but my transcript is not in hg37" anyway, and I'd rather be sure the remainder are right! 😊

dnil (changes requested):
Here I'm not certain - it is also getting a little late - so feel very free to convince me! But it looks a like we find the right transcript for many, but a little by coincidence, and that the is_canonical may not always be right? This would then cause the annotated is_primary to not be flagged and the canonical to be?

scout/server/blueprints/cases/templates/cases/utils.html
northwestwitch: It's not crystal clear to me either. I've tested it on several cases on stage and it looked fine but I can't be sure that it will always capture the right one. Because I'm not sure of the whole thing that is. I don't know, we could close this one and keep it as it is!

dnil: Would there be any issue with just dropping the `and transcript.is_canonical` and leaving it at starts with NM_ or XM_, then matching to the actual primary id?

northwestwitch: I don't know about the result. Let's leave it out of the next release and let's test your suggestion with calm

northwestwitch:
Shall we close this one?

dnil:
Hm, or we just give it some more thought? I wouldn't mind trying it out with my last suggestion. It could be totally naive, but it kind of captures a lot of the spirit the clinicians have for this "Is it an NM transcript? Yeah, good". Then we actually do check that it is equal to the HGNC primary. Seems legit?

northwestwitch:
> Hm, or we just give it some more thought? I wouldn't mind trying it out with my last suggestion. It could be totally naive, but it kind of captures a lot of the spirit the clinicians have for this "Is it an NM transcript? Yeah, good". Then we actually do check that it is equal to the HGNC primary. Seems legit?

OK, I can fix it with your suggestion!

northwestwitch:
@dnil I'm trying to fix this one and I've realized that I don't know how to do it because I don't have clear in mind how it should be. Could you explain what a user would like to see and I can translate it into code?

northwestwitch:
> @dnil I'm trying to fix this one and I've realized that I don't know how to do it because I don't have clear in mind how it should be. Could you explain what a user would like to see and I can translate it into code? I also think that what a user sees in the report shouldn't differ from what he finds in the variant page, otherwise this would lead to questions..

dnil:
I think your second picture above sums it up very nicely: the primary symbol (and report) on the correctly versioned separate refseq transcript rather than the mixed ensemble approximation.

dnil:
But safely on the primary and not just when canonical refseq and primary happen to coincide, if you follow me.

northwestwitch:
> But safely on the primary and not just when canonical refseq and primary happen to coincide, if you follow me.

Ok, I 'll try

northwestwitch:
> I think your second picture above sums it up very nicely: the primary symbol (and report) on the correctly versioned separate refseq transcript rather than the mixed ensemble approximation.

Thanks

northwestwitch:
I've removed the option collection of the RefSeq from the canonical transcript and I'm collecting it from the primary only. It doesn't look like it's working.

image

The specific variant and test case can be taken from the description of this PR: https://github.com/Clinical-Genomics/scout/pull/2275

But I don't know perhaps I'm missing something @dnil, please feel free to edit this PR to test your way if this change is not what you meant!

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -107,14 +107,27 @@ <h5 class="modal-title" id="exampleModalLabel">Beacon submission</h5>
...
       <!-- increase counter if it has associated refseq -->
       {% if transcript.refseq_identifiers %}
         {% set n_primary_txs.count = n_primary_txs.count + 1 %}
       {% endif %}
 
-        <!-- show transcript info only if it's canonical, disease-associated or primary (print max 5 primary transcripts) -->
+      <!-- Capture HGNC primary transcript from primary transcript or refseq id-->
+      {% if hgnc_primary.refseq is none and transcript.is_primary %}
+        {% if transcript.transcript_id.startswith('NM_') %}
+          {% set hgnc_primary.refseq = transcript.transcript_id %}
```
dnil, line 122: Yep, that was the idea! Fingers crossed it actually works!

dnil (approved):

scout/server/blueprints/cases/templates/cases/utils.html
dnil: Oh right, did you check that the is_primary is actually set on the matching "other" refseq transcript as well? Lets see..

dnil: Ok, well, it seems the HGNC primary transcripts are NOT versioned in their db. 🙄

northwestwitch: Nope!

dnil: How about allowing multiple primaries then? That way we could flag both the ENS and NM_ one that both sort of match the HGNC primary?

dnil (changes requested):

scout/server/blueprints/cases/templates/cases/utils.html
dnil: That is so silly of them!? Anyway, that also makes me feel that perhaps we are overdoing this. If they dont do this more carefully than yeah, well, that transcript whatever version, we probably also dont need to overdo it. We should just be very careful to write the *actual* transcript on the report, much like in the variant table.

northwestwitch: Yes but the versioned ones might be only the canonical, as in the example, and you would still miss them

northwestwitch: What about closing this PR and forgetting it ever existed? 😄

dnil: Right, the issue is really that the full ID of a versioned transcript will not match the unversioned HGNC id.

dnil: Yeah, I agree with closing. Your previous PR #2275 fixed the underlying issue we had. Then there are the two 1) ENSEMBL hg37 non-MANE mapping and 2) HGNC non-versioned. I don't have the heart to broach 1) with their devs again, and it is really inevitable that we switch to hg38 anyway. I think I will try pinging HGNC about their non-versioned transcripts. That is really bad form?!

northwestwitch: Perhaps they don't include the version in HGNC because it might change all the time and they would get complaints that stuff downloaded a month before is already old? I don't know otherwise

dnil: I sent them a question - lets see!

PR #2281, opened 2020-12-28
Fixed javascript code and moved css code in case page

### LINKED ISSUE #2280: Broken javascript code on case page
Test locally by opening the demo case and trying to assign a phenotype or an omim diagnosis. Or perhaps by adding a gene to the dynamic gene panel --> the javascript code handling these fields is not responsive any more.
Case sidebar collapse is also broken

Fortunately is some change introduced in current master after deploying to production.

northwestwitch:
Fix #2280 

**How to test**:
1. From master branch, make sure that javascript code doesn't work (see bug issue description).
2. Install this branch and confirm that the errors are fixed

**Expected outcome**:

**Review:**
- [x] code approved by mikaell
- [x] tests executed by mikaell


scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -14,6 +14,30 @@
 {% block css %}
 {{ super() }}
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-select@1.13.9/dist/css/bootstrap-select.min.css">
+<style>
```
northwestwitch, line 17: This code is not new, I've just moved here because it makes more sense to have it under the css styles instead of under the javascript code

northwestwitch:
This is a quick review that fixes a bug. If anybody is looking for a soft task to get back on track after the holidays this is the one!

mikaell:
> This is a quick review that fixes a bug. If anybody is looking for a soft task to get back on track after the holidays this is the one!

This still seems to be open. I'll have a look.

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -519,12 +536,8 @@ <h5 class="modal-title" id="exampleModalLabel">Edit synopsis</h5>
 
       $('select[multiple]').selectpicker({
         width: '100%'
-        });
```
northwestwitch, line 522: This was the error @mikaell

mikaell (approved):
Works!

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -519,12 +536,8 @@ <h5 class="modal-title" id="exampleModalLabel">Edit synopsis</h5>
...
       });
 
-      $('select[multiple]').multiselect({
-        buttonWidth: '100%'
-      });
```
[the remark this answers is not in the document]
northwestwitch: No, check my comment above. But this code it's not longer used because the multiselect now is using the selectpicker and has its own css specifics, that's why I'm removing it

PR #2290, opened 2021-01-14
Reset all dismissed variants button

### LINKED ISSUE #2073: removal of dismissed variants
Can we have a botton to clear up the dismissed variants?

Ex. I start dismissing variant in a case, and for any reason I want to start again. Or someelse had look at the case and I want to do it again from scratch. Or we forgotten to ask to remove the dismissed variants when we ordered a re-run. Or we orderer a re-run in scout instreasd of making a ticket.

Without the need to re-run the case 👍  just pressind a botton

northwestwitch:
Reset all dismissed variants for a case by clicking on a button (fix #2073):

image

**How to test**:
1. Dismiss one or more variants (SNVs, SVs, cancer SNVs, cancer SVs) for a case
2. Reset dismissed status of all the dismissed variants by clicking on the yellow button at the bottom right corner of the variantS page

**Expected outcome**:

**Review:**
- [x] code approved by @mikaell 
- [x] tests executed by @mikaell


northwestwitch:
This is not ready, I need to add a test first!

northwestwitch:
Something went wrong when I tried to fix a conflict due to merging with master. I need to fix it before it can be reviewed

northwestwitch:
oh no this PR has yet another conflict, I'm starting to think that it would be wiser and less messy to close it and open a fresh one

northwestwitch:
Fixed!

mikaell (approved):
Looks good!

northwestwitch:
Thanks for the review @mikaell !

Before merging I have a doubt and I'd like to ask for feedback. The current situation in Scout (on master already, not this branch only) is that if you dismiss a variant and then you reset the dismissed status than the actions recorded in the database are the same: a variant dismissed, twice:

image

Shouldn't we change this behavior and introduce another verb, like "reset dismissed status" or something? It would be more appropriate I think. Unfortunately the new status would't work retroactively.. But better than nothing!

What do you say?

dnil:
Good point. A very quick fix would be changing the dismiss verb to Changed dismiss status on or similar. Your two different events is better and more informative in the long run but will potentially be confusing for a little while. I think both work really, so artistic freedom applies.

dnil:
Oh, and if it is for all variants and not just the selected I would have preferred to have the buttons clearly separated like with the dismiss near the dropdown and the reset all on its own? Possibly also different label, like "Reset all case dismiss status" or something?

northwestwitch:
> Oh, and if it is for all variants and not just the selected I would have preferred to have the buttons clearly separated like with the dismiss near the dropdown and the reset all on its own? Possibly also different label, like "Reset all case dismiss status" or something?

Good point, better to be clear on this functionality! One thing I didn't understand is where you want to move the yellow button (after renaming it to "Reset all case dismiss status")

northwestwitch:
Changed the PR so the buttons look like this: 
image

And the dismiss/reset dismiss like this:
image

dnil:
But yeah, sure the current also works - that way it is really far between the buttons, even if yellow sort of follows very directly on the check-box column. 🤷‍♂️  Potato-potato!

northwestwitch:
> But yeah, sure the current also works - that way it is really far between the buttons, even if yellow sort of follows very directly on the check-box column. 🤷‍♂️ Potato-potato!

I can switch them back in no time, but then the buttons would be more close..

dnil:
> > I can switch them back in no time, but then the buttons would be more close..
> 
> I would have tried a column wide space between the buttons, but on the same side, if nothing else since it is a little hard to understand that the Dismiss selected pertains to the checkboxes anyway..

And sorry, meant to reply instead of edit - that was a quite confusing experience.

northwestwitch:
> And sorry, meant to reply instead of edit - that was a quite confusing experience.

I understood, no worries. what about this?
image

northwestwitch:
Merging this in 3, 2...

PR #2298, opened 2021-01-22
Make get_coding_intervals() in load_variants() aware of genome build.

bjhall:
My first scout hacking in 8 months turned into a 3 hour debugging session with minimal code changes in the end... :)

We've occasionally had compounds appear as "not loaded" (and links not working) although they are certainly loaded. Turns out get_coding_intervals() would always use hg19 intervals, thus sometimes causing variants from the same gene to be processed in different "bulks". This fixes that.

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -587,6 +587,8 @@ class CaseHandler(object):
 
         old_sanger_variants = self.case_sanger_variants(case_obj["_id"])
 
+        genome_build = str(parsed_case.get("genome_build", 37))
+
         if old_case:
             LOG.info(
                 "Update case id for existing case: %s -> %s",
@@ -644,6 +646,7 @@ class CaseHandler(object):
                     case_obj=case_obj,
                     variant_type=variant_type,
                     category=category,
+                    build=genome_build,
                     rank_threshold=case_obj.get("rank_score_threshold", 5),
                 )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Google OAuth2 login setup in README file
 - Redirect to 'missing file'-icon if configured file is missing
 - Javascript error in case page
+- Fix compound matching during variant loading for hg38
 ### Changed
 - Save case variants count in case document and not in sessions
 - Style of gene panels multiselect on case page
--- a/scout/adapter/mongo/variant_loader.py
+++ b/scout/adapter/mongo/variant_loader.py
@@ -390,7 +390,7 @@ class VariantLoader(object):
         genes = [gene_obj for gene_obj in self.all_genes(build=build)]
         gene_to_panels = self.gene_to_panels(case_obj)
         hgncid_to_gene = self.hgncid_to_gene(genes=genes, build=build)
-        genomic_intervals = self.get_coding_intervals(genes=genes)
+        genomic_intervals = self.get_coding_intervals(genes=genes, build=build)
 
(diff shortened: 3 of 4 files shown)


northwestwitch:
Nice to have you back! 😃

scout/adapter/mongo/case.py
```diff
@@ -587,6 +587,8 @@ def load_case(self, config_data, update=False, keep_actions=True):
 
         old_sanger_variants = self.case_sanger_variants(case_obj["_id"])
 
+        genome_build = str(parsed_case.get("genome_build", 37))
```
northwestwitch, line 590: Seeing these recurrent problems makes me think once again that we should have defined a more rigid structure in the code when it comes to passing genome builds. This code is working if you have either 37 or 38 as genome_build at the case level. But it would probably fail if you have "GRCh37" or "GRCh38" (we have a few of these cases in our database).

What about introducing an additional control to force the build to be in [37,38]? Doesn't have to be here, perhaps even before, like here: https://github.com/Clinical-Genomics/scout/blob/e32ef43598191d2110ffa6e26700c2f87fe75772/scout/parse/case.py#L495

northwestwitch (approved):
Suggestion: let's force the case parser to accept either 37 or 38.

scout/adapter/mongo/case.py
bjhall: I completely agree. It's a bit of a mess throughout the code base. "human_genome_build" is inconsistently treated as a string or an integer (I blame the yaml file format for this (death to yaml!)), and sometimes it defaults to 37 and other times not.

I've added a commit that enforces that "human_genome_build" is either 37 or 38. But this breaks loading cases without the yaml file (ie. with --vcf/--ped), since there is no way to specify the genome_build. Not sure what the best solution to this whole situation is. Any suggestions?

northwestwitch: Yes, I see your point here. But how does it work now to upload cases from ped file now? (never done it myself). Do these cases get a default genome build 37? Because that would be conceptually wrong anyway

northwestwitch: No, they get genome build None, see the demo case uploaded using master branch and command: `scout --demo load case --owner cust000 --ped scout/demo/643594.ped --vcf scout/demo/643594.clinical.vcf.gz`
image

northwestwitch: I don't think having cases without genome build makes a lot of sense. We could solve this issue by adding an extra param here https://github.com/Clinical-Genomics/scout/blob/e32ef43598191d2110ffa6e26700c2f87fe75772/scout/commands/load/case.py#L44

To specify the genome build (this click param could accept only 37 or 38 for instance.
And then in the parser we could have 
```
allowed_genome_builds = ["37", "38"]
    if str(config.get("human_genome_build", **build**)) not in allowed_genome_builds:
        raise ConfigError(f"human_genome_build must be one of {allowed_genome_builds}")
```

Where **build** is passed from the CLI.

Unless we have some reason to allow loading cases without genome build, but I am not aware of any. @dni or @moonso ?

northwestwitch: But I think for simplicity's sake we could revert to the code of your original PR and merge, and then fix this issue in another one!

bjhall: Sounds like a good idea. I've reverted the last commit. Feel free to merge.

Yes genome_build=None  is the reason the code is littered with various forms of `config.get(build, 37)`. Would be nice to get rid of that mess at some point.

northwestwitch: I think this is fine to merge now. I've opened a new issue to address the issue we discussed here: #2300

PR #2306, opened 2021-01-26
Fix #2305 - IGV zoom length contig

### LINKED ISSUE #2305: Structural variants alignment in IGV shows entire genome
Hi,
When looking at structural variants, I can view the individual breakpoints in IGV but when I click the IGV button to view the entire SV, I get the whole genome (it used to be an overview of the region with the SV including both breakpoints). This is not case-specific and it's been like this for a couple of weeks.
It's obviously not a big issue, but I hope it can be reverted to the previous settings. Thank you!

dnil:
This PR fixes a bug.

**How to test**:
1. on a case with cram files available, verify that opening IGV view on the whole length of a SV variant (del, dup, etc) opens IGV on all configs
Screenshot 2021-01-26 at 15 09 24

1. after applying patch, note that IGV instead opens on the correct chromosome and region
 
Screenshot 2021-01-26 at 15 09 03

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/utils.html
+++ b/scout/server/blueprints/variant/templates/variant/utils.html
@@ -271,7 +271,7 @@
           <input type="hidden" name="sample" value="{{case.sample_names|join(',')}}">
           <input type="hidden" name="build" value="{{case.genome_build}}">
 
-          <input type="hidden" name="contig" value="{% if zoom == "start" or zoom == "length" %}{{ variant.chromosome }}{% elif zoom == "end" %}{{ variant.end_chrom }}{% endif %}">
+          <input type="hidden" name="contig" value="{% if zoom in ["start", "length"] %}{{ variant.chromosome }}{% elif zoom == "end" %}{{ variant.end_chrom }}{% endif %}">
           {% if zoom == "start" %}
             <input type="hidden" name="start" value="{{variant.position - 500}}">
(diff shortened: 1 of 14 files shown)


dnil:
Ready for a quick review, me hopes! 😅

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -271,7 +271,7 @@
           <input type="hidden" name="sample" value="{{case.sample_names|join(',')}}">
           <input type="hidden" name="build" value="{{case.genome_build}}">
 
-          <input type="hidden" name="contig" value="{% if zoom == "start" %}{{ variant.chromosome }}{% elif zoom == "end" %}{{ variant.end_chrom }}{% endif %}">
+          <input type="hidden" name="contig" value="{% if zoom == "start" or zoom == "length" %}{{ variant.chromosome }}{% elif zoom == "end" %}{{ variant.end_chrom }}{% endif %}">
```
northwestwitch, line 274: you could also have if zoom in ["start", "length"], but it works!

northwestwitch (approved):
👍

scout/server/blueprints/variant/templates/variant/utils.html
dnil: That is cleaner - I will change!

PR #2315, opened 2021-01-28
Silence tp53 link synonymous

dnil:
This PR  fixes a bug.

**How to test**:
1. note that synonymous variants in TP53 still have a `p.=` `HGVS_p` string and thus get a mutatnTP53 link.
2. apply patch and see it dissapear

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -522,7 +522,7 @@ def cbioportal(hgnc_symbol, protein_sequence_name):
 def mutantp53(hgnc_id, protein_variant):
     if hgnc_id != 11998:
         return None
-    if not protein_variant:
+    if not protein_variant or protein_variant.endswith("="):
         return None
 
     url_template = "http://mutantp53.broadinstitute.org/?query={}"
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -522,7 +522,7 @@ def cbioportal(hgnc_symbol, protein_sequence_name):
 def mutantp53(hgnc_id, protein_variant):
     if hgnc_id != 11998:
         return None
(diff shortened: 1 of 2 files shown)


northwestwitch:
Can be tested on stage with this variant: https://scout-stage.scilifelab.se/cust000/wholerhino/df0326bb8a3b0d485c6ce7793d1cb954

dnil:
Before: 
Screenshot 2021-01-28 at 09 56 21

After:
Screenshot 2021-01-28 at 10 11 31

scout/server/links.py
```diff
@@ -522,7 +522,7 @@ def cbioportal(hgnc_symbol, protein_sequence_name):
 def mutantp53(hgnc_id, protein_variant):
     if hgnc_id != 11998:
         return None
-    if not protein_variant:
+    if not protein_variant or protein_variant.endswith("=") or protein_variant.endswith("%3D"):
```
northwestwitch, line 525: Why does it end like an html-encoded "=" ?

northwestwitch (approved):

scout/server/links.py
dnil: Its from VEP, presumably since certain special characters override `"`-escapes/quotes in VCFs. We actually miss that in display in several places that don't use the prettified variant versions..

dnil: At this time we could actually skip the literal `=` test, but I dunno, better safe than sorry?

northwestwitch: Yes I agree

northwestwitch:
merging this @dnil?

Issue #2322, opened 2021-01-29
ClinVar submission doesn't work for a specific variant


northwestwitch:
There is something odd in this variant for instance:

https://scout.scilifelab.se/cust003/15001/52b97af467ec8966f6fb497b1aedb79a

northwestwitch:
Working on this. I'm trying to reproduce it in a test environment by adding another gene to the variant, but in this case it works!

This could could give an indication of what's wrong anyway:

![image](https://user-images.githubusercontent.com/28093618/106435712-862c5100-6473-11eb-85bb-3ddd85f70efd.png)

northwestwitch:
It's not a matter of double gene because another variant hitting the same two genes work: https://scout-stage.scilifelab.se/cust003/20192/25e253c05b52241bc18c64f9dd336f45/clinvar

dnil:
And its not just that one of them has no aa change?

northwestwitch:
Perhaps, I'll try to remove it from TECR in the demo case

dnil:
I haven't tested, just clicked the link once,  so more on the hunch level than usual... 😊

northwestwitch:
No but also the demo case has the aa change in only one gene. 😕

dnil:
Mm, but only one of them; the non-working one only has a splice change in one, the other blank?

dnil:
Sorry, I compared to your test, not the live one! :+1:

dnil:
dbSNP then?

northwestwitch:
The weird thing is that the javascript alert complains for the dbsnp id, that is missing in both cases!

northwestwitch:
And it doesn't work even if you provide a fake dbSNP id (like rs1111)

dnil:
In the spirit of random guessing: next row then is a classification, and the non-working one has some ACMG stuff. Anything there that might interact?

dnil:
Like especially the updated classification events?

northwestwitch:
I've classified the demo one in the same way and still works, so it shouldn't be that. I'm starting to think that is not something in the form you see, but something that fails in the form that is displayed immediately after when you click 
![image](https://user-images.githubusercontent.com/28093618/106448721-7288e680-6483-11eb-8288-7df3c54102e1.png) button

dnil:
Or like the rows before: the non-working appears to have a slightly broken option for the selection for AR_hom, which it is:
(L188-189)

dnil:
Like here:

[...]

PR #2327, opened 2021-02-01
Fix ClinVar submission inheritance option and multiple gene display

northwestwitch:
Fixes a couple of problems in the ClinVar submission page.
- For variants having multiple genes for instance this one: https://scout-stage.scilifelab.se/cust003/15001/01fc608966a1fb60233a119eb651de00
- for all variants, since one of the inheritance models select options was broken

- PS: this PR is not fixing #2322,  because so far I haven't understood why the submission for that variant (same variant as in this PR) is not working.

**How to test**:
1. Check the difference in a ClinVar submission between master and this branch for the variant above

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/clinvar.html
+++ b/scout/server/blueprints/variant/templates/variant/clinvar.html
@@ -179,8 +179,7 @@
                             <option value="Autosomal recessive inheritance" selected>Autosomal recessive inheritance</option>
                             <option value="Autosomal dominant">Autosomal dominant inheritance</option>
                             <option value="X-linked inheritance">X-linked inheritance</option>
-                            <option value="
-                            -limited autosomal dominant">Sex-limited autosomal dominant</option>
+                            <option value="Sex-limited autosomal dominant">Sex-limited autosomal dominant</option>
                             <option value="Mitochondrial inheritance">Mitochondrial inheritance</option>
                             <option value="Y-linked inheritance">Y-linked inheritance</option>
                           {% elif 'XR' in pinned.genetic_models or 'XR_dn' in pinned.genetic_models %}
--- a/scout/server/blueprints/variant/templates/variant/clinvar.html
+++ b/scout/server/blueprints/variant/templates/variant/clinvar.html
@@ -135,14 +135,8 @@
                           {% set gene_symbols = [] %}
                           {% for gene in pinned.genes %}
                             {% do gene_symbols.append(gene.hgnc_symbol) %}
-                            {% if pinned.genes|length == 1 %}
-                              <strong>{{ pinned.genes[0].hgnc_symbol }}</strong><br>
-                            {% else %}
-                              {% for gene in pinned.genes %}
-                              <strong>{{ gene.hgnc_symbol+";" }}</strong>
-                              {% endfor %}
-                            {% endif %}
                           {% endfor %}
+                            <strong>{{gene_symbols|join(", ")}}</strong>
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/variant/templates/variant/clinvar.html
```diff
@@ -135,14 +135,8 @@ <h5>
...
                             {% do gene_symbols.append(gene.hgnc_symbol) %}
-                            {% if pinned.genes|length == 1 %}
-                              <strong>{{ pinned.genes[0].hgnc_symbol }}</strong><br>
-                            {% else %}
-                              {% for gene in pinned.genes %}
-                              <strong>{{ gene.hgnc_symbol+";" }}</strong>
-                              {% endfor %}
-                            {% endif %}
                           {% endfor %}
+                            <strong>{{gene_symbols|join(", ")}}</strong>
```
dnil, line 139: Remind me, we did not enable SV submission yet, right? Because this could potentially grow really, really long then. Optionally truncate at a max len and add like an ellipsis with the number of genes to future proof.

scout/server/blueprints/variant/templates/variant/clinvar.html
```diff
@@ -185,8 +179,7 @@ <h5>
                             <option value="Autosomal recessive inheritance" selected>Autosomal recessive inheritance</option>
                             <option value="Autosomal dominant">Autosomal dominant inheritance</option>
                             <option value="X-linked inheritance">X-linked inheritance</option>
-                            <option value="
-                            -limited autosomal dominant">Sex-limited autosomal dominant</option>
+                            <option value="Sex-limited autosomal dominant">Sex-limited autosomal dominant</option>
```
dnil, line 182: 😄

scout/server/blueprints/variant/templates/variant/clinvar.html
```diff
@@ -238,7 +231,7 @@ <h5>
                     {% if pinned.category == 'snv'%}
                       <tr>
                         <td>var. identifier</td>
-                        <td><input type="text" id="Variation identifiers_{{pinned._id}}" name="variations_ids@{{pinned._id}}" value="{{ pinned.dbsnp_id.split(':')[0] if pinned.dbsnp_id else ''}}" size=14 pattern="rs[0-9]+"></td>
+                        <td><input type="text" id="Variation identifiers_{{pinned._id}}" name="variations_ids@{{pinned._id}}" value="{{ pinned.dbsnp_id.split(':')[0] if pinned.dbsnp_id else ''}}" size=14 pattern="rs[0-9]+" ></td>
```
dnil, line 234: Not sure we need that space, but I take it you rewrote that line a few times. 😬

dnil (approved):

scout/server/blueprints/variant/templates/variant/clinvar.html
northwestwitch: We did enable it actually. It could grow big but perhaps let's leave it to the users if they want to include them or not? I'm not really sure how they want it at ClinVar

scout/server/blueprints/variant/templates/variant/clinvar.html
northwestwitch: I'm not really sure how it ended up there, but shouldn't affect anything

northwestwitch:
Merging in 3, 2 ..

PR #2329, opened 2021-02-02
Fix #2323 - throttle back IGV.js to 2.7.0 until a fix is available

### LINKED ISSUE #2323: IGV gene track
Hej, when looking at the GCH1 gene in IGV I'm now missing the gene (the blue exon/intron) in the IGV gene track. See attached file with screenshot before and after the update(?) - Update IGV.js to latest (v2.7.4). (It's the same result when my colleagues check the same region in IGV for other cases). 

[GCH1_IGV.pdf](https://github.com/Clinical-Genomics/scout/files/5894306/GCH1_IGV.pdf)

dnil:

See issue for details.

**How to test**:
1. select IGV for a variant on master, verify that the gene is not visible at ordinary zoom
2. verify that it is directly visible with patch.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Broken or missing link in the documentation
 - Multiple gene names in ClinVar submission form
 - Inheritance model select field in ClinVar submission
+- IGV.js >2.7.0 has an issue with the gene track zoom levels - temp freeze at 2.7.0
 ### Changed
 
 ## [4.28]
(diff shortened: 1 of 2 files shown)


northwestwitch (approved):
👍

dnil:
Like that..
Screenshot 2021-02-02 at 17 05 22

dnil:
So far we could get away with a patch change, so not a bad idea!

I was thinking of connecting the loqusdb REST version next, but that will require server config changes as well when deploying for real so can block stuff.

PR #2333, opened 2021-02-03
Revert CORS anywhere

dnil:
This PR tempfixes a bug.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
@@ -32,8 +32,6 @@
   <div class="container-fluid" id="igvDiv" style="padding:5px; border:1px solid lightgray"></div>
 </body>
 
-{% set cors_anywhere_prefix = "https://cors-anywhere.herokuapp.com" -%}
-
 <script type="text/javascript">
   $(document).ready(function () {
       var div = $("#igvDiv")[0],
@@ -72,8 +70,8 @@
                           name: "{{ custTrack.name }}",
                           type: "{{ custTrack.type }}",
                           format: "{{ custTrack.format }}",
-                          url: "{{cors_anywhere_prefix}}/{{custTrack.url}}",
-                          indexURL: "{{cors_anywhere_prefix}}/{{custTrack.indexURL}}"
+                          url: "{{custTrack.url}}",
+                          indexURL: "{{custTrack.indexURL}}"
                         },
                       {% endfor %}
                       {% for wtrack in rhocall_wig %}
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
@@ -70,8 +70,8 @@
                           name: "{{ custTrack.name }}",
                           type: "{{ custTrack.type }}",
                           format: "{{ custTrack.format }}",
-                          url: "{{custTrack.url}}",
-                          indexURL: "{{custTrack.indexURL}}"
+                          url: "{{ url_for('alignviewers.remote_cors', remote_url=custTrack.url) }}",
+                          indexURL: "{{ url_for('alignviewers.remote_cors', remote_url=custTrack.indexURL) }}"
                         },
                       {% endfor %}
                       {% for wtrack in rhocall_wig %}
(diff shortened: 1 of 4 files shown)


scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
```diff
@@ -72,8 +70,8 @@
                           name: "{{ custTrack.name }}",
                           type: "{{ custTrack.type }}",
                           format: "{{ custTrack.format }}",
-                          url: "{{cors_anywhere_prefix}}/{{custTrack.url}}",
```
northwestwitch, line 75: Tested this and it's causing this other error:

image

Because the dbVar track can't be reached without the right CORS settings either. I think you should comment out the whole custTrack loop block until we find a solution.

northwestwitch (changes requested):
See my comment!

scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
dnil: Yes, foreseen, and sounds good. I was actually looking at proxying the remote track ourselves, via your link, but in a meeting now for a couple of h. 😊

northwestwitch:
This one: https://www.npmjs.com/package/cors
Available as a library from here ---> https://www.jsdelivr.com/package/npm/cors

dnil:
In the spirit of sharing I used this:
https://stackoverflow.com/questions/6656363/proxying-to-another-web-service-with-flask#36601467
Seems to be working. What do you think?

scout/server/blueprints/alignviewers/views.py
```diff
@@ -27,6 +30,30 @@
 LOG = logging.getLogger(__name__)
 
 
+@alignviewers_bp.route("/remote/cors/<path:remote_url>", methods=["OPTIONS", "GET"])
+def remote_cors(remote_url):
```
northwestwitch, line 34: Yes! It works very nicely:

image

northwestwitch (approved):
Good job! 👍 
You could perhaps mention the stackoverflow thread in a comment, as a thank you to the user that saved us. Nice not to be dependent of an external server any more!

scout/server/blueprints/alignviewers/views.py
northwestwitch: This should also be testable, but when we have time!

dnil: Good call, will do!

dnil: Acknowledging that is, but yeah, testable, only a little involved. Monkey patching comes to mind, but 84 likes on stack overflow must mean something.. 😉

northwestwitch: It basically becomes a file hosted on our server. I could try to write a test if you want, and if it takes too much time we just merge and that's it

dnil: Yes, exactly: one could either fake its env to control that it does exactly the right thing, or just give it a url that will always be up and see that it gets a response..

northwestwitch: I was thinking google

dnil: It crossed my mind.. 😁

tests/server/blueprints/alignviewers/test_alignviewers_views.py
```diff
@@ -1,7 +1,26 @@
...
+    # GIVEN a valid user and institute
+    with app.test_client() as client:
+        # GIVEN that the user could be logged in
+        resp = client.get(url_for("auto_login"))
+        assert resp.status_code == 200
+
+        # WHEN the remote cors endpoint is invoked with an url
+        resp = client.get(url_for("alignviewers.remote_cors", remote_url=cloud_track_url))
+        # THEN it should return success response
+        assert resp.status_code == 200
```
northwestwitch, line 21: I don't know, this is a bit unspecific but perhaps it helps going up in coverage

dnil: Sure, I approve! If you'd like something less generic, you could also use http://www.clinicalgenomics.se/scout/ or something, but uptime is perhaps not as guaranteed. 😸

northwestwitch: I trust more google!

dnil: Lets go with that. I think we can merge this now? Or wait, the changelog is now a little obsolete..

PR #2346, opened 2021-02-10
Center igv on variant start

### LINKED ISSUE #2345: IGV button is not linking to the right position
Hi! 
hope someone can help me! 
The button for IGV once you click on a variant does not take you to the correct position. Instead you have to "drag" your way to the right position. Can this be solved? 

BEst Regards, HEro

dnil:

Fix #2345 by setting SNV variant view IGV initial window end to center from variant start rather than end.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Introduced a `reset dismiss variant` verb
 - Button to reset all dismissed variants for a case
 ### Fixed
+- Center initial igv view on variant start with SNV/indels 
 ### Changed
 
 
--- a/scout/server/blueprints/variant/templates/variant/variant.html
+++ b/scout/server/blueprints/variant/templates/variant/variant.html
@@ -463,7 +463,7 @@
                   <input type="hidden" name="build" value="{{case.genome_build}}">
                   <input type="hidden" name="contig" value="{{variant.chromosome}}">
                   <input type="hidden" name="start" value="{{variant.position - 50}}">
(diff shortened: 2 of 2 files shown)


northwestwitch (approved):
Sounds like a good solution! And I guess that centering the pointer at the beginning is better then centering it in the middle. It was us after all!

dnil:
> Sounds like a good solution! And I guess that centering the pointer at the beginning is better then centering it in the middle. It was us after all!

Yeah, that is sort of what we have now, and I have the feeling for some time it makes users confused. We be like: why is the igv dotted line not directly on my variant but a few nt off? This is bad. Before realising the variant was larger or whatnot. It is particularly weird for ins, with the way the end coord is stored in the db. Not a huge fan of the latter, though it may make them fit the short name lookups better.

northwestwitch:
> Yeah, that is sort of what we have now, and I have the feeling for some time it makes users confused. We be like: why is the igv dotted line not directly on my variant but a few nt off? This is bad. Before realising the variant was larger or whatnot. It is particularly weird for ins, with the way the end coord is stored in the db. Not a huge fan of the latter, though it may make them fit the short name lookups better.

Quite sure that after merging this you can close at least a couple of issues! 🥇

dnil:
Heres a small test variant on stage:
https://scout-stage.scilifelab.se/cust002/F0013248-3/160782076e44fb9553360a6abf638ad0
Screenshot 2021-02-10 at 10 14 13

and after patch:
Screenshot 2021-02-10 at 10 21 14

PR #2347, opened 2021-02-10
Fix DECIPHER link to work for both GRCh37 and GRCh38 after their migr…

bjhall:
DECIPHER recently (?) moved to GRCh38. This means that the link on the sv-variant.html page has been sending GRCh37 users to the wrong region for some time.

This fixes that as good as possible. They do retain some support for GRCh37 by making a liftover when you perform a search with a grch37-prefix. But I cannot find a way to link directly to the nice browser page that worked for GRCh37 before and now works for GRCh38. 

But I still think it is better to link to a search page of the proper region, than to a nice view over the wrong region.

For example. GRCh38 cases would end up on this page:
https://decipher.sanger.ac.uk/browser#q/X:28632456-28634196/location/X:28632456-28634196

And GRCh37 cases here (the coordinates are wrong, I just to show the page where you'd end up):
https://decipher.sanger.ac.uk/search/patients/browser?q=grch37%3AX%3A28632456-28634196

More reason for everyone to move to GRCh38 asap :)


dnil:
Thank you @bjhall! Super!

scout/server/blueprints/variant/templates/variant/sv-variant.html
```diff
@@ -264,9 +264,12 @@
 	    {% endif %}
           </td>
 	  <td>
-	    <a class="btn btn-outline-secondary form-control" href="https://decipher.sanger.ac.uk/browser#q/{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}%20/location/{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}" target ="_blank">DECIPHER {% if case.genome_build == 38 %} (GRCh37){% endif %}</a>
-
-	  </td>
+	    {% if case.genome_build == 38 %}
```
dnil, line 267: This should actually also use links.py, but it was already like this. Feel free to unify, but not required!

scout/server/blueprints/variant/templates/variant/sv-variant.html
```diff
@@ -264,9 +264,12 @@
...
+	      <a class="btn btn-outline-secondary form-control" href="https://decipher.sanger.ac.uk/browser#q/{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}/location/{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}" target ="_blank">DECIPHER</a>
+	    {% else %}
+	      <a class="btn btn-outline-secondary form-control" href="https://decipher.sanger.ac.uk/search/patients/browser?q=grch37:{{ variant.chromosome }}:{{ variant.position }}-{{ variant.end }}" target ="_blank">DECIPHER</a>
```
dnil, line 270: 🥇

dnil (approved):

dnil:
Tested on stage; looks good!

dnil:
You know what: let me show you here (where it is probably my oversight in the first place) what I mean by putting the links among the others on the decorated variant display object, and you can do the same for the other PR (verification email)?

PR #2348, opened 2021-02-10
Additional information in verification emails

bjhall:
This PR adds some additional information to the SNV verification emails. This is a step towards unifying the Lund fork with upstream.

* Adds chromosomal position (previously only printed in the form of (1_121324642_T_C_clinical)
* Adds strand information to transcripts
* Marks primary transcript in the list of transcripts
* Adds a link to ThermoFisher's primer ordering page (if GRCh38). It feels a bit strange to add this, since it somehow feels like Scout endorses ThermoFisher primers... Any thoughts? We need the link in Lund obviously. Should it be disabled by default and activated with some config flag?

This image hightlights the additions (should probably decode the name in "Ordered by" properly, but that's another issue):
bild


dnil:
Yeah, having a link there is not bad, but I like your idea to make it a config option. Just a variable for link - and possibly link label if you don't just make the label a generic "Order primers here" or such?

scout/server/blueprints/variant/verification_controllers.py
```diff
@@ -99,11 +103,15 @@ def variant_verification(
...
     ]
     tx_changes = []
+    thermo_primer_link = ""
 
     if category == "snv":  # SNV
         view_type = "variant.variant"
         tx_changes = []
 
+        if case_obj["genome_build"] == 38:
+            thermo_primer_link = f"https://www.thermofisher.com/order/genome-database/searchResults?searchMode=keyword&CID=&ICID=&productTypeSelect=ceprimer&targetTypeSelect=ceprimer_all&alternateTargetTypeSelect=&alternateProductTypeSelect=&originalCount=0&species=Homo+sapiens&otherSpecies=&additionalFilter=ceprimer-human-exome&keyword=&sequenceInput=&selectedInputType=&chromosome={chromosome}&chromStart={position}&chromStop={position}&vcfUpload=&multiChromoSome=&batchText=&batchUpload=&sequenceUpload=&multiSequence=&multiSequenceNames=&priorSearchTerms=%28NR%29"
```
dnil, line 113: Consider moving this to the links.py helpers!

dnil: (and yeah, probably make it like external_primer_order_link and the target configurable.. 😅)

scout/server/blueprints/variant/verification_controllers.py
```diff
@@ -71,7 +71,11 @@ def variant_verification(
     category = variant_obj.get("category", "snv")
     display_name = variant_obj.get("display_name")
     chromosome = variant_obj["chromosome"]
+    position = variant_obj["position"]
     end_chrom = variant_obj.get("end_chrom", chromosome)
+    chr_position = (
```
dnil, line 76: :+1:

dnil (changes requested):
Well, you actually suggested the changes yourself.. 😉 
But have a look at the general link templates - would make some sense to keep it similar, and perhaps even in the same place.

scout/server/blueprints/variant/verification_controllers.py
bjhall: Sorry for the delayed response! I blame the virus!

I've finally pushed some changes to this branch that moves the primer ordering link to config.py and the link contstruction to links.py. So now you can add a variable to config.py like this.

```
EXTERNAL_PRIMER_ORDER_LINK_38 = "https://www.primercompany.com/order?chr={chromosome}&pos={position}"
```

And a link will show up in the verification email for SNVs. `chromosome` and `position` are the only supported parameters at the moment. If the config variable is not set, things will look the same as before.

EXTERNAL_PRIMER_ORDER_LINK_37 is also supported for grch37. I don't know if it makes sense to have two different config variables or if it should generalized to something like this:

```
EXTERNAL_PRIMER_ORDER_LINK = [
    {"url":"http://some.url", genome_build=38},
    {"url":"http://another.url", genome_build=37}
]
```

Rather not change it unless you have a strong opinion. :)

scout/server/blueprints/variant/verification_controllers.py
```diff
@@ -99,11 +105,14 @@ def variant_verification(
         for sample_obj in variant_obj["samples"]
     ]
     tx_changes = []
+    external_primer_link = ""
 
     if category == "snv":  # SNV
         view_type = "variant.variant"
         tx_changes = []
 
+        external_primer_link = external_primer_order_link(variant_obj, case_obj["genome_build"])
```
dnil, line 114: 👍

dnil (approved):

scout/server/links.py
```diff
@@ -607,3 +608,18 @@ def spidex_human(variant_obj):
         return "medium"
 
     return "high"
+
+
+def external_primer_order_link(variant_obj, build=None):
```
dnil, line 613: Oh, and this wants to have a docstring!

bjhall: I added docstrings for this function, and for some other functions in the same file!

PR #2356, opened 2021-02-12
Update igv.js to 2.4.7 with fix on genes track

northwestwitch:
This PR updated IGV.js from version 2.7.0 to 2.7.4 and removes the visibility window parameter in gene track that caused this following [bug](https://github.com/Clinical-Genomics/scout/issues/2323).

See the thread with a developer of igv.js [here](https://github.com/igvteam/igv.js/issues/1261)

**How to test**:
1. Install the branch on stage, and go to this variant: https://scout-stage.scilifelab.se/cust002/F0013248-3/17981289e34c2c706ecf53f8de78b0d1
2. Click on the igv button and then look in igv for this gene interval: chr14:55,369,194-55,369,443
3. The gene "GCH1" should be displayed directly in the gene track

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Interesting! I only tried lowering the visibility window, not removing it entirely! Do the genes still show on a very zoomed out view with this?

northwestwitch:
> Interesting! I only tried lowering the visibility window, not removing it entirely! Do the genes still show on a very zoomed out view with this?

Genes look fine even if you zoom out a lot, give it a try, this branch is on stage now!

dnil:
Yay, perfect! Well caught!

I found another thing, related to the zoom level: range queries for negative coords (or generally out of bounds?) seem not to terminate any longer (if they ever did). Eg https://scout-stage.scilifelab.se/cust000/643594-300M/sv/variants/1b8aa31ab526f4ef5d1181f523373414. But that is a new issue. After teaching. 😄

dnil (approved):
Well found!

northwestwitch:
This was a igv.js bug after all. They're fixing it now and we should switch to the new release when it out!

dnil:
Yes, seems they are merged-but-not-released yet. Still, this PR works for now as well!

PR #2362, opened 2021-02-16
Load clinical and research gene fusion reports on case upload or load/update them later using the CLI

### LINKED ISSUE #2343: Add transcript fusion report link to reports section in existing case page
**Background**
When analysing an individual it is common to start with a DNA analysis. However, often a causative variant is not found and the same individual also are subjected to a RNA analysis. 

**Challenge**
Currently, there is no way to combine the output from the standard DNA analysis with the selected output from a RNA analysis in Scout. It would be great to be able to update the DNA case view with information on transcript fusions in the form of a pdf report that can be viewed/downloaded from Scout.

**Suggested solution**
- Add the option in `scout case update` to also be able to upload a link pointing to a transcript fusion report from a secondary analysis to an already existing case.
- Also support this from the scout load config
- Add the optional transcript fusion report to the reports section in an already existing DNA analysis case.

northwestwitch:
Fix #2343.

- This PR allows loading gene fusion reports (clinical and research) along with other info on case load, when the key/value 'gene_fusion_report' and 'gene_fusion_report_research' is present in the Scout yaml config file.
- Allows also to add/update the report of an existing case in the database by running the command: `scout load gene-fusion-report`:

image

  

**How to test**:
- [ ] Test on the demo case by running `scout setup demo` or uploading the case again with `scout --demo load case path_to_config_file -u`
- [ ] Try to update the clinical fusion report with any PDF of your choice by following the cli instructions
- [ ] Try to load a research fusion report (any PDF of your choice) from the command line by following the cli instructions
- [ ] Make sure that the links to these 2 reports are correctly visualized on the case sidebar and that clicking on them shows the files

**Expected outcome**:
1. The demo case should show a link to the fusion report on the side bar. Clicking on this link should download the report, given that the file exists in that path on the disk

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
This is ready for review

northwestwitch:
image

And when you click on the link you get this:

image

henrikstranneheim:
Beautiful!

northwestwitch:
> Beautiful!

Thanks I've copied it from Arriba. But is it like this for real?

henrikstranneheim:
> > Beautiful!
> 
> Thanks I've copied it from Arriba. But is it like this for real?

yes

northwestwitch:
Ok, fixed this PR with demo files I got from @jemten. Ready for review again!

jemten:
Oh, I just struck me. There will be one report per sample in the case. Sorry if this messes things up

northwestwitch:
> Oh, I just struck me. There will be one report per sample in the case. Sorry if this messes things up

OK, can't we somehow merge these reports in a unique pdf?

jemten:
Possibly. But since the reports doesn't contain any sample names to indicate which fusion belongs to which sample it won't be as easy as just concatenating them.

northwestwitch:
> Possibly. But since the reports doesn't contain any sample names to indicate which fusion belongs to which sample it won't be as easy as just concatenating them.

I see. Then it's more complicated, yes. Let's think about it!

northwestwitch:
I'm working on a program that will merge and bookmark the files after writing on them the name of the original PDF. Hopefully I'll be finished with it later today

henrikstranneheim:
> I'm working on a program that will merge and bookmark the files after writing on them the name of the original PDF. Hopefully I'll be finished with it later today

Cool! Should we run this in MIP in the arriba step or where are you planning to inject it?

northwestwitch:
> Cool! Should we run this in MIP in the arriba step or where are you planning to inject it?

I think it should be part of the pipeline, before the injection. How do you prefer to feed the pdf files to the program:
- invoke the program and give it the list of all files to merge
- invoke the program with the directory where these files are located (in this case it will merge everything that finds with a .pdf suffix, after ordering them by name)

?

henrikstranneheim:
Let's go with supplying a list of files to merge via the CLI

jemten:
> Let's go with supplying a list of files to merge via the CLI

I agree

northwestwitch:
@henrikstranneheim @jemten  this one: https://github.com/northwestwitch/pdfmerger

I could use some tester with the arriba files and some improvement suggestion!

jemten:
Great work @northwestwitch. 
I'll check it out and add it to the pipeline

northwestwitch:
@jemten @henrikstranneheim I've created a Dockerfile to be pulled and runned on the fly to merge outfiles from arriba (or any other pdf report).

Command:

```
docker run -d --rm \
-v /home/username/documents:/home/pdfmerger/data \
northwestwitch/pdfmerger --orientation landscape \
-f /home/pdfmerger/data/infile1.pdf -f /home/pdfmerger/data/infile2.pdf \
--outfolder /home/pdfmerger/data (--outfile outfilename.pdf)
```

See more info [here](https://github.com/northwestwitch/pdfmerger). 

By default the program doesn't write the file names on each page of each file before merging. If you wish to add the names you can use the additional tag `--write-filenames`

northwestwitch:
This is now fixed according to the discussion we had yesterday @henrikstranneheim and @jemten : 
- You can load clinical gene fusion report and research gene fusion report directly when you load the case (keys: `gene_fusion_report` and `gene_fusion_report_research`)
- You can update a case with the path one of the 2 reports successively. 

On the case page:

image

scout/build/case.py
```diff
@@ -1,9 +1,8 @@
 import logging
-
 from datetime import datetime
 
+from scout.constants import CUSTOM_CASE_REPORTS
```
dnil, line 4: 👍

scout/commands/load/base.py
```diff
@@ -37,4 +38,5 @@ def load():
 load.add_command(delivery_report_command)
 load.add_command(cnv_report_command)
 load.add_command(coverage_qc_report_command)
+load.add_command(gene_fusion_report_command)
```
dnil, line 41: We probably want a load_custom_report at some point (major =), but this is perfect now!

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -449,34 +460,69 @@ <h5 class="modal-title" id="archiveCaseLabel">Confirm archive case</h5>
...
 {% endmacro %}
 
+{% macro gene_fusion_reports(institute, case) %}
+{% set fusion_reports = ["gene_fusion_report", "gene_fusion_report_research"] %}
+{% for report_type in fusion_reports %}
+  {% if case[report_type] %}
+    <div href="#" class="bg-dark list-group-item text-white">
+      <div class="d-flex flex-row flex-fill bd-highlight">
+        <div class="text-nowrap">
+          <span class="menu-collapsed ml-2">{{ report_type|replace("_report","")|replace("_"," ")|capitalize }}</span>
```
dnil, line 500: 😸

scout/server/blueprints/cases/views.py
```diff
@@ -873,6 +873,18 @@ def cnv_report(institute_id, case_name):
     return send_from_directory(out_dir, filename)
 
 
+@cases_bp.route("/<institute_id>/<case_name>/gene-fusion-report/<report_type>")
+def gene_fusion_report(institute_id, case_name, report_type):
+    """Download gene fusion report"""
+    _, case_obj = institute_and_case(store, institute_id, case_name)
+    report_path = case_obj.get(report_type)
```
dnil, line 880: This is not really that dangerous, I think - as it is sort of limited to what is on the case_obj. And we have not really gone over everything like it, but since Im reading. I suppose you could if you really, really wanted add a path into like the synopsis or tissue type fields, then retrieve it by asking for report_type="synopsis" or such. Perhaps limit it to the valid keys, eg the report ones from constants?

dnil (approved):

scout/commands/load/base.py
northwestwitch: I totally agree, it would be so nice to generalize and allow people to provide and visualize any report they want. We could already start thinking about it

scout/server/blueprints/cases/views.py
northwestwitch: Especially true if we start generalizing the available reports. Very good suggestion, I'll fix it!

dnil:
Did you want a test on this as well? Kind of looks like that is already done above, but if you make a minor change now, I could test afterwards.

northwestwitch:
> Did you want a test on this as well? Kind of looks like that is already done above, but if you make a minor change now, I could test afterwards.

I'll do that change in the gene_fusion_report view. Feel free to give it a try later, thanks!

northwestwitch:
Tested after the change, it works: 

image

scout/server/blueprints/cases/views.py
```diff
@@ -873,6 +873,18 @@ def cnv_report(institute_id, case_name):
...
+    if report_path is None or report_type not in CUSTOM_CASE_REPORTS:
```
dnil, line 881: 💯

dnil (approved):

PR #2368, opened 2021-02-17
Added GENS integration

mhkc:
This PR integrates Gens to scout. When the `GENS_HOST` variable i set in the config links to Gens are displayed in case and structural variant views.

You could optionally verify the integrity of the links by installing Gens v1.1.1 to a local vm. I would recommend using the docker image.

**How to test**:
1. goto cust000/643594, verify that "View CN profile" button is absent
2. goto cust000/643594/sv/variants/77d69d4d78a8e272365bdabe4f607327
3. verify that copy number profile section is absent
4. set GENS_HOST and/or GENS_PORT variables in configuration
5. goto previous pages and verify that button is present

**Expected outcome**:
The button is only present when the HOST is configured.

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch (changes requested):
Works fine!
Button show on case view: 

image


SV variant page:
image

I have a number of small suggestions but it's already working fine.

What we should fix before merging:

- One or more tests are not passing. I suggest moving the get_gens_information to scout/utils, and make it as a class to be called from the app as an extension. Otherwise the tests not passing could be fixed so the tested functions are called while being in the app context

- If you go for registering gens as an extension, it would be nice to write a test, showing that the extension contains the right values if the config file provides the right params. I could help you with that.

northwestwitch:
There is a test failing but it's not related to the code in this PR. In fact it's trying to find a resource that is not available:

http://compbio.charite.de/jenkins/job/hpo.annotations/lastSuccessfulBuild/artifact/util/annotation/genes_to_phenotype.txt.

I'll try to patch that test, it's not good that it looks for real resources outside Scout.

northwestwitch (approved):
Good job! 🥇 
I think this is fine to merge as soon as the tests pass!

PR #2371, opened 2021-02-18
Add rerunner integration

mhkc:
This PR integrates a service for toggling a reanalysis of a sample using updated pedigree metadata

**How to test**:
1. goto cust000/643594, verify that "correct metadata" button is absent
2. set RERUNNER_HOST and/or RERUNER_PORT and RERUNNER_API_KEY  variables in configuration
3. goto previous pages and verify that button is present

**Expected outcome**:
The correct metadata button should only be displayed when the `RERUNNER_HOST` is set in the config

**Review:**
- [x] code approved by CR
- [x] tests executed by MJ, CR


mhkc:
I think we could ignore the CodeFactror warnings since they relate to js functions being called in a jinja templates

northwestwitch:
> I think we could ignore the CodeFactror warnings since they relate to js functions being called in a jinja templates

I agree!

northwestwitch:
Also this one, has tests that fail on the download HPO resources, but it's weird because the resource is mocked in the test

northwestwitch:
Unfortunately I won't be able to review this PR soon as I took a vacation next week. I could review when I'm back if nobody else picks it fist

northwestwitch:
Whoops, merging the Gens PR messed up this one, sorry!

mhkc:
On second thought. This solution relies on synchronous calls to an external service through an rest API. If the connection from the service to the pipeline host is slow it will freeze the GUI until either it times out or it gets a response. Perhaps it would be better to handle these types of requests by using a message-queue pattern instead.

What do you think? 
Do you have any preference on the type of system?

I have some experience with RabbitMQ and Celery.

dnil:
> On second thought. This solution relies on synchronous calls to an external service through an rest API. If the connection from the service to the pipeline host is slow it will freeze the GUI until either it times out or it gets a response. Perhaps it would be better to handle these types of requests by using a message-queue pattern instead.
> 
> What do you think?
> Do you have any preference on the type of system?
> 
> I have some experience with RabbitMQ and Celery.

We have eyed Kafka and RabbitMQ but not locked in on anything yet. Definately agree - perhaps take that as a point on that next meeting?

northwestwitch:
Nice job!

Testing the PR. Basically you allow users to edit the case details:

image

image

And then sends a request to the rerunner.

This needs a bit of more thinking I am afraid because if our Stockholm users touch these buttons and edit the case it's going to result in a chain of **complains that the reanalysis doesn't work**:
image

**And that the changed individual features don't reflect the analysis (because of course the case was not re-analysed).**

The easiest way out would be to display the "correct metadata" only if the config file contains RERUNNER_HOST.
But I feel like this would need to be changed in the future because of [this](https://github.com/Clinical-Genomics/scout/issues/2401).

Things that might greatly improve the PR and make it directly compatible with us, I think:

- Add a key/value in scout settings that describes the type of rerun service, for instance:
```
RERUN_TYPE = "link" # rerun buttons redirects to another webpage that requires user login
RERUN_TYPE = "api" # rerun executed directly using an api
```

- Move the changes to the case individuals from the case page to a middle layer (the modal?) and display this form if the  RERUN_TYPE = "api". Then send the data to the reanalysis view. This could be fixed in a dedicate modal to use if RERUN_TYPE = "api".

- Otherwise call the modal we currently use for the reruns when you click on the button on the side panel (and the button on the side panel might actually be removed).

What do you think?

Good point with the limitation of using synchronous calls to the api. I'm not really familiar with queue systems. Yes, perhaps let's have a discussion on next meeting.

dnil:
> The easiest way out would be to display the "correct metadata" only if the config file contains RERUNNER_HOST.
> But I feel like this would need to be changed in the future because of [this](https://github.com/Clinical-Genomics/scout/issues/2401).
> 
> Things that might greatly improve the PR and make it directly compatible with us, I think:
> 
> * Add a key/value in scout settings that describes the type of rerun service, for instance:
> 
> ```
> RERUN_TYPE = "link" # rerun buttons redirects to another webpage that requires user login
> RERUN_TYPE = "api" # rerun executed directly using an api

Sounds super! What do you think @moonso and @patrikgrenfeldt: would it make sense to add a compatible api on cgweb - or just redirect for now and set up queues first?

mhkc:
Sounds good. As this is not vital at for us at the moment I think we should think this through and decide on a queuing system and a simple api definition based on a few common use cases.

Perhaps it would be worth making some template service with containing the boilerplate code for basing other services on.

northwestwitch:
@mhkc I think this PR has really useful code and I'd like to try to merge it. What do you think about my suggestions, basically postpone the case editing to an additional step and not directly on the case page?

mhkc:
That sounds like a good idea. Should we also try to discuss the message broker system in a upcoming meeting so we can move forward with this PR?

Would the GMS interpretation GUI meeting group be suitable or should I join one of your scout dev meetings?

dnil:
> That sounds like a good idea. Should we also try to discuss the message broker system in a upcoming meeting so we can move forward with this PR?
> 
> Would the GMS interpretation GUI meeting group be suitable or should I join one of your scout dev meetings?

I don't know what that is, so you would be a better judge; depends on who is on that group? Is that the group where you discussed using Scout nationally @henrikstranneheim? I would include at least also @moonso and @Mropat from Stockholm regarding choice of broker/queue!

Regardless, this will be a larger project with several systems touched, even if we lead with only a few. For this PR, I would suggest staying on a direct REST connection for now, but make sure not to wait for long computation on the other end; like in the re-runner service respond back that the case will be restarted first, and terminate the connection, then actually do the slow stuff?

northwestwitch:
@mhkc let's dedicate some time to merge this one next week

mhkc:
@northwestwitch I have some time from now on

scout/server/blueprints/cases/views.py
```diff
@@ -748,6 +759,43 @@ def rerun(institute_id, case_name):
...
+            url,
+            params=payload,
+            json=edited_metadata,
+            timeout=rerunner.timeout,
+            headers={"Content-Type": "application/json"},
+            auth=auth,
+        )
+        if resp.status_code == requests.codes.ok:
+            LOG.info(f"Reanalysis was successfully started; case: {case_name}")
+            flash(f"Reanalysis was successfully started; case: {case_name}", "info")
```
northwestwitch, line 786: Apart from this flash message that you see only once, there is no saved information that the rerun is pending at the case level. What about triggering an event that will update the case "rerun_requested" and set it to True, and will also display which user ordered a rerun? Consider calling [request rerun](https://github.com/Clinical-Genomics/scout/blob/5485aa5f9731bffa02fdfebbcd42783d50350071/scout/adapter/mongo/case_events.py#L338) from here.

northwestwitch (changes requested):
Tested (without the rerun API) and looks like it's working really nicely, at least from the Scout side.

I think your solution is the way to go also for us in Stockholm, once we have the rerunner API in place of course! :)

I have a few suggestions for improvement (like moving the code from the view to the controller, and reusing an existing module to send the request).

I think it would also be useful to trigger a case rerun event that will be recorded in the case events, so other users will know who ordered a rerun and when.

It would also be nice to write some tests for the new code, mocking the the rerunner API. I could help with that if you want!

scout/server/blueprints/cases/controllers.py
```diff
@@ -794,6 +795,39 @@ def rerun(store, mail, current_user, institute_id, case_name, sender, recipient)
...
 
+def call_rerunner(store, institute_id, case_name, metadata):
+    """Call rerunner with updated pedigree metadata."""
+    LOG.info("Building query for reanalysis")
+
+    # define the data to be passed
+    payload = {"case_id": case_name, "sample_ids": [m["sample_id"] for m in metadata]}
+
+    cnf = rerunner.connection_settings
+    url = f"http://{cnf.get('host')}/v1.0/rerun"
```
northwestwitch, line 806: This line makes the service really customized for your facility. When we are going to have our rerunner service in place we will be forced to have its url ending in /v1.0/rerun otherwise it won't work. I think url should be what you pass in the config file and set in the rerunner extensions.

northwestwitch (changes requested):
Nice!

I've got just a couple of comments, I'm writing some tests in the meanwhile!

scout/server/blueprints/cases/controllers.py
northwestwitch: The simplest way to fix this would be to pass in the config file the complete url: "http://someurl:port/v1.0/rerun", so you won't need to fix the port in the rerunner extensions settings and it would work directly

scout/server/blueprints/cases/views.py
```diff
@@ -748,6 +752,26 @@ def rerun(institute_id, case_name):
...
+def reanalysis(institute_id, case_name):
+    """Toggle a rerun by making a call to RERUNNER service."""
+    edited_metadata = json.loads(request.form["sample_metadata"])
+
+    try:
+        controllers.call_rerunner(store, institute_id, case_name, edited_metadata)
+    except ReadTimeout as err:
+        msg = f"Error processing request: {err.__class__.__name__}"
+        flash(msg, "danger")
+    except RerunnerError as err:
```
northwestwitch, line 765: It's looking like this now: 

image

scout/server/app.py
```diff
@@ -114,10 +114,16 @@ def configure_extensions(app):
         LOG.info("MatchMaker Exchange enabled")
         extensions.matchmaker.init_app(app)
 
+    if app.config.get("RERUNNER_HOST") and app.config.get("RERUNNER_API_KEY"):
+        LOG.info("Rerunner service enabled")
+        # setup LoqusDB
```
northwestwitch:
```suggestion
        # setup Rerunner
```

northwestwitch:
Tested (without the API) and looks fine. I think we are good with this one and I'd merge as soon as it passes the tests!

scout/server/blueprints/cases/controllers.py
```diff
@@ -794,6 +795,40 @@ def rerun(store, mail, current_user, institute_id, case_name, sender, recipient)
...
+    resp = requests.post(
+        url,
+        params=payload,
+        json=metadata,
+        timeout=rerunner.timeout,
+        headers={"Content-Type": "application/json"},
+        auth=auth,
+    )
+
+    if resp.status_code == 200:
```
northwestwitch, line 818: I changed it back to the previous code, since codefactor didn't like the if True:

scout/server/blueprints/cases/controllers.py
```diff
@@ -794,6 +795,40 @@ def rerun(store, mail, current_user, institute_id, case_name, sender, recipient)
         LOG.error("Cannot send rerun message: no recipient defined in config.")
 
 
+def call_rerunner(store, institute_id, case_name, metadata):
+    """Call rerunner with updated pedigree metadata."""
+    # define the data to be passed
+    payload = {"case_id": case_name, "sample_ids": [m["sample_id"] for m in metadata]}
+
+    cnf = rerunner.connection_settings
+    url = cnf.get("entrypoint")
```
northwestwitch, line 804: 👍🏻

northwestwitch (approved):
It's looking good now, let's merge it!

northwestwitch:
@mhkc I haven't merged this yet since I'm having a little hard time to write a test for the reanalysis endpoint. In the beginning I was thinking that the form data passed was the usual dictionary, but it looks like it's a string representation of a dictionary? The test will fail until I find a solution. Today anyway!

PR #2373, opened 2021-02-18
Load STRs regardless of rank score threshold

dnil:

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
This is the one based on 4.29 right? let's not update the branch with current master!

dnil:
Actually, Im not so sure we should merge it at all? The branch does exist, so perhaps we should rather create a release from it? Sorry for being slow..

northwestwitch (changes requested):
No we shouldn't merge, you are right, but we need to fix the scout version as well before creating the release from this

No worries I'm slower! :)

dnil:
Tested on stage:
Screenshot 2021-02-19 at 08 32 07

dnil:
The speed part is very context sensitive I'd say - typically we tend to end up with the same solution around the same second?

northwestwitch (approved):
Why not version 4.29.1?

dnil:
Because you are on vacation next week!

dnil:
(But the hotfix could theoretically be renamed 4.29.1 and we bump this to 4.29.2 when it's time..)

dnil:
Why not actually - it is just a name anyway and not really a merge graph representation. It is better in line with semantic versioning which would otherwise rather ask for calling this 4.30-rc or something. I like. I’ll update it.

northwestwitch:
The can be either closed or merged. Perhaps close it since it's based on an old master, but doesn't make much difference I guess. You choose @dnil

PR #2372, opened 2021-02-18
Load STRs regardless of rank score threshold

dnil:
This PR adds a functionality or fixes a bug.

**How to test**:
1. try loading an STR file produced with Stranger 0.7 and note how few variants are loaded if rank score threshold is at 5 or higher
2. try with patch version and find variants loaded

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #2374, opened 2021-02-19
Update master version with hotfix.

dnil:
This PR retroactively marks a new Scout release. We apply semantic versioning. This is a hot fix patch release since it does a minor internal change to how variants are loaded, specifically allowing all STR variants to load regardless of the case rank score threshold. 

This should probably have been included in PR #2372 already. 😊 

**How to test**:
1. no test, only doc

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch (approved):
👍

PR #2388, opened 2021-02-23
Freeze actions ubuntu version during rollout

dnil:

Unblocking stalled PR Actions due to warning during ubuntu-20.04 rollout.

Screenshot 2021-02-23 at 16 52 30

**How to test**:
1. automatic test on GitHub Actions

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/.github/workflows/tests_and_cov.yml
+++ b/.github/workflows/tests_and_cov.yml
@@ -5,7 +5,7 @@ on: ["push", "pull_request"]
 jobs:
   build:
     name: Coverage
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         python-version: [3.7]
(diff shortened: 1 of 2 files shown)


dnil:
By the way, Python 3.8 is now default on Ubuntu 20.04. Perhaps time to change - perhaps do 3.6 and 3.8. Thoughts?

moonso:
@dnil I think it can cause problems on clinical-db with python 3.8, I've tried that with another package. As soon as we move to vm ot would not be any problems

moonso (approved):
Feels like almost all GitHub actions will have this warning, perhaps worth to wait it out? But that is your call!

dnil:
Sure, then we keep it for the tests as well. Going three way version matrix seems a little slow.. We will soon need more mongo versions as well.

dnil:
And yeah, likely: but thanks, this is good - if we need to merge sth it is now possible again. And 20.04 is likely going to be latest for quite some time now.

moonso:
Another way to reduce the number of test cycles is to remove `"push"` from 

```yaml
on: ["push", "pull_request"]
```

In that way it will still run for all pushes but only for opened PRs, which is most likely what one wants.

dnil:
It did auto resolve, so closing, but was nice to have at hand. Good tip with the `pull_request`only; I remember a lot of coveralls confusion spawns with those changes, so lets do that when its time to add in more env to the matrix.

PR #2395, opened 2021-02-24
Update variants list view layout for cancer cases

mhkc:
**Changes**
•	Split Gene:Transcript:Exome to separate columns
•	Moved “observed in” badges to separate column
•	Removed rank column
•	Display both HGVSc and HGVSp
•	Display only the gene with most severe consequence
•	Fix “dismissed selected variants” to bottom window

cancer variants list view

**How to test:**
    1. load demo cancer case
    2. Navigate to, cust000/cancer_case/cancer/variants?variant_type=clinical

**Expected outcome**:
 - Should display as separate column
 - Ensembl link should work

**Review:**
- [ ] code approved by
- [ ] tests executed by


mhkc:
I think we can disregard the coverall report as I doesnt relate to changes I have made.

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -75,17 +77,38 @@
...
+                {% endif %}
+              {% endfor %}
+            </td>
             <td>
               <a href="{{ url_for('variant.variant', institute_id=institute._id, case_name=case.display_name,
                                   variant_id=variant._id, cancer='yes') }}">
-              {% for annotation in variant.canonical_transcripts %}
-                <div>{{ annotation }}</div>
-              {% endfor %}
+                {% for gene in variant.genes %}
```
dnil, line 103: This sort of gets untenable for multiple genes. You will get a bunch of gene symbols in one column on the far left, and a bunch of hgvs strings on the right. Pairing is anyones guess. I would suggest either adding in a "symbol:" like for the consequences, or to save space keep a strict line order within the cells so one can read the row of gene->transcript->exon->hgvs->consequence. Perhaps group them in columns next to one another as well to make it easy to read?

dnil: Screenshot 2021-02-25 at 12 07 54
And then stretch the window size to watch the symbols and transcripts reorder.. 😊

dnil (changes requested):
Good start - that column is not visually appealing!

mhkc:
To simplify the variant overview further we should consider,
1. only display the canonical transcript for a given gene as it would be the transcript most clinicians would be most familiar with. All transcripts should be displayed in the detailed view.
2. only display the gene with the variant with greatest negative consequence. There are many times several genes are inconsequential and just adds noise. 
Capture
3. Add HGVSc and HGVSp designations in the same column

dnil:
1. Our clinicians are generally speaking blissfully ignorant about the difference between canonical, primary or disease associated transcripts, transcript versioning and ge. They want "The NM Transcript". To an extent it becomes our role to feed them good stuff. Canonical is practical and a good go-to, primary is best for reporting and disease associated is what we should all strive for in the end.
2. hm, that is slightly scary; given that the gene from the panel / of interest could only have like the second worst consequence. But anyway, that may be good enough for click-bait to get the user to actually check the variant page. A tooltip/popover could be considered as well?
3. 💯

bjhall:
Great discussion here @dnil and @mhkc. I'll allow myself to make a suggestion based on your discussion and our previous experiences.

1. Select one gene and one transcript for every variant. Yes it is theoretically possible that this obscures valuable information (suggestions for mitigations below), but I think we need to apply some heuristics to make the list view as approachable and easily digestable as possible. Users can always click the variant to get all the details.
    * Select gene with most severe consequence
    * Select canonical/primary/MANE select/disease associated transcript. This is more tricky, and I don't have a perfect solution. If we ignore ancient genome builds such as GRCh37 I think it would be fairly safe to pick either "MANE Select" or "MANE Plus Clinical" (the one with most severe consequence) or the disease associated transcript if defined.
 2. In the gene column: show the selected gene on top. However, also include the other genes, but visually indicate that they are "secondary" by for instance graying the text. This would mitigate the problem that @dnil brought up where the gene in the panel may not be the one with the most severe consequence. Potentially one could even show the additional information for these genes (HGVS, consequence etc) on hover. 
 3. Remove the transcript column. I think users typically don't want to see this, but rather trust that we select the relevant transcript (or check the variant details in unclear situations)
 4. Show HGVSc and HGVSp only for the selected gene/transcript. Preferrably in the same column.
 5. Show the consequence for the selected transcript for either only the selected gene or possible all genes (again graying the non-selected ones). 


Unless anyone disagrees with anything else I think the main discussion point is how to select the transcript. I suppose we still have to consider GRCh37... We'll probably have to use different selection criteria for GRCh37 and 38.

bjhall:
Related:

I think it would be nice to show one-letter amino acids for HGVSp in the list view. Both because the three-letter strings become unreasonably long for slightly larger indels and I also think most clinicians (and myself) are so used to seeing recurrent variants such as `V600E` or `T790M` that their brains have developed pattern matching systems for them, which work less efficiently if shown `Val600Glu` or `Thr790Met`.

Unless you already have code for converting this in Scout here is the template filter I've used in coyote.
```
@app.template_filter()                                                                                                                                                                                             
def one_letter_p(st):                                                                                                                                                                                              
    aa = {'Cys': 'C', 'Asp': 'D', 'Ser': 'S', 'Gln': 'Q', 'Lys': 'K',                                                                                                                                              
          'Ile': 'I', 'Pro': 'P', 'Thr': 'T', 'Phe': 'F', 'Asn': 'N',                                                                                                                                              
          'Gly': 'G', 'His': 'H', 'Leu': 'L', 'Arg': 'R', 'Trp': 'W',                                                                                                                                              
          'Ala': 'A', 'Val': 'V', 'Glu': 'E', 'Tyr': 'Y', 'Met': 'M', 'Ter': '*' }                                                                                                                                 
                                                                                                                                                                                                                   
    pattern = re.compile('|'.join(aa.keys()))                                                                                                                                                                      
    new = pattern.sub(lambda x: aa[x.group()], st)                                                                                                                                                                 
    return new   
```

dnil:
Thanks @bjhall!
- the selection of genes must at least include ones on the search criteria: we would have geneticists in panic every time they find a gene on the results that is not on they panel. We have learned to cope with multiple genes, but it is not ideal. Graying out or hovering out those that don't match the query would be extremely neat, but a little technically complex.
- removing transcripts sounds great - I dont miss it on the RD view - but ask at least one experienced cancer analyst first
- there would also be some merit to selecting the primary transcript (which is what you want on the report anyway). This will in the future converge towards one of the MANE - HGVS has a collaboration in works - but they are not quite there yet.

dnil:
> I think it would be nice to show one-letter amino acids for HGVSp in the list view. Both because the three-letter strings become unreasonably long for slightly larger indels and I also think most clinicians (and myself) are so used to seeing recurrent variants such as `V600E` or `T790M` that their brains have developed pattern matching systems for them, which work less efficiently if shown `Val600Glu` or `Thr790Met`.
>
> Unless you already have code for converting this in Scout here is the template filter I've used in coyote.

At least my pattern matching works differently for different genes. There is a lot of historical context, but I much agree with saving space. 

There is a similar think for one letters in the link.py context. Feel free to introduce yours / merge them! Also, there is at least one VEP plugin for that, so we could just read it off annotation, although we don't have it enabled at the moment.

mhkc:
Great discussion.

So  can we settle on this functionality of which genes to display?
- Show the gene / variant with the most severe consequence as primary.
- If the primary gene is not in the selected gene panel, display these underneath as secondary variants
It seems to me that this solution would reduce clutter while not removing essential information.

Also, what there any reason for displaying both rank and score as rank is the sort order of the score and the score is the more informative of the two metrics.

dnil:
Sounds good!

The reason for showing both was probably pedagogical. I find them redundant. We toyed with removing it in another PR that sort of blocked waiting for db server migration. Most ok to hide it here!

bjhall:
> the selection of genes must at least include ones on the search criteria

Would it be a problem if the queried gene would be one of the "gray" ones and the HGVS information shown would be for a different "primary" gene? It would take some getting used to maybe, but if you see that the gene is gray you will know that you should click or hover to see more details. 

> but ask at least one experienced cancer analyst first

In Lund I already know they don't care about them.

bjhall:
>there would also be some merit to selecting the primary transcript (which is what you want on the report anyway). This will in the future converge towards one of the MANE - HGVS has a collaboration in works - but they are not quite there yet.

Where can I find the the primary transcript definitions? I can't seem to find them by googling... I'd like to make a comparison between primary and MANE.

mhkc:
> Would it be a problem if the queried gene would be one of the "gray" ones and the HGVS information shown would be for a different "primary" gene? It would take some getting used to maybe, but if you see that the gene is gray you will know that you should click or hover to see more details.

We could include all information but greyed out although I dont think there would be an issue if we didn't. Since we would always display the most severe variant, wouldn't the variant be investigated regardless of what the secondary gene would be? 
Or are there certain variants that are frequent false positives that you would normally just disregard that you would not be able to spot?

bjhall:
> 
> 
> > Would it be a problem if the queried gene would be one of the "gray" ones and the HGVS information shown would be for a different "primary" gene? It would take some getting used to maybe, but if you see that the gene is gray you will know that you should click or hover to see more details.
> 
> We could include all information but greyed out although I dont think there would be an issue if we didn't. Since we would always display the most severe variant, wouldn't the variant be investigated regardless of what the secondary gene would be?

I think the problem Daniel is alluding to is when the queried gene (through gene panel or search filter) is not the one with the most severe consequence. It could be confusing when you search for one gene, but get results for another. However I think showing the other genes grayed out would solve that from my perspective.

mhkc:
What do you think about moving the database related flags out of the Pop Freq column and either into a dedicated "Observed" column?

We could also include a "loqusdb" flag if the variant has been previously seen in another sample.

dnil:
> What do you think about moving the database related flags out of the Pop Freq column and either into a dedicated "Observed" column?
> 
> We could also include a "loqusdb" flag if the variant has been previously seen in another sample.

Sounds good for the long haul: I imagine we will get a lot more of these found-in-significant-db annotations for these!

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -73,21 +73,22 @@
...
+            </td>
+            <td>{{ gene_cell(variant) }}</td>
             <td>
               <a href="{{ url_for('variant.variant', institute_id=institute._id, case_name=case.display_name,
                                   variant_id=variant._id, cancer='yes') }}">
-              {% for annotation in variant.canonical_transcripts %}
-                <div>{{ annotation }}</div>
-              {% endfor %}
+                <div>{{ variant.primary_gene.hgvs_identifier }}</div>
+                <div>{{ (variant.primary_gene.hgvsp_identifier or '') |url_decode }}</div>
```
dnil, line 84: Hm, perhaps we should fix this "None" (large indel from demo case) when it is the primary link to the variant:
Screenshot 2021-04-19 at 18 05 35

dnil (changes requested):
Overall very nice! This will be much cleaner on the cancer view, and ripple over with good stuff for the rare track as well.

In detail, there were some partly unintended consequences with changes in macros used in display for other variant types. I dont have a strong opinion on the best kind of polymorphism here. jinja templates do support a form of inheritance, with polymorphism, but that may be overkill. Simply trying to call all with the same macro and checking for category could work, or use separate ones. The most common solution in the rest of the code is to keep a common base function and have category specific functions/macros call that for the reused parts and add there own flair for what is extra.

I have a slight worry about the secondary_gene, but lets try it out - and perhaps you have already?

I'll pick "request changes" as the rare variants list is currently broken, and it might be good with another check before merging. I think we can merge as soon as it is back on its feet again. The remaining actual cancer page fixes - e.g. adding back comments - could be considered minor.

dnil:
Oh, and don't forget the failing tests obviously! 😊

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -73,21 +73,22 @@
           {% else %}
             <tr>
           {% endif %}
-            <td><input type="checkbox" value="{{variant._id}}" name="dismiss" onclick="enableDismiss()"></td>
-            <td class="text-left">{{cell_rank(variant, institute, case, form, manual_rank_options)}}</td>
```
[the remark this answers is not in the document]
mhkc: What comments are you referring to?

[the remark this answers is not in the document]
dnil: Ah, not code comments: comments from users - a type of event associated with variants that contain free-text information about variants either specific to the case or globally for the variant. They were previously shown in the cell_rank box and are now missing from the cancer variants? Like here: https://github.com/Clinical-Genomics/scout/blob/d94efc68fae47334694a6c1a46f74600fcf7b5ef/scout/server/blueprints/variants/templates/variants/utils.html#L712. A suggestion would be to break those into a macro comment_badges that could be included in different columns?

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -119,10 +150,16 @@
     <span class="badge badge-dark">MITOMAP</span>
     <br>
   {% endif %}
+  </div>
```
dnil, line 153: Ah, but then the tooltip will not be on anything really if there is no badge in the cell? Before there was at the least an `freq else: ~`. I tested and could not trigger the tooltip for for rare or cancer variants without badges on Safari at least?

scout/server/blueprints/institutes/templates/overview/gene_variants.html
```diff
@@ -1,5 +1,5 @@
 {% extends "layout_bs4.html" %}
-{% from "variants/components.html" import gene_cell, frequency_cell %}
+{% from "variants/components.html" import gene_cell, frequency_cell_general %}
```
dnil, line 2: 🥇 Good Enough - better than many other solutions we have!

scout/server/blueprints/variants/controllers.py
```diff
@@ -362,6 +369,21 @@ def parse_variant(
     variant_obj["str_source_link"] = str_source_link(variant_obj)
     # Format clinvar information
     variant_obj["clinsig_human"] = clinsig_human(variant_obj) if variant_obj.get("clnsig") else None
+    # Set the gene with most severe consequence as being representavie
```
dnil, line 372:
```suggestion
    # Set the gene with most severe consequence as being representative
```

scout/server/blueprints/variants/controllers.py
```diff
@@ -581,11 +603,25 @@ def cancer_variants(store, institute_id, case_name, variants_query, variant_coun
...
+    # get gene panel lookup table
+    gene_panel_lookup = store.gene_to_panels(case_obj)  # build variant object
     variants_list = []
-
     for variant in variant_res:
-        elem = parse_variant(store, institute_obj, case_obj, variant, update=True)
-        variants_list.append(elem)
+        # parse variant information
+        variant_obj = parse_variant(store, institute_obj, case_obj, variant, update=True)
+        # if one or more gene panles is selected assigne the gene present
```
dnil, line 612:
```suggestion
        # if one or more gene panels are selected, assign the gene present
```

scout/server/blueprints/variants/controllers.py
```diff
@@ -581,11 +603,25 @@ def cancer_variants(store, institute_id, case_name, variants_query, variant_coun
 
     variant_res = variants_query.skip(skip_count).limit(per_page)
 
+    # get gene panel lookup table
+    gene_panel_lookup = store.gene_to_panels(case_obj)  # build variant object
```
dnil, line 607: Hm, isn't this still relevant though:
https://github.com/Clinical-Genomics/scout/pull/2395#discussion_r615918310
It is a very good start as is, but I imagine the manual gene symbol search box and HPO / dynamic panel users will be confused once in a while when their searched variant is shown with a different gene name?

dnil (changes requested):
Super! Almost there!

scout/server/blueprints/variants/templates/variants/components.html
mhkc: Good point. I have divided the information over the pop freq column and observed column

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -43,27 +43,28 @@
...
+              <th title="Genomic coordinate" style="width:10%">Chr pos</th>
+              <th title="Population frequency">Pop Freq</th>
+              <th title="Observed" style="width:7%">Observed</th>
               <th title="Variant type">Type</th>
               <th title="Functional consequence annotation" style="width:18%">Consequence</th>
               <th data-toggle="tooltip" data-placement="top" title="Tumor alt. AF. &#013; Alt. allele count | Ref. allele count">Tumor</th>
               <th data-toggle="tooltip" data-placement="top" title="Normal alt. AF. &#013; Alt. allele count | Ref. allele count">Normal</th>
             </tr>
       </thead>
+      {{ variants[5] }}
```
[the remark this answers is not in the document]
mhkc: 🤦‍♂️

scout/server/blueprints/variants/templates/variants/cancer-variants.html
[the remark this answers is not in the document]
mhkc: I have refactored the badges into `indicators.html` which could be populated more in the future. Now the assessments of different kinds show up the the assessments column (renamed tier)

dnil (approved):
I had a few issues when testing it locally, but nothing of conceptual magnitude, and all fixable. Since we've had a few cycles of this one, and its looking good already I say lets merge it and deal with corrections. I still had two things: one was the representative gene from gene panel and not from hpo etc, one was page resetting to 1 when dismissing from a page further on. The latter is not your fault though, so lets just make an issue or two with them and move on!

PR #2397, opened 2021-02-25
Fix STR hover reference size

dnil:

STR reference size was duplicated on the hover, and also showing the wrong number. It has a column of its own, so we remove it for now.
Screenshot 2021-02-25 at 12 33 13

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch (approved):
👍

PR #2399, opened 2021-02-26
Add case_group aln to igv button

dnil:

Allow display of bam files from other cases in alignviewer / IGV.js. Testing the waters with case groups.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/demo/cancer.load_config.yaml
+++ b/scout/demo/cancer.load_config.yaml
@@ -13,7 +13,7 @@ samples:
     phenotype: affected
     sex: male
     expected_coverage: 90
-    mt_bam: ~/tmp/19050-I-2A.mt.bam
+    mt_bam: scout/demo/reduced_mt.bam
     tmb: 7
     msi: 15
     tumor_purity: 1
--- a/scout/demo/cancer.load_config.yaml
+++ b/scout/demo/cancer.load_config.yaml
@@ -13,7 +13,7 @@ samples:
     phenotype: affected
     sex: male
     expected_coverage: 90
-    mt_bam: scout/demo/reduced_mt.bam
+    mt_bam: ~/tmp/19050-I-2A.mt.bam
     tmb: 7
     msi: 15
     tumor_purity: 1
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -145,18 +145,13 @@ def variant(
     if add_case:
         variant_case(store, case_obj, variant_obj)
 
-    # Fetch ids for grouped cases and prepare alignment dispay
+    # Fetch ids for grouped cases and prepare alignment display
     case_groups = {}
-    case_group_label = {}
     if case_obj.get("group"):
         for group in case_obj.get("group"):
             case_groups[group] = list(store.cases(group=group))
-            case_group_label[group] = store.case_group_label(group)
-            LOG.debug("Populate aligments for entire case group {}:".format(group))
             for grouped_case in case_groups[group]:
-                LOG.debug("Populate case {}".format(grouped_case))
                 case_append_alignments(grouped_case)
-                LOG.debug("Ok: {}".format(grouped_case))
 
     # Collect all the events for the variant
     events = list(store.events(institute_obj, case=case_obj, variant_id=variant_id))
@@ -253,7 +248,6 @@ def variant(
         "institute": institute_obj,
         "case": case_obj,
         "case_groups": case_groups,
-        "case_group_label": case_group_label,
         "variant": variant_obj,
         variant_category: True,
         "causatives": other_causatives,
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -21,6 +21,7 @@ from scout.constants import (
 from scout.parse.variant.ids import parse_document_id
 from scout.server.links import ensembl, get_variant_links
 from scout.server.utils import (
+    case_append_alignments,
     institute_and_case,
     user_institutes,
     variant_case,
@@ -142,6 +143,19 @@ def variant(
     if add_case:
         variant_case(store, case_obj, variant_obj)
 
+    # Fetch ids for grouped cases and prepare alignment dispay
+    case_groups = {}
+    case_group_label = {}
+    if case_obj.get("group"):
+        for group in case_obj.get("group"):
+            case_groups[group] = list(store.cases(group=group))
+            case_group_label[group] = store.case_group_label(group)
+            LOG.debug("Populate aligments for entire case group {}:".format(group))
+            for grouped_case in case_groups[group]:
+                LOG.debug("Populate case {}".format(grouped_case))
+                case_append_alignments(grouped_case)
+                LOG.debug("Ok: {}".format(grouped_case))
+
     # Collect all the events for the variant
     events = list(store.events(institute_obj, case=case_obj, variant_id=variant_id))
     for event in events:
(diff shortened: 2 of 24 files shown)


dnil:
Procrastination.. I don't know, perhaps it is better as a separate button, but as long as the case groups are used sparingly and not too wildly with different build versions (which would make the shadowing assessments part bail violating assumptions as well) it is kind of nice. Sometimes it is soo tempting just to compare directly to that other similar case..

dnil:
Oh, and if it wasn't obvious, this is intended to answer some of the cancer view and RD additional directed carrier testing issues with manually looking for sightings of single other read on the same position as the original case.

northwestwitch (commented):
It works perfectly! tested locally by loading the clone of the demo case (643595) and grouping it with the demo case:

image

Alignment on a variant of MT chr, after adjusting the tracks height:
image

Impressions:
- It's a very helpful functionality and I'm sure it will be appreciated
- What if you don't want to see alignments from other cases from the group, or perhaps alignments from a few cases of the group? What about transforming the igv button thingy into a form and adding a select with the cases in the group to show alignments for?
image

- The list of tracks can become very long and might be complicated to understand which tracks corresponds to which individual of which case. We could investigate if it's possible to modify the label of the IGV tracks from `individual` `case - individual`, perhaps it's possible!

scout/demo/cancer.load_config.yaml
```diff
@@ -13,7 +13,7 @@ samples:
     phenotype: affected
     sex: male
     expected_coverage: 90
-    mt_bam: scout/demo/reduced_mt.bam
+    mt_bam: ~/tmp/19050-I-2A.mt.bam
```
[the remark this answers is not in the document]
dnil: 😸 Indeed! Thanks!

dnil:
Thanks for testing!

> * What if you don't want to see alignments from other cases from the group, or perhaps alignments from a few cases of the group? What about transforming the igv button thingy into a form and adding a select with the cases in the group to show alignments for?

Yes, not bad; I might have a look - but it could also rest until another PR given that the case groups are not very many or large yet.

> * The list of tracks can become very long and might be complicated to understand which tracks corresponds to which individual of which case. We could investigate if it's possible to modify the label of the IGV tracks from `individual` to  `case - individual`, perhaps it's possible!

Right: I think it might indeed be possible, and would actually make it easier to keep track of several open windows from different families as well. Will do!

dnil:
Like so:
Screenshot 2021-03-02 at 17 24 37

northwestwitch:
> Yes, not bad; I might have a look - but it could also rest until another PR given that the case groups are not very many or large yet.

Yes, good point! I'll re-review!

scout/server/blueprints/variant/controllers.py
```diff
@@ -144,6 +145,19 @@ def variant(
     if add_case:
         variant_case(store, case_obj, variant_obj)
 
+    # Fetch ids for grouped cases and prepare alignment dispay
```
northwestwitch, line 148:
```suggestion
    # Fetch ids for grouped cases and prepare alignment display
```

scout/server/blueprints/variant/controllers.py
```diff
@@ -144,6 +145,19 @@ def variant(
...
+    case_group_label = {}
+    if case_obj.get("group"):
+        for group in case_obj.get("group"):
+            case_groups[group] = list(store.cases(group=group))
+            case_group_label[group] = store.case_group_label(group)
+            LOG.debug("Populate aligments for entire case group {}:".format(group))
+            for grouped_case in case_groups[group]:
+                LOG.debug("Populate case {}".format(grouped_case))
+                case_append_alignments(grouped_case)
+                LOG.debug("Ok: {}".format(grouped_case))
```
northwestwitch, line 159: this is producing a lot of unused text, can we remove it?

image

scout/server/blueprints/variant/controllers.py
```diff
@@ -144,6 +145,19 @@ def variant(
     if add_case:
         variant_case(store, case_obj, variant_obj)
 
+    # Fetch ids for grouped cases and prepare alignment dispay
+    case_groups = {}
+    case_group_label = {}
+    if case_obj.get("group"):
+        for group in case_obj.get("group"):
+            case_groups[group] = list(store.cases(group=group))
```
northwestwitch, line 153: This dictionary contains the complete info of all cases in each group. It's quite a lot of data but only the alignment paths are used successively in the igv_button macro. Wouldn't be better to capure only the tracks?

scout/server/blueprints/variant/controllers.py
```diff
@@ -144,6 +145,19 @@ def variant(
     if add_case:
         variant_case(store, case_obj, variant_obj)
 
+    # Fetch ids for grouped cases and prepare alignment dispay
+    case_groups = {}
+    case_group_label = {}
```
northwestwitch, line 150: Do we really need this variable? I don't see it used later in the templates 🤔

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -537,6 +501,78 @@ <h6>No matching causative or managed variants</h6>
   </div> <!--  end of card div -->
 {% endmacro %}
 
+{% macro igv_button() %}
```
northwestwitch, line 504: Nice with moving the code to a dedicated macro! 👍

scout/server/utils.py
```diff
@@ -145,7 +145,11 @@ def case_append_alignments(case_obj):
     ]
 
     for individual in case_obj["individuals"]:
-        append_safe(case_obj, "sample_names", individual.get("display_name"))
+        append_safe(
```
northwestwitch, line 148: 🥇

northwestwitch (changes requested):
Nice work! Especially with finding out how to display case name in igv.

I've got some suggestions in the comments!

scout/server/blueprints/variant/controllers.py
dnil: No, the label is used on the case pages to organise the groups, but it is not used downstream really - yet. It should be safe to remove it!

scout/server/blueprints/variant/controllers.py
dnil: I figured its not that much data, and we still catch a few other things as well like display name, other tracks (wigs, beds etc). The alternative would basically be to make a more lightweight case-group-member object for this purpose. Or a general aln-display-case object that would take the unwrapping data added to the case_objs by the alnviewer utils helpers. Both are doable, and probably on the same complexity level as the current reuse-the-case solution.

How does python actually solve the memory management here - does it really deepcopy them, or pass partial references? Languages where you don't do the malloc calls yourself encourages you to think less about that. 😊

northwestwitch: > How does python actually solve the memory management here - does it really deepcopy them, or pass partial references?

I don't know, you should ask a programmer 😆 Fine, keep it like this!

northwestwitch (approved):
👍

scout/server/blueprints/variant/controllers.py
dnil: Ok, I dont feel we would particularly benefit elsewhere right now from either of the other two constructs I came up with, so I'll leave it like this. But you do have a good point in general: we could be more explicit (and sparse) with how much we populate the objects for display when we pull them from db. Might be worth thinking about another layer for the purpose of feeding the templates. Perhaps this will be very evident if we go more towards react / json-delivering views.

dnil:
Just remembered we had a similar button on the STR-variantS view as well - made them the same:
Screenshot 2021-03-03 at 12 08 34
I think that would have been missed.

PR #2409, opened 2021-03-02
Search cases by assignee name, surname of full name

### LINKED ISSUE #2405: Searching for cases assigned to user by full name yields no hits
When you search for cases assigned to users by full name yields no results while searching using either first name or surname works.

**With surname**

image

**Full name**

image

northwestwitch:
Fix #2405 

**How to test**:
1. switch to this branch on a local instance of Scout or on a stage server
1. Assign yourself to a case 
1. Search cases by assigned user name, user surname or full name

**Expected outcome**:
All the types of searches should return results

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -213,16 +213,11 @@ class CaseHandler(object):
         if query_field == "user":
             LOG.debug(f"Search for cases with assignee '{query_term}'.")
             query_terms = query_term.split(" ")
-            query = {"$or": [{"name": {"$regex": term, "$options": "i"}} for term in query_terms]}
-            users = self.user_collection.find(query)
-            nr_users = sum(
-                1
-                for i in self.user_collection.find(
-                    {"name": {"$regex": query_term, "$options": "i"}}
-                )
-            )
+            user_query = {
+                "$or": [{"name": {"$regex": term, "$options": "i"}} for term in query_terms]
+            }
+            users = self.user_collection.find(user_query)
             query["assignees"] = {"$in": [user["email"] for user in users]}
-            LOG.error(query["assignees"])
 
         return order
 
@@ -367,7 +362,7 @@ class CaseHandler(object):
         if yield_query:
             return query
 
-        LOG.info(f"Get cases with query {query}")
+        LOG.error(f"Get cases with query {query}")
         if order:
             return self.case_collection.find(query)
 
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -213,9 +213,9 @@ class CaseHandler(object):
         if query_field == "user":
             LOG.debug(f"Search for cases with assignee '{query_term}'.")
             query_terms = query_term.split(" ")
-            query = {"$or": [{"name": {"$regex": term, "$options": "i"}} for term in query_terms]}
-            LOG.error(query)
-            users = self.user_collection.find(query)
+            users = self.user_collection.find(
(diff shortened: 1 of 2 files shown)


northwestwitch:
Tested locally using "Clark": 🆗 

image
Ready for review.


Tested locally using "Kent": 🆗 

image

Tested locally using "Clark Kent": 🆗 

image

Tested locally using "Kent Clark": 🆗 

image

northwestwitch:
@mhkc this one should fix your issue. Feel free to review if you have time!

scout/adapter/mongo/case.py
```diff
@@ -156,7 +156,7 @@ def _populate_name_query(self, query, name_query, owner=None, collaborator=None)
         """
         order = None
         query_field = name_query.split(":")[0]  # example:status
-        query_term = name_query[name_query.index(":") + 1 :].replace(" ", "")  # example:active
+        query_term = name_query[name_query.index(":") + 1 :].strip()
```
dnil, line 159: 👍

scout/adapter/mongo/case.py
```diff
@@ -212,13 +212,11 @@ def _populate_name_query(self, query, name_query, owner=None, collaborator=None)
...
-            users = self.user_collection.find({"name": {"$regex": query_term, "$options": "i"}})
-            nr_users = sum(
-                1
-                for i in self.user_collection.find(
-                    {"name": {"$regex": query_term, "$options": "i"}}
-                )
-            )
+            query_terms = query_term.split(" ")
+            user_query = {
+                "$or": [{"name": {"$regex": term, "$options": "i"}} for term in query_terms]
```
dnil, line 217: 👍

dnil (approved):

scout/adapter/mongo/case.py
dnil: I guess it will also behave a wee bit differently if you search for a specific Anna, but now it can find them - even if you get some non specific hits as well. 😸

northwestwitch:
That was fast @dnil, Thanks for the review!

dnil:
Sorry, could have left it for you @mhkc if I had read the comments first. Anyway, there are plenty to choose from. 😅

northwestwitch:
> Sorry, could have left it for you @mhkc if I had read the comments first. Anyway, there are plenty to choose from

I'm not complaining here! 😄

scout/adapter/mongo/case.py
northwestwitch: If you searched for "Anna" before and now you should get exactly the same results. Same if you searched by surname. What you get now if you search by "Name Surname" is All users where "Name" or "Surname" is in the name.

northwestwitch: And among these users there is also "Name Surname"

northwestwitch: Do we want to match exactly both terms? Then we substitute the "$OR" with the "$AND"

dnil: Agreed, that’s how I read it. What someone might expect is only getting Name Surname, rather than all, but hey, also not complaining - this is better by a long shot than not getting a result at all!

northwestwitch: But then I fix it to get only Name Surname, since I'm already working on it!

northwestwitch:
Latest change test: 

2 users: `Clark Kent` and `Clark Another`, each assignee of a case:

image

Searching for `Clark Another` returns 1 case:

image

Searching for `Clark` returns both cases:

image

Searching for `Kent` returns 1:

image

Do I merge?

PR #2411, opened 2021-03-03
Make less visible XM refseq ids in variant tx overview if they don't overlap with tx refseq_id

### LINKED ISSUE #2404: Limit Refseq transcripts displayed?
Hello Scout team!

Since one of the latest updates, way more Refseq transcripts are included in the Transcript overview for each variant. This makes it considerably harder to quickly identify the relevant transcript. Could you please consider not displaying the XM transcripts in the Transcript overview? If they're still included in the Transcripts lists at the bottom of the variant page no information is lost, but it would make things much cleaner.

Example from variant https://scout.scilifelab.se/cust003/21048/1b812e5af199db0eb3a7ccace5185311
image

Thank you :)

northwestwitch:
Possible  fix #2404.
Make less visible XM transcripts which don't overlap with transcript refseq_id. example:

Before:

image

After:

image

**How to test**:
1. [Variant](scout/server/blueprints/variant/templates/variant/tx_overview.html) used in the example above

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/tx_overview.html
+++ b/scout/server/blueprints/variant/templates/variant/tx_overview.html
@@ -53,9 +53,9 @@
                   </td> <!-- end of gene symbol col-->
                   <td>
                     {% set refseq_identifiers = [] %}
-                    {% for tr_id in transcript.refseq_identifiers %}
-                      {% if not (tr_id.startswith('XM') and transcript.refseq_id != tr_id) %}
-                        {{ refseq_identifiers.append(tr_id)|default("", true) }}
+                    {% for tx_id in transcript.refseq_identifiers %}
+                      {% if not (tx_id.startswith('XM') and transcript.refseq_id != tx_id) %}
+                        {{ refseq_identifiers.append(tx_id)|default("", true) }}
                       {% endif %}
                     {% endfor %}
                     {{ refseq_identifiers|join(", ")}}</td> <!-- Refseq IDs -->
--- a/scout/server/blueprints/variant/templates/variant/tx_overview.html
+++ b/scout/server/blueprints/variant/templates/variant/tx_overview.html
@@ -28,8 +28,8 @@
 
 {% macro transcripts_overview(variant) %}
   <div class="card panel-default">
-    <div data-toggle='tooltip' class="panel-heading" title="Displays all transcripts with refseq ids. One ensembl transcript can have multiple refseq ids. Blue color indicates that the ensembl transcript is mapped to a transcript that is canonical according to HGNC">
-      Transcript overview</div>
+    <div data-toggle='tooltip' class="panel-heading" title="Displays all transcripts with RefSeq id. The complete list of transcripts is available below. One ensembl transcript can have multiple RefSeq ids. Blue color indicates that the ensembl transcript is mapped to a transcript that is canonical according to HGNC">
+      RefSeq transcripts</div>
     <div class=card-body>
       <table id="transcript_overview_table" class="table table-bordered table-hover table-sm">
         <thead class="thead-light">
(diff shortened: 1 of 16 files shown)


scout/server/blueprints/variant/templates/variant/tx_overview.html
```diff
@@ -51,7 +51,14 @@
                       {{ gene.common.hgnc_symbol if gene.common else gene.hgnc_id }}
                     </a>
                   </td> <!-- end of gene symbol col-->
-                  <td>{{ transcript.refseq_identifiers|join(", ") }}</td> <!-- Refseq IDs -->
+                  <td>
+                    {% set refseq_identifiers = [] %}
+                    {% for tr_id in transcript.refseq_identifiers %}
+                      {% if not (tr_id.startswith('XM') and transcript.refseq_id != tr_id) %}
+                        {{ refseq_identifiers.append(tr_id)|default("", true) }}
```
northwestwitch, line 58: In case the reviewer is wondering what |default("", true) does: 

https://support.sendwithus.com/jinja/default/

hassanfa:
Chiara, if you decide to hide XM transcripts, is it possible to notify user that "there are hidden XM transcripts"? Just wondering... Sometimes XM transcripts can be informative.

dnil:
Perhaps add in like an ellipsis with a popover for the other transcripts?

northwestwitch:
Thinking about this. Do you think many users would still prefer seeing those XM transcripts? Because otherwise we could simply print the "uninteresting" ones in a lighter color instead..

dnil:
Would work for me, but then I'm probably not the target audience for this. The thing we want to avoid though (which kind of led to this display) is making it look like the ENSEMBL transcript *is* identical to the one and only RefSeq one when it occasionally isn't.

hassanfa:
Reason i am asking for it: VEP's only-matched options sometimes leaves only XM transcript and some other non canonical one. It might break Scout

hassanfa:
> Reason i am asking for it: VEP's only-matched options sometimes leaves only XM transcript and some other non canonical one. It might break Scout

But then again, this will be user problem not Scout

northwestwitch:
@dnil having these XM transcripts in a popover wouldn't be so good as the code is also used in the general report. And also wouldn't be nice to have a different list of transcripts in the variant and in the variant section of the report I think. So it's about having them or not I think. Perhaps we can just highlight the most important ones, I've asked the user.

dnil:
Agreed - reuse serves a dual purpose here! Highlight sounds good to me, but then I am not very perturbed by long stretches of alphanumeric jumbles. We could also try asking one of the users with the incorrect-transcripts-on-the-report issue as well so they feel the separation is still clear?

scout/server/blueprints/variant/templates/variant/tx_overview.html
```diff
@@ -51,7 +51,18 @@
                       {{ gene.common.hgnc_symbol if gene.common else gene.hgnc_id }}
                     </a>
                   </td> <!-- end of gene symbol col-->
-                  <td>{{ transcript.refseq_identifiers|join(", ") }}</td> <!-- Refseq IDs -->
+                  <td>
+                    {% set decorated_txs = [] %}
+                    {% for tx_id in transcript.refseq_identifiers %}
+                      {% if tx_id.startswith('XM') and transcript.refseq_id != tx_id %}
```
dnil, line 57: 👍

scout/server/blueprints/variant/templates/variant/tx_overview.html
dnil: Ehm, you mean for not only printing None? Or did you rather mean append(tr_id|default())?

dnil (approved):

dnil:
Looks good on stage:
Screenshot 2021-03-03 at 13 06 53

scout/server/blueprints/variant/templates/variant/tx_overview.html
northwestwitch: Without |default("", true) it prints first "none" then the transcript id

northwestwitch:
I've changed the transcripts overview like this, so it's more clear that these are the refseq transcripts:

image

dnil:
Good enough! Canonical is according to ENSEMBL via VEP, primary is from HGNC.

PR #2413, opened 2021-03-03
New minor version 4.30

northwestwitch:
### Added
- Introduced a `reset dismiss variant` verb
- Button to reset all dismissed variants for a case
- Add black border to Chromograph ideograms
- Show ClinVar annotations on variantS page
- Added integration with GENS, copy number visualization tool
- Added a VUS label to the manual classification variant tags
- Add additional information to SNV verification emails
- Tooltips documenting manual annotations from default panels
- Case groups now show bam files from all cases on align view
### Fixed
- Center initial igv view on variant start with SNV/indels
- Don't set initial igv view to negative coordinates
- Display of GQ for SV and STR
- Parsing of AD and related info for STRs
- LoqusDB field in institute settings accepts only existing Loqus instances
- Fix DECIPHER link to work after DECIPHER migrated to GRCh38
- Removed visibility window param from igv.js genes track
- Updated HPO download URL
- Patch HPO download test correctly
- Reference size on STR hover not needed (also wrong)
- Introduced genome build check (allowed values: 37, 38, "37", "38") on case load
- Improve case searching by assignee full name
- Populating the LoqusDB select in institute settings
### Changed
- Cancer variants table header (pop freq etc)
- Only admin users can modify LoqusDB instance in Institute settings
- Style of case synopsis, variants and case comments
- Switched to igv.js 2.7.5
- Do not choke if case is missing research variants when research requested
- Improve create new gene panel form validation
- Make XM- transcripts less visible if they don't overlap with transcript refseq_id in variant page
- Color of gene panels and comments panels on cases and variant pages

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Checklist before merging/releasing:

### Added
- [ ] Introduced a `reset dismiss variant` verb
- [ ] Button to reset all dismissed variants for a case
- [ ] Add black border to Chromograph ideograms
- [ ] Show ClinVar annotations on variantS page
- [ ] Added integration with GENS, copy number visualization tool
- [ ] Added a VUS label to the manual classification variant tags
- [ ] Add additional information to SNV verification emails
- [ ] Tooltips documenting manual annotations from default panels
- [ ] Case groups now show bam files from all cases on align view
### Fixed
- [ ] Center initial igv view on variant start with SNV/indels
- [ ] Don't set initial igv view to negative coordinates
- [ ] Display of GQ for SV and STR
- [ ] Parsing of AD and related info for STRs
- [ ] LoqusDB field in institute settings accepts only existing Loqus instances
- [ ] Fix DECIPHER link to work after DECIPHER migrated to GRCh38
- [ ] Removed visibility window param from igv.js genes track
- [ ] Updated HPO download URL
- [ ] Patch HPO download test correctly
- [ ] Reference size on STR hover not needed (also wrong)
- [ ] Introduced genome build check (allowed values: 37, 38, "37", "38") on case load
- [ ] Improve case searching by assignee full name
- [ ] Populating the LoqusDB select in institute settings
### Changed
- [ ] Cancer variants table header (pop freq etc)
- [ ] Only admin users can modify LoqusDB instance in Institute settings
- [ ] Style of case synopsis, variants and case comments
- [ ] Switched to igv.js 2.7.5
- [ ] Do not choke if case is missing research variants when research requested
- [ ] Improve create new gene panel form validation
- [ ] Make XM- transcripts less visible if they don't overlap with transcript refseq_id in variant page

CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
...
 ### Added
+### Fixed
+### Changed
+
+## [4.30]
+### Added
 - Introduced a `reset dismiss variant` verb
 - Button to reset all dismissed variants for a case
-- Add black border to ideograms
+- Add black border to Chromograph ideograms
```
dnil, line 16: 👍

dnil (approved):
I'll approve as the code is ok, and it is in principle tested, but quite agree we shall tick off your list before actually releasing!

CHANGELOG.md
northwestwitch: > quite agree we shall tick off your list before actually releasing!

It's on clinical-db stage now. I'll do some checks and then continue tomorrow

dnil: It would be nice with re-uploading an up-to-date MIP case, for the CLI side changes like STR AD..

northwestwitch:
OMG what happened to the gene panel? 😆 

image

northwestwitch:
> OMG what happened to the gene panel? 😆

I know that it was me initially to introduce that yellow for the comments, but since already have requests for dark theme why don'y we tone it down a bit?

dnil:
Yeah, agreed. The previous odd/even colors would work fine for comments as well methinks?

northwestwitch:
> Yeah, agreed. The previous odd/even colors would work fine for comments as well methinks?

There was no differential color before for the comments. We could keep the blue and instead of yellow we could have the gray background color perhaps?

dnil:
Sure! Or just the beige background interleaved with gray? Try one and pick one that looks ok with both comments and panels!

northwestwitch:
Mmm looks like that after resetting the dismissed status of the variants (that works) it's not showing that a user (name) has done it, in the case events, but it should because we have created the verb. I'll look into it.

northwestwitch:
> Mmm looks like that after resetting the dismissed status of the variants (that works) it's not showing that a user (name) has done it, in the case events, but it should because we have created the verb. I'll look into it.

Ok that's because the reset dismissed is showed in the variant page, not in the case one. Perhaps it would be better to create an additional event, specific for the case when the reset happens. But we can take in another PR perhaps!

dnil:
> > Mmm looks like that after resetting the dismissed status of the variants (that works) it's not showing that a user (name) has done it, in the case events, but it should because we have created the verb. I'll look into it.
> 
> Ok that's because the reset dismissed is showed in the variant page, not in the case one. Perhaps it would be better to create an additional event, specific for the case when the reset happens. But we can take in another PR perhaps!

Well spotted, but yeah, even if we add one on the case level sometime, we probably still want to keep the variant level ones anyway.

northwestwitch:
Testing the upload of a recent case on hasta stage:

image
image

Loading the case keenviper: 
image

image

PR #2424, opened 2021-03-08
fix to the fix (missing end_chrom)

northwestwitch:
Thanks @dnil for the previous review. Immediately after I made it available for review I've realized that this is a better solution, sorry!

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -420,7 +420,7 @@ def ensembl_link(variant_obj, build=37):
     """Compose (sv) variant link to ensembl"""
 
     my_end = variant_obj["end"]
-    if variant_obj["chromosome"] != variant_obj.get("end_chrom"):
+    if variant_obj["chromosome"] != variant_obj.get("end_chrom", variant_obj["chromosome"]):
         my_end = variant_obj["position"]
 
     if build == 37:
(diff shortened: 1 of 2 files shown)


scout/server/links.py
```diff
@@ -420,7 +420,7 @@ def ensembl_link(variant_obj, build=37):
     """Compose (sv) variant link to ensembl"""
 
     my_end = variant_obj["end"]
-    if variant_obj["chromosome"] != variant_obj.get("end_chrom"):
+    if variant_obj["chromosome"] != variant_obj.get("end_chrom", variant_obj["chromosome"]):
```
dnil, line 423: Eh, sorry, I'm missing something? Are you trying to avoid some implicit None comparison issue or what?

dnil: Ah ok. Fine. That is probably better - anything would do really, but yes.

dnil (approved):

scout/server/links.py
dnil: You could also check if variant_obj.get("end_chrom") was something first, which is sort of what we normally do?

dnil: ..but this works.

northwestwitch: I don't know, I just assumed that when end_chrom is missing it should be end_chrom = chromosome. This expression is more clear i think, but you are right, the PR would have worked anyway.

The expression should implicitly first check if variant_obj.get("end_chrom") and of it's not then end_chrom = chromosome, no?

dnil: I think both would have worked really: if end_chrom == None, my_end=pos, if end_chrom != chr, my_end = pos but if end_chrom == chr my_end=end? In a way I guess the first one is then an easier read, but it is fine this way too.

PR #2425, opened 2021-03-08
Patch release 4.30.2

northwestwitch:
### Fixed
- Use VEP RefSeq ID if RefSeq list is empty in RefSeq transcripts overview
- Bug creating variant links for variants with no end_chrom
### Changed

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
+
 ## [X.X.X]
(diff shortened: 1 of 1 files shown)


northwestwitch:
Version is already fixed:

image

dnil (approved):

dnil:
We could just merge this, right? Creating the release in a way that doesn't mess with pypi may be a tad complicated, but first things first.

northwestwitch:
> We could just merge this, right? Creating the release in a way that doesn't mess with pypi may be a tad complicated, but first things first.

Yes, I think so!

PR #2428, opened 2021-03-09
A more compact case groups panel

### LINKED ISSUE #2418: Improve the looks of group of cases
Group label and "bind other case" button, that is only showing "bind other" (chrome)

image

northwestwitch:
Fix #2418 

Changed
- Added number of group of cases in panel header
- Made a scrollable div so it doesn't expand so much if there are many cases
- Moved the "add new group" button to the panel header (top right corner)
- Made the whole thing more compact
- Every time a change happens the page redirects to the panel as an anchor

Before:

image

After:

image

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -257,7 +257,7 @@
               </div>
               <input type="text" class="form-control" name="label" id="case_group_label-{{group_id}}" placeholder="{{ case_group_label[group_id] }}">
             </div>
-            <button type="submit" class="btn btn-secondary btn-sm mb-2">save label</button>
+            <button type="submit" class="btn btn-secondary btn-sm mb-2">Save label</button>
             <button type="button" class="btn btn-secondary btn-sm mb-2 ml-3 text-right" data-toggle="modal" data-target="#add-case-group-{{group_id}}">
               <i class="fas fa-user-plus" data-toggle='tooltip' title="Add other case to this group"></i>
             </button>
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -243,7 +243,7 @@
       <span>
         <i class="fa fa-users mr-1"></i>Group connected cases ({{case_groups|length}} groups)
       </span>
-      <span class="float-sm-right" data-toggle='tooltip' title="Create a new group containing case {{case.display_name}}"><a href="{{ url_for('cases.add_case_group',institute_id=institute._id, case_name=case.display_name) }}" class="text-white"><span class="fas fa-plus text-white mr-1"></span>New group</a></span>
+      <span class="float-sm-right" data-toggle='tooltip' data-placement="right" title="Create a new group containing case {{case.display_name}}"><a href="{{ url_for('cases.add_case_group',institute_id=institute._id, case_name=case.display_name) }}" class="text-white"><span class="fas fa-plus text-white mr-1"></span>New group</a></span>
     </div>
     <div class="card-body">
       <div class="list-group" style="max-height:200px; overflow-y: scroll;">
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -241,7 +241,7 @@
   <div class="card panel-default" id="case_groups">
     <div class="panel-heading">
       <span data-toggle='tooltip' title="Group cases to share user variant assessments like comments, classification and dismissal. A case can be in multiple groups.">
-        <i class="fa fa-users mr-1"></i>Group connected cases ({{case_groups|length}})
+        <i class="fa fa-users mr-1"></i>Group connected cases ({{case_groups|length}} groups)
       </span>
       <span class="float-sm-right"><a href="{{ url_for('cases.add_case_group',institute_id=institute._id, case_name=case.display_name) }}" class="text-white"><span class="fas fa-plus text-white mr-1"></span>new group</a></span>
     </div>
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -265,15 +265,15 @@
 
           <div class="col-12">
             {% for grouped_case in case_group %}
-              <div class="mb-2">
+              <span class="badge badge-pill badge-light">
                 <a href="{{ url_for('cases.case', institute_id=grouped_case.owner, case_name=grouped_case.display_name) }}"">{{ grouped_case.display_name }}</a>
-                    <span class="pull-right">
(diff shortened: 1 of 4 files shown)


northwestwitch:
This is ready for review now. In case the reviewer is wondering why I include a request "referer" in the request header of the tests and but we always redirect to the request.referrer (with double r), don't blame me but Phillip Hallam-Baker, check it on [wikipedia](https://en.wikipedia.org/wiki/HTTP_referer)

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -238,49 +238,48 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
-            <div class="col-12">
-              <ul class="list-group">
-              {% for grouped_case in case_group %}
-                <li class="list-group-item">
-                    <a href="{{ url_for('cases.case', institute_id=grouped_case.owner, case_name=grouped_case.display_name) }}"">{{ grouped_case.display_name }}</a>
-                    <span class="pull-right">
-                      <a href="{{ url_for('cases.remove_case_group', institute_id=institute._id, case_name=grouped_case.display_name, case_group=group_id) }}" class="btn btn-link btn-sm">
+            <button type="submit" class="btn btn-secondary btn-sm mb-2">save label</button>
+            <button type="button" class="btn btn-secondary btn-sm mb-2 ml-3 text-right" data-toggle="modal" data-target="#add-case-group-{{group_id}}">
+              <i class="fas fa-user-plus" data-toggle='tooltip' title="Add other case to this group"></i>
```
dnil, line 262: 👍

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -238,49 +238,48 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
                         <span class="fa fa-remove"></span></a>
-                    </span>
-                </li>
-              {% endfor %}
-              </ul>
-            </div>
+                  </span>
+              </span>
+            {% endfor %}
+            <hr style="height:2px;">
```
dnil, line 276: Perhaps make that ruler join like and show it only if there is a next element (eg before all except the first)? It looks a little lonely over the card border for the last one?

scout/server/blueprints/cases/views.py
```diff
@@ -993,15 +993,15 @@ def add_case_group(institute_id, case_name):
 
     controllers.add_case_group(store, current_user, institute_id, case_name, group_id)
 
-    return redirect(request.referrer)
+    return redirect(request.referrer + "#case_groups")
```
dnil, line 996: :+1: We should really do that for more elements further down on the page! I was pleasantly surprised when testing. 😸

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -334,13 +334,21 @@ def test_add_case_group(app, case_obj, institute_obj):
         # GIVEN that the user could be logged in
         resp = client.get(url_for("auto_login"))
 
+        # Given a page referrer (request to add case group starts from case page)
```
dnil, line 337: 😄

dnil (approved):

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -238,49 +238,48 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
 {% endmacro %}
 
 {% macro group_panel() %}
-  <div class="card panel-default">
-    <div data-toggle='tooltip' class="panel-heading" title="Group cases to share user variant assessments like comments, classification and dismissal. A case can be in multiple groups."><span class="fa fa-users"></span>&nbsp;Group connected cases</div>
+  <div class="card panel-default" id="case_groups">
+    <div class="panel-heading">
+      <span data-toggle='tooltip' title="Group cases to share user variant assessments like comments, classification and dismissal. A case can be in multiple groups.">
```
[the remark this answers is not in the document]
northwestwitch: I actually did this on purpose, because I didn't want the tooltip to activate when you hover over the "+ new group", since that would have been very confusing. If you really don't like it I can try to find a better solution. Ideas?

scout/server/blueprints/cases/views.py
northwestwitch: I agree, little by little!

scout/server/blueprints/cases/templates/cases/case.html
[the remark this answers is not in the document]
dnil: Hm, you think so? The text kind of fits? But anyway, what happens if you add a tooltip to the "+ New group" link - would it override?

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -238,49 +238,48 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
-            </div>
-            <div class="col text-right">
-              <div class="btn-group">
-                <button type="button" class="btn btn-outline-secondary form-control" data-toggle="modal" data-target="#add-case-group-{{group_id}}">
-                  <span class="fa fa-plus-square-o"></span>&nbsp;Bind other case
-                </button>
+          <form class="form-inline mt-2" action="{{ url_for('cases.case_group_update_label', case_group=group_id)}}" method="POST">
+            <div class="input-group mb-2 mr-sm-2">
+              <div class="input-group-prepend">
+                <div class="input-group-text"><i class="fa fa-users"></i></div>
```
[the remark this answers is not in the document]
northwestwitch: I did it following the [bootstrap form-inline example](https://getbootstrap.com/docs/4.3/components/forms/#inline-forms)

scout/server/blueprints/cases/templates/cases/case.html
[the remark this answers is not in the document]
northwestwitch: I'll try!

northwestwitch:
I've fixed according to your suggestions @dnil (good feedback btw!).

- Moving the tooltip back to the panel header makes it behave as before. Adding an additional tooltip on the "+ New group" and hovering only over it, triggers both tooltips, but perhaps it is fine:

image

- Printing the divider (hr) on top of the group item from the second loop on does the trick and if you have one group looks like this now:

image


If you have more groups:

image

dnil:
> * Moving the tooltip back to the panel header makes it behave as before. Adding an additional tooltip on the "+ New group" and hovering only over it, triggers both tooltips, but perhaps it is fine:

How about showing the New group tooltip on like the right side?

northwestwitch:
A small improvements so the 2 tooltips don't overlap: I've moved the new group one on the right:

image

Ready to merge?

northwestwitch:
> How about showing the New group tooltip on like the right side?

OMG we've been working together for too long and starts to be creepy 😆

PR #2434, opened 2021-03-11
Export primary HGNC transcript in variantS export file

### LINKED ISSUE #2398: Add VAF info exported variant excel file?
Hi! 
 When you export variants from the variant page to excel, the HGNC transcript does not come up, instead some other prefereed transcript comes up. Can this be changed, or can a column be added for the HGNC transcript? 

Hälsningar Hero

### LINKED ISSUE #2429: "Filter and export" excludes most information for variants that overlap multiple genes
When a variant overlaps more than one gene, like this:

bild

The row in the export csv is almost empty

bild

A quick look at the code suggests that this should work, but I haven't debugged it yet. I'll try to find some time later today.

northwestwitch:
This PR adds a functionality:
- merges 3 columns relative to the canonical transcript into one: `Canonical_transcript|HGVS|protein_change`
- creates another column for the primary transcript:  `Primary_transcript|HGVS|protein_change`
- Fixes a bug caused by using default genome build 37 when exporting these info from variants page

Partial fix #2398 and possible fix #2429 

**How to test**:
1. Go to variantS page, filter and export variants.
2. Check exported transcripts against transcripts specified on each variant's page

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -514,7 +514,7 @@ def variant_export_genes_info(store, gene_list, genome_build="37"):
         primary_txs += gene_primary_txs
 
     for item in [gene_ids, gene_names, canonical_txs, primary_txs]:
-        gene_info.append(";".join(str(x) for x in item))
+        gene_info.append("-".join(str(x) for x in item))
 
     return gene_info
 
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -419,7 +419,7 @@ def variant_export_lines(store, case_obj, variants_query):
 
         # if variant is in genes
         if gene_list is not None and len(gene_list) > 0:
-            gene_info = variant_export_genes_info(store, gene_list)
+            gene_info = variant_export_genes_info(store, gene_list, case_obj.get("genome_build"))
             variant_line += gene_info
         else:
             empty_col = 0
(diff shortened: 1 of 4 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -445,47 +444,77 @@ def variant_export_lines(store, case_obj, variants_query):
...
-                protein_change = transcript_obj.get("protein_sequence_name") or "-"
-        hgvs_c.append(hgvs_nucleotide)
-        pt_c.append(protein_change)
+        canonical_txs += gene_canonical_txs
+        primary_txs += gene_primary_txs
 
-    for item in [gene_ids, gene_names, canonical_txs, hgvs_c, pt_c]:
-        gene_info.append(";".join(str(x) for x in item))
+    for item in [gene_ids, gene_names, canonical_txs, primary_txs]:
+        gene_info.append("-".join(str(x) for x in item))
```
bjhall, line 517: I think you need to join with `" - "` (with spaces). Since gene symbols can contain dashes things like `CYP1B1-CYP1B1-AS1` become ambiguous.

bjhall:
This looks much better now! It even works in Swedish Excel! 

However I am getting almost nothing in the canonical field. And when there is no primary transcript I get nothing at all (which is expected, but a bit sad :) ).

bild

northwestwitch:
Right! I chose the dumbest separator ever, sorry! Regarding the very few canonical transcripts I am quite sure that this is due to the genome build bugfix. I'll think about it and get back to you with a fix, thanks!

northwestwitch:
@bjhall I've changed the inner separators to "/" and the gene separators to " | ". I've also changed how the canonical transcripts are collected and you should be able to see more of them now!

bjhall:
Yes now this looks much better! I still see very few (maybe 1 out of 100) variants where I get nothing in either the canonical or primary column. But these appear to be when 1) there is no defined primary/canonical transcript or 2) when the canonical transcript is some weird non-coding thing. 

Maybe leaving those corner cases unfixed is fine, since the export function is "for research only", right?

northwestwitch:
> Maybe leaving those corner cases unfixed is fine, since the export function is "for research only", right?

Yes! I see the same with build 37 and I think it's fine! If you think it's fine could you approve the PR? So we'll have it in the next release!

bjhall (approved):

northwestwitch:
Thanks @bjhall, merging!

PR #2437, opened 2021-03-15
Create a case-level event when a user resets all dismissed variants

### LINKED ISSUE #2416: Create an event at the case level when a user resets all dismissed variants
Now it's shown on the variant page only (version 4.30)

northwestwitch:
This PR adds a functionality:
-  fix #2416 
- Makes sure that the activity events are displayed with the newest first on case and variant pages

**How to test**:
1. Dismiss one or more variants for a page
2. In master branch, reset all dismissed variants from one of the variantS page
3. Make sure that the reset dismissed events are visible only on the involved variants pages
4. Switch to this branch and repeat. It should now display that your specific user did the reset of all dismissed variants

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


scout/server/blueprints/variants/controllers.py
```diff
@@ -928,6 +928,14 @@ def reset_all_dimissed(store, institute_obj, case_obj):
         case_obj(dict): a case dictionary
     """
     evaluated_vars = store.evaluated_variants(case_id=case_obj["_id"])
+    user_obj = store.user(current_user.email)
```
dnil, line 931: 👍

scout/server/templates/utils.html
```diff
@@ -141,7 +141,7 @@ <h5 class="modal-title">Edit comment</h5>
   <div class="card panel-default">
     <div class="panel-heading"><span class="fa fa-book"></span>&nbsp;Activity</div>
     <ul class="list-group" style="max-height:400px; overflow-y: scroll;">
-      {% for event in events %}
+      {% for event in events|sort(attribute="created_at", reverse=True) %}
```
dnil, line 144: You could also sort on the cursor / db call. I'm guessing the db implementation is slightly faster. But since they are few and all get shown anyway I doubt the performance hit is in any way noticeable.

dnil: "This is more of a comment than a question".. 😉

dnil (approved):
Looks tip-top!

dnil:
Screenshot 2021-03-22 at 11 03 44

PR #2441, opened 2021-03-15
Include also genes from dynamic_gene_list in HPO gene list export

### LINKED ISSUE #2440: HPO gene panel download offline?
Something is off with HPO gene panel downloads; reported by Hafdis H for https://scout.scilifelab.se/cust002/F0038096 and reproduces with just adding a single gene on stage for https://scout-stage.scilifelab.se/cust002/F0013248-3.

🕵🏻‍♂️ HPO gene table? pdf generation?

northwestwitch:
fix #2440 

The error was probably due to the fact that only genes from HPO panels created based on HPO terms are exportable. Whenever a user adds a custom gene not associated to a specific HPO term, then this genes is added into case["dynamic_gene_list"], that is not actually contemplated in the export.

This PR fixes this

**How to test**:
1. Add one or more genes to the HPO panel by using the dedicated form:
image

1. Export the panel and see that the genes are written to the export file

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -655,7 +655,8 @@ def phenotypes_genes(store, case_obj):
 
     if case_obj.get("dynamic_gene_list"):
         gene_list = [
-            gene.get("hgnc_symbol") or gene["hgnc_id"] for gene in case_obj["dynamic_gene_list"]
+            gene.get("hgnc_symbol") or str(gene["hgnc_id"])
+            for gene in case_obj["dynamic_gene_list"]
         ]
         hpo_genes["Case dynamic gene list"] = {
             "description": "custom",
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -630,7 +630,7 @@ def phenotypes_genes(store, case_obj):
     # Loop over the dynamic phenotypes of a case
     for hpo_id in case_obj.get("dynamic_panel_phenotypes", []):
         hpo_term = store.hpo_term(hpo_id)
-        # Check that HPO term exists in dat§abase
+        # Check that HPO term exists in database
         if hpo_term is None:
             LOG.warning(f"Could not find HPO term with ID '{hpo_id}' in database")
             continue
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -660,6 +660,7 @@ def phenotypes_genes(store, case_obj):
             gene.get("hgnc_symbol") or str(gene["hgnc_id"])
             for gene in case_obj["dynamic_gene_list"]
         ]
+        unique_genes = set(gene_list)
         by_phenotype = False
 
     if by_phenotype is False:
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -655,6 +655,7 @@ def phenotypes_genes(store, case_obj):
             "genes": ", ".join(sorted(gene_list)),
         }
 
+    # Case where dynamic_panel_phenotypes is empty, perhaps because user has added custom genes to HPO panel
     if not hpo_gene_list and case_obj.get("dynamic_gene_list"):
         gene_list = [
             gene.get("hgnc_symbol") or str(gene["hgnc_id"])
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -626,10 +626,11 @@ def phenotypes_genes(store, case_obj):
     by_phenotype = True  # display genes by phenotype
     unique_genes = set()
     hpo_genes = {}
+
     # Loop over the dynamic phenotypes of a case
     for hpo_id in case_obj.get("dynamic_panel_phenotypes", []):
         hpo_term = store.hpo_term(hpo_id)
-        # Check that HPO term exists in database
+        # Check that HPO term exists in dat§abase
         if hpo_term is None:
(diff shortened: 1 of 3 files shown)


northwestwitch:
Tested with https://scout-stage.scilifelab.se/cust002/F0013248-3, works 🆗 


image

dnil (approved):
Right! So this function is sort of balancing between being a general thing and a very specialised controller method for one view, right? This kind of pushes it over to the latter. Nothing really wrong with that - just update the docstring and even possibly the function name. Or we could think harder and split it into perhaps two different general methods - they would almost certainly be useful from other places pertaining to the hpo gene relations - but not mandatory.

Any reason for keeping the add to `gene_list` from the dynamic_panel_phenotypes (L648) if you trust the L657 to do the work if it has been generated? How about a set?

This is certainly approved: it is a working bugfix - but it does point to something that could use a refactor for reuse and function clarification.

There is also the spectre of some incorrect gene lists with reports out there. Not sure how much further we can get in the archeology than giving a version range?

scout/server/blueprints/cases/controllers.py
```diff
@@ -652,6 +653,16 @@ def phenotypes_genes(store, case_obj):
...
         }
 
+    if case_obj.get("dynamic_gene_list"):
+        gene_list = [
+            gene.get("hgnc_symbol") or str(gene["hgnc_id"])
+            for gene in case_obj["dynamic_gene_list"]
+        ]
+        hpo_genes["Case dynamic gene list"] = {
+            "description": "custom",
+            "genes": ", ".join(sorted(gene_list)),
```
dnil, line 663: Actually what you really would want is sets and subtract the genes that came from `dynamic_panel_phenotypes` right? The difference would be either `custom` adds or stuff no longer concurrent due to later day changes in the HPO-to-gene mapping. Yiekes, that is actually scary as well: let's absolutely not print any gene to that document that has not really been on the panel for analysis. 😊

northwestwitch: > let's absolutely not print any gene to that document that has not really been on the panel for analysis

But then this check should also be done for the genes coming from HPO terms, right?

northwestwitch: > Actually what you really would want is sets and subtract the genes that came from `dynamic_panel_phenotypes` right?

Right! I didn't notice that when you add genes from a phenotypes they get also added in dynamic_gene_list. I'll fix, thanks!

northwestwitch: Ok this is a bit more complicated than I thought. For instance imagine you have have an HPO panel composed of genes associated to one or more HPO terms. If you export it then all good. 

But if you add a custom gene, then `dynamic_panel_phenotypes` field becomes empty, and that's why it didn't print anything in the first place. So there is really not much to subtract here, because only the dynamic_gene_list is populated.. What I'd to fix it then is checking if dynamic_panel_phenotypes is empty, and if it is print the list of genes in dynamic_gene_list only. But let me know if you have a better suggestion..!

northwestwitch: Mmm the more I think about this and the more it looks like a bug. dynamic_panel_phenotypes shouldn't be erased when you add a non-hpo custom gene. I'll try to fix it and propose a solution on this PR

dnil: Ok, that sounds well in keeping with previous mode of operation then! 
I very much like that, and I would like it even better if it had turned out we do the same thing (clear the intermediate dynamic HPO term list) every time the dynamic gene list was generated.

I will right back say before I read you PR, I would not have remembered we stored the `dynamic_panel_phenotypes` for use beyond repopulating the `dynamic_gene_list`. The two should really be kept well in sync with how they are presented on the interface, otherwise we can end up in version issues with the HPO-gene mapping. Assume for instance a user Alice does some filtering with a set of HPO terms, generating a dynamic panel and using it. Then for some reason they wait a bit with generating the gene list pdf for the report - e.g. while waiting for a doctor Chiara to respond to a comment about one of the genes. In the meantime Bob at CG does a monthly (cough) update of HPO terms. Alice may now generate a pdf with genes that were actually not checked, and doctor Chiara sees their new favourite gene checked, while in fact it wasn't.

northwestwitch: Yes I see your point! So it isn't a bug but a way to ensure that the HPO genes are up to date with the analysis. I'll revert!

dnil: Good stuff! Yes, or perhaps rather out-of-date, but completely in line with the analysis.

On the account of reading-before-ranting-next-time, L645 actually kills most of my second part worries by checking that filtered out (or as it were, novel) genes are not shown anyway. So all good now methinks?

northwestwitch: I think so! I've added a comment to make it more easy to remember next time we touch this code

scout/server/blueprints/cases/controllers.py
```diff
@@ -655,6 +655,7 @@ def phenotypes_genes(store, case_obj):
             "genes": ", ".join(sorted(gene_list)),
         }
 
+    # Case where dynamic_panel_phenotypes is empty, perhaps because user has added custom genes to HPO panel
```
dnil, line 658: A set catching extras may still have been a more elegant way to do this, but yeah, I do think this covers it. 💯

dnil (approved):

scout/server/blueprints/cases/controllers.py
northwestwitch: Thanks for the review!

Issue #2324, opened 2021-01-29
 Unify codebase with molecuar diagnostics Lund


mhkc:
We in Lund has made the following changes to Scout code base. For more information please look at [our fork](https://github.com/Clinical-Genomics-Lund/scout/tree/production).
- [x] Removed genome build from DECIPHER button
- [x] Added button that opens [GENS ](https://github.com/Clinical-Genomics-Lund/gens)to case view and SV variant view. https://github.com/Clinical-Genomics/scout/pull/2368
- [ ] Added button for correcting family metadata and submit a reanalysis of VCF files. https://github.com/Clinical-Genomics/scout/pull/2371
- [x] Added Dockerfile. Refactor LoqusDB settings parsing and introducing Loqus queries via REST API #2349 
- [x] Added link for primer ordering to Sanger sequencing request email
- [ ] Custom variant evaluation terms (manual rank and dismiss options) 
**NOTE: there might be more to come**

Most things are minor changes and could be solved by allowing to add custom fields. The "edit metadata and submit for reanalysis" is more elaborate and should be an optional feature. However it would require that we define a "standard RESTapi" which Scout uses to communicate with a custom service that each site is responsible for setting up. This might be a feature that could be implemented in the future.

northwestwitch:
OK! I was afraid it would me much more different, but this is very mergeable I think! 
That note on the other hand .. 😆

mhkc:
How do you think the very Lund specific changes (Gens & rerunner) should be implemented?

I guess the easiest solution would be to only display the integration if related variables are set in the config. 
Another, more involved, method could be to expand the ability to define custom data fields in the case yaml file.

dnil:
Display-if-available sounds the absolute best! That way everyone else can also use the integration as soon as they get off their butts and start using Gens & rerunner! 😸

northwestwitch:
I wouldn't mind start using Gens in Stockholm as well, and fixing rerunner to be general enough to send requests to our [cg](https://github.com/Clinical-Genomics/cg) software

dnil:
Same here.. Gems is a little complicated for us as it is right now with how cases are stored, but it would be changeable if someone has the interest. We could have a a project student on that one - @moonso, wouldn't you like to supervise one formally before you are defending perhaps? Or you C? You know array data well enough!

northwestwitch:
Yes, that would be fun!

bjhall:
[...]

PR #2444, opened 2021-03-16
Show variant QUAL in summary.

bjhall:
As part of https://github.com/Clinical-Genomics/scout/issues/2324 this PR adds the QUAL value to the "variant summary" section of the variants detail page.

bild

At least for Sentieon DNAscope this value is much more informative than GQ, which is always 99...


northwestwitch (approved):
👍 Looks good!

Issue #2417, opened 2021-03-04
Increasing the size of a case page with Chromograph images breaks the layout


northwestwitch:
See for instance this example: https://scout-stage.scilifelab.se/cust002/F0026339-2

Suggestion: insert chromograph pics into a scrollable div

northwestwitch:
this is what I see on Chrome, zoom 100% of the page:

![image](https://user-images.githubusercontent.com/28093618/109933123-5c14bb80-7ccb-11eb-8b90-b85aff434e44.png)

dnil:
Its ok in Safari, at least when the page is not made very narrow, but we did have issues with exactly the above some time back. Scrollable divs sound perfect!

dnil:
The core thing for both this and the #2451 is the fixed collapsible size div/nav that goes with the relative size columns in the grid. Any other size limitations beyond the navbar really messes causes the column wrap when the viewport is small enough. Try eg resizing your browser window when you have a Madeline svg on there. Same story, only less apparent due to smaller picture.

I've toyed a bit with different layouts for the sidebar; flex is slightly better, going table solves the issue completely, but both also removes some of the responsive aspects so cards don't shrink or restack well. Relative size sidebar also solves the issue, but is rather ugly. I think I'll just try updating the width on the Chromograph image, then we should probably have a discussion about what screen sizes to support / relayout for..

northwestwitch:
Adding class="img-fluid" to the SVG images should help, but you'd still need a scroller for the div

dnil:
Right, img-fluid will make one of the widths relative again, so will look much like the table or stretch global layout options!

northwestwitch:
And there is also that javascript function that arranges the chromosomes into 2 or 3 columns according to the page width . Perhaps adding an additional option to lay them all in one col id the zoom is high?

dnil:
Yeah, I'll just do that (and add a y-scroll). It will now not be chromograph messing with the layout until somewhere down below 1080px display.

northwestwitch:
> Yeah, I'll just do that (and add a y-scroll). It will now not be chromograph messing with the layout until somewhere down below 1080px display.

Sounds like a good solution. We are going to change these pages in the future anyway!

PR #2455, opened 2021-03-18
Introduce spaces into long strings of comments that break the page view

### LINKED ISSUE #2451: odd graphis for case 21011
it looks  like the case is not loaded
image

but if I scroll down i can see it is under in the page

image

This happens only with this case. When I've activated and analyzed this case it look as usual. 

This is not a real problem but good to know as you can think the case is not shown, so I think it could be good to report it.

northwestwitch:
Fix #2451.

We observed this before, but we had it fixed only in the report page, see https://github.com/Clinical-Genomics/scout/pull/1976/files

**How to test**:
1. Open a case and add this comment: 
`A,very,long,comment,with,no,spacesonlyaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryveryaveryverylong string. and, then,some,more,genes,then,some,more,genes,then,some,more,genes,then,some,more,genes,then,some,more,genes,then,some,more,genes.`
1. Case page should break on master
1. Case page should be OK using this branch
1. Test also by adding the same comment on a variant

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Case:

image


Variant: 

image

northwestwitch:
I wanted to include in this PR falso the fix for #2417 but fixed width image is not so easy scrollable as text apparently and it might take some time. Let's have it in another PR then.

This is ready for review!

scout/server/templates/utils.html
```diff
@@ -43,7 +43,7 @@ <h6 class="comments-title"></i><i class="far fa-comments"></i> Comments ({{ comm
     <style>.odd { background-color: #d5e8f0 }</style>
     {% for comment in comments %}
       <div class="row ml-1 alert {% if loop.index0 % 2 %}even{% else %}odd{% endif %}">
-          <div class="col-7">{{ comment.content }}</div>
+          <div class="col-7 text-break">{{ comment.content|fix_punctuation }}</div>
```
dnil, line 46: 👍

dnil (approved):

PR #2457, opened 2021-03-19
One col chromograph images

### LINKED ISSUE #1510: Hide Chromograph PNGs when page is too narrow
Too narrow browser screen will make Chromograph pictures line up in a strange way and make the feature useless.

Solution
* Hide images
* Link to separate page?
* Info text (keeping user informed of hidden pictures -always be kind to the end user

### LINKED ISSUE #2417: Increasing the size of a case page with Chromograph images breaks the layout

dnil:
This PR adds a functionality or fixes a bug. Fix #2417. Effectively fix #1510. 

Screenshot 2021-03-19 at 10 39 42

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
You can assign me as a reviewer of this one when it's ready!

scout/server/blueprints/cases/static/case_images.js
```diff
@@ -1,4 +1,5 @@
 const WIDTH_BREAKPOINT = 1955  // less will remove one column
+const SINGLE_COL_WIDTH_BREAKPOINT = 1555
```
northwestwitch, line 2: why exacly 1555?

scout/server/blueprints/cases/static/case_images.js
```diff
@@ -162,6 +164,10 @@ function clear_svg(svg_element){
  *
  */
 function set_svg_size(svg_element, number_of_columns){
+		if(number_of_columns == 1){
```
northwestwitch, line 167: This looks like the perfect situation to introduce a [switch](https://www.w3schools.com/js/js_switch.asp)!

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -333,15 +333,15 @@ <h5 class="modal-title" id="case_pick_dialog">Add other case to case group ({{ l
...
+    <div class="col-12">
       <div id={{"collapse_" + id_tag}}>
 	<div class="card panel-default">
 	  <div class="panel-heading" >
       <div class="panel-heading" href={{"#collapsing_" + id_tag}} data-toggle="collapse" datatarget={{"#collapsing_" + id_tag}} aria-expanded="true" aria-controls={{"collapsing_" + id_tag}} data-parent={{"#collapse_" + id_tag}}>Images {{display_name}} &gt	<span class="pull-right clickable"><span class="glyphicon glyphicon-chevron-up"></span></span></div>
           </div>
-          <div id={{"collapsing_" + id_tag}} class="collapse" data-parent={{"#collapse_" + id_tag}}>
-            <div class="card-body">
-              <svg id={{"svg_"+id_tag}} width="1200" height="1700"> </svg>
+          <div id={{"collapsing_" + id_tag}} style="width:100%" class="collapse" data-parent={{"#collapse_" + id_tag}}>
```
northwestwitch, line 342: 👍

scout/server/blueprints/cases/templates/cases/case_sma.html
```diff
@@ -2,7 +2,6 @@
 {% from "cases/collapsible_actionbar.html" import action_bar, research_modal, rerun_modal %}
 {% from "utils.html" import comments_panel, activity_panel %}
 {% from "cases/utils.html" import individuals_table %}
-{% from "cases/case_ideograms.html" import case_ideograms %}
```
northwestwitch, line 5: An ancient relic from the past?

northwestwitch (approved):
Works perfectly! 🥇 

image

I've got a small suggestion if you like it but it's fine already!

scout/server/blueprints/cases/templates/cases/case_sma.html
dnil: Yes, it did not seem to be doing much, and there was a clear phylogenetic similarity to the case_images.js that is actually in use. I actually made the edits in that one first.. 🙄

scout/server/blueprints/cases/static/case_images.js
dnil: It is based on scientifically exact calculation; unfortunately this window is too narrow to hold it, so the rest is black magic. It is based off the static image sizes given to the svg, which in turn depend on the relative chromosome sizes, also possible sidebar sizes. The part with ballparking common resolutions is the most magic; given a certain resolution or range I think we could settle on a formula. As it is, changing them slightly is advised if they misbehave.

scout/server/blueprints/cases/static/case_images.js
dnil: Hihi, thanks - it is like a Pascal dialect!

PR #2458, opened 2021-03-19
Handle VEPs new MANE annotations

bjhall:
Since VEP 103 and MANE 0.92 the MANE annotations in the CSQ has changed from `MANE` to `MANE_SELECT` and `MANE_PLUS_CLINICAL`.

We now load both these and mark them in the transcript list, using `M` for `MANE_SELECT` and `M+` for `MANE_PLUS_CLINICAL`. The colors for these badges feel pretty random, so I could change them if someone has a better idea.

bild

The old annotation format is still supported for backwards compatibility.


ehre:
Great news, @bjhall! This will be an important addition for us, and it would be good to be able to visually distinguish "M" and "M+" from the other annotations, with "M" and "M+" grouped together. How about a green background for instance? (But I am no UI designer ...)

bjhall:
> it would be good to be able to visually distinguish "M" and "M+" from the other annotations, with "M" and "M+" grouped together. How about a green background for instance? (But I am no UI designer ...)

Personally I find the badges to be more useful than adding more random background colors (if too many things are highlighted it will lose effect). 

Anyway, for the absolute majority of genes the primary and MANE Select transcript are the same, and will already be highlighted. Hightlighting it in a different color in the rare cases where primary and MANE are different would probably cause more confusion than clarity.

How about highlighting primary, MANE Select and MANE Plus Clinical with the same color (blue)? Most of the time one row will be highlighted. When several rows are hightlighted, simply move your eye to the badges.

ehre:
For the record, I was discussing badge color, I thought that was your question? So I meant "badge background color", sorry for any confusion.

bjhall:
Sorry, I was confusing it with the discussion we had the other day about highlighting the row...

I'll change the badge color!

bjhall:
I changed the badge color to green. Unless anyone objects this is ready to be merged.

bild

dnil (approved):
Looks good to me; tests are in place as well, and I kind of assume you have tested locally as well? If so, feel free to merge, otw we will rerun VEP on one of the test cases eventually.. 😸

ehre:
Looks nice, @bjhall ! But what happened to the Refseq transcript version number? In your picture, the "Mane Select" transcript NM_000335 appears on a different row from the "canonical" NM_000335.5 ?

bjhall:
@ehre That is unrelated to these changes. Not sure when/why that was changed. Maybe @dnil or @northwestwitch knows?

dnil:
It's the same as before, only less hidden since a little while back. The versioned "native" refseq transcript has always had a separate row since VEP started annotating it. The ENSEMBL transcripts had rather long lines of "closest" matching refseq transcripts, but these have never been versioned (although in principle that info is available for the MANE data). What happened is the

northwestwitch:
I start to wonder if it's the case to add a popover message to explain where each transcript comes from, I see that there is a bit of confusion among the users since the latest changes in this panel

dnil:
Good idea - also let's perhaps have a look at VEP and its references. If they had proper version numbers on there, things would be one step clearer.

ehre:
So, you should probably add me to your list of confused users, @dnil  ;-) This thread leaves me with two, perhaps separate, questions:
1. The primary MANE data (https://ftp.ncbi.nlm.nih.gov/refseq/MANE/MANE_human/) includes version numbers for all transcripts, and it would of course be beneficial if that annotation is visible in Scout for the users. So in this SCN5A example NM_000335.5 (ENST00000423572.7) for MANE Select and NM_001099404.2 (ENST00000413689.6) for MANE Plus Clinical.
2. To reduce clutter, it would also be beneficial if each transcript gets one and only one row in Scout.

dnil:
Yeah, you could add me there as well if you want - I don't get why VEP/ENSEMBL would ever not use versioned transcripts. Especially on hg38 and with MANE this should not be an issue.

We could restart the discussion on the respective pipelines or on the VEP git. I think it would be scary to try to merge these post hoc in Scout without versions. 

Again, yes, the view has changed to a slightly more readable, but those refseq native and ENSEMBL native ones have been there for a long time before that.

bjhall:
Since VEP gives version numbers for MANE transcripts we can think of of treating these differently in the table. I'll try to come up with a suggested solution soon!

northwestwitch:
Hi all, we had a little discussion at CG Stockholm today and we came up with the idea of having 2 distinct tables, one for genome build 37 (basically leaving it as it is) and one for build 38 (to display all the goodies from MANE perhaps?). And then according to the genome build of the case display one or the other. How do you like this?

bjhall:
Yes I was thinking along the same lines! I'll make a first attempt and we can dicuss the details in that PR!

PR #2465, opened 2021-03-23
DRAFT: Feature configparse 210217

mikaell:
Taking too long time. Must merge to Master. Pytests passes, then fails :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,11 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Parsing of AD and related info for STRs
 - LoqusDB field in institute settings accepts only existing Loqus instances
 - Fix DECIPHER link to work after DECIPHER migrated to GRCh38
+- Removed visibility window param from igv.js genes track
 ### Changed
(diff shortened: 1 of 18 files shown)


dnil:
Yeah I know the feeling.. Try breaking out smaller parts if possible and work on them one by one.. Also ping someone (best bet @northwestwitch?) if you want hints/help from someone who knows the day to day changes with merging conflicts!

northwestwitch:
> who knows the day to day changes with merging conflicts!

Life brings so much suffering, yes! 😄

northwestwitch:
This PR is introducing changes in the parsing of every key/value of the config file and for this reason we have to test it super carefully with a real case. But I was wondering if it would be possible somehow possible to leave the current test as they are, at least in a first moment, so we'd know that the result of the parsing is the same as before?

mikaell:
> This PR is introducing changes in the parsing of every key/value of the config file and for this reason we have to test it super carefully with a real case. But I was wondering if it would be possible somehow to leave the current test as they are, at least in a first moment, so we'd know that the result of the parsing is the same as before?

Yes. That is possible. In fact, that is what I've done and it works.

Issue #2454, opened 2021-03-18
gnomAD for mtDNA variants


bjhall:
Fairly recently gnomAD introduced data for mtDNA variants (https://gnomad.broadinstitute.org/blog/2020-11-gnomad-v3-1-mitochondrial-dna-variants/)

It would be nice to show these data for mt variants in Scout. One could of course simply use the same field as for normal variants (gnomAD_AF), but due to the nature of mt variants I propose to introduce two new VCF fields that would be loaded into scout:

`gnomAD_AF_mthom` and for homoplasmic frequencies
`gnomAD_AF_mthet` and for heteroplasmic frequencies

And then simply show them in the "Pop freq" column in the variant list and in the "Frequencies" panel of the variant details page.

We can start implementing this if no one has any objections.

hassanfa:
gnomad VCF file tags are `AF_hom` and `AF_het`, can we preserve the original tag and use `gnomAD_mt_AF_hom` and `gnomAD_mt_AF_het`?

bjhall:
@hassanfa Absolutely! I have no opinion about the names :).

PR #2470, opened 2021-03-24
Add gnomAD mitochondrial frequencies

bjhall:
Solves https://github.com/Clinical-Genomics/scout/issues/2454

This PR adds support for mitochondrial gnomAD frequencies. It can parse them from VEP's CSQ string or normal INFO fields. Either way, the fields should be called: `GNOMAD_MT_AF_HOM` and `GNOMAD_MT_AF_HET`

The frequency parsing/building/viewing in Scout is very messy (due to legacy stuff), and this adds even more complexity. However I've been careful not to disrupt anything, and have done fairly extensive testing to make sure everything is shown correctly.

It should change nothing for non-mitochondrial variants. And for mitochondrial variants it will show this instead:

Variant list:
bild

Variant details:
bild


northwestwitch:
I can review @bjhall!

scout/parse/variant/frequency.py
```diff
@@ -22,6 +22,7 @@ def parse_frequencies(variant, transcripts):
     # Gnomad have both snv and sv frequencies
     gnomad_keys = ["GNOMADAF", "GNOMAD_AF", "gnomad_svAF"]
     gnomad_max_keys = ["GNOMADAF_popmax", "GNOMADAF_POPMAX", "GNOMADAF_MAX"]
+    gnomad_mt_keys = ["GNOMAD_MT_AF_HOM", "GNOMAD_MT_AF_HET"]
```
northwestwitch, line 25: 👍

scout/parse/variant/frequency.py
```diff
@@ -59,6 +60,15 @@ def parse_frequencies(variant, transcripts):
             frequencies["gnomad_max"] = gnomad_max
             break
 
+    # For mitochondrial variants, keep both "hom" and "het" freqs
+    gnomad_mt_hom = parse_frequency(variant, "GNOMAD_MT_AF_HOM")
+    if gnomad_mt_hom is not None:
```
northwestwitch, line 65:
```suggestion
    if gnomad_mt_hom:
```

scout/parse/variant/frequency.py
```diff
@@ -59,6 +60,15 @@ def parse_frequencies(variant, transcripts):
...
+        frequencies["gnomad_mt_homoplasmic"] = gnomad_mt_hom
+
+    gnomad_mt_het = parse_frequency(variant, "GNOMAD_MT_AF_HET")
+    if gnomad_mt_het is not None:
```
northwestwitch, line 69:
```suggestion
    if gnomad_mt_het:
```

scout/parse/variant/transcript.py
```diff
@@ -154,6 +154,12 @@ def parse_transcripts(raw_transcripts, allele=None):
                 float(fractmatch) for fractmatch in superdups_fractmatch.split("&")
             ]
 
+        # Get mitochondrial gnomAD frequencies
+        if "GNOMAD_MT_AF_HOM" in entry and entry.get("GNOMAD_MT_AF_HOM") != "":
```
northwestwitch, line 158:
```suggestion
        if entry.get("GNOMAD_MT_AF_HOM", "") != "":
```

scout/parse/variant/transcript.py
```diff
@@ -154,6 +154,12 @@ def parse_transcripts(raw_transcripts, allele=None):
...
+            transcript["gnomad_mt_homoplasmic"] = float(entry.get("GNOMAD_MT_AF_HOM"))
+        if "GNOMAD_MT_AF_HET" in entry and entry.get("GNOMAD_MT_AF_HET") != "":
```
northwestwitch, line 160:
```suggestion
        if entry.get("GNOMAD_MT_AF_HET", "") != "":
```

northwestwitch (approved):
Nice new feature, and looks like it's working perfectly from your tests!

I've got a few (untested) small suggestions that you can accept if you want, but otherwise ot's fine to merge I think!

scout/parse/variant/frequency.py
bjhall: I was just sticking to the theme. Every other place in this file uses `is not None`. I could change it for all of them though, but I don't feel like rocking the boat too much :)

scout/parse/variant/frequency.py
bjhall: Same as above

scout/parse/variant/frequency.py
northwestwitch: Ok, but then let's stick to is not None, as it's also less prone to give unwanted results!

PR #2474, opened 2021-03-25
Fix SNP ids and links in variant page, general report and clinvar submission form

### LINKED ISSUE #2473: ClinVar - submission page freeze :(
There are some variants I can not submit as the submission page freeze :(

when i click the  
image

I send you the link of the 2 variants:

https://scout.scilifelab.se/cust003/19143/c8dce83866007943d3b2e4661290a8a2
https://scout.scilifelab.se/cust003/19091/4987140b170ac969865cfa93bf54fd2d

in both cases 19143 and 19091 there are 2 pinned variants per case but only one freeze :(

could you help me with this? I have an old memory that this has happend before but not how it was solved 
thanks :)

northwestwitch:
The current situation is that SNP IDs might be multiple. For instance [this variant](https://scout.scilifelab.se/cust003/19143/72f6fe76c1807ccbd6c31884e4574002) has the following SNP ID: `rs150429680;431849`

- The fist part of that string is the dbSNP id: https://www.ncbi.nlm.nih.gov/snp/rs150429680
- The second part part is the ClinVar variation ID: https://www.ncbi.nlm.nih.gov/clinvar/variation/431849

This PR:
- Fixes the ClinVar submission form by replacing the wrong SNP format with the dbSNP part only (fix #2473 )
- Fixes the variant page by splitting the wrong Id into 2 distinct ID with their links to the respective pages
- Fixes in the same way also the general report

**How to test**:
1. Locally, take a SNP variant and in the database replace its dbsnp_id with 'rs150429680;431849'
2. Check that the distinct links on the variant page work
3. Check that the distinct links on the general report work
4. Pin this variant and another, and on master branch and recreate the situation observed in the clinvar bug (https://github.com/Clinical-Genomics/scout/issues/2473#issuecomment-806023821)
5. Make sure that the submission using this branch works as it should

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -453,12 +453,21 @@ <h6>No matching causative or managed variants</h6>
...
           <li class="list-group-item">
-            SNPedia
+            SNP info
             <span class="float-right">
-              <a target="_blank" href="http://snpedia.com/index.php/{{ variant.dbsnp_id }}">
-                {{ variant.dbsnp_id }}
-              </a>
+              {% for snp in dbsnp_ids %}
+                {% if "rs" in snp %}
+                <a target="_blank" href="https://www.ncbi.nlm.nih.gov/snp/{{ snp }}" data-toggle="tooltip" data-placement="top" title="NCBI dbSNP ID">
```
dnil, line 462: `links.py`!

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -464,7 +464,18 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-secondary">Pa
             <td style="word-break: break-all;">{{variant.reference}} → {{variant.alternative}}</td>
             <td>
               {% if variant.dbsnp_id %}
-                {{variant.dbsnp_id}}
+                {% set dbsnp_ids = variant.dbsnp_id.split(';') %}
+                {% for snp in dbsnp_ids %}
+                  {% if "rs" in snp %}
+                    <a target="_blank" href="https://www.ncbi.nlm.nih.gov/snp/{{ snp }}">
```
dnil, line 470: `links.py`

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -464,7 +464,18 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-secondary">Pa
...
+                      dbSNP:{{ snp }}
+                    </a>
+                  {% else %}
+                    <a target="_blank" href="https://www.ncbi.nlm.nih.gov/clinvar/variation/{{ snp }}">
```
dnil, line 474: 🔁

dnil (approved):
Super! Interesting how the same underlying issue pops up. There is also the issue of the "new" Cosmic identifiers. I *think* they should be safely split off from the ID column to their own field during loading, but lets test a cancer case or two as well.. 😊

scout/server/blueprints/variant/templates/variant/variant.html
northwestwitch: Good idea, I'll move all links there!

northwestwitch:
Fixed according to the review: moved links creation to links.py and added a test. Tested again the modified pages and they look fine, I'd merge if you agree @dnil

northwestwitch:
> Super! Interesting how the same underlying issue pops up. There is also the issue of the "new" Cosmic identifiers. I _think_ they should be safely split off from the ID column to their own field during loading, but lets test a cancer case or two as well.. 😊

Cosmic IDs should be fine, because the're collected from a field named `cosmic_ids`. Check the code here: https://github.com/Clinical-Genomics/scout/blob/031042db30be06fe0c50d3e3649daaaf5c8b06a7/scout/server/links.py#L483

dnil:
> > Super! Interesting how the same underlying issue pops up. There is also the issue of the "new" Cosmic identifiers. I _think_ they should be safely split off from the ID column to their own field during loading, but lets test a cancer case or two as well.. 😊
> 
> Cosmic IDs should be fine, because the're collected from a field named `cosmic_ids`. Check the code here:
> 
> https://github.com/Clinical-Genomics/scout/blob/031042db30be06fe0c50d3e3649daaaf5c8b06a7/scout/server/links.py#L483

Yes, I agree; see e.g. here https://github.com/Clinical-Genomics/scout/blob/e05a51d5a037af88e5a740f5cdbe77508da6a7eb/scout/parse/variant/transcript.py#L229. Especially since this is on the transcript (CSQ) level for Balsamic, things should be all fine.

PR #2483, opened 2021-03-26
Populate variants filter chromosomes choices according to genome build of the case

### LINKED ISSUE #2481: Mitochondria contig name hg19 vs hg38
In hg38 genome builds contig name for mitochondria is named chrM/M whereas in hg19 it's named chrMT/MT.

Currently scout does not support the M notation and as such we cannot filter against the contig. 

Would it be possible to support both notations?

northwestwitch:
fix #2481

Default chromosomes list contains chromosome "MT" and whenever this is selected in the variants filters page it doesn't return any results for cases in genome build 38 (because its variants have chromosome = "M")

This PR moves the code that populate the choices of the chromosomes outside the Form and into the controllers, so that the chromosome choices reflect the case genome build.

**How to test**:
1. It would be ideal to have one case in genome build 38 and one in build 37.
2. Check that the form chromosomes select contains values and the filtering by chromosome work 

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Ready for review!

ViktorHy:
I became a bit hesitant, it would seem different software cannot handle M whereas some cannot handle MT. For instance, I believe VEP (latest version) does not annotate M, but does annotate MT. But CADD wants MT. Don't really now the best way going forward from here. What does your own hg38 look like?

northwestwitch:
Oh no, in that case my PR wouldn't work. I'm not really sure, we don't have any case in our database in genome build 38 yet. Ping @henrikstranneheim and @hassanfa: what is the MT chromosome in a VCF created using hg38, "M" or "MT"?

hassanfa:
@northwestwitch It should be M

henrikstranneheim:
To make matters even worse - MIP uses `chrM`

ViktorHy:
I think scout handles both chromosomes with chr and without though? We only use without and it has worked for years

hassanfa:
I meant M as in chrM and not chrMT. In balsamic, we will also have it as chrM.

What does the .fai for hg38 say? Standard fasta för hg38 seems to be chrM. We could all start using a single standard of chrM starting with hg38.

henrikstranneheim:
> We could all start using a single standard of chrM starting with hg38.

Sounds like a good idea!

northwestwitch:
Just checked the code and as suggested by @ViktorHy Scout removes the "chr" prefix when variants are parsed. Code is here: https://github.com/Clinical-Genomics/scout/blob/ab9eed67dc2eda514632ce471286fb741a7d2ad7/scout/parse/variant/coordinates.py#L144

where `CHR_PATTERN = re.compile(r"(chr)?(.*)", re.IGNORECASE)`

So it's fine to use either "chrM" or "M"

dnil (approved):
Good - much better than before! I miss a populate for the filter from managed variants though. Export looks fixed already, but there are some calls for e.g. exons from biomart that use CHROMOSOME. Does biomart handle that, or are they not fetched for hg38?

northwestwitch:
> Export looks fixed already, but there are some calls for e.g. exons from biomart that use CHROMOSOME. Does biomart handle that, or are they not fetched for hg38?

Genes and transcripts from chromosome build "38" contain chrom "MT" and not "M" (37 documents each):
image

image


Tried the command `scout load exons`, and  38 exons are also using "MT":
image

I think it's fine, as long as they're all using the same system.

I think I'd skip adding the build to the managed variants, since it would make a useful (necessary?) feature to add apart.

northwestwitch:
Is this one ready to merge?

PR #2485, opened 2021-03-26
Various mitochondria related things

bjhall:
This PR adds and fixes a number of things related to mitochondrial variants:

1. Parses and loads the "MITOMAP associated diseases" from VCF. We are annoatating this using https://github.com/robertopreste/HmtNote), and they call the INFO field "MitomapAssociatedDiseases".

This is shown in the variant list as a badge and in the tooltip:
bild

And in the summary of the variant page
bild

2. Load variant frequencies (for all variants, is that problematic?) and show them in various places (only for MT variants)

In the GT call table
bild

And for heteroplasmic variants in the variant list as badges. Heteroplasmies are arbirarily defined as VAF<90%
bild

It feels a bit strange that the badge is placed in the chromosome column, but I could not find a better place. Please suggest! :)

3. Finally it fixes the MitoMap link in the "External links". However if the variant doesn't exist i mitomap you'll get to an error page. But I think we have to live with that.

I'm happy to discuss any of these changes!


ehre:
Looks really good, @bjhall ! Of course, in the GT call table only maternal inheritance is relevant, but if for technical reasons it is easier to show the entire trio I can certainly live with that. And BTW, I would also be perfectly fine with seeing VAF% for all variants here and not only for mtDNA-variants (perhaps easier with a consistent visual interface, VAF is always relevant for somatic analyses and could yield some form of quality control even for constitutional samples, especially in mosaic cases).

bjhall:
@ehre I agree that it might be nice to always show VAF%. I don't know what the rational for not showing it for RD cases (they are already shown for somatic variants). Maybe because it's not very accurate at 30x?

@dnil @northwestwitch Is there anything about this PR that needs to be discussed/changed, of can we merge it?

northwestwitch:
> Is there anything about this PR that needs to be discussed/changed, of can we merge it?

HI, I'm on vacation and without computer at the moment. I could review when I'm back, on Tuesday next week. Looks like it might some black reformatting to see that tests pass

dnil:
> @ehre I agree that it might be nice to always show VAF%. I don't know what the rational for not showing it for RD cases (they are already shown for somatic variants). Maybe because it's not very accurate at 30x?

Showing both counts and ratios would be redundant. 
I think you nail it @bjhall. More digits will generally be needed to get the same precision with ratios for small numbers; for large numbers the ratio estimates help as they can be safely rounded and use fewer digits to convey the relevant info. 
It is common to want to know the exact count anyway if the nominator is small, so even then the VAF tends to be a little less useful on its own.

> @dnil @northwestwitch Is there anything about this PR that needs to be discussed/changed, of can we merge it?

Looks super from the images and description! I also left my regular computer for service over the holiday. Just let me know if you are in a hurry and I can scare up an editor on my clinic windows machine and review, but otherwise we'll be back next week! Happy Easter!

ehre:
My point was mainly that it might be good with visual consistency in the different SNV modules (RD, mtDNA and somatic), and that I see no harm in including VAF% also for 30x cases, as long as we also show counts (i.e. exactly as it was shown now in the GT call table that @bjhall posted).

scout/parse/variant/variant.py
```diff
@@ -248,6 +248,13 @@ def parse_variant(
     if somatic_score:
         parsed_variant["somatic_score"] = int(somatic_score)
 
+    ################# Add mitomap info #################
+    mitomap_associated_diseases = variant.INFO.get("MitomapAssociatedDiseases")
```
northwestwitch, line 252: 👍

scout/build/variant/variant.py
```diff
@@ -234,6 +234,10 @@ def build_variant(
     if "str_source" in variant:
         variant_obj["str_source"] = variant["str_source"]
 
+    # Mitochondria specific
+    if "mitomap_associated_diseases" in variant:
```
northwestwitch, line 238: From reading the code below, looks like this might be ".", why don't you check at this level and collect the value only if it's different from ".", so you avoid all the checks later in the templates?

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -599,7 +608,7 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
           {% endif %}
           <a href="{{ gene.ppaint_link }}" class="btn btn-outline-secondary" target="_blank">Protein Paint</a>
           {% if variant.chromosome == 'MT' %}
-            <a href="https://www.mitomap.org/foswiki/bin/view////Main/SearchAllele" target="_blank">MitoMap</a>
```
northwestwitch, line 602: Wow that looks weird!

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -599,7 +608,7 @@ <h4 class="modal-title">Cancel Sanger sequencing</h4>
...
+            <a href="{{ variant.mitomap_link }}" class="btn btn-outline-secondary" target="_blank">MitoMap</a>
```
northwestwitch, line 611: 👍

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -21,6 +21,9 @@
             {% elif variant.category == "str" %}
               <th rowspan="2" colspan="1">ExpansionHunter support</th>
             {% else %}
+	      {% if variant.chromosome == "MT" %}
```
northwestwitch, line 24:
```suggestion
	      {% if variant.chromosome in ["MT","M"] %}
```

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -48,6 +51,16 @@
                   <td class="text-right"><small>N/A</small></td>
                   <td class="text-right"><small>N/A</small></td>
               {% endif %}
+
+              {% if variant.category == "snv" and variant.chromosome == "MT" %}
```
northwestwitch, line 55:
```suggestion
              {% if variant.category == "snv" and variant.chromosome in ["MT","M"] %}
```

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -93,6 +93,9 @@
       <div>
         <strong>Local (arch.)</strong>: {{ variant.local_obs_old or 0 }} obs.
       </div>
+      {% if variant.mitomap_associated_diseases and variant.mitomap_associated_diseases != "." %}
```
northwestwitch, line 96: Yes, it would be nice not to save it at all in the database if it doesn't have valid data

scout/server/blueprints/variants/templates/variants/variants.html
```diff
@@ -80,7 +80,12 @@
                 <td><input type="checkbox" value="{{variant._id}}" name="dismiss" onclick="enableDismiss()"></td>
                 <td class="text-left">{{ cell_rank(variant, institute, case, form, manual_rank_options) }}</td>
                 <td class="text-right">{{ variant.rank_score|int }}</td>
-                <td>{{ variant.chromosome }}</td>
+                <td>
+                  {{ variant.chromosome }}
+                  {% if variant.chromosome == "MT" %}
```
northwestwitch, line 85:
```suggestion
                  {% if variant.chromosome in ["MT","M"] %}
```

scout/server/links.py
```diff
@@ -621,6 +622,12 @@ def alamut_link(variant_obj, build=None):
     return url_template.format(this=variant_obj, build_str=build_str)
 
 
+def mitomap_link(variant_obj):
```
northwestwitch, line 625: 👍

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Display on case activity the event of a user resetting all dismissed variants
 - Support gnomAD population frequencies for mitochondrial variants
 - Anchor links in Casedata ClinVar panels to redirect after renaming individuals
+- Load and show MITOMAP associated diseases from VCF (INFO field: MitomapAssociatedDiseases)
```
northwestwitch, line 25:
```suggestion
- Load and show MITOMAP associated diseases from VCF (INFO field: MitomapAssociatedDiseases, via HmtNote)
```

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -1,6 +1,17 @@
 {% import "bootstrap/wtf.html" as wtf %}
 {% from "utils.html" import comments_table %}
 
+{% macro mark_heteroplasmic_mt(individuals, samples) %}
+  {% for ind in individuals if ind.phenotype == 2 %}
+    {% for gt in samples|selectattr("sample_id", "equalto", ind.individual_id) %}
```
northwestwitch, line 6: It works nicely:


Perhaps make the badge a bit smaller so it fits better the column? (tested with chrome)

northwestwitch (changes requested):
Amazing new features!

Haven't tested the VCF parsing because we don't annotate using HmtNote, but looks tested and working fine in your instance. It would be super if you could add a small test for this perhaps?

I have some suggestions for improvement, mostly the dot thing, but looks really good!

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: If the badge looks OK on your browser is fine, could be the laptop I'm using today!

scout/build/variant/variant.py
bjhall: Good idea, fixed. I was annoyed with checking for the dot everywhere...

scout/server/blueprints/variants/templates/variants/utils.html
bjhall: I made the badge smaller by removing the `Het:` prefix to make it look good on smaller screens.

bjhall:
Thanks for the review @northwestwitch! I've fixed your requests and added a small test for the parsing of MitomapAssociatedDiseases.


Unrelatedly I noticed that `test_parse_cadd` in `tests/parse/test_parse_variant.py` is not working as indented (it can never fail)

```
    for variant in variants:
        # WHEN score is present
        if "CADD" in variant.INFO:
            cadd_score = float(variant.INFO["CADD"])
            parsed_variant = parse_variant(variant, case_obj)
            # THEN make sure that the cadd score is parsed correct
            assert parsed_variant["cadd_score"] == cadd_score
```

` if "CADD" in variant.INFO:` will never be true, so it will never get to the assertion. In the MitoMap test in this PR I used ` if variant.INFO.get("MitomapAssociatedDiseases"):`, which solved the problem. Not sure why... I haven't fixed this, just notifying about it here. Maybe other tests are affected as well. Maybe it's a bad idea to have tests where it is possible that the assertion is not reached.

northwestwitch:
> `if "CADD" in variant.INFO:` will never be true, so it will never get to the assertion.

OK! I've never noticed! I will check, could be due to an old way of parsing CADD info that is no more in use. Don't worry we can fix in another PR, but good that you found it! I'll re-review this PR now, good stuff!

northwestwitch (approved):
Looks good! I've noticed that you modified the demo VCF file and index in this PR. Consider to revert if you didn't intend to merge these modified files, otherwise feel free to merge!

PR #2487, opened 2021-03-26
Splice junctions view

### LINKED ISSUE #2344: Display RNA alignment data in snv/indels variants view on existing case
**Background**
When analysing an individual it is common to start with a DNA analysis. However, often a causative variant is not found and the same individual also are subjected to a RNA analysis. 

**Challenge**
Currently, there is no way to combine the output from the standard DNA analysis with the alignment output from a RNA analysis in Scout. It would be great to be able to update the DNA snv/indel view with alignment information for the RNA analysis in the form of a Sashimi plot: https://software.broadinstitute.org/software/igv/Sashimi. So when clicking the IGV (RNA) button you can go directly to a Sashimi plot of the gene(s) in question to look at the splice junctions across the samples in the case.

**Suggested solution**
- [ ] Add the option in `scout update case` to also be able to upload a link pointing to an alignment file (BAM|CRAM + index) from a secondary analysis to an already existing case.
- [ ] Also support this from the scout load config
- [ ] Add button for RNA Sashimi plots in IGV on the snv/indel variants view

### LINKED ISSUE #2439: Update case individuals with RNA-seq tracks
Update one or more individuals of a case from the command line, by adding another option to the `scout update individual `command:

image

northwestwitch:
This PR adds a functionality:
- Adds 2 types of tracks to update case individuals (fix #2439) with:
  -  splice_junctions_bed
  -  rna_coverage_bigwig
- Adds a button in variants page with a link to a splice junctions view (if at least one individual has this type of tracks) (fix #2344):
image

- Displays the sashimi plot with these tracks in a new window 

image

**How to test, locally**:
1. Download the demo track data from here: https://figshare.com/articles/dataset/Splice_Junctions_test_data/14381657
1. Update one or more individuals of the test case with the bigwig track:
```scout --demo update individual -c internal_id -n NA12882 rna_coverage_bigwig /path_to/Downloads/Downloads/junctions_test_data/sampleA.chr15-92835700-93031800.bigWig```
1. Update the same individual of the test case with the bed track:
```scout --demo update individual -c internal_id -n NA12882 splice_junctions_bed /path_to/Downloads/junctions_test_data/sampleA.chr15-92835700-93031800.SJ.out.bed.gz ```
1. Go to any variant of the case and click on the splice junctions link: you won't see any data because the demo tracks contain data only in the interval: **chr15:92835700-93031800**. Navigate to these coordinates and you'll see the sashimi plot

**Expected outcome**:

**Review:**
- [x] code approved by DN

### CODE REVIEW

**Comment by northwestwitch** [comment]
This is ready for review now!

File: scout/commands/update/individual.py [hunk 1]
```diff
@@ -13,6 +13,8 @@
     "tiddit_coverage_wig",
     "upd_regions_bed",
     "upd_sites_bed",
+    "splice_junctions_bed",  # An indexed junctions .bed.gz file obtained from STAR v2 aligner *.SJ.out.tab file.
```
dnil, line 16: These comments (and the new keys) would look really neat in the individual model on models/case/case.py!

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -1,15 +1,69 @@
...
+        gene_obj = store.hgnc_gene(hgnc_identifier=gene_id, build=build)
+        if gene_obj is None:
+            continue
+        gene_symbols.append(gene_obj.get("hgnc_symbol") or gene_obj["hgnc_id"])
+        locus_start_coords.append(gene_obj["start"])
+        locus_end_coords.append(gene_obj["end"])
+
+    locus_start = min(locus_start_coords)
+    locus_end = max(locus_end_coords)
+    locus = f"{variant_obj['chromosome']}:{locus_start}-{locus_end}"  # Locus will span all genes the variant falls into
```
dnil, line 39: 🥇

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -1,15 +1,69 @@
...
+
+    # Populate tracks for each individual with splice junction track data
+    _, case_obj = institute_and_case(store, institute_id, case_name)
+    for ind in case_obj.get("individuals", []):
+        if not all([ind.get("splice_junctions_bed"), ind.get("rna_coverage_bigwig")]):
+            continue
+
+        coverage_wig = ind["rna_coverage_bigwig"]
+        splicej_bed = ind["splice_junctions_bed"]
+        splicej_bed_index = f"{splicej_bed}.tbi" if os.path.isfile(f"{splicej_bed}.tbi") else None
```
dnil, line 50: Display will fail if the index is missing, but silently. I guess the target page has no flash, and it's perhaps not end user actionable info anyway: perhaps log something so it's apparent things are not quite well with the case?

dnil: Also note the index naming convention in docs somewhere!

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta content="Jim Robinson" name="author">
```
dnil, line 5: 😸

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -0,0 +1,82 @@
...
+    <link rel="shortcut icon" href="//igv.org/web/img/favicon.ico">
+    <title>igv.js</title>
+
+    <!-- IGV JS -->
+    <script src="https://cdn.jsdelivr.net/npm/igv@2.7.5/dist/igv.min.js"></script>
```
dnil, line 10: What do you say, is it starting to get time for a constant for this perhaps? I wouldn't mind that for all versioned js we have actually - they are not too many, but hunting around to update them is a little silly.

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -0,0 +1,82 @@
...
+                        cytobandURL: "https://s3.amazonaws.com/igv.broadinstitute.org/genomes/seq/hg38/hg38.fa.fai"
+                    },
+                locus: '{{ locus}}',
+                tracks:
+                    [
+                        {
+                            name: "Genes",
+                            type: "annotation",
+                            format: "gtf",
+                            url: "https://s3.amazonaws.com/igv.broadinstitute.org/annotations/hg38/genes/Homo_sapiens.GRCh38.80.sorted.gtf.gz",
```
dnil, line 39: We actually have those in constants; HG38GENES_URL etc - let's use them for all the URLs!

scout/server/blueprints/variant/controllers.py
```diff
@@ -45,6 +45,22 @@
 LOG = logging.getLogger(__name__)
 
 
+def has_rna_tracks(case_obj):
```
dnil, line 48: 💯

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -1,6 +1,6 @@
 {% macro igv_track_selection(igv_tracks, current_user) %}
 <div id="accordion">
-  <a data-toggle="collapse" data-parent="#accordion" href="#track_settings">Settings</a>
+  <a data-toggle="collapse" data-parent="#accordion" href="#track_settings"><i class="fas fa-cogs text-secondary"></i></a>
```
dnil, line 3: Nice! As it is a little to the side, of the button, and drops down to get close to the RNA button, probably keep a label anyway, but make it like "Settings DNA" or something that makes it distinct from the junctions track?

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -146,7 +150,7 @@
       {% if case.upd_sites_beds %}
         <input type="hidden" name="upd_sites_bed" value="{{case.upd_sites_beds|join(',')}}">
       {% endif %}
-      <button class="btn btn-outline-secondary btn-sm" name="align" value="bam" type="submit">IGV viewer</button>
+      <button class="btn btn-secondary btn-sm" name="align" value="bam" type="submit">IGV viewer</button>
```
dnil, line 153: Probably drop the btn-sm here as well - it looks a little odd next to the RNA track button.

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -90,7 +94,7 @@
     <input type="hidden" name="stop" value="{{variant.position + 50}}">
     <!-- end of general hidden form fields -->
     {% set aln = namespace(sample_names=[],mt_bams=[], mt_bais=[], bam_files=[], bai_files=[]) %}
-    {% if variant.chromosome == "MT" %}
+    {% if variant.chromosome in ["MT","M"] %}
```
dnil, line 97: 😄

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -1,3 +1,7 @@
+{% macro splice_junctions_button(institute_id, case_name, variant_id) %}
+  <a class="btn btn-secondary text-white mt-1" href="{{url_for('alignviewers.sashimi_igv', institute_id=institute_id, case_name=case_name, variant_id=variant_id)}}" target="_blank"
+    data-toggle="tooltip" data-placement="top" title="Gene splicing is only available in build GRCh38">Gene splicing</a>
```
dnil, line 3: Gene splicing makes my thoughts go to early biotech efforts. 😄  What do you think about "RNA splicing" or "Splice junctions"?

dnil (approved):
🚀 Super cool! This is a really nice step forward!

scout/server/blueprints/variant/templates/variant/buttons.html
dnil: Screenshot 2021-04-08 at 11 32 33

scout/server/blueprints/variant/templates/variant/buttons.html
northwestwitch: Now that you mention it, yes! "RNA splicing" sounds good!

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
northwestwitch: Sounds like a good idea, we could import it from somewhere!

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
northwestwitch: It's only fair I think!

scout/commands/update/individual.py
northwestwitch: Agreed!

scout/server/blueprints/alignviewers/controllers.py
northwestwitch: It will work without an index, I tried. Perhaps it will be a bit slower. But an alert would be nice to have, I agree!

dnil: Interesting; mine didn't - perhaps some library version differences or browser timeout thresholds somewhere? Anyways, even better, but a small alert here wouldn't hurt - we are verbose enough with other stuff.. 😝

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
dnil: Yes, like a constants file where we have URLs anyway? Seems like a Good Thing! Could be a separate PR if you tire of it.

scout/server/blueprints/alignviewers/controllers.py
northwestwitch: I've tried not with the files downloadable from figshare, but some complete profiles from a real analysis. Could be that affects the result as well!

northwestwitch: fixed:

image

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
northwestwitch: We can have it in here, I'm going to put it in the same file if the the igv tracks

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
northwestwitch: I've removed him no since this file is very original from the one I've copied from igv.js

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
northwestwitch: In the end it was easier to create a macro in a new utils.html file and have it there

scout/server/blueprints/variant/templates/variant/buttons.html
northwestwitch: Fixed it like this:

image

northwestwitch:
I think I've addressed all your comments. Thanks for the review btw! Do you want to give a final check @dnil ?

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -0,0 +1,100 @@
+{% from "alignviewers/utils.html" import igv_script %}
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <link rel="shortcut icon" href="//igv.org/web/img/favicon.ico">
+    <title>Scout: Splice junctions viewer</title>
+
+    {{ igv_script() }}
```
dnil, line 10: 👍

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -1,23 +1,27 @@
...
-            </select>
-          </div>
-          <div class="col-3">
-            <button type="submit" class="btn btn-sm btn-outline-secondary">Save</button>
-          </div>
-      </div>
-    </form>
+  <div class="row">
+    <div class="col-4">
+     <p><a data-toggle="collapse" data-parent="#accordion" href="#track_settings" class="text-secondary">settings DNA <i class="fas fa-cogs text-secondary"></i></a></p>
```
dnil, line 5:
```suggestion
     <p><a data-toggle="collapse" data-parent="#accordion" href="#track_settings" class="text-secondary">Settings DNA <i class="fas fa-cogs text-secondary"></i></a></p>
```

dnil: Otw looks really good! So cool!

PR #2502, opened 2021-04-07
Fix tests for CADD score parsing functions

northwestwitch:
As noticed by @bjhall the current test for cadd parsing is not working. This is a fix

**How to test**:
1. Automatic tests, only code review is required!

**Expected outcome**:
All tests should pass

**Review:**
- [ ] code approved by
- [x] tests executed by pytest


northwestwitch:
An easy review!

tests/parse/test_parse_deleteriousness.py
```diff
@@ -0,0 +1,29 @@
+from scout.parse.variant.deleteriousness import parse_cadd
+
+
+def test_parse_cadd_info(cyvcf2_variant, case_obj):
+    """Test parsing a variant that has CADD score in the info field"""
+
+    variant = cyvcf2_variant
+    assert "CADD" not in variant.INFO
```
dnil, line 8: Not wrong, but since you actually assign over it afterwards, not much use in checking either? Did you perhaps mean to double check that there are no other sources of CADD info on the variant? You do have control over that here as well, but..

tests/parse/test_parse_deleteriousness.py
```diff
@@ -0,0 +1,29 @@
...
+    cadd_score = parse_cadd(variant, [])
+
+    # THEN the parse cadd function should return the expected value
+    assert cadd_score == float(variant.INFO["CADD"])
+
+
+def test_parse_cadd_transcripts(cyvcf2_variant, transcript_info):
+    """Test parsing a variant that has CADD score in the transcript dictionary"""
+
+    assert "cadd" not in transcript_info
```
dnil, line 21: same here?

tests/parse/test_parse_transcripts.py
```diff
@@ -177,6 +177,20 @@ def test_parse_superdups_fractmatch():
         assert transcript["superdups_fracmatch"] == fract_match
 
 
+def test_parse_cadd(vep_csq_header, vep_csq):
+    """Testing parsing of CADD_PHRED score from VEP-annotated transcripts"""
+
+    # GIVEN a transcript with the CADD score in th CSQ
+    header = [word.upper() for word in vep_csq_header.split("|")]
+    raw_transcripts = [dict(zip(header, entry.split("|"))) for entry in vep_csq.split(",")]
```
dnil, line 185: I would be more keen to know that we actually saw CADD here first? Still, if it wasn't there the test would fail, so that is a pure convenience.

dnil (approved):
Looks good and needed! The current approach is a little hopeful, and prone to silently break when the example files change. I wouldn't mind the csq test being a little bit more explicit about how the csq info is stored / parsed, but it works and is already an improvement!

tests/parse/test_parse_deleteriousness.py
northwestwitch: The idea was to simply test the parse_cadd function in scout.parse.variant.deleteriousness, not what's in variant.INFO when the cyvcf2_variant is created. To do that it is perfect that I assign the cadd in variant info, so I have control over the input and the output of the function

tests/parse/test_parse_deleteriousness.py
northwestwitch: yes :)

tests/parse/test_parse_deleteriousness.py
dnil: Yes, but `variant.INFO["CADD"] = 23` is already overwriting whatever would have been found with `assert "CADD" not in variant.INFO`..

dnil: It look as if you check for something else?

tests/parse/test_parse_transcripts.py
northwestwitch: I didn't get it 🤔 I picked vep_csq_header and vep_csq because they contain a transcript with CADD score that can be shown to be parsed?

tests/parse/test_parse_deleteriousness.py
dnil: It would make sense to check that the transcripts on that test variant have no `"cadd"`, because otherwise the test may not do what was intended..

northwestwitch: > Yes, but `variant.INFO["CADD"] = 23` is already overwriting whatever would have been found with `assert "CADD" not in variant.INFO`..

Ah now I get it, it looks like the test was meant to check something in variant.INFO because I wrote`assert "CADD" not in variant.INFO`. It's not, it was just to make sure I started from a "blank" state of variant.INFO. I'll remove that line, so it will be less confusing!

tests/parse/test_parse_transcripts.py
dnil: It's fine - just a little implicit, compared to the very explicit ones above! 😄 
It is all good as long as no-one pokes with the fixtures - and if someone does the test likely just fails as it should. You could assert that there was actually an appropriate "cadd" in the dict after explicitly making the raw transcript locally though, since that part is not calling a function in the main code? See what I mean?

tests/parse/test_parse_deleteriousness.py
northwestwitch: > It would make sense to check that the transcripts on that test variant have no `"cadd"`, because otherwise the test may not do what was intended..

I'm not passing any transcript to the test, it's an empty list.

tests/parse/test_parse_transcripts.py
northwestwitch: Mmm I'm not sure I get what you mean. There are other 6 tests in that file that use the same fixtures exactly the same way and I took inspiration (ready copy) from them:

- test_parse_functional_annotation
- test_parse_optional_hgnc_annotation
- test_parse_vep_freq_thousand_g_alt
- test_parse_vep_freq_gnomad
- test_parse_vep_freq_mtgnomad
- test_parse_hg38_mane_transcripts

This test, like the other ones, is calling `parse_transcripts` (a function in the main code) to create parsed transcripts.

You want me to check that raw_transcripts contains a cadd value and that corresponds to the same value of the parsed transcripts?

northwestwitch: I made it a bit more explicit, I think!

tests/parse/test_parse_transcripts.py
```diff
@@ -184,11 +184,14 @@ def test_parse_cadd(vep_csq_header, vep_csq):
     header = [word.upper() for word in vep_csq_header.split("|")]
     raw_transcripts = [dict(zip(header, entry.split("|"))) for entry in vep_csq.split(",")]
 
+    tx_cadd = float(raw_transcripts[0].get("CADD_PHRED"))
```
dnil, line 187: 👍

tests/parse/test_parse_transcripts.py
dnil: 💯

PR #2506, opened 2021-04-08
Remove cyvcf2 restraint

dnil:

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Well, at least the tests ran some 25% faster. I dunno, maybe that is partly load dependent, but I had the subjective feeling of  distinctly more responsive variantS pages.

northwestwitch:
> at least the tests ran some 25% faster

Say no more

northwestwitch (approved):

dnil:
I'll try it out on stage. Fingers crossed...

dnil:
Mm, it works just fine on `hasta` `S_scout`, but not on `clinical-db` `stage`. The cyvcf2 build fails on a cyvcf2.c inclusion stage, but thats for python3.6. For python3.7 on hasta, there is a nice dowloadable manylinux wheel, which just bypasses build and installs. 🙄

northwestwitch:
> Mm, it works just fine on `hasta` `S_scout`, but not on `clinical-db` `stage`. The cyvcf2 build fails on a cyvcf2.c inclusion stage, but thats for python3.6. For python3.7 on hasta, there is a nice dowloadable manylinux wheel, which just bypasses build and installs. 🙄

OK then we should probably wait until we move to the new server. Told you that the software on clinical-dB was from the Mesozoic!

northwestwitch:
And when we merge this I guess it will become a major, since python 3.6 will be no longer supported? Another checkbox to add to the major release changes!

dnil:
I don't really think its really py36; its more that there is a shortcut to not need to have all the build stuff for cyvcf, which is kind of .c-inclusion-heavy. If it was py36, notice you would have our py36 test matrix half failing, or at least pulling old cyvcf, but it seems fine, right?

dnil:
But we should for sure not merge this until we have a plan. Since we have no eta for new server, best would be to separate out scout into its "own" env on clinical (like possibly together with chanjo, maybe even loqus etc but not with like cg).

northwestwitch:
> I don't really think its really py36; its more that there is a shortcut to not need to have all the build stuff for cyvcf, which is kind of .c-inclusion-heavy. If it was py36, notice you would have our py36 test matrix half failing, or at least pulling old cyvcf, but it seems fine, right?

Right, the matrix is testing against py36 and it's not failing!

Dockerfile
```diff
@@ -9,10 +9,33 @@ LABEL about.license="MIT License (MIT)"
 # Install required libs
 RUN apk update
 RUN apk --no-cache add make automake gcc g++ linux-headers libffi-dev zlib-dev \
-   jpeg-dev libressl-dev cairo-dev pango-dev gdk-pixbuf ttf-freefont bash
+   jpeg-dev libressl-dev cairo-dev pango-dev gdk-pixbuf ttf-freefont bash \
+   autoconf musl-dev perl bzip2-dev xz-dev curl-dev
 RUN pip install numpy Cython
 
+# Install cyvcf2
+RUN apk add git
```
northwestwitch, line 17: Ah yes, we could add --no-cache here.

Dockerfile
```diff
@@ -9,10 +9,33 @@ LABEL about.license="MIT License (MIT)"
 # Install required libs
 RUN apk update
 RUN apk --no-cache add make automake gcc g++ linux-headers libffi-dev zlib-dev \
-   jpeg-dev libressl-dev cairo-dev pango-dev gdk-pixbuf ttf-freefont bash
+   jpeg-dev libressl-dev cairo-dev pango-dev gdk-pixbuf ttf-freefont bash \
+   autoconf musl-dev perl bzip2-dev xz-dev curl-dev
 RUN pip install numpy Cython
```
northwestwitch, line 14: Perhaps we could follow Codefactor's suggestion?

Dockerfile
```diff
@@ -9,10 +9,33 @@ LABEL about.license="MIT License (MIT)"
...
+WORKDIR /tmp
+RUN git clone --recursive https://github.com/brentp/cyvcf2
+WORKDIR /tmp/cyvcf2/htslib
+RUN autoheader
+RUN autoconf
+RUN ./configure --enable-libcurl
+RUN make
+WORKDIR /tmp/cyvcf2
+RUN pip install -r requirements.txt
+RUN CYTHONIZE=1 pip install -e .
```
northwestwitch, line 27: I had seen this solution [here](https://github.com/brentp/cyvcf2/pull/143#issuecomment-606077539) as well. It means recompile Cython? Then it should install it on clinical-db right?

northwestwitch (approved):
Nice solution! We should test it on clinical-db and if everything works fine let's merge!

Dockerfile
northwestwitch: We would probably need to change the scripts of clinical-db to have CYTHONIZE=1 as well right?

dnil: That might actually just work; it still needs htslib, which I thought was the issue here (needs a --recursive pull from git, as it is only linked). Actually now that I think about it, we should probably change to a versioned release as well - it's all good to use cyvcf2-master for development, but maybe not ideal if this turns into the preferred deployment method

Dockerfile
dnil: Yes for apk, all hands down. I'm not quite sure about pip, but I suppose it will still use the previously installed ones anyway, so probably all good.

Dockerfile
dnil: For reference, this gets stuck wanting `bzip2-devel` on clinical-db. Not insurmountable, but tedious and complex since it is not available prepackaded from conda. (Also, the autoconf result is a little weird; looks like autotools are  not quite perfectly set up either?)

dnil:
Ok lets close this for now, but revisit soon. Its silly to have an old and slower version going just because a system library is missing on clinicaldb..

PR #2509, opened 2021-04-09
Simple Docker compose dev file

dnil:
This PR adds a simple modified docker compose to use local build dir instead of docker hub images for scout web and cli.

**How to test**:
1. `docker-compose -f docker-compose-dev.yml up`
2. Scout with a demo instance should now be available on port 5000

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,0 +1,51 @@
+version: '3'
+# usage:
+# (sudo) docker-compose up -d
+# (sudo) docker-compose down
+services:
+
+  mongodb:
+    image: mvertes/alpine-mongo
+    container_name: mongodb
+    ports:
+      - '27017:27017'
+    expose:
+      - '27017'
+    networks:
+      - scout-net
+
+  scout-cli:
+    build: .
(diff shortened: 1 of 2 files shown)


docker-compose-dev.yml
```diff
@@ -0,0 +1,51 @@
...
+    container_name: mongodb
+    ports:
+      - '27017:27017'
+    expose:
+      - '27017'
+    networks:
+      - scout-net
+
+  scout-cli:
+    build: .
```
northwestwitch, line 18: Was this one (and the same change later) the only difference from the non-dev docker-compose?

northwestwitch (approved):
👍

docker-compose-dev.yml
dnil: Yep. So far. It might be possible to do an editable install, with mounting from outside the container, but lets play it safe with small rebuilds for now.

dnil: As long as requirements are not touched, the cached rebuilds should be fast.

PR #2513, opened 2021-04-12
Open variant in a new window

### LINKED ISSUE #2512: Open variant in a new window
When clicking on a variant from variantS page

northwestwitch:
- Open variants in new a new window (from variants page) . Fix #2512 
- Open overlapping variants (SNVs and SVs) in a new window, when clicked on a variant page.

**How to test**:
1. From variantS page (snvs, svs, strs, cancer vars), when a user clicks on variant link, the variant page is opened in a new window.
2. Check that overlapping variants in variant page (SNVs and SVs) open correctly in a new window. This will be better tested on clinical-db

**Expected outcome**:
- Try to open variants from all types.

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Refactor according to CodeFactor - mostly reuse of duplicated code
 - Phenomodels language adjustment
-- Open variants in a new window from variants list
+- Open variants in a new window (from variants page)
+- Open overlapping and compund variants in a new window (from variant page)
 
 ## [4.31.1]
 ### Added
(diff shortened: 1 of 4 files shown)


northwestwitch:
This is ready. An easy review!

dnil (approved):
Works for me, but since it is not "pop-under" by default in all browsers (I just noticed Safari shifts focus directly to the new tab), some users may still prefer to use a context menu and open in background. Lets try this anyway and see what they say?

Perhaps add it to the compounds as well - feels a little odd when the overlapping ones go to blanks but not the compounds?

northwestwitch:
> Perhaps add it to the compounds as well - feels a little odd when the overlapping ones go to blanks but not the compounds?

Good idea, will do!

PR #2515, opened 2021-04-12
gnomAD 2 --> 3 for MT variants

### LINKED ISSUE #2479: Fix gnomAD link for MT variants in build 37
> > How is the situation in current master regarding the MT gnomAD frequency for genome build 37? Always None? Tested a few and looks like it. Both the values parsed from the VCF and the link.
> > While there isn't much to do for the values parsed from the VCF, especially because we are all moving to build 38 anyway, the link to gnomAD can be adjusted to use gnomAD version 3.11 regardless of the genome build, right? Perhaps in another pull request?
> 
> That sounds right to me; perhaps we double check that with HS & AJ before changing so we are very sure the links are correct, and document that so no-one else goes 37 with some old MT build and gets the wrong position. But there is no need to keep them for 37:
> 
> ```
> $ zgrep MT /home/proj/production/rare-disease/references/references_9.0/grch37_gnomad_reformated_-r2.1.1-.vcf.gz
> $ 
> ```
> 
> and
> Screenshot 2021-03-25 at 15 09 35

Moving this to a dedicated issue!

_Originally posted by @northwestwitch in https://github.com/Clinical-Genomics/scout/issues/2475#issuecomment-806829179_

northwestwitch:
Fix #2479 --> Change the link from gnomAD v.2 (build 37) to gnomAD 3 (GRCh38) for mitochondrial variants from cases sequenced in build 37. Since MT variants are not aligned using genome build 37, but 38 (kindof).

**How to test**:
1. Locally or on clinical-db. 
2. On master, under frequency, notice that all MT variants contain links to gnomAD and that this link doesn't work (it tells you that `Mitochondrial variants are not available in gnomAD v2.1.1`
3. Switch to this branch and see that the gnomAD link works because it points to gnomAD 3 instead of gnomAD 2.1.1

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -457,20 +457,16 @@ def exac_link(variant_obj):
 
 
 def gnomad_link(variant_obj, build=37):
-    """Compose link to gnomAD website."""
+    """Compose link to gnomAD website for a variant."""
+    url_template = (
+        "http://gnomad.broadinstitute.org/variant/{this[chromosome]}-"
+        "{this[position]}-{this[reference]}-{this[alternative]}"
+    ).format(this=variant_obj)
 
     if build == 38 or variant_obj["chromosome"] in ["M", "MT"]:
-        url_template = (
-            "http://gnomad.broadinstitute.org/variant/{this[chromosome]}-"
-            "{this[position]}-{this[reference]}-{this[alternative]}?dataset=gnomad_r3"
-        )
-    else:
-        url_template = (
-            "http://gnomad.broadinstitute.org/variant/{this[chromosome]}-"
-            "{this[position]}-{this[reference]}-{this[alternative]}"
-        )
+        url_template += "?dataset=gnomad_r3"
 
-    return url_template.format(this=variant_obj)
+    return url_template
 
 
 def swegen_link(variant_obj):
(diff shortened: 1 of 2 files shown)


northwestwitch:
Noooo, complex code! I'll try to simplify

northwestwitch:
Meh I changed my mind, I don't feel like changing this whole file that has complex code for such a small change. Hope the reviewer agrees (🙏 ). Ready for review

scout/server/links.py
```diff
@@ -457,20 +457,16 @@ def exac_link(variant_obj):
...
-        url_template = (
-            "http://gnomad.broadinstitute.org/variant/{this[chromosome]}-"
-            "{this[position]}-{this[reference]}-{this[alternative]}?dataset=gnomad_r3"
-        )
-    else:
-        url_template = (
-            "http://gnomad.broadinstitute.org/variant/{this[chromosome]}-"
-            "{this[position]}-{this[reference]}-{this[alternative]}"
-        )
+    if build == 38 or variant_obj["chromosome"] in ["M", "MT"]:
```
dnil, line 466: 👍

dnil (approved):

Issue #2507, opened 2021-04-09
coverage file missing


4WGH:
![image](https://user-images.githubusercontent.com/12525201/114151131-1926b300-991d-11eb-9c55-9c9821c1343e.png)

fam 18150
there is not coverage file....
bug?

Michela

northwestwitch:
We recently made a change because users analyzing cancer cases didn't want to see coverage and MT reports on the sidebar, so we are showing them only cases with `track:"rare"`.

This case which should belong to the track "rare" doesn't have a key/value `track:"rare"` in the database. I can see this difference comparing for instance with [this other case](https://scout.scilifelab.se/cust003/20166). 

Now I don't know why it is that and we'll need to check if there are more. Should be fixed now that I've added manually the right track to the case. Thanks for reporting the error!

4WGH:
thank you for such a fast fix :)

northwestwitch:
Also missing the coverage and MT report: [15056](https://scout.scilifelab.se/cust003/15056), uploaded the day after 18150. Cases before and after these 2 look fine. I'm fixing the 15056 as well and closing this, but please reopen if you find any other of this.

I don't think we need to take further action because current cases looks fine. Must have been some bug in the process that has been fixed immediately after these 2 cases

PR #2535, opened 2021-04-14
Display mitochondrial and rare diseases coverage report in cases with missing track

### LINKED ISSUE #2536: Missing reports in Scout - coverage and mtDNA
I just noticed that some reports are suddenly missing in many rerun cases in Scout.

One example is case 18114, where both coverage and mtDNA reports are missing. I know the mtDNA report was there the 1st of April when I worked with the case and exported the report.

Other cases are for example 18187, 17119, 18005, 16043, 14029 but probably many other cases are affected.

Could you please check what has happened.

northwestwitch:
This PR fixes 2 things:
- Display rare disease coverage report and MT report for RD cases with no saved track
- Fix the bug that causes the case track not to be saved in the database when a case is re-uploaded (fix #2536 and #2507)

**How to test**:
1. Locally, upload the demo case (has "rare" track)
2. From mongo compass, remove the track key from demo case
3. Using master branch, go to demo case page and make sure that MT report is not available in the side panel.
4. From master branch, force reupload the case with the command: 
```scout --demo load case scout/demo/643594.config.yaml -u```
This will cause the case to be reuploaded without a track key, even if the key exists in case the config file
5. MT reports should still be not available from master branch
6. Switch to this branch and make sure that MT is available for the case
7. On this branch, repeat the force upload of the case. This time track : "rare" will be saved in the database for this case
8. MT report should still be present on the sidebar of the case

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### CODE REVIEW

File: scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html [hunk 1]
```diff
@@ -85,7 +85,7 @@
         </div>
       </div>
     {% endif %}
-    {% if config.SQLALCHEMY_DATABASE_URI and case.track == "rare" %}
+    {% if config.SQLALCHEMY_DATABASE_URI and case.track != "cancer" %}
```
dnil, line 88: 👍

dnil (approved):

scout/adapter/mongo/case.py
```diff
@@ -689,7 +689,11 @@ def update_case(self, case_obj, keep_date=False):
         The following will be updated:
             - collaborators: If new collaborators these will be added to the old ones
             - analysis_date: Is updated to the new date
-            - analyses: The new analysis date will be added to old runs
+            - delivery_report: path to the static delivery report file
```
northwestwitch, line 692: I've rearranged this comment to reflect the order of the items that are updated later in the code. Some items where missing from here.

scout/adapter/mongo/case.py
```diff
@@ -775,6 +782,7 @@ def update_case(self, case_obj, keep_date=False):
                     "mme_submission": case_obj.get("mme_submission"),
                     "status": case_obj.get("status"),
                     "variants_stats": case_obj.get("variants_stats"),
+                    "track": case_obj.get("track"),
```
northwestwitch, line 785: track was not updated on re-upload!

northwestwitch:
When removing case "track" in database:
- [x] master branch doesn't display MT report
- [x] this branch does

- When force-reuploading the case from master:
-  [x] track is **not** saved in database
- [x] master branch doesn't display MT report

- When force-reuploading the case from this branch:
- [x] rare track is saved in database
- [x] master branch displays MT report

northwestwitch:
I feel ready with this one. Are you still approving it @dnil?

scout/adapter/mongo/case.py
dnil: 😅  There was this concept in computer science called OOD. It could have been useful here when propagating info  for this object here. A bit of inheritance with polymorphism could have made this a little easier. Anyway, this is in the adapter, so all good I suppose. Was that the only missing one?

dnil: Maybe we should at the very least sort according to either alphabet or the order in the "models" object. Pydantic may help a bit as well later.

scout/adapter/mongo/case.py
dnil: Good! Ideally reorder with models and/or alphabet, but it's all good otherwise.

scout/adapter/mongo/case.py
northwestwitch: I can sort alphabetically, and sort the model as well. I'll check if we forgot anything with the model object

northwestwitch: Did the alphabetical sorting and found out that using models for checking that all the keys are taken into account in adapters is actually not the best way to act --> In fact I've found 2 keys that are used in case adapter but not found in case model (i've added them now).

I think a more comprehensive check of all the possible keys of a case (some might still be missing in the model) and their use in this adapter function is beyond the purpose of this fix. Hopefully using pydantic will help find more of these issues!

PR #2539, opened 2021-04-15
Removed OMIM and updated causative panel for cancer cases

mhkc:
This PR updates the layout of the case view for cancer cases to be better suited to that type of analysis

image

**How to test**:
1. load the demo cancer case
2. navigate to the case view

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
The diagnosis list is under discussion for all really, but perfectly good to start turning it off here already. OMIM isn't very detailed on cancer. But enlighten an RD and inherited cancer person: why no causatives?

scout/server/blueprints/cases/templates/cases/individuals_table.html
```diff
@@ -2,7 +2,7 @@
 {% macro cancer_individuals_table(case, institute, tissues) %}
 <form method="POST" action="{{ url_for('cases.update_cancer_sample', institute_id=institute._id, case_name=case.display_name) }}">
   <div class="card panel-default">
-    <div class="panel-heading">Individuals</div>
+    <div class="panel-heading">Samples</div>
```
dnil, line 5: 👍

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -124,21 +126,23 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
...
-                <div class="card-body d-flex flex-row">
-                  <div class="col-sm-6">
-                    {{ diagnosis_phenotypes(case, institute, omim_terms) }}
-                  </div>
-                  <div class="col-sm-6">
-                    {{ diagnosis_genes(case, institute) }}
-                  </div>
-                </div><!--end of card body-->
-            </div><!--end of card-->
+        {% if not case.track == 'cancer' %}
```
dnil, line 129: :+1:

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -91,7 +91,9 @@ <h4 class="card-header">Case: {{case.display_name}}</h4>
       {% endif %}
 
       <div class="row ">
-        <div class="col-xs-12 col-md-6">{{ causatives_list(causatives, partial_causatives, institute, case) }}</div>
+        {% if not case.track == 'cancer' %}
+          <div class="col-xs-12 col-md-6">{{ causatives_list(causatives, partial_causatives, institute, case) }}</div>
```
dnil, line 95: We will need to differentiate further to remove this methinks? At least for inherited cancer and hematology it makes a lot of sense to work with "causative" variants. If you worry about the wording for a complex solid tumor, perhaps we could introduce concepts like "driver, putative" or qualify the case highlighted variants with "actionable", "biomarker" - or make it "phenotypically significant" variants?

dnil (changes requested):

bjhall:
@dnil Personally I don't see the point of having the causative concept for somatic variants in general, but please let me how you are using it for hematology! We often find 2-3 Tier I variants, and I don't know what the point of also marking them as causative (currently it is only possible to select one causative, right?). Sure, marking them as "therapeutic", "prognostic", "actionable" etc would be relevant, but why squeeze that into the "causative" concept?

(As always I think it would be great if geneticists at different sites could reach a consensus about these thinks).

Btw, are you using the "cancer track" in scout for inherited cancer? Otherwise maybe it should be renamed to "somatic" rather than "cancer"? We are currently using the RD track för inherited cancers.

mhkc:
One option could be to add the tier and variant tag information to pinned variants to differentiate variants that are interesting from diagnostic or technical perspective.

image

Or we could add a panel that contains all Tier I and II variants

dnil:
> @dnil Personally I don't see the point of having the causative concept for somatic variants in general, but please let me how you are using it for hematology! 

The important thing I guess is wether to retain three categories - I think we can find a naming that works. I quite agree "causative" is a bit uncomfortable if you have only somatic variants, and even sometimes for germline - e.g. if presymptomatic or otherwise prognostic. How about "Clincal significance"/"Clinically significant"? We would have to check the reception of it obviously,  as "causative" has been with us for many years.

To us, pinned suspects are a mechanism more to flag things for local discussion or to keep an eye on for later than to say they are of clinical interest to report back. Once deemed strong enough - eg after a phenotype has been double checked or a few clinicians consulted - they can graduate from pinned suspect or go back to the hundreds of thousands not reported. 

Also, thats where we currently have the option to differentiate the effects of several different variants (partial causatives currently), and the ones that we primarily track for recurrence - though the classifications could be altered and used for that as well. 

We would need a new view for looking at all old classifications - currently we have the "causatives" view for that.

I guess it may be easier to adhere to the workflow from RD and inherited cancer in hematology where in general, like brain tumors, pediatric tumors etc, usually has a low number of even actual somatic variants. And even fewer of actual clinical interest naturally.

> We often find 2-3 Tier I variants, and I don't know what the point of also marking them as causative is (currently it is only possible to select one causative, right?).

I kind of agree: it makes little sense to mark them "suspect" for later rounds and discussion: I would more or less immediately flag them "causative" if they are clear Tier I. We could always auto-include them on a "clinically sinificant" list if Tier I/II or ACMG P/LP.

> Sure, marking them as "therapeutic", "prognostic", "actionable" etc would be relevant, but why squeeze that into the "causative" concept?

As long as we have a category. We do not have to mark them specifically either - that can always go into comments. 
> (As always I think it would be great if geneticists at different sites could reach a consensus about these thinks).

Have you been to one of our meetings yet? 😉 I think one can really raise an eyebrow when there is unison on any issue, large or small. It is a field still very much in development, despite its age. We have much to learn.
 
> Btw, are you using the "cancer track" in scout for inherited cancer? Otherwise maybe it should be renamed to "somatic" rather than "cancer"? We are currently using the RD track för inherited cancers.

We have a mixed bag there, and the current best-guess roadmap is to mix them even more, and have both germline and somatic views on the same case.

> @dnil One option could be to add the tier and variant tag information to pinned variants to differentiate variants that are interesting from diagnostic or technical perspective.
>
> Or we could add a panel that contains all Tier I and II variants

@mhkc Yes, it seems reasonable to auto-qualify them! I feel there could sometimes be variants of somewhat unknown significance that don't quite fit the category that could be of interest to report, as with VUS in RD or inherited cancer, but I will not press that one. Need to ask around for a consensus. Couldn't like "novel" mutations in established driver genes end up in Tier III sometimes? Or will they always be considered prognostic?

dnil:
tl;dr:
How about renaming "causatives" to "clinically significant" and possibly auto-mark tier I, tier II as causative?

mhkc:
I have fixed those things now so it should be fine to merge it

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -173,33 +173,47 @@ <h5 class="modal-title" id="exampleModalLabel">Beacon submission</h5>
...
+
+{% macro variant_validation_badge(variant) %}
+  {% if variant.sanger_ordered and not variant.validation in ['True positive','False positive'] %}
+    <span class="badge badge-default">Verification ordered</span>
+  {% elif variant.sanger_ordered %}
+    <span class="badge badge-success">Validated</span>
+  {% endif %}
+{% endmacro %}
+
+{% macro tumor_allele_freq(variant) %}
```
dnil, line 185: 👍

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -285,6 +299,27 @@ <h5 class="modal-title" id="exampleModalLabel">Beacon submission</h5>
       {% if not case.causatives|length and not case.partial_causatives|length %}
         No variants marked causative
       {% endif %}
+      <!-- variants with cancer tier I or II -->
+      {% if case.track == 'cancer' %}
+        {% for variant in evaluated_variants %}
+          {% set exp_tiers =  ['1A', '1B', '2C', '2D'] %}
```
dnil, line 305: Also at least P/LP in ACMG right?

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -218,7 +232,7 @@ <h5 class="modal-title" id="exampleModalLabel">Beacon submission</h5>
       <!-- Partial causative variants -->
       {% for variant_phenotypes in partial_causatives %}
         {% set variant = variant_phenotypes.variant %}
-        <div class="list-group-item">
+        <div class="list-group-item list-group-item-action">
```
dnil, line 235: 👍

scout/server/blueprints/cases/templates/cases/utils.html
dnil: It would be less confusing to show each variant only once, but with all badges etc. Consider e.g. if there would be a slightly ambiguous DEL call, that is both Tier 1B, LP and marked causative - one could get the impression there were several similar variants.

dnil (approved):
This is an improvement already, but do consider adding ACMG categories and collapsing the list to display variants only once. There is a risk of confusing the users otherwise.

dnil:
Ok, have a look now! I noticed what in my somewhat starved state seems an odd issue with the RD case suspect and causative list display. I just hid it under if cancer.track tags, but if you immediately see why the assessments are populated for the RD case view but not the cancer one, feel free to fix it. That would perhaps have been a better solution to that whole thing anyway. 🙄 But it can be a later thing. We will anyway want to merge your nice new cancer variantS rank independent view to the RD side, and that would be a good time to also move the assessments fully into cancer.

Screenshot 2021-05-21 at 13 02 44

Screenshot 2021-05-21 at 13 02 32

PR #2540, opened 2021-04-15
Improved check if case was rerunned

### LINKED ISSUE #2534: Almost all prod cases are rerun? -- CG Solna issue --
Check for instance cases from cust003: https://scout.scilifelab.se/cust003. How can it be that almost all their cases are rerun??? Same for cust002.
Cases on scout-stage look OK so should be some parameter used when loading them?

Thanks @emmser for noticing! 🙏

northwestwitch:
Fix #2534 

The database document of a real rerunned case looks like this (example: https://scout.scilifelab.se/cust003/15029):

image

A non-rerunned case that gets the same `rerun` label looks like this (example: https://scout.scilifelab.se/cust003/46):

image

This PR is fixing this behavior.

**How to test**:
- Locally, on master branch, load a case and re-create the database situation shown in the second figure above.
- Make sure that this branch doesn't show the case as rerun in the case list page.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -171,7 +171,14 @@ def cases(store, case_query, prioritized_cases_query=None, limit=100):
         case_obj["assignees"] = [
             store.user(user_email) for user_email in case_obj.get("assignees", [])
         ]
-        case_obj["is_rerun"] = len(case_obj.get("analyses", [])) > 0
+        # Check if case was re-runned
+        analyses = case_obj.get("analyses", [])
+        now = datetime.datetime.now()
+        case_obj["is_rerun"] = (
+            len(analyses) > 1
+            or analyses
+            and analyses[0].get("date", now) < case_obj.get("analysis_date", now)
+        )
         case_obj["clinvar_variants"] = store.case_to_clinVars(case_obj["_id"])
         case_obj["display_track"] = TRACKS[case_obj.get("track", "rare")]
         return case_obj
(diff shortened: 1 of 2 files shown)


northwestwitch:
Local test, after modifying the case like this:

image


Master branch:

image

This branch: 🆗 
image

scout/server/blueprints/institutes/controllers.py
```diff
@@ -171,7 +171,14 @@ def populate_case_obj(case_obj):
         case_obj["assignees"] = [
             store.user(user_email) for user_email in case_obj.get("assignees", [])
         ]
-        case_obj["is_rerun"] = len(case_obj.get("analyses", [])) > 0
+        # Check if case was re-runned
+        analyses = case_obj.get("analyses", [])
+        now = datetime.datetime.now()
+        case_obj["is_rerun"] = (
+            len(analyses) > 1
+            or analyses
```
dnil, line 179: Nice! I worry about the order of checking for existence and use of analyses[0] though; it may be ok but did you test with an old/fake case with empty analyses?

dnil (approved):

scout/server/blueprints/institutes/controllers.py
northwestwitch: One could change the `or analyses` into `or len(analyses) == 1`

northwestwitch: If you have empty analyses 'analyses' is False and you don't get into the condition

northwestwitch: Test with an empty "analysis" array: 

image

Does not crash and of course does not show 'rerun' badge: 🆗 

image

PR #2544, opened 2021-04-15
iSort check only

dnil:
This PR removes the somewhat naive attempt at a modifying iSort action. We may return to that at some point, but checks only for now.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


.github/workflows/linting_and_fixing.yml
```diff
@@ -40,8 +40,3 @@ jobs:
         # stop the build if there are Python syntax errors or undefined names
         flake8: true
         flake8_args: "scout/ --count --select=E9,F63,F7,F82 --show-source --statistics"
-
-    - name: Run isort
```
northwestwitch, line 44: 👍

northwestwitch (approved):
👍

PR #2553, opened 2021-04-19
Deploy docs action

dnil:
This PR adds a functionality: auto deploy docs on release

**How to test**:
1. how to test it: release a release and check that the docs are updated (or test by moving it to trigger on push once if it looks good to the reviewer?)

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -59,13 +59,14 @@ jobs:
    name: Deploy Docs to GitHubIO
    runs-on: ubuntu-latest
    steps:
-    - name: Checkout main
+    - name: Checkout repo
       uses: actions/checkout@v2
+    - name: Setup environment for docs deployment
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.x
+    - name: Install mkdocs
+      run: pip install mkdocs-material
+    - name: Deploy documentation
+      run: mkdocs gh-deploy --force
 
-    - name: Deploy docs
-      uses: mhausenblas/mkdocs-deploy-gh-pages@master
-       # Or use mhausenblas/mkdocs-deploy-gh-pages@nomaterial to build without the mkdocs-material theme
-      env:
-       GITHUB_TOKEN: ${{ secrets.github_token }}
-       CONFIG_FILE: mkdocs.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_publish.yml
```diff
@@ -55,3 +55,17 @@ jobs:
...
 
+ deploy-docs:
+   name: Deploy Docs to GitHubIO
+   runs-on: ubuntu-latest
+   steps:
+    - name: Checkout main
+      uses: actions/checkout@v2
+
+    - name: Deploy docs
+      uses: mhausenblas/mkdocs-deploy-gh-pages@master
```
northwestwitch, line 66: The sh file runned by this action looks quite complicated. I wonder if it wouldn't be easier to just run the usual commands as "run" in here instead. But fine if it works!

northwestwitch (approved):
Hard to test unless we deploy. I think it looks harmless enough to merge. Good to have an action to update the docs! 👍

dnil:
> Hard to test unless we release. I think it looks harmless enough to merge. Good to have an action to update the docs! 👍 

Yes, perhaps not as bad as some other similar ones for pdf as well. Do you have a good trick for uploading though - the regular gh-deploy doesn't take a GitHub.secret does it?

northwestwitch:
> Yes, perhaps not as bad as some other similar ones for pdf as well. Do you have a good trick for uploading though - the regular gh-deploy doesn't take a GitHub.secret does it?

if you have a secret already set I think you're good, check here: https://squidfunk.github.io/mkdocs-material/publishing-your-site/

dnil:
> > Yes, perhaps not as bad as some other similar ones for pdf as well. Do you have a good trick for uploading though - the regular gh-deploy doesn't take a GitHub.secret does it?
> 
> if you have a secret already set I think you're good, check here: https://squidfunk.github.io/mkdocs-material/publishing-your-site/

Lets try - that would be super nice! I still have a nagging feeling you would have to do what the mhausenblas script does to pass the token indirectly to MkDocs via setting it on the remote repo. But lets see, perhaps there is some good behind the scenes automagic that will set it for us!

PR #2559, opened 2021-04-21
Add dimissed compound info

northwestwitch:
OR

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Seriously @moonso ??? 😩 

image

northwestwitch:
@moonso very good to know which dependencies we should checked!
But for some of them, like weasyprint, we already know that they're older than my granpa, and they should be kept to have the case general report working as it should. Current situation is that that test will always fail, which is very annoying. Isn't there a way to pass the test with warning messages only?

northwestwitch:
@moedarrah @JessicaHillert this one

northwestwitch:
Mark this PR as ready for review when you are done, so someone reviews!

scout/server/blueprints/variants/controllers.py
```diff
@@ -300,11 +300,10 @@ def parse_variant(
     has_changed = False
     compounds = variant_obj.get("compounds", [])
     if compounds and get_compounds:
-        # Check if we need to add compound information
-        # If it is the first time the case is viewed we fill in some compound information
-        if "not_loaded" not in compounds[0]:
```
northwestwitch, line 305: This change is likely to slow down the loading of the variants page. This is because before update_variant_compounds was called just once if the not_loaded key wasn't present in compound dict. Now it is called every time a variants page is loaded. The thing is you can'y know a priory if the compound is dismissed, or if the dismissed status has changed since last time you were on variantS page. 

We should test it on stage and see how much slower it gets. Any other solution is welcome @dnil

dnil: Testing is good. Without profiling, it does seem a bit overkill to have the compounds be re-added each time the page is loaded. Perhaps add like a populate_compounds to the controller instead that just quickly fills the relevant data based on existing compound variant ids rather than always going to the case load version? Im fairly sure that one was intended as a one-off.. The dismiss status may have changed, but the gene symbol and compound variant id would still be the same.

If we index the events collection we can go that way instead as well. There will just be a handful of dismiss events for each case, and anyway we know the variant id from the compound already..

northwestwitch: Yes, I like the second solution better, with the events should be faster! I'll fix this branch accordingly. Brilliant! 👍

scout/adapter/mongo/event.py
```diff
@@ -203,7 +203,6 @@ def add_phenotype(
             else:
                 raise ValueError("Must supply either hpo or omim term")
         except ValueError as ex:
-            ## TODO Should ve raise a more proper exception here?
```
dnil, line 193: Yes, never leave a TODO for codefactor... 😅

northwestwitch: Coveralls is still down! If it wasn't for the security breach of few days ago I'd be tempted to switch to codecov.. I like it quite much!

northwestwitch: Perhaps it is Coveralls that is breached now? 😆

dnil: Yes.. Their status was still green, but the queue-time graph peaks so I guess it is real.

dnil: I don't mind switching - while I'm kind of used to coveralls, if you think codecov is better.. 👍🏻
But let's not do both.

northwestwitch: I think we have more urgent stuff to fix. But it might be a valid alternative if Coveralls keeps having problems

dnil: We were fairly lucky to get the release out on time, before whatever hit codecov went here:
Screenshot 2021-04-26 at 16 44 21
Maybe someone was finally fed up with having to write another test? 😉

dnil: Hm, yes, actually sounds like a new Solarwind. Maybe they actually did both at the same time, only Coveralls reacted a little slower.

northwestwitch: Indeed!

northwestwitch:
I've changed the python code of this one to use the events to collect the list of dismissed variants for a case, to be compared with with the compounds later in controllers. It took some more code then the first version, but now it won't be updating the compounds every time the variants page is loaded. Could you review when you have time @dnil?

northwestwitch:
Oh no, this has a conflict now. I'll fix it!

dnil (commented):
Ehrm, but now it broke, didn't it? Or do you see it working locally somehow? 😊

dnil:
> Ehrm, but now it broke, didn't it? Or do you see it working locally somehow? 😊

Sorry, my bad - old branch running. 🙄

northwestwitch:
> Ehrm, but now it broke, didn't it? Or do you see it working locally someh

Haven't tested after merging the conflict. could be that. I'll check!

northwestwitch:
> > Ehrm, but now it broke, didn't it? Or do you see it working locally somehow? 😊
> 
> Sorry, my bad - old branch running. 🙄

Ah OK!

dnil:
> > Sorry, my bad - old branch running. 🙄
> 
> Ah OK!

I juggled two different installations while comparing changes to the cancer variantS branch, and forgot to switch the install back..

scout/adapter/mongo/event.py
```diff
@@ -155,6 +155,19 @@ def events(
...
+        """Return events with a specific verb for a case of an institute
+        Args:
+          institute (dict): an institute id
+          case (dict): a case id
+          verb (dict): an event action verb, example: "dismiss_variant"
+
+        Returns:
+            pymongo.Cursor: Query results
+        """
+        query = {"institute": institute["_id"], "case": case["_id"], "verb": verb}
```
dnil, line 168: We should include category here for two reasons: 1) it will generally keep the number of hits down - no reason to search different categories for dismiss here at least, 2) there is already an index on events for category-verb, which will buy some speedup at least. Also consider adding a very specific index for this query. It will have quite a use count, and events are collecting quickly - not as fast as variants obviously, but still many per case.

scout/server/blueprints/variants/controllers.py
```diff
@@ -270,8 +287,39 @@ def get_manual_assessments(variant_obj):
     return assessments
 
 
+def compounds_need_updating(compounds, dismissed):
```
dnil, line 290: 👍

scout/server/blueprints/variants/controllers.py
dnil: Good compromise!

dnil (approved):
Works really well locally! I think for the long haul we should at least use, and better yet add at least one more index to events.

scout/adapter/mongo/event.py
northwestwitch: Good idea! I've included the category in the query now.

There exist already these 2 indexes for the events collection:

"event": [
        IndexModel([("category", ASCENDING), ("verb", ASCENDING)], name="category_verb"),
        IndexModel([("variant_id", ASCENDING)], name="variant_id"),
    ],

I'm not very skilled with indexes, something like this?
IndexModel([("institute", ASCENDING), ("case", ASCENDING), ("category":ASCENDING), ("verb", ASCENDING)], name="case_verb"),

dnil: Yes like so! We could also try with just the added category as it will hit one of the indices you listed, and add your full suggestion to that old index-overhaul branch. I'll brush that branch up any week now, since we got a nice new contribution to the cancer variantS view, but I don't think we should merge until around when we can accept some downtime/slowness for rebuilding.

northwestwitch: Ok, let's do that instead. Merging this one, thanks for the review!

PR #2560, opened 2021-04-22
/api/v1/genes queries all genome builds by default

mhkc:
This PR fixes a bug where the user could not search for genes from grch38 in `/genes`

**How to test**:
1. goto `/genes`
2. search for "NOTCH2NLA" (in grch37 and 38)
3. search for "NOTCH2NLC" (in grch38)

**Expected outcome**:
- Genes should be found regardless of genome build
- Should only show one hit per gene, regardless if a gene is present in multiple genome builds

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -117,8 +117,8 @@ class GeneHandler(object):
         """
         LOG.debug("Fetching genes with symbol %s" % hgnc_symbol)
         build_query = {}
-        if str(build) in ['37', '38']:
-            build_query['build'] = str(build)
+        if str(build) in ["37", "38"]:
+            build_query["build"] = str(build)
 
         if search:
             # first search for a full match
@@ -127,8 +127,10 @@ class GeneHandler(object):
             if nr_genes != 0:
                 return self.hgnc_collection.find(query_full_match)
 
-            return self.hgnc_collection.find({'aliases': {"$regex": hgnc_symbol, "$options": "i"}, **build_query})
-        return self.hgnc_collection.find({'aliases': hgnc_symbol, **build_query})
+            return self.hgnc_collection.find(
+                {"aliases": {"$regex": hgnc_symbol, "$options": "i"}, **build_query}
+            )
+        return self.hgnc_collection.find({"aliases": hgnc_symbol, **build_query})
 
     def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
         """Find one hgnc genes that match a hgnc symbol. Replaces depricated
@@ -154,7 +156,7 @@ class GeneHandler(object):
                 {"hgnc_id": int(hgnc_symbol) if hgnc_symbol.isdigit() else None},
             ],
         }
-        if build in ['37', '38']:
+        if build in ["37", "38"]:
             query["build"] = str(build)
         return query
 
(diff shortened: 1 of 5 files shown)


northwestwitch:
I've installed this PR on out stage server (Stockholm) so it's ready for testing also today, Otherwise I can test it and review tomorrow!

northwestwitch:
@mhkc tried a quick search for NOTCH2NL and I get no suggestions for the one in build 38 only:
image

I'm pretty sure we should have genes in both builds there but I will check better tomorrow!

mhkc:
The query is working but it is a bit wonky at times. I think it is the result of the initial query to find perfect matches. Note that NOTCH2NL is one of the aliases of NOTCH2NLA.

gene_query

Should I remove the initial perfect matches?

dnil:
Sorry @northwestwitch - that is probably simply due to my aborted update test? It seems we don't have build:"38" on stage at the moment. Will fix!

northwestwitch:
> Sorry @northwestwitch - that is probably simply due to my aborted update test? It seems we don't have build:"38" on stage at the moment. Will fix!

Yes I suspected it, immediately after writing this comment. Thanks!

dnil:
Works after updating genes! Sorry about that!

Screenshot 2021-04-23 at 10 44 28

scout/adapter/mongo/hgnc.py
```diff
@@ -116,20 +116,21 @@ def hgnc_genes(self, hgnc_symbol, build="37", search=False):
             pymongo.cursor
         """
         LOG.debug("Fetching genes with symbol %s" % hgnc_symbol)
+        build_query = {}
+        if str(build) in ["37", "38"]:
+            build_query["build"] = str(build)
```
dnil, line 121: :+1:

scout/server/blueprints/genes/controllers.py
```diff
@@ -52,16 +52,16 @@ def gene(store, hgnc_id):
     return res
 
 
-def genes_to_json(store, query):
+def genes_to_json(store, query, build):
     """Fetch matching genes and convert to JSON."""
-    gene_query = store.hgnc_genes(query, search=True)
-    json_terms = [
-        {
+    gene_query = store.hgnc_genes(query, build, search=True)
```
dnil, line 57: 👍

dnil (approved):
Works just fine! As you have identified a problem that could perhaps recur, a test would be super nice. That is no absolute requirement, just good practice!

scout/server/blueprints/genes/controllers.py
```diff
@@ -52,16 +52,16 @@ def gene(store, hgnc_id):
...
+    json_terms = {
+        gene["hgnc_id"]: {
```
dnil, line 59: The solutions appear very similar to me - was there some weird duplication going on, or was this just part of debugging? 😊

scout/server/blueprints/genes/controllers.py
mhkc: its basically passing on the build argument. I debugged the code at some time so there might appear to be more different than it actually is

scout/server/blueprints/genes/controllers.py
mhkc: its basically passing on the build argument. I debugged the code at some time so there might appear to be more different than it actually is

mhkc:
@dnil I have added a few tests as well. Feel free to merge when you have the time

tests/adapter/mongo/test_gene_handler.py
```diff
@@ -204,6 +204,58 @@ def test_get_genes_regex(real_adapter):
     assert sum(1 for i in res) == 3
 
 
+def test_get_genes_per_build(real_adapter):
```
dnil, line 207: 👍

PR #2567, opened 2021-04-26
Minor version 4.33

northwestwitch:
New minor version:

### Fixed
- Command line crashing when updating an individual not found in database
- Dashboard page crashing when filters return no data
- Cancer variants filter by chromosome
- /api/v1/genes now searches for genes in all genome builds by default
- Upgraded igv.js to version 2.8.1 (Fixed Unparsable bed record error)
### Added
- Autodeploy docs on release
- Documentation for updating case individuals tracks
- Filter cases and dashboard stats by analysis track
### Changed
- Changed from deprecated db update method
- Pre-selected fields to run queries with in dashboard page
- Do not filter by any institute when first accessing the dashboard

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil (approved):

northwestwitch:
It's now on clinical-db stage:

### Fixed
- [ ] Command line crashing when updating an individual not found in database
- [ ] Dashboard page crashing when filters return no data
- [ ] Cancer variants filter by chromosome
- [ ] /api/v1/genes now searches for genes in all genome builds by default
- [ ] Upgraded igv.js to version 2.8.1 (Fixed Unparsable bed record error)
### Added
- [ ] Autodeploy docs on release
- [ ] Documentation for updating case individuals tracks
- [ ] Filter cases and dashboard stats by analysis track
### Changed
- [ ] Changed from deprecated db update method
- [ ] Pre-selected fields to run queries with in dashboard page
- [ ] Do not filter by any institute when first accessing the dashboard

northwestwitch:
Checking the dashboard changes and got this (perhaps) peculiar result when filtering for cases with a non-existent phenotype:

image


Weird thing is that I get results also from cases page:

image

dnil:
> Checking the dashboard changes and got this (perhaps) peculiar result when filtering for cases with a non-existent phenotype:
> Weird thing is that I get results also from cases page:

My memory is not perfect there, but I would think we never picked a cutoff for similarity really - which is kind of fine in a list sorted by similarity but a tad odd for a stats module.

northwestwitch:
> My memory is not perfect there, but I would think we never picked a cutoff for similarity really - which is kind of fine in a list sorted by similarity but a tad odd for a stats module.

Isn't it weird that with that search it is returning 24/95 cases in cases page anyway? I mean, who has that phenotype. Perhaps it is returning case with any phenotype.

dnil:
Yes in falling order of similarity to whatever term matched best to the query?! Or to None?

northwestwitch:
> one valid hpo term perhaps

Yes, that looks like the way to go. Should we add to this release or a later one?

dnil:
> > one valid hpo term perhaps
> 
> Yes, that looks like the way to go. Should we add to this release or a later one?

Yeah, its all cases with phenotypes now. Its not that bad for blabla on statistics, possibly worse on the case page with a subtly malformed hpo-term.. But, not really a big churn on that feature, so fine to make an issue and go on methinks.

northwestwitch:
Deployed now:

image

PR #2572, opened 2021-04-26
Minor mt fixes

bjhall:
This PR fixes MitoMap link for hg38 cases and adds a link to HmtVar for mitochondrial variant (if the INFO-field HmtVar is defined in the VCF. We're using HmtNote to annotate the VCF with this field)

Also added a test for HmtVar parsing.


scout/server/links.py
```diff
@@ -625,6 +626,12 @@ def mitomap_link(variant_obj):
     return url_template.format(this=variant_obj)
 
 
+def hmtvar_link(variant_obj):
```
northwestwitch, line 629: 👍

scout/server/links.py
```diff
@@ -625,6 +626,12 @@ def mitomap_link(variant_obj):
...
+    """Compose a link to a variant in HmtVar"""
+    url_template = "https://www.hmtvar.uniba.it/varCard/{id}"
```
northwestwitch, line 631: Works perfectly: 
image

northwestwitch: An improvement would be to link to the actual position of the variant in HmtVar: https://www.hmtvar.uniba.it/varCard/39192

tests/parse/test_parse_variant.py
```diff
@@ -103,3 +103,14 @@ def test_parse_mitomapassociateddiseases(variants, case_obj):
             parsed_variant = parse_variant(variant, case_obj)
             # THEN make sure that it is parsed correct
             assert parsed_variant["mitomap_associated_diseases"] == mitomap_associated_diseases
+
+
+def test_parse_hmtvar(variants, case_obj):
+    # GIVEN some parsed variant dicts
+    for variant in variants:
+        # WHEN "mitomap associated diseases" is present
+        if variant.INFO.get("HmtVar"):
```
northwestwitch, line 112: Tests are much appreciated! 🥇  

For how it's written I realize that if one day we modify scout/demo/643594.clinical.vcf.gz by removing the MT variant containing HmtVar and MITOMAP info, the relative tests will still be successful because you'll never enter the condition in this line, and line 101 in the MITOMAP test.

Perhaps a better way of testing the parsing would be something like this ?
```
def test_parse_hmtvar(cyvcf2_variant, case_obj):
    """Test parsing HmtVar value from variant annotated with HmtNote"""

    # GIVEN a variant containing HmtVar key in the INFO field:
    cyvcf2_variant.INFO["HmtVar"] = "39192"

    # THEN make sure that it is parsed correctly
    hmtvar_variant_id = int(cyvcf2_variant.INFO["HmtVar"])
    parsed_variant = parse_variant(cyvcf2_variant, case_obj)    
    assert parsed_variant["hmtvar_variant_id"] == hmtvar_variant_id
```

northwestwitch (approved):
Nice!
Works as it should and code is neat! I have a couple of small suggestions. Feel free to fix the real link to HmtVar or not, but I think the tests (the one from from this PR and the MITOMAP one) could be improved. But I really appreciated the effort to include the test 💯

scout/server/links.py
bjhall: Hmm, I don't see the difference? At least that was what I was trying to achieve, and it works here. What does the link look like for you?

tests/parse/test_parse_variant.py
bjhall: I agree that silently "failing" tests should be avoided, but I also think it feels more realistic to parse an actual file. But I don't have a better suggestion, so I'll change it according to your suggestion (tomorrow)! :)

northwestwitch: An alternative would be do the parsing of the real file and then run a mongo query to look for a variant with those specifics..

scout/server/links.py
northwestwitch: I agree, It doesn't really matter. I just noticed that the link for this demo variant:

image

leads to another variant on HmtVar: https://www.hmtvar.uniba.it/varCard/12345, probably because you used a random number (12345) as hmtvar ID. The id for that specific variant is another, but it's fine, really, since it's just test data.

tests/parse/test_parse_variant.py
bjhall: I took the easy way out and implemented them according to your first suggestion :)

scout/server/links.py
bjhall: Yes sorry, I was lazy and just put a random number there...

bjhall:
Hopefully this should be ready to merge now!

northwestwitch:
> Hopefully this should be ready to merge now!

I think so! Good job with this PR!

PR #2573, opened 2021-04-27
Sort ClinVar submission variants and case data by ascending "Last Evaluated" date

### LINKED ISSUE #2519: Order ClinVar submission sub-items by date
what is the order the variants get listed?
image

- it is not the order they are added (which will be the best in case I want to check the variant  and the case just added to the submission went in
- not the order of coordinates
- no gene name
- no family name

It would be helpful to know  so I can cross check the data without having to look (randomly ) for it  
The longer the list of variants  (and cases) the more I need to look up and down to find the newly added variants and case (I need to change the name to each case so it is very important to do it correctly :)

northwestwitch:
Fix #2519

Introduced changes:
- Expose link to Clinvar submissions in Institute page sidebar. In current master you need to expand sidebar "shared data" to see it.
- Order Submissions variants by ascending "last_evaluated" date.
- Order case data according to the order of variants 

**How to test**:
1. Pin and submit (with casedata) at least 2 variants without modifying the "last_evaluated" field in the submission form.
1. Note the order in which these variants are displayed in the submission object list, for instance this:
image

1. In the database, edit the second variant with a date older than today ("clinvar" collection).
1. Make sure that both the second variant and its case data are shown first now in the submission object

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


scout/server/blueprints/institutes/templates/overview/institute_sidebar.html
```diff
@@ -35,14 +35,13 @@
               </div>
           </a>
 
-          <!-- shared data menu item -->
-          <a href="#submenu1" data-toggle="collapse" aria-expanded="false" class="bg-dark list-group-item list-group-item-action flex-column align-items-start">
+          <!-- ClinVar data menu item -->
+          <a href="{{ url_for('overview.clinvar_submissions', institute_id=institute._id) }}" class="bg-dark list-group-item list-group-item-action flex-column align-items-start">
```
dnil, line 39: 👍

tests/adapter/mongo/test_clinvar_handler.py
```diff
@@ -7,6 +7,21 @@
 LOG = logging.getLogger(__name__)
 
 
+def test_sort_case_data(adapter):
```
dnil, line 10: 👍

dnil (approved):
💯

PR #2574, opened 2021-04-27
Configurable variant options

mhkc:
This PR enables storage of variant dismiss terms in mongo instead being hard coded in constants files.

It is currently only functional for variant.html

**How to test**:
1. add a dismiss term with `scout load evaluation_terms`
2. goto http://10.0.224.64:5000/cust000/643594/4c7d5c70d955875504db72ef8e1abe77
3. dismiss the variant

**Expected outcome**:
Retain current functionality

**Review:**
- [ ] code approved by
- [ ] tests executed by


mhkc:
I dont think the codefactor issues relates to changes I have made

northwestwitch:
> I dont think the codefactor issues relates to changes I have made

No, they were there since before! I took the liberty to fix them. 😃

northwestwitch:
Hi, I haven'r really started looking at the code of this PR (it's quite a big one) and I figured it's better to understand how it is supposed to work first. I've got a few questions:

1. 

1. Custom Evaluations terms will be bulk-loaded from file and saved to database when the `scout setup demo` or `scout setup database` cli command is invoked
1.  Same for variant dismiss options right?
1.  Single variant evaluation terms can also be loaded one at the time using the cli.
1. Is there a command to bulk-load all evaluation terms when you basically have a server already with data, like we have now in production?
1. Is your PR description actual (it works only for variant.html (not cancer variants, no strs, no sv?)

A consideration: 
Evaluation terms and variant dismiss options provided in don't correspond to the terms we (Stockholm) have in production at the moment:

Our prod:
image

This PR:
image


Same for dismiss options:
Our prod:
image

This PR:
image

So I think we should sit together (possibly after having reviewed and tested this PR) and decide what is standard and should be included as default terms before proceeding. What do you say @dnil?

northwestwitch:
Testing the PR. I'm not done with the review (haven't started to look at the code yet) but I have some comments that can be addressed already.

Testing the PR locally in this case scenario:

1. Pre-existent instance of Scout with rare disease case that has dismissed (with one or more reasons) and evaluated (for instance likely benign).
1. Switch to this branch and load the terms in the database:
There exist **2 commands to load terms in the database: `scout load batch-evaluation-term` and `scout load evaluation-term`.** I find a bit confusing that they have the same explanation:

image

-[ ] **Suggestion: Could you be a bit more specific with the help of these command so it's clearer what is the difference?**

1. The command to use would be `scout load batch-evaluation-term`. Loading the evaluation and dismissed terms using the provided resource():
```
scout --demo load batch-evaluation-term -f scout/resources/variant_manual_rank_terms.json
scout --demo load batch-evaluation-term -f scout/resources/variant_dismissal_terms.json
```
The commands work 🆗 

Looking at the variant in question: looks fine on variantS list: 

image


- [ ] But doesn't show any dismissed and manual evaluated tags on the variant page:

image

- [ ] Looks like there is something wrong with how the terms are collected for variant tags

image

- [ ] And dismiss options:
image

- [ ] Please check that these lists correspond to the terms that are currently present in master branch

dnil:
> * [ ]  Please check that these lists correspond to the terms that are currently present in master branch

We will likely have to live with at least two sets here for starters, since Lund have a different set. To me it would be very ok to include json for both current master and current Lund for an alternative. This would solve the more pressing issue of allowing us to work off one codebase properly. 

Getting a user group together to evaluate all terms - or send a list around for comments to all - would be very nice. But the idea is that this ought to be mergeable without a full consensus.

northwestwitch:
> We will likely have to live with at least two sets here for starters, since Lund have a different set. To me it would be very ok to include json for both current master and current Lund for an alternative.

Yes I agree, 2 lists under resources is perfect. But I think that the default one should correspond to the one we have on master

mhkc:
> We` will likely have to live with at least two sets here for starters, since Lund have a different set. To me it would be very ok to include json for both current master and current Lund for an alternative. This would solve the more pressing issue of allowing us to work off one codebase properly.

I have only included the terms you are using. They are divided into two different list depending, one for dismissal terms and one for manual rank terms. Both are loaded by default when one setup either a demo or normal database.

northwestwitch:
Testing and starting the review of this PR now. It's a fantastic refactoring work, but it touches so many files and functionalities that I am a bit scared of where breaking changes might occur. Note  to @mhkc for next time: try to break down the tasks as much as you can: It is is much easier to test and review smaller PRs. This one for instance could have been divided in sub-tasks like this

- Move variant tags to the database and establish system to load them all together from resources yaml file
- Move dismiss tags to the database using the same system as variant tags
- Add/remove one custom term to the variants evaluations using the cli
- Add the cli command to view the evaluation terms
- Institute-specific variant evaluation terms (page)


Checklist of things to test that come to my mind:

- [ ] All previous terms loaded into database.
- [ ] Previous dismissed/ranked variants still get the same evaluation
- [ ] Evaluations / ranking shows in general report
- [ ] Evaluation of old evaluated cases should show the same evaluations in general report
- [ ] Document the procedure in docs
- [ ] Add one custom evaluation term using the cli and see that it shows and works in the portal
- [ ] Try by adding one term that has the same "rank" and see what happens (it shouldn't be possible to have duplicates)
- [ ] Check institute settings page when loading a custom eval term

northwestwitch (changes requested):
Nice with the fixes, it's getting there! Once again, I haven't started the review of the code since I think there are some tested things that don't work as they should, yet.

I'm copying the text of a previous comment and post it as review. I've included a checklist of tests and some need a bit of attention.

It's a fantastic refactoring work, but it touches so many files and functionalities that I am a bit scared of where breaking changes might occur. Note  to @mhkc for next time: try to break down the tasks as much as you can: It is is much easier to test and review smaller PRs. This one for instance could have been divided in sub-tasks like this

- Move variant tags to the database and establish system to load them all together from resources yaml file
- Move dismiss tags to the database using the same system as variant tags
- Add/remove one custom term to the variants evaluations using the cli
- Add the cli command to view the evaluation terms
- Institute-specific variant evaluation terms


Checklist of things to test that come to my mind - pics from this branch on the left - :

- [x] All previous tag terms loaded into database when setting up a database with `scout setup demo/database`
- [ ] All variant tags and  are OK for RD variant (they are, but sorted they other way around)
image
- [x] All dismiss terms are OK for RD variant
image
- [ ] All variant tags are OK for cancer variant (they are, but sorted they other way around)
image
- [ ] All dismiss terms are OK for cancer variant. **NOPE, check pic!**
image

- [ ] Previous dismissed/ranked RD variants still get the same evaluation. **Tags are not shown on variant's page, see pic**
image

- [ ] Previous dismissed/ranked cancer variants still get the same evaluation. **No evaluation shown in this branch**
image

- [x] Correct evaluation of old evaluated cases should show the same evaluations in general report (rare)
- [ ] Correct evaluation of old evaluated cases should show the same evaluations in general report (cancer). **The page crashes when using cancer-specific dismiss tags**:
```
  File "/Users/chiararasi/Documents/work/GITs/scout/scout/server/blueprints/cases/templates/cases/case_report.html", line 779, in template
    <li>{{dismissed_options[reason|int]['description']}}
  File "/Users/chiararasi/miniconda3/envs/py38/lib/python3.8/site-packages/jinja2/environment.py", line 452, in getitem
    return obj[argument]
jinja2.exceptions.UndefinedError: list object has no element 46
```
- [x] Batch load evaluation terms: `scout --demo load batch-evaluation-term -f scout/resources/variant_manual_rank_terms.json`
- [x] Batch load dismiss terms: `scout --demo load batch-evaluation-term -f scout/resources/variant_dismissal_terms.json`
- [ ] Document the procedure in docs. **Documentation on the 2 commands above is missing. Document also how to add a single term**.
- [ ] Add one custom evaluation term using the cli and see that it shows and works in the portal. The cli command works fine, but the syntax is too relaxed. There is only one mandatory field (name). If you provided just name you end up with a useless term added to the database: 
image

    - Suggestion: make description and term-category also mandatory. 
    - Define strict term categories that can be assigned in the constants, and check that term has a valid category when created
    - What is label and what is it used for?
    - Which field is displayed on the variant page? Mention this in the help of the CLI command perhaps so user knows at the moment of term creation?
    - Description of institute option look incomplete:
    
image

    - Improve description of the `scout load batch-evaluation-term,` because it's identical to the `scout load evaluation-term` one
     
image

- [x] Try by adding one term that has the same "rank" and see what happens (it shouldn't be possible to have duplicates)
- [ ] Check institute settings page when loading a custom eval term. **The field looks a but weird on institute settings page:**

image

- [x] Check a variant with custom institute terms (should show the term)

northwestwitch:
Do you need help with this one @mhkc ? I could find time today and tomorrow to help fixing stuff

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -11,7 +11,7 @@
         <select name="manual_rank" class="form-control">
           <option value="-1">Select a tag...</option>
           {% for data in manual_rank_options %}
-            <option {% if data.internal_id == variant.manual_rank %} selected {% endif %} value="{{ data.internal_id }}">
+            <option {% if data.rank == variant.manual_rank %} selected {% endif %} value="{{ data.rank }}">
```
northwestwitch, line 14: The unrecognized old rankings and dismissed terms was due to this. The variant updated with new manual rank and dismissed terms looks like this on master:

image

mhkc:
>  All variant tags and are OK for RD variant (they are, but sorted they other way around)

I think this is acceptable. Currently `manual rank` and `dismissal term` have the same sort order (ascending rank) having different sort orders would be confusing. If we were to change the rank value of the manual rank terms then we would break backward compatibility without prior migration of the database.

scout/server/blueprints/variant/controllers.py
```diff
@@ -253,12 +251,21 @@ def variant(
...
-            **DISMISS_VARIANT_OPTIONS,
-            **CANCER_SPECIFIC_VARIANT_DISMISS_OPTIONS,
-        }
+    # get dismiss_variant_options
+    evalutation_terms = store.evaluation_terms(
+        term_category="dismissal_term",
+        analysis_type="cancer" if case_obj.get("track") == "cancer" else None,
+        institute_id=institute_obj["internal_id"],
+    )
+    dismiss_options = build_variant_evaluation_terms(evalutation_terms)
```
northwestwitch, line 260:
```suggestion
    dismiss_options = build_variant_evaluation_terms(evaluation_terms)
```

scout/server/blueprints/variant/controllers.py
```diff
@@ -253,12 +251,21 @@ def variant(
...
+
+    # get manual rank options
+    evalutation_terms = store.evaluation_terms(
```
northwestwitch, line 263:
```suggestion
    evaluation_terms = store.evaluation_terms(
```

scout/server/blueprints/variants/views.py
```diff
@@ -376,6 +410,15 @@ def cancer_variants(institute_id, case_name):
         form,
         page=page,
     )
+
+    # get dismiss_variant_options
+    evalutation_terms = store.evaluation_terms(
```
northwestwitch, line 415:
```suggestion
    evaluation_terms = store.evaluation_terms(
```

scout/server/blueprints/variants/views.py
```diff
@@ -376,6 +410,15 @@ def cancer_variants(institute_id, case_name):
...
+        term_category="dismissal_term",
+        analysis_type="cancer",
+        institute_id=institute_obj["internal_id"],
+    )
+    dismiss_variant_options = build_variant_evaluation_terms(evalutation_terms)
```
northwestwitch, line 420:
```suggestion
    dismiss_variant_options = build_variant_evaluation_terms(evaluation_terms)
```

scout/server/blueprints/variants/views.py
```diff
@@ -446,19 +486,31 @@ def cancer_sv_variants(institute_id, case_name):
         store, institute_obj, case_obj, variants_query, result_size, page
     )
     expand_search = request.method == "POST" and request.form.get("expand_search") in ["True", ""]
+
+    # get dismiss_variant_options
+    evalutation_terms = store.evaluation_terms(
```
northwestwitch, line 491:
```suggestion
    evaluation_terms = store.evaluation_terms(
```

scout/server/blueprints/variants/views.py
```diff
@@ -446,19 +486,31 @@ def cancer_sv_variants(institute_id, case_name):
...
+        term_category="dismissal_term", institute_id=institute_obj["internal_id"]
+    )
+    dismiss_variant_options = build_variant_evaluation_terms(evalutation_terms)
```
northwestwitch, line 494:
```suggestion
    dismiss_variant_options = build_variant_evaluation_terms(evaluation_terms)
```

scout/server/blueprints/variants/views.py
```diff
@@ -446,19 +486,31 @@ def cancer_sv_variants(institute_id, case_name):
...
+
+    # get manual_rank_options
+    evalutation_terms = store.evaluation_terms(
+        term_category="dismissal_term",
+        analysis_type="cancer",
+        institute_id=institute_obj["internal_id"],
+    )
+    manual_rank_options = build_variant_evaluation_terms(evalutation_terms)
```
northwestwitch, line 502:
```suggestion
    manual_rank_options = build_variant_evaluation_terms(evaluation_terms)
```

tests/adapter/mongo/test_evaluation_terms.py
```diff
@@ -0,0 +1,57 @@
...
+    ## override an exclusive param
+    generic_term = base_term.copy()
+    obj_id = adapter.add_evaluation_term(generic_term)
+
+    ## institute
+    exclusive_term_obj = base_term.copy()
+    exclusive_term_obj[param_name] = "foobar"
+    obj_id = adapter.add_evaluation_term(exclusive_term_obj)
+
+    ##THEN assert that the generic transcript is there
```
northwestwitch, line 51: Transcript?

northwestwitch (changes requested):
Nicely done! I think we are getting closer now!

I have a few comments on the code.

Also: the order in which the terms are displayed is not important so it's fine to have it reverted. But I have a concern regarding the format used to save rank and dismiss info in the variant document.

On master (and on millions of variants in our prod database with have the following format:

image


This branch is changing the format into:

image

I actually like your way because it's intuitive but think we should avoid introducing such radical changes (if possible). We will end up with a mixture of strings and numbers and it will be much more complicated to maintain. Is there a reason why you want to change the format?

northwestwitch:
@mhkc I've merged the conflicts but there is a test that is not passing. Could be due to the change that you mentioned (return error when the app loads if there are no evaluation terms in database). I can take a look at this later if you want!

mhkc:
@northwestwitch please do. I have been swamped with issues with the scout update

northwestwitch:
What about about splitting this one in smaller PRs instead? I'm having a hard time testing everything it contains and I am a bit scared to introduce all these changes at once. The branch at the moment crashes if there are previous evaluations but no terms loaded in the database.

I could start doing the splitting and then we could review and test one feature at the time. What do you think?

mhkc:
I dont really se how we can go smaller since one builds on top of another.

I guess we could add the changes to the institute config page later but that is a minor thing.

Also we kind of need this one at latest in late August due to the accreditation

northwestwitch:
> @northwestwitch I dont really se how we can go smaller since one builds on top of another. We need the core functionality of having the terms in the database. In order to make it usable we need both a set of default terms to be loaded during setup and the appropriate cli commands for working with the terms
> 
> I guess we could add the changes to the institute config page later but that is a minor thing.
> 
> Also we kind of need this one at latest in late August due to the accreditation

Have you checked my #2675? It's a basic version of this one that customizes also other types of terms. We could merge it easily but I need a reviewer

PR #2496, opened 2021-04-01
Feature configparse 210401

mikaell:
Redoing the work from home…

PR #2581, opened 2021-04-28
Test saved case/individual keys on case load -- Automatic tests --

northwestwitch:
When loading a case using `scout load case` command:

- Check that reference key values are saved in case document
- Check that reference key values are saved in case individuals

**OBS: keys used in these test are those saved when loading a demo case, and might not be comprehensive of all possible keys that are saved when loading a real case**. That is a much bigger task. This PR is just a start, meant to help finding and fixing missing keys when loading cases with pydantic (#2496).

These tests are different from case parsing tests because:
- tests here are checking that key/values exist after loading from yaml file (using the command line).
- parsing tests are testing the specific values parsed for a number of keys (far from all possible keys) receiving a config file as input, thus skipping the initial load case command (and the pydantic validation).

**How to test**:
1. Automatic tests

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by pytest


northwestwitch:
@mikaell in this PR I've fixed a couple of tests to fail if the expected keys are not found (both in case and case individuals) when loading cases with the command line. 
The idea was to help fixing the issues on your PR. You could be a reviewer of this if you have time!

mikaell:
> @mikaell in this PR I've fixed a couple of tests to fail if the expected keys are not found (both in case and case individuals) when loading cases with the command line.
> The idea was to help fixing the issues on your PR. You could be a reviewer of this one if you have time!

Hi,
I kinda get the idea for this test. But to me it seems to include a too wide range of internal functionality. Wouldn't it be better to test that the case-object created is OK as early on as possible. Before writing it the the database (because that should always be the same independent of case-object creation).

northwestwitch:
Yes, that would also work fine. Feel free to modify this PR or you could
include this type of tests in your pydantic one. I think the important is
to check that all the expected keys are present after the case parsing and
before (or in this case after) the case is saved to database

tests/commands/load/test_load_case_cmd.py
```diff
@@ -45,25 +45,104 @@ def test_load_case_from_ped(mock_app, institute_obj, case_obj):
...
+        # Some fields might not have values:
+        elif key in ["multiqc", "cnv_report", "coverage_qc_report", "gene_fusion_report_research"]:
+            key_value = key_value_template.format(key=key, bool_value=new_case[key] is None)
+
+        # Some other keys might be empty lists
+        elif key in ["dynamic_gene_list", "group"]:
+            key_value = key_value_template.format(key=key, bool_value=len(new_case[key]) == 0)
+
+        else:
+            key_value = key_value_template.format(key=key, bool_value=bool(new_case[key]))
```
mikaell, line 94: I think these checks are too tightly coupled with the example demo file.  For example if calling `bool()` on an empty string or zero will result in `False` breaking the test. This make the test implementation dependent on the config.

As I understand the purpose is to check that keys exists in the dict read from the DB. Would this be enough?


```
    # WHEN a case is loaded using a yaml config file
    result = runner.invoke(cli, ["load", "case", load_path])

    # THEN the command should be successful
    assert result.exit_code == 0

    # AND the case should be saved in database
    new_case = store.case_collection.find_one()
    assert new_case

    # WHEN case is loaded it should contain the expected fields:
    for key in demo_case_keys:
        assert key in new_case
```

tests/commands/conftest.py
```diff
@@ -21,6 +21,90 @@ def fixture_demo_files():
...
+        "analysis_type",
+        "tissue_type",
+        "is_sma",
+        "is_sma_carrier",
+        "smn1_cn",
+        "smn2_cn",
+        "smn2delta78_cn",
+        "smn_27134_cn",
+        "splice_junctions_bed",
+        "rna_coverage_bigwig",
```
mikaell, line 103: Alphabetical order would be nice.

tests/commands/conftest.py
```diff
@@ -21,6 +21,90 @@ def fixture_demo_files():
...
+        "coverage_qc_report",
+        "gene_fusion_report",
+        "gene_fusion_report_research",
+        "vcf_files",
+        "delivery_report",
+        "has_svvariants",
+        "has_strvariants",
+        "is_migrated",
+        "track",
+        "group",
```
mikaell, line 65: Alphabetical order would be nice.

mikaell (changes requested):
If the sole purpose is to check that keys exist I think the code could be more decoupled.

tests/commands/load/test_load_case_cmd.py
northwestwitch: > This make the test implementation dependent on the config.

That's the intended behavior. The demo file is the only config file that reflects all the possible loadable values from the start for a case. It's fairly complete in providing an example for the demo case. If any of these tests start to fail it would means we have disrupted a function that should be working instead. 

> As I understand the purpose is to check that keys exists in the dict read from the DB. Would this be enough?

The demo case has a number of keys that exist but are None and it's fine. For some other keys it's different: if they exist but they are none after loading the demo case it means that something has gone wrong during the case loading. For instance the pedigree

This is the key after you load the case using master branch:
image

This is the same key with None value when using the pydantic PR:
image

That's why I think it's important that we check that these keys not only exist, but have the expected values

tests/commands/conftest.py
northwestwitch: Good idea!

tests/commands/load/test_load_case_cmd.py
mikaell: Yes, I agree that we should test for mandatory params and that they are set if needed (there are indeed some bugs in the pydantic-PR). But I think it could be done even better.

* Fail early. ie. as soon as we notice parsing is broken. Not after data has been written and read from DB
* Independent of underlying representation 

For example small changes in the config file might cause the tests to fail. Try setting 
`rank_model_version: ''`

northwestwitch: That's fine to check them earlier. Feel free to close this PR and introduce these changes in your Pydantic PR if you want! 👍🏻

northwestwitch:
Closing this as it will not be used!

PR #2584, opened 2021-04-28
Tentative fix #2580

### LINKED ISSUE #2580: Link to other cases from loqus erratic
The sv loqus box reports the correct number of cases, but links to other cases are only made seemingly stochastically. See e.g. https://scout.scilifelab.se/cust002/F0037887-2/sv/variants/5046c0363f2daa27494a98ece7f99733. Here several of the other cases actually do belong to cust002 and should be shown, although a couple don't and shouldn't.

Appears the case list might break off when there is even a single case that belongs to another institute - a misplaced continue or comprehension if perhaps? I could have a look tomorrow earliest..

northwestwitch:
- Skip calling parse_document_id to create a variant["variant_id] to look for the same variant in other cases, since the other variants share the same document_id with the variant in question.
- A more precise Loqus case counter, that when counting total cases takes into account only cases having that specific type of variant (snv/sv).

A bit complicated to test on stage, since it's running queries to loqus prod, but at least the stats for SNVs seem to work!

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -330,7 +330,7 @@ def observations(store, loqusdb, case_obj, variant_obj):
         LOG.debug("Could not find any observations for %s", composite_id)
         obs_data["total"] = loqusdb.case_count(
             variant_category=variant_obj["category"]
-        )  # count only case having the specifuc type of variant (snv/sv)
+        )  # count only cases having the specific type of variant (snv/sv)
         return obs_data
 
     user_institutes_ids = set([inst["_id"] for inst in user_institutes(store, current_user)])
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -352,7 +352,7 @@ def observations(store, loqusdb, case_obj, variant_obj):
             # If the user does not have access to the information we skip it
             continue
 
-        other_variant = store.variant(document_id=variant_obj[""])
+        other_variant = store.variant(document_id=variant_obj["variant_id"])
         # If the other variant is not loaded we skip it
         if not other_variant:
             continue
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -352,7 +352,9 @@ def observations(store, loqusdb, case_obj, variant_obj):
             # If the user does not have access to the information we skip it
             continue
 
-        other_variant = store.variant(document_id=variant_obj["variant_id"])
+        other_variant = store.variant(
+            case_id=other_case["_id"], document_id=variant_obj["variant_id"]
+        )
         # If the other variant is not loaded we skip it
         if not other_variant:
             continue
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -328,7 +328,9 @@ def observations(store, loqusdb, case_obj, variant_obj):
     obs_data = loqusdb.get_variant(variant_query, loqusdb_id=institute_obj.get("loqusdb_id")) or {}
     if not obs_data:
         LOG.debug("Could not find any observations for %s", composite_id)
-        obs_data["total"] = loqusdb.case_count()
+        obs_data["total"] = loqusdb.case_count(
+            variant_category=variant_obj["category"]
+        )  # count only case having the specifuc type of variant (snv/sv)
         return obs_data
 
     user_institutes_ids = set([inst["_id"] for inst in user_institutes(store, current_user)])
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -1,7 +1,7 @@
 import logging
 from datetime import date
 
-from flask import url_for
+from flask import flash, url_for
 from flask_login import current_user
 
 from scout.constants import (
@@ -335,6 +335,7 @@ def observations(store, loqusdb, case_obj, variant_obj):
     user_institutes_ids = set([inst["_id"] for inst in user_institutes(store, current_user)])
 
     obs_data["cases"] = []
+    flash(obs_data)
     for i, case_id in enumerate(obs_data.get("families", [])):
         if i > 10:
             break
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -1,7 +1,7 @@
 import logging
 from datetime import date
 
-from flask import flash, url_for
+from flask import url_for
 from flask_login import current_user
 
 from scout.constants import (
@@ -19,7 +19,6 @@ from scout.constants import (
     MOSAICISM_OPTIONS,
     VERBS_MAP,
 )
-from scout.parse.variant.ids import parse_document_id
 from scout.server.extensions import cloud_tracks, gens
 from scout.server.links import ensembl, get_variant_links
 from scout.server.utils import (
@@ -335,7 +334,6 @@ def observations(store, loqusdb, case_obj, variant_obj):
     user_institutes_ids = set([inst["_id"] for inst in user_institutes(store, current_user)])
 
     obs_data["cases"] = []
-    flash(obs_data)
     for i, case_id in enumerate(obs_data.get("families", [])):
         if i > 10:
             break
@@ -353,8 +351,8 @@ def observations(store, loqusdb, case_obj, variant_obj):
         if user_institutes_ids.isdisjoint(other_institutes):
             # If the user does not have access to the information we skip it
             continue
-        document_id = parse_document_id(chrom, str(pos), ref, alt, var_type, case_id)
-        other_variant = store.variant(document_id=document_id)
+
+        other_variant = store.variant(document_id=variant_obj[""])
         # If the other variant is not loaded we skip it
         if not other_variant:
             continue
--- a/scout/server/extensions/loqus_extension.py
+++ b/scout/server/extensions/loqus_extension.py
@@ -236,7 +236,7 @@ class LoqusDB:
         except AttributeError:
             raise ConfigError("LoqusDB id not found: {}".format(loqusdb_id))
 
-    def case_count(self):
+    def case_count(self, variant_category="snv"):
         """Returns number of cases in loqus instance
 
         Returns:
@@ -244,7 +244,7 @@ class LoqusDB:
         """
         nr_cases = 0
         case_call = self.get_command()
-        case_call.extend(["cases", "--count"])
+        case_call.extend(["cases", "--count", "-t", variant_category])
         output = ""
         try:
             output = execute_command(case_call)


northwestwitch:
I'll try to write an automatic test to check that this works

scout/server/extensions/loqus_extension.py
```diff
@@ -236,15 +236,15 @@ def get_configured_version(self, loqusdb_id=None):
...
+    def case_count(self, variant_category="snv"):
         """Returns number of cases in loqus instance
 
         Returns:
             nr_cases(int)
         """
         nr_cases = 0
         case_call = self.get_command()
-        case_call.extend(["cases", "--count"])
+        case_call.extend(["cases", "--count", "-t", variant_category])
```
northwestwitch, line 247: https://github.com/moonso/loqusdb/blob/a6b7ee426357dd493eb31777ded4d6246753c484/loqusdb/commands/view.py#L22

dnil: 👍

scout/server/blueprints/variant/controllers.py
```diff
@@ -352,8 +353,10 @@ def observations(store, loqusdb, case_obj, variant_obj):
         if user_institutes_ids.isdisjoint(other_institutes):
             # If the user does not have access to the information we skip it
             continue
-        document_id = parse_document_id(chrom, str(pos), ref, alt, var_type, case_id)
-        other_variant = store.variant(document_id=document_id)
+
+        other_variant = store.variant(
```
dnil, line 357: Agreed, that is convenient, but still not going to work for all returned SVs, right? loqusdb is destructive in its encoding, so similar, but not identical SVs from different cases will be mapped to the same loqusdb entry, and can not be easily recovered. An approach can be to return links to the exact SV variant iff it exists with the same id in the other case(s), but return the case name and link to the case page if the variant could not be easily identified, like just a jinja template fix. Let me know if you want me to try to draft something for you to check instead!

dnil (changes requested):

scout/server/blueprints/variant/controllers.py
northwestwitch: I had the feeling the matter wasn't so easy to fix (therefore the title of the PR!). Checked now the [code on loqus](https://github.com/moonso/loqusdb/blob/a6b7ee426357dd493eb31777ded4d6246753c484/loqusdb/commands/view.py#L120) and as you say, all SVs matching those coordinates are returned, regardless of the SV type. I'll try to fix accordingly!

northwestwitch:
@dnil this is also ready for a new review

scout/adapter/mongo/variant.py
```diff
@@ -292,6 +292,26 @@ def variant(
...
+            category (str): "sv" or "snv"
+            chrom (str): variant chromosome
+            chrom_end (str): end chromosome
+            start (int): start position
+            end (int): end position
+
+        Returns:
+            variant_obj (Variant): a variant object dictionary
+        """
+        query = dict(case_id=case_id, category=category, chromosome=chrom, position=start, end=end)
```
dnil, line 309: You'll need variant type for this as well to get a boost from the index `caseid_category_chromosome_start_end`!

scout/server/blueprints/variant/controllers.py
```diff
@@ -306,7 +305,9 @@ def observations(store, loqusdb, case_obj, variant_obj):
         obs_data(dict)
     """
     chrom = variant_obj["chromosome"]
+    end_chrom = variant_obj.get("end_chrom", chrom)
```
dnil, line 308: 👍

scout/server/blueprints/variant/controllers.py
```diff
@@ -365,10 +365,19 @@ def observations(store, loqusdb, case_obj, variant_obj):
...
             # If the user does not have access to the information we skip it
             continue
-        document_id = parse_document_id(chrom, str(pos), ref, alt, var_type, case_id)
-        other_variant = store.variant(document_id=document_id)
+
+        other_variant = store.variant(
+            case_id=other_case["_id"], document_id=variant_obj["variant_id"]
+        )
+
+        # IF variant is SV variant, look for variants with different sub_category occurring at the same coordinates
```
dnil, line 373: Hm, why so? Are you thinking like BND vs DEL for different callers?

dnil: This could happen, though e.g. DEL and DUP at the same pos are definately distinct (and would not normally occur in the same case or in a case with a match, so probably safe). More importantly, the start and end coordinates do not have to match exactly, but just be in a vicinity, with the exact interval depending on the size of the variant (rules in loqusdb). In these cases it would be good to return a link to the matched case, even if the variant was not found.

scout/server/blueprints/variant/controllers.py
```diff
@@ -365,10 +365,19 @@ def observations(store, loqusdb, case_obj, variant_obj):
...
+        if other_variant is None and category == "sv":
+            other_variant = store.case_variant_by_coordinates(
+                other_case["_id"], category, chrom, end_chrom, pos, end
+            )
+
         # If the other variant is not loaded we skip it
-        if not other_variant:
+        if other_variant is None:
```
dnil, line 380: Like here: we should append other case and a set variant None. There is even code for that in the template already:
https://github.com/Clinical-Genomics/scout/blob/df3020ef7a119f7b1f1d76f86c78c73c518c9aa8/scout/server/blueprints/variant/templates/variant/variant_details.html#L183

My current feeling is that the whole issue would be fixed by deleting L380-381 here, but is is probably also good to keep the search-for-different-subtypes part to catch a few more variants exactly?

dnil (changes requested):

scout/server/blueprints/variant/controllers.py
northwestwitch: No you are right, looking at the loqus code, the query if performed only on variants with the same sub-category. I'll fix! I've checked the rules in loqus (as you suggested) and I see how the query is runned now (not exact match, whoch makes a lot of sense for SVs). I might end up with a slightly different solution but I'm aware of the thing now!

scout/adapter/mongo/variant.py
```diff
@@ -292,6 +292,52 @@ def variant(
...
+                query,
+            )
+        )
+        if not overlapping_svs:
+            return None
+        if len(overlapping_svs) == 1:
+            return overlapping_svs[0]
+
+        # If more than one SV is overlapping with this variant
+        # return the one with most similar size
```
northwestwitch, line 331: It ended up very different from the [loqus algorithm](https://github.com/moonso/loqusdb/blob/a6b7ee426357dd493eb31777ded4d6246753c484/loqusdb/plugins/mongo/structural_variant.py#L232). If you don't like it @dnil I can use the code in LoqusDB instead. The thing is, it's not really an exact match so any criterion is good as long as one like it..

scout/server/blueprints/variant/controllers.py
dnil: Agree: it makes sense to run a more complex query, but chances are that will also come up negative, so I still think leaving it with case=other_case, variant=None is much better than just hiding the original loqus hit and implicitly pretending the variant was from another institute..

scout/adapter/mongo/variant.py
dnil: 😄 Indeed, you will find quite a handful of others over time in SVDB, as well as in the loqusdb project. I think the salient point is having some agreement between # found and the number reported.

This works for me. Anything with a lambda in it is already very high up on my liking, and reusing the existing scout overlap definition is very nice. To get the exact same result as loqusdb with the loqusdb code you also need to be using the data from its clustering. It will also create a slightly uncomfortable dependency. 
The results should be very similar when it comes to del / dup recall, but will have some issues with BNDs on different chr, potentially also with inversions.

dnil (approved):
We should test this some more before merging, but my most important concern is just that we keep an opening for forwarding the counted matches number of cases from loqus to the user, as long as the user is allowed to view those cases. Going to the variant directly is important, but less so than being able to ascertain the phenotype.

scout/server/blueprints/variant/controllers.py
northwestwitch: I agree, better than nothing!

northwestwitch:
I'm going to install this branch on stage to give it a try there now

northwestwitch:
Testing with one SNV: https://scout-stage.scilifelab.se/cust000/643594-300M/88f6e9015fe13d1255fbe28e198c4c97

Master:
image


This branch: 🆗 

image


--------

Testing with one SV: https://scout-stage.scilifelab.se/cust000/643594-300M/sv/variants/2923e3e133fdc0ced728cbca91c22bd0

Master branch:
image


This branch: 
image

dnil:
Nice! Much better, though one gets curious about what the other three cases were? Simply not loaded into stage, or something else? We should also confirm the other way around that when logging in with a regular user, only collaborators are shown.

dnil:
Oh, ok, sorry didn't read the last comment. It does need to honor the variant category.

dnil:
🤔 Now that is worth some consideration. Absolute transitivity is not expected, but some level at least. Since the obs count is 0 I dont think that new issue is primarily with Scout? This looks like a case for checking first what the variant looks like in loqus, that the command line tool returns the same, and then passing on an issue or checking the loqus overlap code directly? Perhaps there is a mistake in the "contained in" case?

northwestwitch:
Yes the total number of cases = 0 is strange.. I'm very confused.

northwestwitch:
The variant should be found by loqus, if Scout sends the right info for building the query:

16052 is superbbison:

image

northwestwitch:
While being on the page of the [problematic variant](https://scout-stage.scilifelab.se/cust003/16052/sv/variants/4eeed3ef06cc25cf3da2e3e79f06d01a)

The query is sent like this:
`2021-05-10 15:06:52 clinical-db.scilifelab.se scout.server.extensions.loqus_extension[19773] ERROR ['/home/hiseq.clinical/miniconda/envs/S_loqusdb/bin/loqusdb', '--config', '/home/hiseq.clinical/servers/config/clinical-db.scilifelab.se/loqusdb.yaml', 'variants', '--to-json', '--variant-id', '11_14870001_T_<DUP>', '--case-count', '-t', 'sv', '-c', '11', '-s', '14870001', '-e', '14905000', '--end-chromosome', '11', '--sv-type', 'DUP']`

And returns the correct info:
```
2021-05-10 15:06:54 clinical-db.scilifelab.se scout.server.extensions.loqus_extension[19773] ERROR {'chrom': '11', 'end_chrom': '11', 'sv_type': 'DUP', 'pos_sum': 133834403, 'end_sum': 134144836, 'observations': 9, 'length': 34492, 'families': ['fondkoi', 'happyclam', 'neededant', 'idealtetra', 'stableorca', 'simpleearwig', 'mutualtrout', 'sacredcaiman', 'superbbison'], 'pos_left': 14868489, 'pos_right': 14872489, 'end_left': 14902981, 'end_right': 14906981, 'total': 4432}
2021-05-10 15:06:59 clinical-db.scilifelab.se scout.server.extensions.loqus_extension[19773] ERROR Could not find a Loqus instance with id:
2021-05-10 15:06:59 clinical-db.scilifelab.se scout.server.extensions.loqus_extension[19773] ERROR Could not find a Loqus instance with id:
```

But check at the last 2 lines. There is an error somewhere in Scout!

dnil:
Did you do the ordinary-user test as well to be sure we don't suddenly give away too much?

northwestwitch:
> Did you do the ordinary-user test as well to be sure we don't suddenly give away too much?

No I've used my admin user. I can check!

PR #2587, opened 2021-04-30
Remove parsing of case `genome_version`, since it's not used anywhere downstream

### LINKED ISSUE #2578: Case 'genome_version' key, are we using it?
Found here in the code:
```
/tests/conftest.py:    case["genome_version"] = None
./scout/models/case/case.py:    genome_version=float,  # What version of the build
./scout/adapter/mongo/case.py:            - genome_version: - || -
./scout/adapter/mongo/case.py:                    "genome_version": case_obj.get("genome_version"),
./scout/build/case.py:        genome_version = float, # What version of the build
./scout/build/case.py:    case_obj["genome_version"] = case_data.get("genome_version")
```
 
It should be a floating poing number defining the specific version of the genome build, but I don't think we ever used it. 
Since we have enough problems already to handle the two official genome builds, wouldn't be better to just remove it?

northwestwitch:
`genome_version` is a field that gets populated every time a case is created from parsing a yaml file. Our prod cases have a None value for this key and the key/value is not used anywhere downstream in the app anyway. Fix #2578

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Add link to HmtVar for mitochondrial variants (if VCF is annotated with HmtNote)
 ### Changed
+- Remove parsing of case `genome_version`, since it's not used anywhere downstream
 
 ## [4.33.1]
 ### Fixed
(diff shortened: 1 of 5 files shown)


dnil (approved):
Genome build (sub)version does obviously exist, but we don't really use. 37 has reached its terminal patch version since many years. To be fair 38 also has not been patched since 2019 (p13). Since it's not actually in use, and could be confusing, do go ahead: we can very easily reintroduce it if we ever need to use it!

PR #2590, opened 2021-05-04
 Redirect to last visited variantS page when dismissing variants from variants list

### LINKED ISSUE #2588: Dismiss multiple variants from variantS pages reloads from page 1
When dismissing variants from a later page, it is annoying to have to paginate back again..

northwestwitch:
Fix #2588

**How to test**:
1. Go to variantS page 2 and dismiss some variants. When page redirects you should still be on variantS page 2

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -31,7 +31,7 @@
 </div>
 {% endmacro %}
 
-{% macro dismiss_variants_block(dismiss_variant_options, institute, case) %}
+{% macro dismiss_variants_block(dismiss_variant_options, institute, case, page) %}
 <div class="row d-flex align-items-center mb-2 ml-1">
   <div class="col-2 ">Dismiss selected variants:</div>
   <div class="col-4 align-items-start">
@@ -735,6 +735,7 @@
     <input type="submit" name="page" id="paginate-first" value=1 hidden="true">
     <input type="submit" name="page" id="paginate-previous" value={{page-1}} hidden="true">
     <input type="submit" name="page" id="paginate-next" value={{page+1}} hidden="true">
+    <input type="hidden" name="page" value="{{page}}"> <!-- In order to work, this input should be placed AFTER the submit form items with the same name -->
   </div>
 {% endmacro %}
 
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -735,6 +735,7 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
     <input type="submit" name="page" id="paginate-first" value=1 hidden="true">
     <input type="submit" name="page" id="paginate-previous" value={{page-1}} hidden="true">
     <input type="submit" name="page" id="paginate-next" value={{page+1}} hidden="true">
+    <input type="hidden" name="page" value="{{page}}"> <!-- In order to work, this input should be placed AFTER the submit form items with the same name -->
```
dnil, line 738: 😮 I didn't know that! Nice solution! Works fine in Safari; is it the same for all browsers?

dnil (approved):

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: I think so. The explanations can be found in the HTML rules: https://html.spec.whatwg.org/multipage/forms.html#configuring-a-form-to-communicate-with-a-server

It says: `If there are multiple matching items, then a RadioNodeList object containing all those elements is returned.` which means a list. And the list is populated according to the occurrence of the items. 

When no submit button is pressed the the list has one element (the hidden field). If you press a button it will have 2 elements (the button and the hidden div). 

If you want to see both elements of this list you should do something like form.page.getList, but since in the controller we use only form.get("page") it returns only the first element. 

Perhaps not the clearest way of make it work, but looks like it does!

northwestwitch: Are still satisfied with the solution after knowing how it works ? 😆

dnil: Certainly; we have only had a little bit of complaints about this so far, and if this works for at least most browsers it's all good.

PR #2593, opened 2021-05-05
Update README.md

dnil:
The month average does not seem to be ready yet. Lets start with weekly?

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [![GitHub issues-closed][closed-issues-img]][closed-issues-url]
 [![Average time to resolve an issue][ismaintained-resolve-img]][ismaintained-resolve-url]
 [![Percentage of issues still open][ismaintained-open-rate-img]][ismaintained-open-rate-url]
-[![GitHub commit rate](https://img.shields.io//github/commit-activity/m/Clinical-Genomics/scout)](https://GitHub.com/Clinical-Genomics/scout/pulse/)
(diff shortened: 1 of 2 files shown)


northwestwitch:
The proposed link to the badge says 21 week, what's that?

image

README.md
```diff
@@ -11,7 +11,7 @@
 [![GitHub issues-closed][closed-issues-img]][closed-issues-url]
 [![Average time to resolve an issue][ismaintained-resolve-img]][ismaintained-resolve-url]
 [![Percentage of issues still open][ismaintained-open-rate-img]][ismaintained-open-rate-url]
-[![GitHub commit rate](https://img.shields.io//github/commit-activity/m/Clinical-Genomics/scout)](https://GitHub.com/Clinical-Genomics/scout/pulse/)
```
northwestwitch, line 14: It's not found because there are two // after img.shields.io. But It makes sense to use a weekly button since the pulse default activity is the weekly one

northwestwitch (approved):

northwestwitch:
> The proposed link to the badge says 21 week, what's that?
> 
> image

Never mind, I understood 😄

README.md
dnil: Mm, that is normally not an issue with URLs: could be for the cached image though! But actually also not generated if you check the link..

dnil: Ah, ok, now the m also works. Maybe its just some caching issues. Anyway, w is good - lets see if the cache doesn't update soon.

PR #2594, opened 2021-05-05
Filter cancer variants by "is in ClinVar" and "is in Cosmic"

### LINKED ISSUE #2591: "in COSMIC and in Clinvar" filter in variant list view.
Emma T and us were discussing filtering variants based on if they are present in COSMIC and/or ClinVar. This is to provide an option to look into variants that have hits in ClinVar and COSMIC before doing further investigation.

PS: I don't have Emma's GitHub account and can't tag her here for further input.

northwestwitch:
Fix #2591

Introduced a couple of checkboxes to filter cancer variants and get those in ClinVar and Cosmic:

image

**How to test**:
1. Switch to this branch locally and load demo cancer case
2. Filter using the new checkboxes

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -161,7 +161,7 @@ class QueryHandler(object):
 
         return mongo_variant_query
 
-    def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
+    def build_query(self, case_id, query=None, variant_ids=None, category="snv") -> dict:
         """Build a mongo query
 
         These are the different query options:
@@ -189,6 +189,8 @@ class QueryHandler(object):
                 'size_shorter': boolean,
                 'gene_panels': list(str),
                 'mvl_tag": boolean,
+                'clinvar_tag': boolean,
+                'cosmic_tag' boolean,
                 'decipher": boolean,
                 'hide_dismissed': boolean
             }
@@ -268,7 +270,7 @@ class QueryHandler(object):
         # there is only 'clinsig' criterion among the primary terms right now
         primary_terms = False
 
-        # gnomad_frequency, local_obs, clingen_ngi, swegen, spidex_human, cadd_score, genetic_models, mvl_tag
+        # gnomad_frequency, local_obs, clingen_ngi, swegen, spidex_human, cadd_score, genetic_models, mvl_tag, clinvar_tag, cosmic_tag
         # functional_annotations, region_annotations, size, svtype, decipher, depth, alt_count, control_frequency, tumor_frequency
         secondary_terms = False
 
@@ -514,7 +516,6 @@ class QueryHandler(object):
                             ]
                         }
                     )
-                LOG.debug("Adding gnomad_frequency to query")
 
             if criterion == "local_obs":
                 local_obs = query.get("local_obs")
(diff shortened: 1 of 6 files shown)


scout/server/blueprints/variants/forms.py
```diff
@@ -145,7 +145,9 @@ class CancerFiltersForm(VariantFiltersForm):
     tumor_frequency = BetterDecimalField(
         "Tumor alt AF >", places=2, validators=[validators.Optional()]
     )
-    mvl_tag = BooleanField("In Managed Variant List")
+    clinvar_tag = BooleanField("ClinVar hits")
+    cosmic_tag = BooleanField("Cosmic hits")
+    mvl_tag = BooleanField("Managed Variants hits")
```
northwestwitch, line 150: I don't know about the naming for these 3, perhaps this was better as it was

dnil: Potato potato!

scout/adapter/mongo/query.py
```diff
@@ -161,7 +161,7 @@ def build_variant_query(
 
         return mongo_variant_query
 
-    def build_query(self, case_id, query=None, variant_ids=None, category="snv"):
+    def build_query(self, case_id, query=None, variant_ids=None, category="snv") -> dict:
```
dnil, line 164: ☮️

scout/adapter/mongo/query.py
```diff
@@ -623,41 +619,41 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
...
 
             if criterion == "tumor_frequency":
-                LOG.debug("add minimum VAF filter")
                 mongo_secondary_query.append(
                     {"tumor.alt_freq": {"$gt": float(query.get("tumor_frequency"))}}
                 )
 
             if criterion == "mvl_tag":
-                LOG.debug("add managed variant list filter")
                 mongo_secondary_query.append({"mvl_tag": {"$exists": True}})
```
dnil, line 649: Do you think this might also need a `$ne: None` soon, or is that just being extra careful for cosmic etc?

dnil (approved):
🚒 💨 🚑 💨

dnil:
Works perfectly locally. Maybe roll it on stage as well, but I don't see any reason for a problem there either.

scout/adapter/mongo/query.py
northwestwitch: I had to add it for cosmic otherwise the query didn't work: many vars had cosmic_ids:none. I'm not sure how mvl_tag looks like

dnil: It should be fine, unless there is a hit. Not sure if one would be able to get to an existing but None key with adding and removing MVs. Or by a slightly mislead pydantic application..

northwestwitch: I think it's fine: mvl_tag exists only if there is a hit, for instance the demo cases have no variants with mvl_tag.

dnil: It will be evident then, dont worry.

dnil: Just my stream of consciousness speaking - and checking if I had missed something.

northwestwitch:
Checked on stage with this case: https://scout-stage.scilifelab.se/cust000/balsamic-validation-gmsmyeloid-tumor-only-2/cancer/variants and works fine. 

The thing is: if you use the clinvar and the cosmic checkbox together you return only the variants that are in both, not the variants that are in either or.. Let's hope this is fine for the users!

Issue #2598, opened 2021-05-09
Make Scout inclusive 


northwestwitch:
Suggested actions:

- Adapt colors for users with visual impairment
- Check terminology and remove non-inclusive words (perhaps via a github action)
- Rename "master" to "main"
Other??

mhkc:
I dont know if you do that at the moment but I think it would be a good idea to verify if the GUI behaves nice at different zoom levels or if we run into overflow and alignment issues.

Do you know if there is a need to support screen readers or similar aids for people with visual disabilities?

dnil:
I haven't heard, but if this would be a site accessible to the public, we would definitely be required to support screen readers anyway. As it is now, I suppose it exists in a bit of limbo. I don't think e.g. our LIMS does. But as soon as a new coworker arrives - or develops vision problems - that could change. It is really good practice to have good text alts for any image, icon etc anyway.

As for sizes, we have discussed it a little, but not reached a conclusion. It would be very useful to have at least a min required size established, especially for like plugins with graphics.

mhkc:
> As for sizes, we have discussed it a little, but not reached a conclusion. It would be very useful to have at least a min required size established, especially for like plugins with graphics.

I agree. Also I do believe this would have the highest impact on the current user base as peoples eyesight tends to worsen with age.

PR #2600, opened 2021-05-10
Use woke to check for non-inclusive language in repo

northwestwitch:
Partial fix for #2598

### This PR adds a github action that is using  [woke](https://github.com/get-woke/woke) for checking for non-sensitive language:
- Action that checks the whole repo (action doesn't fail with errors but generates warnings)

**How to prepare for test**:
- Tests are executed via the github actions

### Expected outcome:
- [x] Get warnings if non-sensitive expressions are found in this repo

**Review:**
- [ ] code approved by
- [x] tests executed by github


dnil:
👀 I expected it would find something at least? Aside from the horror of "Merge branch 'master' into woke_action".. 😸

dnil:
Hm, it still missed mongo..

northwestwitch:
> Hm, it still missed mongo..

Perhaps it's not a universally-blamed word. In Italy for instance it is slightly different 🤔

northwestwitch:
Aha! Found the one that blocks everything!

northwestwitch:
Better now:

image

northwestwitch:
I think I'm satisfied enough with this for now. It checks but doesn't fail if it finds expressions it doesn't like.

dnil (approved):
The report on the action flagged green is not very easy to find, but if the "check annotations" comments on the code stays it should be very obvious anyway. Let's go with that!

northwestwitch:
> The report on the action flagged green is not very easy to find, but if the "check annotations" comments on the code stays it should be very obvious anyway. Let's go with that!

I know, but the option was to fail (which we don't want). I've just merged the same action in the MatchMaker repo, but I've set that one to fail instead (now it doesn't find anything). We could do the same in Scout after fixing all the the issues and adding the words we would like to skip in a ahem "allowlist" in the scout settings.

PR #2613, opened 2021-05-12
Fix causatives view

dnil:

To apply to an existing installation, one will also need to
```
scout index --update
```

I haven't checked yet, but something reminds me that perhaps it is best just to add the same indices manually to prod when the time comes. We should perhaps also take a quick peek at current indices before running an update on stage. These ones should be fine, but there could be some other pending change that will slow things down more for a bit. Over the long weekend would actually not be at all bad for that.

**How to test**:
1. test user events view
2. test causatives view
3. test a case_verb intense view, like hovering over compounds to shade for dismissed ones

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 
 BeautifulSoup4
 werkzeug
-Flask>=0.10
+Flask>=0.10,<2
 Flask-DebugToolbar
 Flask-Bootstrap
 path.py
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 
 BeautifulSoup4
 werkzeug
-Flask>=0.10,<2
+Flask>=0.10
 Flask-DebugToolbar
 Flask-Bootstrap
 path.py
--- a/scout/constants/indexes.py
+++ b/scout/constants/indexes.py
@@ -89,6 +89,16 @@ INDEXES = {
     "event": [
         IndexModel([("category", ASCENDING), ("verb", ASCENDING)], name="category_verb"),
         IndexModel([("variant_id", ASCENDING)], name="variant_id"),
+        IndexModel(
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@
 
 BeautifulSoup4
 werkzeug
-Flask>=0.10
+Flask>=0.10,<2
```
northwestwitch, line 5: Come on, give me some credit, I can fix that! 🤩

dnil: Hehe, I did't see your solution until now. I'll revert and look at that first. Also, this didn't go all the way anyway. 😆

northwestwitch (approved):
Looks OK but we should test it on stage before merging I think!

scout/constants/indexes.py
```diff
@@ -84,7 +84,7 @@
     "hpo_term": [
         IndexModel([("description", ASCENDING)], name="description"),
         IndexModel([("description", TEXT)], default_language="english", name="description_text"),
-        IndexModel([("hpo_number", ASCENDING)], name="number"),
+        IndexModel([("hpo_number", ASCENDING)], name="number", background=True),
```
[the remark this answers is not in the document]
dnil: > What the background?

Runs non blocking and populates the index over time. It makes more sense for the bigger ones, but I just kept it.

dnil:
I'll do a manual test on stage since a full branch install would upgrade to Flask2 without freezing.

dnil:
Tested the indices on stage. They built quickly there. Obviously there is more of everything on prod, but should not be a huge issue methinks. It is obviously a little hard to test function currently, but at least they immediately take usage.

PR #2616, opened 2021-05-14
Fix #2614 - always use hgnc_id for gene filter

### LINKED ISSUE #2614: MTRFR not coming up in IEM gene panel
The gene MTRFR is not coming up the in the clinical list of variants in our IEM panel in patient 19253-I-2A.  When I search specifically for variants in this gene I get up variants that have a high rank score (27) and everything looks normal on the variant page etc.  This gene has recently changed name from C12orf65 to MTRFR but the HGNC number has not changed.  Why are these variants not coming up when a panel is selected?

dnil:
This PR fixes a bug.

Gene filter since v4.21 used gene panel hgnc symbols for filtering, instead of the panels assigned on load. This harmonises well with gene symbol queries and arbitrary lists. However, if a gene symbol changes, even if hgnc_id stays the same, queries will miss the gene. Manually added search symbols go through validation checks. To avoid similar issues in the future, always gene filter with hgnc_id.

**How to test**:
1. Filter demo case with the test1 panel. Note a gene.
2. Re-symbol the gene in a panel, e.g. 
Screenshot 2021-05-14 at 18 02 23
3. Before checking out this branch, filter with the gene of choice and note that you no longer find the gene with the panel, but can find it with manual lookup of its actual hgnc symbol (e.g. "POT1")
4. Apply patch and note that the gene list filter is working again
5. Ideally retest with different scenarios: combine with manual hgnc symbols, use different case builds, hpo panels, possibly different variant categories...
 
**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -161,7 +161,9 @@ class QueryHandler(object):
 
         return mongo_variant_query
 
-    def build_query(self, case_id, query=None, variant_ids=None, category="snv") -> dict:
+    def build_query(
+        self, case_id, query=None, variant_ids=None, category="snv", build="37"
+    ) -> dict:
         """Build a mongo query
 
         These are the different query options:
@@ -240,9 +242,9 @@ class QueryHandler(object):
             # the rest of the query content is clear.
 
             if criterion in ["hgnc_symbols", "gene_panels"]:
-                gene_query = self.gene_filter(query, mongo_query)
+                gene_query = self.gene_filter(query, build=build)
                 if len(gene_query) > 0 or "hpo" in query.get("gene_panels", []):
-                    mongo_query["hgnc_symbols"] = {"$in": gene_query}
+                    mongo_query["hgnc_ids"] = {"$in": gene_query}
                 continue
 
             if criterion == "chrom" and query.get("chrom"):  # filter by coordinates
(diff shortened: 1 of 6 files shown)


scout/adapter/mongo/query.py
```diff
@@ -240,9 +242,9 @@ def build_query(self, case_id, query=None, variant_ids=None, category="snv") ->
             # the rest of the query content is clear.
 
             if criterion in ["hgnc_symbols", "gene_panels"]:
-                gene_query = self.gene_filter(query, mongo_query)
+                gene_query = self.gene_filter(query, build=build)
                 if len(gene_query) > 0 or "hpo" in query.get("gene_panels", []):
-                    mongo_query["hgnc_symbols"] = {"$in": gene_query}
+                    mongo_query["hgnc_ids"] = {"$in": gene_query}
```
northwestwitch, line 247: Nice

northwestwitch (approved):
Tested as suggested by the PR comment, Works and it's very nicely handled!

Master, where the gene disappears when filtering:

image


This branch: 🆗 
image

I'm just a bit concerned about the search by genome build, not but the code itself but for the weird situation we already have in the database. It would be great to force build into "37" or "38" after checking that case build exists and contains "37" or "38", see my comment in the code!

tests/adapter/mongo/test_query.py
```diff
@@ -61,16 +61,23 @@ def test_gene_symbols_query(adapter, case_obj):
     """Test variants query using HGNC symbol"""
 
     # WHEN hgnc_symbols params is provided to the query builder
-    test_gene = "POT1"
-    query = {"hgnc_symbols": [test_gene], "gene_panels": []}
+    gene_obj = {
+        "hgnc_id": 17284,
+        "hgnc_symbol": "POT1",
+        "build": "37",
+        "aliases": ["POTTA", "POT1"],
```
northwestwitch, line 68: Nice naming BTW! 😆

scout/server/blueprints/variants/views.py
```diff
@@ -356,7 +356,7 @@ def cancer_variants(institute_id, case_name):
     controllers.populate_chrom_choices(form, case_obj)
 
     form.gene_panels.choices = controllers.gene_panel_choices(institute_obj, case_obj)
-    genome_build = case_obj.get("genome_build")
+    genome_build = "38" if "38" in str(case_obj.get("genome_build")) else "37"
```
northwestwitch, line 359: 👍🏻

PR #2619, opened 2021-05-17
Fix mme match external

### LINKED ISSUE #2618: Manual Matching against external MatchMaker nodes fails
image

But matching works when a patient is saved to database.

northwestwitch:
Fix #2618 

Fix a typo causing the display of an error every time a patient is matched against external MatchMaker nodes. It's not really a bug in the matching (which occurs), it's just saying an error occurred.

**How to test**:
1. Install the branch on scout stage (clinical-db)
2. Go to any case as a user that has MatchMaker permissions (grant yourself permissions using the command line if you are not one)
3. Add a case with variant and/or phenotypes to matchmaker and match it again one or more external nodes
4. The matching should not return error

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -1218,7 +1218,7 @@ def matchmaker_match(request, match_type, institute_id, case_name):
             nodes = [node["id"] for node in matchmaker.connected_nodes]
             for node in nodes:
                 json_resp = matchmaker.match_external(patient_id, node)
-                if json_resp.get("status") != 200:
+                if json_resp.get("status_code") != 200:
                     flash(
                         f"An error occurred while matching patient against external node: '{node}' : {json_resp.get('message')}",
                         "danger",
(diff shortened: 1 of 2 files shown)


northwestwitch:
Tested locally with 2 MatchMaker nodes connected to each other and scout connected to one of them:

Master branch:

image


This branch:

image

northwestwitch:
Tested also on clinical-db stage and it's working fine: 

image

scout/server/blueprints/cases/controllers.py
```diff
@@ -1218,7 +1218,7 @@ def matchmaker_match(request, match_type, institute_id, case_name):
             nodes = [node["id"] for node in matchmaker.connected_nodes]
             for node in nodes:
                 json_resp = matchmaker.match_external(patient_id, node)
-                if json_resp.get("status") != 200:
+                if json_resp.get("status_code") != 200:
```
dnil, line 1221: 👍

dnil (approved):

scout/server/blueprints/cases/controllers.py
northwestwitch: Thanks!

PR #2620, opened 2021-05-17
style match-maker table

moedarrah:
This PR:
1. Add styles to the Matchmaker matches table to make it more compact.
2. Add a docker-compose file to be able to test matchmaker locally

Old table:
117405350-c9cba680-af0b-11eb-895d-b3dd55ea9808

New table: 
Screen Shot 2021-05-17 at 08 55 21

How to test locally:
1. Go to matchmaker branch
2. Open a Rare Disease case
3. Click on MatchMaker in the menu
4. Chose a match from the drop-down
5. Click on the Past Matches button to go to the Matchmaker matches table
6. Under Local matches open a match

**Review:**
- [ ] code approved by
- [x] tests executed by MOD


docker-compose-mm.yml
```diff
@@ -0,0 +1,92 @@
+version: '3'
```
northwestwitch, line 1: Really good to keep this version with Scout connected to MatchMaker, we can reuse it for sure when we need to test changes that have an effect on the interaction of the 2 repositories. What I'd do is renaming into something more self-explanatory, like `docker-compose-matchmaker.yml ` and move it where we keep other containers. What about creating a new directory under scout/containers/development (where there are already kubernetes and systemd) and move it there?

scout/server/config.py
```diff
@@ -68,3 +68,9 @@
 RANK_MODEL_LINK_POSTFIX = "-.ini"
 SV_RANK_MODEL_LINK_PREFIX = "https://github.com/Clinical-Genomics/reference-files/blob/master/rare-disease/rank_model/svrank_model_-v"
 SV_RANK_MODEL_LINK_POSTFIX = "-.ini"
+
+
+# MatchMaker
+MME_ACCEPTS = "application/vnd.ga4gh.matchmaker.v1.0+json"
```
northwestwitch, line 74: These lines need to be commented out in master branch since there is no MatchMaker connection in the default Scout. What you could do instead is add them as comments to `docker-compose-mm.yml`, so we remember how config should be modified for that specific docker-compose file to work

northwestwitch (changes requested):
Those pages look much better! 👍🏻 

I've got a couple of suggestions for improvement and before merging we should test it on stage. Let's install it as soon as scout-stage is available!

containers/development /docker-compose-matchmaker.yml
```diff
@@ -0,0 +1,102 @@
...
+  custom-net:
+    driver: bridge
+    ipam:
+      driver: default
+      config:
+        - subnet: 172.21.0.0/24
+
+
+
+# MatchMaker
```
northwestwitch, line 96: 👍🏻

northwestwitch (approved):
Looking good on stage! 👍🏻 

Example: https://scout-stage.scilifelab.se/cust101/A1383/mme_matches

image

northwestwitch:
Let's merge this one!

PR #2632, opened 2021-05-18
Added GATK to SV/ cancer SV list

mhkc:
This adds GATK as a caller for SV and cancer SV variants

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil (approved):

dnil:
Did you test it locally already?

mhkc:
Yes. I have tested it locally with two samples used for validating our MT pipeline. But I have not created any dummy data that does not contain any sensitive information.

dnil:
That's fine! We don't have to have test data for every constant.

PR #2634, opened 2021-05-18
Fixed crash when gene hasnt a primary_transcript

mhkc:
This fixes an issue when the variant view would crash if a gene hasn't a primary transcript

**Review:**
- [x] code approved by CR
- [x] tests executed by MJ

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/tx_overview.html
+++ b/scout/server/blueprints/variant/templates/variant/tx_overview.html
@@ -69,7 +69,7 @@
                   <td class="d-flex align-items-center">
                     <div> <!-- ID col-->
                       <span class="text"> {{ transcript.transcript_id }}</span>
-                      {% if gene.common and transcript.refseq_id in gene.common.primary_transcripts %}
+                      {% if (primary_transcripts in gene.common) and (gene.common and transcript.refseq_id in gene.common.primary_transcripts) %}
                         <a href="#" data-toggle="tooltip" title="hgnc primary"><span class="badge badge-primary" title="hgnc primary">P</span></a>
                       {% endif %}
                       {% if transcript.mane_select_transcript %}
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/variant/templates/variant/tx_overview.html
```diff
@@ -69,7 +69,7 @@
                   <td class="d-flex align-items-center">
                     <div> <!-- ID col-->
                       <span class="text"> {{ transcript.transcript_id }}</span>
-                      {% if gene.common and transcript.refseq_id in gene.common.primary_transcripts %}
+                      {% if (primary_transcripts in gene.common) and (gene.common and transcript.refseq_id in gene.common.primary_transcripts) %}
```
northwestwitch, line 72:
```suggestion
                      {% if gene.common and 'primary_transcripts' in gene.common and transcript.refseq_id in gene.common.primary_transcripts %}
```

northwestwitch: Jinja is slight different from python and to check that a key exists in a dictionary you check it either with

`if 'key' in dict ` or with `if dict.key`

We've found this out with last week, remember @dnil ? 🤣

northwestwitch (commented):
I have a suggestion but haven't tested it. Could test it later!

scout/server/blueprints/variant/templates/variant/tx_overview.html
dnil: 😅 Yes, that was not totally obvious. Flask2 is like the Dragos, tough on the tough?

dnil: https://www.naspbokhandel.se/pub_images/large/9789176217115_1.Jpeg?timestamp=1616910123

mhkc: @northwestwitch Ill try that solution.

mhkc:
@northwestwitch your solution works. I have committed it.

northwestwitch (approved):
👍🏻

PR #2635, opened 2021-05-18
chr pos split js safer for STR view

dnil:

Some issues remained / arose with filter lock with the new release candidate.

1. A js error on the STR filters (no chr_pos to split) disabled both populating chr and filter lock

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/static/form_scripts.js
+++ b/scout/server/blueprints/variants/static/form_scripts.js
@@ -2,7 +2,11 @@
 function populateCytobands(cytobands){
   var chromPosPattern = new RegExp("^(?:chr)?([1-9]|1[0-9]|2[0-2]|X|Y|MT)$");
   var chrom = document.forms["filters_form"].elements["chrom"].value;
-  var chromPos = document.forms["filters_form"].elements["chrom_pos"].value;
+  var chromPos = "";
+  if (typeof document.forms["filters_form"].elements["chrom_pos"] != "undefined") {
+  	chromPos=document.forms["filters_form"].elements["chrom_pos"].value;
+	}
+
   var chromosome = "";
   console.log("Populate cytobands")
   var matchedChrName = chromPos.match(chromPosPattern)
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -2,7 +2,11 @@
 function populateCytobands(cytobands){
   var chromPosPattern = new RegExp("^(?:chr)?([1-9]|1[0-9]|2[0-2]|X|Y|MT)$");
   var chrom = document.forms["filters_form"].elements["chrom"].value;
-  var chromPos = document.forms["filters_form"].elements["chrom_pos"].value;
+  var chromPos = "";
+  if (typeof document.forms["filters_form"].elements["chrom_pos"] != "undefined") {
```
dnil, line 6: It's not just Jinja and Python! Hm, lets add rust next perhaps?

northwestwitch, line 6:
```suggestion
  if (typeof document.forms["filters_form"].elements["chrom_pos"] !== "undefined") {
```

northwestwitch: Nice that you found the problem! You probably tried this solution and it works, so feel free to ignore my suggestion. But I was thinking that if you are doing a type comparison perhaps the triple equals could be more appropriate than then double equal way (check the docs [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) for a description of the differences)

northwestwitch (approved):
Nice solution! 👍🏻

scout/server/blueprints/variants/static/form_scripts.js
dnil: Thanks, yes, that appears correct, though there is no real issue with using the double equal here either - typeof will return a string regardless of what came in.

PR #2646, opened 2021-05-20
Add a new index in hgnc collection for gene_symbol

northwestwitch:
Adds a new index
**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by (index is already in production)

### DIFF
--- a/scout/constants/indexes.py
+++ b/scout/constants/indexes.py
@@ -19,7 +19,7 @@ INDEXES = {
         ),
         IndexModel(
             [("build", ASCENDING), ("hgnc_symbol", ASCENDING)],
-            name="build_symbol",
+            name="build_hgnc_symbol",
             background=True,
         ),
     ],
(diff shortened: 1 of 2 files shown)


scout/constants/indexes.py
```diff
@@ -17,6 +17,11 @@
             name="build_aliases",
             background=True,
         ),
+        IndexModel(
+            [("build", ASCENDING), ("hgnc_symbol", ASCENDING)],
+            name="build_symbol",
```
dnil, line 22:
```suggestion
            name="build_hgnc_symbol",
```

dnil: I actually like your name better, but *cough* for reasons of compatibility with the current db perhaps it is easiest to use mine? 😬

scout/constants/indexes.py
```diff
@@ -17,6 +17,11 @@
...
+            background=True,
```
dnil, line 23:
```suggestion
```
Not worth it - it is instantly made. Better be done with it!

dnil (approved):

PR #2651, opened 2021-05-21
Gene panel dropdown height

moedarrah:
This PR adds:
Fix the height of all dropdowns in Clinical SNV and INDELs view to be able to select an option from a dropdown in search without the new "bar" (static position)  overlap the dropdown.

I tested it on (Region Annotations, Functional Annotations), and its works as it should.

Screen Shot 2021-05-21 at 10 59 45

**Review:**
- [ ] code approved by
- [x] tests executed by MOD

### DIFF
--- a/scout/server/static/bs4_styles.css
+++ b/scout/server/static/bs4_styles.css
@@ -342,7 +342,7 @@ option {
 }
 
 .dropdown-menu {
-  max-height: 420px !important;
+  max-height: 130px !important;
 }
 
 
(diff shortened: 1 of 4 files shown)


northwestwitch (changes requested):
Nice! I think the change should be applied to all types of variants pages: sv_variants.html, cancer-variants.html, cancer-sv-variants.html and str variants.

You might consider adding the style at a higher level ([here](https://github.com/Clinical-Genomics/scout/blob/master/scout/server/static/bs4_styles.css)), so it's shared among all templates.

scout/server/static/bs4_styles.css
```diff
@@ -343,6 +341,10 @@ option {
    transition: all 0.5s;
 }
 
+.dropdown-menu {
```
northwestwitch, line 344: 👍🏻

northwestwitch (approved):
It's looking good also on the other filter of the other variant types!

Good job also fixing the codefactor issues on the styles document! 🥇

Issue #2615, opened 2021-05-14
SV variants with same position not loaded


ViktorHy:
When loading a trio-SV-vcf into scout, we had this situation:

One variant found in parent and child with position chr7:7478750-74813500
and another variant found in the other parent at the same start position chr7:74787501-74810500

Both variants are loaded into the database, but only one is shown in scout. 

Feels like scout should be able to support several variants with the same start position but differentiating end positions. This specific example might not be the most useful, but still displays the issue. This was a variant called in all three individuals that turned out to be an artifact of the gatk-caller. I just imagine if say, manta would make a call at the same position but as a true positive, but have vastly different end position such that it would not overlap 70% with the gatk. 

[bug_report_scout.ped.gz](https://github.com/Clinical-Genomics/scout/files/6479013/bug_report_scout.ped.gz)
[bug_report_scout.vcf.gz](https://github.com/Clinical-Genomics/scout/files/6479014/bug_report_scout.vcf.gz)

dnil:
We tend to rely on svdb merge to solve a lot of these issues for us. 😊

ViktorHy:
It was merged with svdb before.

ViktorHy:
Still, you could have situations where you have identical start positions that are not merged with svdb.

dnil:
Sure, that could happen. We should at the very least warn about it - I certainly don't like that the first variant is silently overwritten. What ends up in the database after parsing is actually just one variant, the last one to be precise, even though the parse output is for two variants, just since they get the same simple ID (same chr, start, ref, alt, type), and the alt is nondescript in the case of the 23k or 26k deletions.

dnil:
Sorry, something is odd with that `insert_many` operation - as I currently understand it, it should really throw a `DuplicateKeyError` and allow us to take it one variant at the time. But possibly that changed when they changed the write_concern things. We were hoping to have something for this release at least, but we'll have to get back to you on this one.

dnil:
There would be several ways to deal with this long term. My current favoured one is 2, but other ideas are most welcome. 

1. in caller, always output `ALT` `BND` string, as per VCF standard (section 5.4 in v4.3). If approximate, indicate this with the `APPROXIMATE` flag. This guarantees effective uniqueness, and requires very limited changes to the code. Old cases can be re-run to update if needed.
[...]

PR #2661, opened 2021-05-25
Warn for #2615 - duplicate insertion in same bulk

dnil:
This PR fixes a missing warning when duplicate variant_id occur in same insertion bulk.

The actual issue with SVs with same start and nondescript ALT getting the same variant_id is left open. Changing the id structure for the entire db would mean an epic major change. We could "cheat" a little with like tweakin the start a tiny bit, which may not mean so much for a large SV, but we are getting into a territory where we really need to sit back and think about the best to do this.

**How to test**:
1. See test data in issue #2615

**Expected outcome**:

Screenshot 2021-05-25 at 09 22 26

**Review:**
- [x] code approved by CR
- [x] tests executed by DN


scout/adapter/mongo/variant_loader.py
```diff
@@ -312,12 +312,11 @@ def upsert_variant(self, variant_obj):
         try:
             result = self.variant_collection.insert_one(variant_obj)
         except DuplicateKeyError as err:
-            LOG.debug("Variant %s already exists in database", variant_obj["_id"])
+            LOG.warning("Variant %s already exists in database - modifying", variant_obj["_id"])
```
northwestwitch, line 315: 👍🏻

scout/adapter/mongo/variant_loader.py
```diff
@@ -338,6 +337,7 @@ def load_variant_bulk(self, variants):
         except (DuplicateKeyError, BulkWriteError) as err:
             # If the bulk write is wrong there are probably some variants already existing
             # In the database. So insert each variant
+            LOG.debug("Bulk insertion failed - attempting separate upsert")
```
northwestwitch, line 340: Why debug and not warning here?

scout/adapter/mongo/variant_loader.py
```diff
@@ -312,12 +312,11 @@ def upsert_variant(self, variant_obj):
         try:
             result = self.variant_collection.insert_one(variant_obj)
         except DuplicateKeyError as err:
-            LOG.debug("Variant %s already exists in database", variant_obj["_id"])
+            LOG.warning("Variant %s already exists in database - modifying", variant_obj["_id"])
             result = self.variant_collection.find_one_and_update(
                 {"_id": variant_obj["_id"]},
                 {"$set": {"compounds": variant_obj.get("compounds", [])}},
             )
-            variant = self.variant_collection.find_one({"_id": variant_obj["_id"]})
```
northwestwitch, line 320: This wasn't used?

scout/adapter/mongo/variant_loader.py
```diff
@@ -488,7 +488,17 @@ def _load_variants(
                     bulk = {}
 
                 current_region = new_region
-                bulk[var_id] = variant_obj
+                if var_id in bulk:
```
northwestwitch, line 491: Perhaps a stupid question but I feel I have a limited understanding the bulks: what if the variant is in another bulk instead?

scout/adapter/mongo/variant_loader.py
```diff
@@ -488,7 +488,17 @@ def _load_variants(
...
+                    LOG.warning(
```
northwestwitch, line 492: 👍🏻

northwestwitch (approved):
The code looks good, and I see that it's tested! 👍🏻

scout/adapter/mongo/variant_loader.py
dnil: Then the warning that was already in the code should work (database throws an exception, and each variant in the bulk is upserted separately to get to the issue and ensure that as many as possible have been inserted.

scout/adapter/mongo/variant_loader.py
dnil: Apparently not.. Timesaving! 😁

scout/adapter/mongo/variant_loader.py
dnil: Good point - I actually ran the "test"with debug. Will fix!

PR #2667, opened 2021-05-25
Clearer warnings in variants filters, gene search

### LINKED ISSUE #2664: Meaning of filter warning "Gene not included in case clinical list" is not at all clear for users
Especially when they filter using HPO genes. In that case they get long list of genes in that warning and they wonder why. 

This is just a warning clarifying that genes used in the search are not in the panels used to analyse the case (nothing weird with it). But the warning is worrisome for many:

Suggestion:
- Change the color of the alert to green, instead of yellow
- Change the text of the alert to be more self-explanatory

northwestwitch:
Fix #2664.

- Change color of message of the warning message when gene is not in gene panels to blue
- Clearer warning messages in general

**How to test**:
1. Install on stage or locally and do filter variants using the HPO list, gene names and gene HGNC Ids

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -1007,19 +1007,19 @@ def check_form_gene_symbols(
 
     errors = {
         "non_clinical_symbols": {
-            "alert": "Gene not included in case clinical list",
+            "info": "Genes not included in case gene panels",
             "gene_list": non_clinical_symbols,
         },
         "not_found_symbols": {
-            "alert": "HGNC symbol not present in genes collection",
+            "alert": "HGNC symbols not present in database genes collection",
             "gene_list": not_found_symbols,
         },
         "not_found_ids": {
-            "alert": "HGNC id not present in genes collection",
+            "alert": "HGNC id not present in database genes collection",
             "gene_list": not_found_ids,
         },
         "outdated_symbols": {
-            "alert": "Clinical list contains a panel with an outdated symbol for genes",
+            "alert": "Case gene panels contains an outdated symbol for genes",
             "gene_list": outdated_symbols,
         },
     }
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -1019,7 +1019,7 @@ def check_form_gene_symbols(
             "gene_list": not_found_ids,
         },
         "outdated_symbols": {
-            "alert": "Case gene panels contains an outdated symbol for genes",
+            "alert": "Case gene panels contain an outdated symbol for genes",
             "gene_list": outdated_symbols,
         },
     }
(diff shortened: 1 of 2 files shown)


northwestwitch:
Do you have time to review this @mhkc?

scout/server/blueprints/variants/controllers.py
```diff
@@ -1007,19 +1007,19 @@ def check_form_gene_symbols(
 
     errors = {
         "non_clinical_symbols": {
-            "alert": "Gene not included in case clinical list",
+            "info": "Genes not included in case gene panels",
```
dnil, line 1010: I came to worry a bit about this one again. For searching, latest is good, but for warning, this should strictly speaking be for the exact versions that the case was run with. I have a feeling we may have lost touch with that at some point?

scout/server/blueprints/variants/controllers.py
```diff
@@ -1007,19 +1007,19 @@ def check_form_gene_symbols(
...
             "gene_list": not_found_symbols,
         },
         "not_found_ids": {
-            "alert": "HGNC id not present in genes collection",
+            "alert": "HGNC ids not present in database genes collection",
             "gene_list": not_found_ids,
         },
         "outdated_symbols": {
-            "alert": "Clinical list contains a panel with an outdated symbol for genes",
+            "alert": "Case gene panels contain an outdated symbol for genes",
```
dnil, line 1022: Maybe this should also stay on the info stage? It could really be from a panel that the user does not maintain, and it can be a bit slow finding the panel in question.

dnil (approved):
I'll approve: its better than before, but I still think we might do a bit more to clarify the conditions under which each is produced.

scout/server/blueprints/variants/controllers.py
northwestwitch: Good point. Check that clinical genes are retrieved from the right panel versions: 

I think we are good: https://github.com/Clinical-Genomics/scout/blob/34627d65284f28375f86db47a051f904d18fbfc3/scout/adapter/mongo/panel.py#L526

image

dnil: Good! Then I think we can also be a bit more precise about that in the info text as well, right, since the gene could *now* be on the panel, but wasn't in the version run with the case - or the other way around..

northwestwitch:
> I'll approve: its better than before, but I still think we might do a bit more to clarify the conditions under which each is produced.

You don't have to approve if you are not convinced 😄  Any suggestion for improvement?

northwestwitch:
Oh no I've merged by mistake!

northwestwitch:
I'll fix as you suggested in another PR!

PR #2668, opened 2021-05-25
Popup to explain why there are missing links in Loqus observations panel

### LINKED ISSUE #2663: Local database: variants not always present in previous, highlighted cases
When previous cases with the same variant are shown under local database, the previous cases are sometimes in grey:
image
https://scout.scilifelab.se/cust003/21126/321f4cb4c1aea36b23f2e50813814b9d

In the grey cases, the variant cannot be found when searching for the gene name (sometimes other variants in the same gene are found). In a few instances we've checked by position in IGV and there is no trace of the variant. Why are these grey cases showing up?

northwestwitch:
Fix #2663.

Hover on the popup and you'll see this:

image

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/variant_details.html
+++ b/scout/server/blueprints/variant/templates/variant/variant_details.html
@@ -152,8 +152,9 @@
 
 {% macro observations_panel(variant, observations, case) %}
   <div class="card panel-default">
-    <div class="panel-heading">
+    <div class="panel-heading d-flex justify-content-between">
       <a class="text-white" href="https://github.com/moonso/loqusdb" target="_blank">Local observations</a>
+      <span data-toggle="tooltip" title="Missing links to case variants might be caused by variants not loaded after a rerun">?</span>
     </div>
     <div class="card-body">
       <table class="table">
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -152,8 +152,9 @@
 
 {% macro observations_panel(variant, observations, case) %}
   <div class="card panel-default">
-    <div class="panel-heading">
+    <div class="panel-heading d-flex justify-content-between">
       <a class="text-white" href="https://github.com/moonso/loqusdb" target="_blank">Local observations</a>
+      <span data-toggle="tooltip" title="Missing links to case variants might be caused by variants not loaded after a rerun">?</span>
```
dnil, line 157:
```suggestion
      <span data-toggle="tooltip" title="Nr of observations is the total seen in this loqusdb instance. Missing links on greyed out case names for shared local case variants might be caused by variants not loaded after a rerun or inexact SV matching. Only names of collaborator cases are shown. Please contact support to establish a contact for cases from other institutes.">?</span>
```

dnil (approved):
I wrote a small essay. Feel free to move or duplicate and expand in docs. 😅

scout/server/blueprints/variant/templates/variant/variant_details.html
northwestwitch: Nooo no docs, let's keep it there. I think it's really helpful. Perhaps we could omit "Please contact support to establish a contact for cases from other institutes" because it's kind of obvious?

dnil: Sure! That's also really the kind of knowledge that one would wish to have a separate channel for, as it is not really much to do with the software as such.

Issue #2665, opened 2021-05-25
Bad gateway error in causatives page


mhkc:
In version 3.35.0 we frequently get "Bad gateway error" in the /causatives page

northwestwitch:
I think this issue was fixed in #2613. You need to update the indexes of your prod Scout. @dnil could help with more detailed instructions

dnil:
Hm, we had out of memory errors with it before. Well, I would strongly suggest updating the indices as per #2613 anyway, but not really for the causatives. For that part I just moved the sort from the db engine to the web server. For us thats a big improvement - do you have more constraints there possibly?

dnil:
Or do you get any nice log messages that you could share from the crashes? Normally we get an admin log email with each crash that caused a Bad gateway, where one can usually see what is going down. I had to fix the sort a few times since there were some unexpected configurations of gene symbols on the variants. Perhaps you have one or two interesting versions of that that we dont have?

dnil:
The immediate issue was solved with #2673. It would be very helpful to the overall index size if we take the time to rewrite this last(?) large use of variant_id queries to _id, but let us call that part an enhancement instead.

PR #2673, opened 2021-05-26
Case index

dnil:

Attempting to solve #2665 for Lund. Adding back an index for variant_id to the constants, that is used when unrolling the query cursor into a list. An alternative might be to try to recast the positional ids to case id containing document ids, but that will take a bit of error prone logic. Solna still has variant_id in prod, though it was so to speak deprecated, and it is confirmed to take hits on causatives search. It is still slow, but at least one order of magnitude faster than in Lund.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Warning on overwriting variants with same position was no longer shown
 - Increase the height of the dropdowns to 425px
+- More indices for the case table as it grows, specifically for causatives queries
 ### Changed
 - Clearer warning messages for gene searches in variants filters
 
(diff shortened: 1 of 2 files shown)


mhkc (changes requested):
This doesnt solve the issue for me. Having an index with just the variant_id as key solved the issue for me.

The slow query only involves the keys `variant_id`and `institute`

Slow query
```
{'variant_id': {'$in': ['397b5a850464ca8a3ba793619f3ec045', '657bf046d4bf1901b75412ca85b9bc3f', ...]}, 'institute': 'klingen_38'}
```

dnil:
I do think it would solve your issue, as the extra gain from adding the institute to the compound index is bound to be very small after the heavy lifting with variant_id is done, but why not - it can still be reused for other queries and be tad bit faster here! Me like!

dnil:
Hehe, I dont know if you changed your mind or I misread you the first time, but it seems we agree it doesn't matter terribly wether its a compound index or not, as long as variant_id is the first one. I noticed a duplicate for the case ones (those will not really do much for the issue, but are small and will help the caseS page a tad). They would also help with exporting causatives via get_causatives, although that looks a little sketchy to be honest.

mhkc (approved):

PR #2689, opened 2021-06-03
Show matching tiered variants

### LINKED ISSUE #2680: Show if a cancer variant has been previously tiered in another case

northwestwitch:
Show if one cancer variant has matching variants with Tier both on the variantS view and on the variant page. Fix #2680

image

image

**How to test**:
1. Create a duplicate load config file from cancer demo case config file (rename family name and case display name)
2. Load both cases.
3. Go to the of the 2 cancer cases and tag one variant with a tier
4. Go to the other variant and make sure that one variant is reported to have matching tiered variants on the cancer variantS page and on the variant page itself.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/adapter/mongo/variant_events.py
```diff
@@ -246,6 +246,44 @@ def validated(self, institute_id=None, user_id=None, case_id=None):
...
+        for tiered_event in self.event_collection.find(query):
+            # Check if variant is still tiered as suggested by the event
+            tiered_matching_variant = self.variant(
+                case_id=tiered_event["case"], document_id=tiered_event["variant_id"]
+            )
+            if (
+                tiered_matching_variant is None
+                or tiered_matching_variant["_id"] == query_variant["_id"]
+            ):
+                continue
```
dnil, line 275: The loophole being if both a somatic and a constitutional variant have been tiered in the same case, but that is a rather odd situation and could potentially be useful. It does allow the list to grow fairly large if it is a common recurrent variant, but that is better handled on output I guess.

dnil: So, 💯, after some consideration!

dnil (approved):
This is much desired, and working well, so approved. But in its current form, it seems it will quickly run into rather long lists for recurrent variants, so consider fixing that already now. As for the implicit global nature, I am not worried, just want to make sure it was intended!

dnil:
Tested locally:
Screenshot 2021-06-08 at 10 48 04

scout/server/blueprints/variant/controllers.py
```diff
@@ -188,6 +188,11 @@ def variant(
 
     managed_variant = store.find_managed_variant_id(variant_obj["variant_id"])
 
+    if variant_obj.get("category") == "cancer":
+        variant_obj["matching_tiered"] = store.matching_tiered(
+            variant_obj, user_institutes(store, current_user)
```
northwestwitch, line 193: Answering here about you comment about this to be a "global" functionality. The matching tiered variants are collected only for cases in institutes the user is allowed to see (see the user_institutes used here). That should be OK right?

dnil: Good; I missed that - I'm still not sure that we decided anything about sharing beyond general sentiments that it would be good to share, but then this lands on the safe side of that. 👍

northwestwitch:
Changed for displaying the previously tiered grouped by tier category.

variantS page:

image


Variant page:
image

The badges on the variant page are clickable and give you the name of the case with the matching tiered variant, which are in turn links to the variant:
image

northwestwitch:
I'd merge this one @dnil, unless you see anything weird in the new code?

dnil:
I can re-review if you wish - just request if you like. 😁 But it was working before, and the latest addition solved any remaining qualms I had about it.

northwestwitch:
> I can re-review if you wish - just request if you like. 😁 But it was working before, and the latest addition solved any remaining qualms I had about it.

Ok but let's merge then, and we'll test on scout stage before the new release! Thanks for the review!

PR #2698, opened 2021-06-07
Spliceai redux

### LINKED ISSUE #1250: from ESHG 2019 spliceAI
SpliceAI
https://www.cell.com/cell/fulltext/S0092-8674(18)31629-5?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS0092867418316295%3Fshowall%3Dtrue
https://github.com/Illumina/SpliceAI

there should already be a precomputed list of scores for SNV (not for indels)

dnil:

Fix #1250 - parse and display SpliceAI.

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN


scout/constants/variant_tags.py
```diff
@@ -57,6 +57,12 @@
     "high": {"neg": [-2, -float("inf")], "pos": [2, float("inf")]},
 }
 
+SPLICEAI_LIMITS = {
```
dnil, line 60: For filtering later.

northwestwitch:
I can review this one!

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -241,6 +241,20 @@
...
           <span class="float-right">{{ variant.spidex_human }}</span>
         </li>
+        <li class="list-group-item">
+          <a hfref="https://github.com/Illumina/SpliceAI">SpliceAI</a> DS max
+          <span class="float-right control-label" data-toggle="tooltip" data-html="true" data-placement="bottom"
+              title="<strong>{% if variant.spliceai_scores %}
+                 {% for entry in variant.spliceai_scores %}
+                    SpliceAI highest delta score {{ entry }} </strong> at position {{ variant.spliceai_positions[loop.index0]}} relative to this variant.
+                    <br>All scores and positions(relative to variant):<br>
+                    {{ variant.spliceai_predictions[loop.index0] }}
```
northwestwitch, line 251: Predictions could be created here instead ?

northwestwitch (approved):
Really nice to have this feature available from now on! 🥇 

Looks like it's working as it should, but I find the parsing code a bit hard to follow. This is mostly due to the fact that the predictions are created when parsing the transcripts spliceai values. 
Since predictions are used in templates and contain html code, wouldn't be easy enough to display them directly there, instead of saving them in the gene object?

northwestwitch:
Haven't tried to load a case with variants containing SpliceAI annotation, did you @dnil? Otherwise we should do it to test it before next release

dnil:
Thanks for the review! I did test loading a case locally, but it would be good form to do so also on stage before deploying.

I'll have a look at the html formatting - I'm sure I could do a placeholder, escape character or possibly do an array, but to avoid repeating the selection logic, it would still have to go on the gene methinks?

northwestwitch:
> I'll have a look at the html formatting - I'm sure I could do a placeholder, escape character or possibly do an array, but to avoid repeating the selection logic, it would still have to go on the gene methinks?

Mmm it's a but hard for me to figure out how the prediction fields are rendered in the template because the demo case doesn't have the spliceAI variant keys. I was reasoning that if predictions are built upon spliceAI values at a given position then you already have these variables in the loop at line 248 in scout/server/blueprints/variant/templates/variant/variant_details.html, but perhaps it's not so simple?

dnil:
> > I'll have a look at the html formatting - I'm sure I could do a placeholder, escape character or possibly do an array, but to avoid repeating the selection logic, it would still have to go on the gene methinks?
> 
> Mmm it's a bit hard for me to figure out how the prediction fields are rendered in the template because the demo case doesn't have the spliceAI variant keys. I was reasoning that if predictions are built upon spliceAI values at a given position then you already have these variables in the loop at line 248 in scout/server/blueprints/variant/templates/variant/variant_details.html, but perhaps it's not so simple?

You can always borrow a recent NIST one from rare-disorders cases prod or stage, or even housekeeper. But loading is kind of slow, so no need. No, only the variables we decide to store from parsing are actually stored, if that makes sense. In this case, we store only the data from the most dire consequence transcript for each gene, and not even there the details of the eight different variables present other than on that "prediction" field. We could do that as well, perhaps to show them in the transcripts table.

northwestwitch:
> You can always borrow a recent NIST one from rare-disorders cases prod or stage, or even housekeeper. But loading is kind of slow, so no need. No, only the variables we decide to store from parsing are actually stored, if that makes sense. In this case, we store only the data from the most dire consequence transcript for each gene, and not even there the details of the eight different variables present other than on that "prediction" field. We could do that as well, perhaps to show them in the transcripts table.

OK looks complicated, so never mind, it's working! One easy thing that could be done is moving the spliceAI parsing code to a dedicated function. That way the parse transcript function would be less complex.

dnil:
> OK looks complicated, so never mind, it's working! One easy thing that could be done is moving the spliceAI parsing code to a dedicated function. That way the parse transcript function would be less complex.

Agreed: its a first, but they all should be really.

PR #2705, opened 2021-06-09
Fix #2245 - managed variants defaults to all categories

### LINKED ISSUE #2245: Default managed variants filter to all categories?
Managed variant add was acting up on stage. Whats up?

dnil:

**How to test**:
1. try adding a couple of managed variants from different categories and make sure all are shown by default
2. also check that the multi select actually works buy checking a few different groups, single and no category and confirming result
3. consider if we should mark the checkboxes filled in on first load. May be more intuitive, while I think users will forgive us as no filtering is done when none is checked. And this is not sensitive. It may at one point become a long list, although thanks to ClinVar and the like - maybe not, but we can deal with that then.

**Expected outcome**:

Managed variant:
Screen Shot 2021-06-11 at 11 05 18

Category and Kind drop-down:
Screen Shot 2021-06-11 at 10 17 21
Screen Shot 2021-06-11 at 10 17 31
Screen Shot 2021-06-11 at 10 17 40
Screen Shot 2021-06-11 at 10 17 54

**Review:**
- [ ] code approved by
- [x] tests executed by MOD


northwestwitch (changes requested):
The functionality works: defaults to all types of variants! 🥇 

BUT since we are already touching the code I think it would be a good time to introduce some improvement to the web page 😃 

Suggestions:

- In this dropdown and in filters, rename `CANCER` to `CANCER SNV`

image

- Rename `Kind` to `Sub-category` in "Add variant" panel and in Filters perhaps? (Up to you, it's fine to keep it like this anyway)

- In current master is it possible to create a managed variant of category SNV and kind BND. This is conceptually wrong because **a SNV variant (and SNV cancer variant) can only be of subtype SNV, INDEL**. On the other hand variants of category **SV and cancer SV can be of these subtypes: INS, DEL, DUP, CNV, INV, BND**. If would be nice if the second select would only show the appropriate choices according to what you have previously selected in the "category" select
 
- Another thing that I've noticed is that in managed variants filters, **cytoband selects** are not populated when you select a chromosome (check how it should work instead on the variantS filters)

dnil:
- In this dropdown and in filters, rename `CANCER` to `CANCER SNV`

Hm, we do use this terminology a lot, but I guess exposed to different user groups e.g. cli users and pipeline devs see this a lot, but perhaps not web users. Sure.

- Rename `Kind` to `Sub-category` in "Add variant" panel and in Filters perhaps? (Up to you, it's fine to keep it like this anyway)

I'd prefer to keep it different if nothing else since its a mixed bag. It would be very simple to change later if there is reason.

- In current master is it possible to create a managed variant of category SNV and kind BND. This is conceptually wrong because **a SNV variant (and SNV cancer variant) can only be of subtype SNV, INDEL**. On the other hand variants of category **SV and cancer SV can be of these subtypes: INS, DEL, DUP, CNV, INV, BND**. If would be nice if the second select would only show the appropriate choices according to what you have previously selected in the "category" select

There is no such thing as a given first and second: a user may first click in either box, and I don't quite see why we would need to drive them to do it in a particular order? Worst case, they would misunderstand, input a filter combo that doesn't give any results, note this and change - or try to input a variant that is already there? Its not actually patient variants on that list, so not quite an issue I think?

- Another thing that I've noticed is that in managed variants filters, **cytoband selects** are not populated when you select a chromosome (check how it should work instead on the variantS filters)

Absolutely, it was never enabled. Don't quite know why the author kept the fields in the form, but so it was. It's a nice feature - I added it!

scout/server/blueprints/managed_variants/forms.py
```diff
@@ -24,19 +24,24 @@
...
+    cytoband_end = NonValidatingSelectField("Cytoband end", choices=[])
 
     description = TextField(label="Description")
 
 
 class ManagedVariantsFilterForm(ManagedVariantForm):
-    chromosome = SelectField(
-        "Chromosome", [validators.Optional()], choices=CHROMOSOME_FILTER_OPTIONS
-    )
+    chromosome = NonValidatingSelectField("Chromosome", choices=CHROMOSOME_FILTER_OPTIONS)
```
northwestwitch, line 44: 👍🏻

northwestwitch (approved):
> There is no such thing as a given first and second: a user may first click in either box, and I don't quite see why we would need to drive them to do it in a particular order?

Filters is fine, in the worst case scenario you don't find any variant. I was more thinking about the panel for creating variants. It doesn't make much sense to me to create a SNV of subtype BND. But if you want to keep it this way on this PR is fine, it's not the main purpose of the PR anyway. Nice with the cytobands: the selects work now! I took the liberty to split a js function in 2 smaller functions so codefactor now is happy!

scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
```diff
@@ -276,6 +276,29 @@ <h5>
     $('select[multiple]').multiselect({
       buttonWidth: '100%'
     });
+
+    /* Changing the Kind drop-down according to Category drop-down */
+    $('select#category').change(function(){
```
northwestwitch, line 281: Nice with the second select changing according to what's selected on the first one. It's working well when you create a new managed variant.

There are a couple of problems however when you are filtering managed variants:
- If you select all category types (SNV, SV, CANCER SNV, CANCER SV) the second select is only showing the oprions for SNV and CANCER SNV:
image

- Category select in filter modifies both "Kind" selects in filters and in "Add variant" panel, so if you by chance had some previous selection in filters you might end up in this situation in the "Add variant" panel:
image


Suggestion: let's have this function modifying the second select only in "Add variant" panel and  let's leave the filters as they are (all options shown and selectable regardless of the category selected). What do you say @moedarrah ?

northwestwitch (changes requested):
Nice with the preselection of subtype according to the variant type when creating a variant. 
The same system doesn't apply so well in managed variants filtering (see my comment in the code) and perhaps it's an overkill to have it there in the first place. What about removing it from there?

scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
moedarrah: My bad, I didn't notice that it affected the filter.

northwestwitch:
I was doing a final review to this PR but I've noticed that there is still some issue with the query:

For instance if you create a SNV managed variant:
image

And then you filter to retrieve all variants of kind SNV, then the query doesn't return the new variant:

image


Same if you try to return a variant of category "SNV" of kind "SNV":

image

@dnil looks busy on these days so I can fix the query part instead!

dnil:
I think I have a local from last week with some fixes (I seem to recall validation of variant data before attempting a db write) but since you have a frontend version that will cover much of that, so just go for it. Possibly consider applying the same to the file upload. I can give you a better estimate of how much time I will have later in the week, but its not terribly much. I will need to do a little over one week of NIPT and at least a few days of routine arrays, then I'm back to development again. If there is anything pressing, just ping me! Maybe also ping me again before merging this so I can have a quick look? Thanks!

northwestwitch:
> I can give you a better estimate of how much time I will have later in the week, but its not terribly much

I figures as much, and no worries, we'll manage! :) We'll let you know when we are ready to merge this one!

northwestwitch:
Cytobands and chromosome values depend on genome build. For instance chromosomes with genome build 38 will have the mitochondrial chromosome named "M" instead of "MT". Managed variants still doesn't support different genome builds so it's a bit tricky here. Now it would be a good moment to introduce the genome build option. I'll draft somethink!

northwestwitch:
Looks like category and sub-categories contain the wrong options in edit managed variant form:

image

will fix!

northwestwitch:
@dnil we have fixed the PR like this:
- add_variant form: suggest subcategory if user pre-selects variant category
- edit_variant form: suggest subcategory if user pre-selects variant category
- Added genome build in variant_filter form because it is necessary for populating the chroms and cytobands selects
- Populating the cytobands according to genome build and chromosome (variant_filter form)
- Populating the start and end positions according to the selected cytoband (variant_filter form)

What we haven't introduced yet is creating variants with a specific genome build (these is some issue already opened about this). It's not complicated but requires some more work, because you need to take into account existing managed variants without build, managed uploaded via file, filtering using the build param. 
I'd like to introduce this change in another PR since this one is already fairly big as it is, what do you say?

northwestwitch (approved):
I'd merge it..

dnil:
Sorry for leaving this one hanging; I had some commits left over for this one, but they largely overlap (cytobands etc), and your version works nicely as far as I can tell! It is a little confusing with the build-option that isn't really doing much currently, but the issues to fix the db part of it exists anyway hidden or no. This is a good point to merge and it should be obvious where to go next.

northwestwitch:
No problem, it's always good to merge something at some point! 🥳

Issue #2702, opened 2021-06-08
Rerunned case has coverage report from another case


northwestwitch:
This one: https://scout.scilifelab.se/cust002/F0009506-2. Check the coverage report, the family name in the report is F0009506-3.

Both F0009506-2 and F0009506-3 have been rerunned recently

northwestwitch:
Another one: https://scout.scilifelab.se/cust003/21119-2

northwestwitch:
This would be easier checked with Hasta up and running. I'll take a look at it in the afternoon

northwestwitch:
Checking the first case : https://scout.scilifelab.se/cust002/F0009506-2

Scout config file contains the right case display name (2017-25393-04)
And sample ID in the config file has the same name used in the link to retrieve the coverage report from chanjo report:

![image](https://user-images.githubusercontent.com/28093618/121352737-6f1fd300-c92d-11eb-9299-37cd64e15af5.png)

![image](https://user-images.githubusercontent.com/28093618/121353218-e81f2a80-c92d-11eb-97f3-4fbadede7443.png)

So I don't really see an error in Scout..

northwestwitch:
If you run a query in chanjo for that sample you get another family: 🤔 

![image](https://user-images.githubusercontent.com/28093618/121355460-32091000-c930-11eb-8eef-897332141955.png)

northwestwitch:
Sample ACC3549A3 is found in 3 cases in status db and looks like the Scout case (second in the list) is pointing to the coverage data of the other case (third in the list). 

![image](https://user-images.githubusercontent.com/28093618/121357037-b8722180-c931-11eb-9951-96c9b352270e.png)

Perhaps it's not an issue at all?


The other case (https://scout.scilifelab.se/cust003/21119-2) has a similar situation.

PR #2727, opened 2021-06-24
Check case (_id, display_name) and case individuals(individual_id, display_name, phenotype) on case reupload

### LINKED ISSUE #2723: Check what happens if a case is reuploaded with different important keys
Check case:
- Reupload with different display name
- Reupload with different internal id (what happens to the variants, do they point to the correct case?)
- Reupload with different individual IDs
- reupload withn different individual display names.

Perhaps Ideally avoid all these situations and force operator to remove the previous case.

See also #2721 and #2702

### LINKED ISSUE #2771: Update case name command has a misleading description
This is the syntax to update a case:
```
Usage: scout update case [OPTIONS] [CASE_ID]

  Update a case in the database

Options:
  -n, --case-name TEXT           Add/update the display name of case
  -i, --institute TEXT           Update what institutes that has access to
                                 case
  -c, --collaborator TEXT        Add a collaborator to the case
  --vcf PATH                     path to clinical VCF file to be added
  --vcf-sv PATH                  path to clinical SV VCF file to be added
  --vcf-cancer PATH              path to clinical cancer VCF file to be added
  --vcf-cancer-sv PATH           path to clinical cancer structural VCF file
                                 to be added
  --vcf-research PATH            path to research VCF file to be added
  --vcf-sv-research PATH         path to research VCF with SV variants to be
                                 added
  --vcf-cancer-research PATH     path to research VCF with cancer variants to
                                 be added
  --vcf-cancer-sv-research PATH  path to research VCF with cancer structural
                                 variants to be added
  --reupload-sv                  Remove all SVs and re upload from existing
                                 files
  --rankscore-treshold TEXT      Set a new rank score treshold if desired
  --rankmodel-version TEXT       Update the rank model version
  --sv-rankmodel-version TEXT    Update the SV rank model version
  --help                         Show this message and exit.
```

According to the command's description, `-n, --case-name` makes you update a case name, but this is actually incorrect. Case name and institute are instead used to understand which case should be updated whenever a case ID is not provided with the command. Relative code is here: https://github.com/Clinical-Genomics/scout/blob/6171fff1f4b68b03f690b9155fcae16abe76b5e8/scout/commands/update/case.py#L86

northwestwitch:
Fix #2723 and fix #2771, check also #2702

Make sure that when a case is re-uploaded it has:
- Same display_name
- Same individuals (cheked keys: individual_id, display_name, phenotype)

Otherwise raise Error.

**How to test**:
1. Load the demo case into a local instance -> `scout setup demo`
2. Modify the demo config file by changing case display name (family_name in config file) and try to update the case with the command ->  `scout --demo load case scout/demo/643594.config.yaml -u` . It should trigger an error
3. Modify the demo config file by changing one of the samples (individual_id, display_name, phenotype) and try to update the case with the command above. It should also trigger error.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR, pytest


northwestwitch:
@mikaell I could use a review for this one whenever you have time!

mikaell:
> @mikaell I could use a review for this one whenever you have time!

I'll have a look.

scout/adapter/mongo/case.py
```diff
@@ -555,6 +557,60 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
+            institute_obj(dict): institute dictionary
+            update(bool): If existing case should be updated
+            keep_actions(bool): If old evaluated variants should be kept when case is updated
+
+        returns:
+            previous_evaluated_variants(list): list of variants evaluated in previous case
+        """
+
+        if existing_case is None:
+            return
```
mikaell, line 576: Function documentation says a list is returned, but here None is returned.

scout/adapter/mongo/case.py
```diff
@@ -555,6 +557,60 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
+                for ind in existing_case.get("individuals")
+            ]
+        )
+        case_inds = set(
+            [
+                (ind["individual_id"], ind["display_name"], ind["phenotype"])
+                for ind in case_obj.get("individuals")
+            ]
+        )
+        if existing_case_inds != case_inds:
```
mikaell, line 605: The other integrity errors will print a hint to which parameter doesn't match. Might be a good idea to do it here as well?

scout/commands/update/case.py
```diff
@@ -12,8 +12,6 @@
 
 @click.command("case", short_help="Update a case")
 @click.argument("case_id", required=False)
-@click.option("--case-name", "-n", help="Add/update the display name of case")
```
mikaell, line 15: Are commands that used to work removed? Or is something else going on? (I asking because I wonder if removing commands will destroy backwards compatibility and make us do a mayor)

mikaell (commented):
Functionality tested locally and works.

Some comments.

mikaell (approved):
Looks good!

northwestwitch:
I'd like to merge this PR soon (today) as it will avoid problems like #2723 and #2771 in the future. I think it's safe to merge also if will case the next release to be a major.

northwestwitch:
Tested to load a case with this branch on hasta stage: 🆗 
image

This is the case: https://scout-stage.scilifelab.se/cust000/F0006358-mip7_clone

Let's remove it after we merge since has way too many variants!

scout/adapter/mongo/case.py
```diff
@@ -555,6 +557,61 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
+                or None if case is not already present in the database.
+        """
+
+        if existing_case is None:
+            return
+
+        # Avoid saving a new case in database with a display_name that already exists for another case from the same customer
+        if existing_case["_id"] != case_obj["_id"]:
+            raise IntegrityError(
+                "A case with display name %s already exists in database for this customer"
```
dnil, line 582: But you are checking _id, not display name here?

northwestwitch: Check later, at line 590

scout/adapter/mongo/case.py
```diff
@@ -555,6 +557,61 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
+            raise IntegrityError("Case %s already exists in database" % case_obj["_id"])
+
+        # Enforce same display name for updated case as existing case
+        if case_obj["display_name"] != existing_case["display_name"]:
+            raise IntegrityError("Updated case name doesn't match existing case name.")
+
+        # Check that individuals from updated case match individuals from existing case in ID, name and affected status
+        existing_case_inds = set(
+            [
+                (ind["individual_id"], ind["display_name"], ind["phenotype"])
```
dnil, line 596: Why is it important to conserve the phenotype here?

northwestwitch: I was reasoning that if case individuals change affected status it doesn't have any sense to keep old info, since it means that the analysis has been rerunned, which is like having an entire new case

scout/adapter/mongo/case.py
dnil: Ah, or rather line 655! How about printing the _ids as well: that might be handy. That is just QoL though.😊  
I would also rather drop the 579 comment, or explicitly state that it is assumed that the only other way the _ids should differ is If the display name already existed.

dnil: Thanks! 😊

northwestwitch:
I don't understand this new error, I'm not even touching that code! 🤔 

```
images = [{'description': 'A very good description', 'height': 100, 'path': PosixPath('scout/demo/images/custom_images/640x480_... 'scout/demo/images/custom_images/640x480_{REPID}.svg', 'str_repid': '{REPID}', 'title': 'Image for replicon {REPID}'}]

    def _parse_images_cnf(images):
        """
        Parse image info from configuration file and process wildcard searches.
    
        Args:
            images: array with image configuration
    
        Returns:
            parsed_images: array with image binary blob
        """
        parsed_images = []
        for image in images:
            img = image["path"]
>           if "{" in img and "}" in img:
E           TypeError: argument of type 'PosixPath' is not iterable

/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/scout/parse/case.py:59: TypeError
```

northwestwitch:
@dnil  I realized that we'll never going to merge this one if it's a major, so I've reintroduced the cli params and made it a minor. I fixed some stuff also in functions not related to this PR because: **a)** codefactor was complaining for some changes introduced after merging last PR in master, **b)** a test parsing the case config params was no longer working (see PosixPath in comment above). I haven't touched the case and individuals check functions introduced by this before.

dnil:
👍 Do you want a second look at it or are you reasonably confident it is all the same as before?

scout/commands/update/case.py
```diff
@@ -12,8 +12,12 @@
...
-@click.option("--institute", "-i", help="Update what institutes that has access to case")
+@click.option(
+    "--case-name", "-n", help="Search case by display name (institute ID should also be provided)"
```
dnil, line 16: 👍

northwestwitch:
> 👍 Do you want a second look at it or are you reasonably confident it is all the same as before?

It should be fine and this time we have to test a re-upload anyway before creating a new release I think..

tests/commands/update/test_update_case_cmd.py
```diff
@@ -14,7 +14,7 @@ def test_update_case_no_args(mock_app):
     result = runner.invoke(cli, ["update", "case"])
 
     ## THEN it should return an error message
-    assert "Please specify which case to update" in result.output
+    assert "Please specify either a case ID or both case name and institute ID" in result.output
```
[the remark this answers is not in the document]
dnil: Nice with the tests as well!

PR #2731, opened 2021-06-28
Edit/create panels with genes in build 38

### LINKED ISSUE #1753: Add reference build to panel pages
It should be possible to create gene panels in genome builds other than 37. Everything now in those pages point to genome build 37. I'm thinking about genes mapping differently among builds, genes which are missing in one build but present in another, "gene panels" consisting of chromosome coordinates instead of genes.

### LINKED ISSUE #2726: Cant add NOTCH2NLC to gene panels
When trying to add the gene "NOTCH2NLC" to a gene panel scout just flashes the warning "HGNC id not found: 53924" despite it being added to the database.

Its caused by scout not passing genome build parameter to `hgnc_gene` which defaults to grch37.

[https://github.com/Clinical-Genomics/scout/blob/b8ef1230651275c77541920eab5ec9116c71c35a/scout/server/blueprints/panels/views.py#L138](https://github.com/Clinical-Genomics/scout/blob/b8ef1230651275c77541920eab5ec9116c71c35a/scout/server/blueprints/panels/views.py#L138)

northwestwitch:
Fix #2726 and fix #1753

How to test:

- On a local instance of Scout containing all genes (37 and 38) edit the demo gene panel with gene NOTCH2NLC (present only with genome build 38).
- Try to load a new gene panel from file after modifying the file to include gene NOTCH2NLC. Tip: replace a gene from this file with 53924 and NOTCH2NLC.
- In both cases the panel should be updated with the gene
- Note that the multiselect containing the options for the disease transcripts now contains more options (tx for both builds and eventual previously selected transcripts). These transcripts may be the same, but I was thinking better safe than sorry..

Expected outcome:

**Review:**
- [x] code approved by MHKC
- [x] tests executed by MHKC, CR


northwestwitch:
This PR should fix your issues @mhkc (Lund users not being able to add to panels genes that are present only in build 38). I could use a reviewer if you have time!

mhkc:
I will have time tomorrow

mhkc (changes requested):
I am not sure this is working.

If I try to add NOTCH2NLC to the dynamic panel (case view) scout does not throw a warning but the gene does not show up in HPO gene panel. This regardless if I have selected a phenotype or not.

I have tried it with sample 643594

northwestwitch:
> NOTCH2NLC to the dynamic panel (case view) scout does not throw a warning but the gene does not show up in HPO gene panel. This regardless if I have selected a phenotype or not.
> 
> I have tried it with sample 643594

I haven't actually touched that code at all. From your issue I understood that users did want to add HGNC genes to custom genes panels, not dynamic genes panels. We can take it in another PR perhaps?

northwestwitch:
> I haven't actually touched that code at all. From your issue I understood that users did want to add build 38 genes to custom gene panels, not dynamic gene panels. We can take it in another PR perhaps?

I'm suggesting this because gene panels and dynamic gene panels are quite different entities, both in the code and in the applications

northwestwitch:
I've realized that fixing dynamic gene panels to accept genes in build 38 (reflecting genome build of the case) wasn't complicated at all and introduced the fix in this PR! Could you try it when you have time @mhkc. We can release a hotfix  even today if this works fine!

mhkc (approved):
Looks good and seems to be working well

northwestwitch:
What do you say, shall we create a new release (hotfix) based on the branch? So you can deploy it this week?

northwestwitch:
If it is not urgent we can have it in the next minor release (August/Beginning or September)

mhkc:
Its not that urgent. I think its better to include it in the next release.

northwestwitch:
> Its not that urgent. I think its better to include it in the next release.

Ok, good, let's merge it to master then!

PR #2732, opened 2021-06-28
Display filter badges in cancer variant list

mhkc:
The relevant content of the filters column in the VCF file is being displayed in cancer variant list as a qc indicator

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
 
 VARIANT_CALL = ("Pass", "Filtered", "Not Found", "Not Used")
 
+# Formatting for different filters in the vcf file
+VARIANT_FILTERS = {
```
northwestwitch, line 53: Cool, it's working nicely: (cancer demo case, all filters pass :)  )

image

northwestwitch (commented):
It's working nicely it's a simple implementation that doesn't alter database entries (variant parsing and saving). I feel this is a smart and clean solution that might work very well until filters don't need to get too custom for different facilities. If one day we'll realize that we need those filters to diverge a lot we'll find another solution!


Before approving and merging, @hassanfa could you give a look to see if you have suggestions for some of these filters or if you want to add more from Balsamic?

hassanfa:
I'll leave comment in the code. Not for review purpose, but to comment :-D

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "pass": {"label": "PASS", "description": "Passed filtering", "label_class": "success"},
+    "germline": {"label": "GERM", "description": "Germline variant", "label_class": "warning"},
+    "germline_risk": {"label": "GERM", "description": "Germline risk", "label_class": "warning"},
```
hassanfa, line 56: These are filtered out in BALSAMIC. We don't report any germline or germline_risk in BALSAMIC.

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "fail_nvaf": {
+        "label": "N",
+        "description": "Too high VAF in normal sample",
+        "label_class": "danger",
+    },
```
hassanfa, line 61: We leave this filter to be applied by Clinicians. Generally speaking, if a sample has high normal_vaf then Scout's filter will be used to filter them out.

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "fail_pvalue": {"label": "P", "description": "Too low P-value", "label_class": "danger"},
```
hassanfa, line 62: What p-value is this? What type of test is used? pvalue for variant calling or strand bias?

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "fail_homopolymer_indel": {
+        "label": "HP",
+        "Variant in homopolymer region": "Passed filtering",
+        "label_class": "danger",
+    },
+    "fail_longdel": {
+        "label": "LD",
+        "description": "Long DEL from vardict",
+        "label_class": "danger",
+    },
```
hassanfa, line 72: We don't apply any length filter in VarDict on this one either. 

Generally, we trust deletions from VarDict as we still have sv calling enabled.

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "fail_no_tvar": {"label": "NO", "description": "No tumor variant", "label_class": "danger"},
```
hassanfa, line 73: If it is not somatic, why should we even upload it to Scout?

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+    "fail_pon_tnscope": {
+        "label": "PON",
+        "description": "Variant in panel of normals",
+        "label_class": "danger",
+    },
+    "fail_pon_vardict": {
+        "label": "PON",
+        "description": "Variant in panel of normals",
+        "label_class": "danger",
+    },
```
hassanfa, line 88: No PoN in TNScope or VarDict for us either.

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+        "label": "HP",
+        "Variant in homopolymer region": "Passed filtering",
+        "label_class": "warning",
+    },
+    "warn_low_tcov": {
+        "label": "COV",
+        "description": "Low tumor coverage",
+        "label_class": "warning",
+    },
+    "warn_novar": {"label": "NO", "description": "No tumor variant", "label_class": "warning"},
```
hassanfa, line 100: Isn't this exact replica of `fail_no_tvar` above?

northwestwitch:
> I'll leave comment in the code. Not for review purpose, but to comment :-D

Please do! 😄

scout/constants/variant_tags.py
```diff
@@ -49,6 +49,79 @@
...
+        "description": "Passed filtering",
+        "label_class": "warning",
+    },
+    "warn_strandbias": {"label": "SB", "description": "Strand bias", "label_class": "warning"},
+    "warn_low_tvaf": {"label": "LO", "description": "Low tumor VAF", "label_class": "warning"},
+    "warn_verylow_tvaf": {
+        "label": "XLO",
+        "description": "Very low tumor WAF",
+        "label_class": "warning",
+    },
```
hassanfa, line 122: Similar to normal_VAF, we leave this for clinicians to decide what cutoff to apply.

This is going to be confusing because when we use UMIs, we can still call very low AF variants. Wouldn't it be confusing to see a UMI variant without a XLO badge?

hassanfa (commented):
Generally, we never upload anything besides PASS to Scout. If we do so, we end up uploading  10k-200k variants per case with all filters.

Honestly, I don't see a circumstance where we will use any of these badges. All of the variants will only be PASS'ed ones.

hassanfa:
To clear out, I though we are talking about tags in `INFO` and not `FILTER`. Thus badges to be taken from `INFO/WORKFLOW_BADGE` or something.

Example: `1 123123 A T . . . PASS DP=100;AF=0.0532;WORKFLOW_BADGE=possible_germline|low_vaf|strand_bias`

northwestwitch:
> To clear out, I thought we are talking about tags in `INFO` and not `FILTER`. Thus badges to be taken from `INFO/WORKFLOW_BADGE` or something.
> 
> Example: `1 123123 A T . . . PASS DP=100;AF=0.0532;WORKFLOW_BADGE=possible_germline|low_vaf|strand_bias`

These filters are those parsed from the FILTER field I think, see [the code](https://github.com/Clinical-Genomics/scout/blob/ae8f5b3c216c1642e987c195a6ee49c061538aad/scout/parse/variant/variant.py#L110). I don't think Scout parses any filter from the INFO field.

hassanfa:
> > To clear out, I thought we are talking about tags in `INFO` and not `FILTER`. Thus badges to be taken from `INFO/WORKFLOW_BADGE` or something.
> > Example: `1 123123 A T . . . PASS DP=100;AF=0.0532;WORKFLOW_BADGE=possible_germline|low_vaf|strand_bias`
> 
> These filters are those parsed from the FILTER field I think, see [the code](https://github.com/Clinical-Genomics/scout/blob/ae8f5b3c216c1642e987c195a6ee49c061538aad/scout/parse/variant/variant.py#L110). I don't think Scout parses any filter from the INFO field.

If that's the case, then there is little in this PR useful for us from Stockholm site.

bjhall:
@hassanfa 

> Generally, we never upload anything besides PASS to Scout. If we don't do that, we end up uploading 10k-200k variants per case with all filters.

The purpose of these tags/filters are to warn the user that something could be fishy about the variant. I assume that you are not 100% certain about every single variant that you upload to scout and have 0 false positives? So why not help the user to quickly spot potentially spurious variants? I added this feature to Coyote 1-2 years ago and it has been very appreciated by the clinicians.

A lot of these filters are probably specific to ours pipeline and not interesting to you (that's why we originally suggested having them in a config file rather than hardcoding them). But if you are not interested in using a specific filter, just don't use it!

Still some comments about the invidual filters:

> These are filtered out in BALSAMIC. We don't report any germline or germline_risk in BALSAMIC.

We don't intentionally load germline variants either. But how can you be sure for tumor only samples? We use this filter to hightlight to the user that the variant could potentially be a germline variant.

> We leave this filter to be applied by Clinicians. Generally speaking, if a sample has high normal_vaf then Scout's filter will be used to filter them out.

Sure, we also use filters for nVAF, but this serves as an additional warning to the user. Generally I think it is dangerous to set too strict filters for nVAF, since variants in certain homopolymers can have fairly high VAFs in normal samples. 

> What p-value is this? What type of test is used? pvalue for variant calling or strand bias?

For us this is simply the PV2 value from TNscope. But again the details are not really important. I think the point is that you should use warnings that are useful to your pipeline.

>We don't apply any length filter in VarDict on this one either.
>Generally, we trust deletions from VarDict as we still have sv calling enabled.

For us VarDict is generating quite a few false positive long deletions/insertions. Don't you see the same thing? It's possible we're holding it wrong :)

> If it is not somatic, why should we even upload it to Scout?

This filter exists purely due to a technical detail in freebayes. Again the details are not relevant...

> No PoN in TNScope or VarDict for us either.

Then simply don't use that filter! Although I highly recommend looking into using PONs! Most useful filter I ever added!

>Similar to normal_VAF, we leave this for clinicians to decide what cutoff to apply.
>This is going to be confusing because when we use UMIs, we can still call very low AF variants. Wouldn't it be confusing to see a UMI variant without a XLO badge?

Again, what's the harm in hightlighting to the user that the variant has low frequency in the tumor? Doesn't mean that the variant in not true. Not sure what you mean with `Wouldn't it be confusing to see a UMI variant without a XLO badge`

>To clear out, I thought we are talking about tags in INFO and not FILTER. Thus badges to be taken from INFO/WORKFLOW_BADGE or something.
>Example: 1 123123 A T . . . PASS DP=100;AF=0.0532;WORKFLOW_BADGE=possible_germline|low_vaf|strand_bias

That's another way to do the same thing, I suppose. However I think this is great fit for the FILTER field and don't see the point in inventing an arbitrary INFO-field to accomplish the same thing.

hassanfa:
It is not that we have 0 false positives. We do have just any other diagnostics workflow. Again, at Stockholm we strive to upload as few variants as possible to Scout for actionable targets aiming for maximum specificity. These filters might come handy in future for sure 👍🏽 

I commented on the code because it was easier to comment there. As per my previous comment above, those are _not_ a code review. 😉  Of course your can go ahead with these changes as you seem fitting.

Most of these filters, besides some being redundant, are just common sense that should and has to be done before interpretation. I see these filters as a wonderful tool for research cases, so 👍🏽 

Regarding VarDict: I don't think you are doing it wrong ;-) For cancer screening or disease monitoring, then high false positive would've been a problematic. But since we are still trying to look for actionable, druggable, and mostly known targets, then sensitivity is _not_ the highest importance for now. We can live with a few false positives. If you have solved both sensitivity and specificity issues at the same time, then that's a big plus!

northwestwitch (approved):
The code works so I'm approving. Feel free to merge whenever you are satisfied with the filters!

bjhall:
@hassanfa  

Just a quick comment again about this, since I think you are misunderstanding the purpose of these flags/warnings (let's not call them filters, I suppose). Obviously you're free to not use this feature, but I think you're dismissing it on incorrects assumptions.

> It is not that we have 0 false positives. We do have those just lik eany other diagnostics workflow. Again, at Stockholm we strive to upload as few variants as possible to Scout for actionable targets aiming for maximum specificity. These filters might come handy in future for sure 👍🏽

We definitely have the same aim in Lund (although we prioritize sensitivity over specificify, when in doubt). For a typical hematological case we present 5-10 variants, out of which maybe 0-2 variants (typically 0) have one of these flags set, the remaining simply being PASS.

Something like this:
bild


> Most of these filters, besides some being redundant

They are not redundant, but may appear to be since they are poorly described (sorry about that, I can try to write better descriptions)... 

>are just common sense that should and has to be done before interpretation. 

Again this assumes that you can do perfect filtering with 100% spec and sens. Obviously we also filter out 99% of the false positives prior to loading variants, but in order to not lose sensitvity we sometimes present *a small number* of variants that are borderline cases (or have certain characteristics) and add one of these flags as a warning to the person interpreting the data. The alternatives being 1) filtering them out and inevitably losing sensitivity or 2) simply presenting them as PASS, both of which are significantly worse!

PR #2734, opened 2021-06-29
Phenotypes (HPO terms )on individuals

### LINKED ISSUE #1952: Phenotypes on individuals

northwestwitch:
Assign phenotype terms at the case level or at individual level. Only HPO terms only for now. 

image

Fix #1952 

**How to test**:
1. Install the branch
2. Try to assign phenotype terms to a case without specifying individuals
2. Try to assign phenotype terms to a case specifying individuals showing phenotype
3. Remove the assigned HPO terms from cases above

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/event.py
+++ b/scout/adapter/mongo/event.py
@@ -181,7 +181,15 @@ class EventHandler(CaseEventHandler, VariantEventHandler):
         return self.event_collection.find(query)
 
     def add_phenotype(
-        self, institute, case, user, link, hpo_term=None, omim_term=None, is_group=False
+        self,
+        institute,
+        case,
+        user,
+        link,
+        hpo_term=None,
+        omim_term=None,
+        is_group=False,
+        individuals=[],
     ):
         """Add a new phenotype term to a case
 
@@ -195,6 +203,7 @@ class EventHandler(CaseEventHandler, VariantEventHandler):
             hpo_term (str): A hpo id
             omim_term (str): A omim id
             is_group (bool): is phenotype term a group?
+            individuals(list): list of case individuals showing a phenotype
 
         """
         hpo_results = []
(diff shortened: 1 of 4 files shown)


northwestwitch:
This is ready for review!

scout/adapter/mongo/event.py
```diff
@@ -223,25 +232,35 @@ def add_phenotype(
...
-                    verb="add_phenotype",
-                    subject=case["display_name"],
-                    content=phenotype_id,
-                )
+            LOG.info(
+                "Creating event for adding phenotype term for case"
+                " {0}".format(case["display_name"])
+            )
+
+            self.create_event(
```
mhkc, line 254: Please considering make the event texts more informative now when there is more granularity in how HPO terms can be applied. The current event text is `<user> added HPO term for 643594 at 2021-07-22`, it could instead be something like

`<user> added HPO term <term> to <individuals> at 2021-07-22`

mhkc (approved):
Looks good! Great work!

northwestwitch:
I've introduced a small improvement: red color for individual icons whenever the individual is affected
image

scout/adapter/mongo/event.py
northwestwitch: That's a really good idea. I'll see what can be done! Thanks for the review 🙏🏻

northwestwitch:
I think I'm going to merge this one as soon as the tests pass. The changes I've made are quite safe and we're going to test it again anyway before deploying!

PR #2701, opened 2021-06-08
Print Stacktrace if Load Fails

mikaell:
This PR adds a functionality.

Print stack trace instead of hiding it when Scout crashes on load command. Print crash data as Error and not Warning.

**How to test**:
1. how to test it, possibly with real cases/data
Break a config file. For example:

```
$ sed -i.bak 's/scout\/demo\/643594\.clinical\.vcf\.gz/scout\/demo\/643594\.clinical\.vcf\.gz\.bad_path/g' scout/demo/643594.config.yaml
$ scout load case scout/demo/643594.config.yaml -u
```

**Expected outcome**:
The functionality should be working

image

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #2736, opened 2021-06-29
Print Stacktrace for More Exceptions When Loading

mikaell:

Print stack trace for even more command line errors. These changes are uniform with #2701.

**How to test**:
_I'm having a hard time figuring out a good way to force a failure naturally, but got stuck. Any ideas?_

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/commands/load/case.py
+++ b/scout/commands/load/case.py
@@ -98,7 +98,7 @@ def case(
         )
     except SyntaxError as err:
         LOG.error(
-            "SyntaxError {} missing when loading '{}' {}".format(
+            "SyntaxError {}: missing when loading '{}' {}".format(
                 err, config.name, traceback.format_exc()
             )
         )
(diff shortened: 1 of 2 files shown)


mikaell:
Changelog already has a line regarding stack trace printing for the upcoming version.

northwestwitch (approved):
It's looking good but I've tried to omit a couple of important keys (family and owner) and looks like that they don't trigger the KeyError (missing key) exception. It doesn't matter much if we are going to go pydantic so fee free to fix it or not.

mikaell:
> It's looking good but I've tried to omit a couple of important keys (family and owner) and looks like that they don't trigger the KeyError (missing key) exception. It doesn't matter much if we are going to go pydantic so fee free to fix it or not.

Yes. And other errors tried one tries to provoke are caught by the yams parser or in similar instances. But sometimes you'll get that strange error and then it's nice with a stack trace 😎

PR #2747, opened 2021-07-06
BUGFIX - Optional institute-specific Alamut Visual Plus API key in Alamut links

### LINKED ISSUE #2745: Connection between SCOUT and new version of Alamut is lost...
We have installed the new version of Alamut - AlmutVisualPlus

Due to this we can not "connect" between SCOUT and the new version of the program.

According to @dnil this could be due changes in HTTP-API (and hopefully only this)

Here is their reply regrading this question:

_Dear Kristina,

Actually API in Alamut Visual Plus requires the user or the third-partysoftware to provide its institution ID and API key (this API key is created upon request, I forward your request to our Sales department). You can find all needed information in the user manual on page 63 (https://extranet.interactive-biosoftware.com/User%20Guide%20Alamut%20Visual%20Plus%20v1.2.pdf).

Best,

Thomas_

northwestwitch:
Fix #2745. Allows providing an institute-specific Alamut Visual Plus key if this software is present.

In that case the link to alamut from variants page will be the following:
http://localhost:10000/search?apikey=blablabla&request=7:124491972C%3EA

If no alamut key is set for the given institute then the link to Alamut will be this:
http://localhost:10000/show?request=7:124491972C%3EA

**How to test**:
2. As an admin user, edit institute settings for an institute who has access to Alamut:
image
3. Go to a variant and click on the Alamut link, check that link works
4. Try to reset the Alamut link to None from the institute settings by providing a blank string for this paremeter

**Expected outcome**:
The alamut link should contain `apikey` args 

**Review:**
- [ ] code approved by
- [x] tests executed by DN


northwestwitch:
Tested this with Nicole. It didn't work, see error screenshot:
image

I think we might to provide also a string for the "institution".

northwestwitch:
This is ready for review now. Assigning @dnil as he has access to the app!

northwestwitch:
It's installed on stage now!

dnil:
Works! 🚀 

It is still perhaps a little confusing, as the new tab contains that key and an empty "error" element. Perhaps we could capture that and avoid displaying, but that will require adding some js magic. Would not be so fun to debug without a local Alamut client. Lets keep it as is. 

Just one thing; currently the Alamut button is disabled if there are no bam-files: no doubt a vestige from an attempt to (optionally) load bams to Alamut. I would suggest replacing those with always enabling the link for variants that could be opened. If we later manage to get bam load working we could have a conditional for that load URL, but always use the search/show link anyway.

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -97,20 +97,29 @@ <h5 class="header">General Institute Settings</h5>
...
                 <select class="selectpicker" name="loqusdb_id">
                   {% for instance in loqus_instances %}
                     <option value="{{instance}}" {% if institute.loqusdb_id == instance %} selected {% endif %}>{{instance}}</option>
                   {% endfor %}
                 </select>
               </div>
+              <!-- End of loqusdb settings -->
+              <!-- Alamut settings -->
+              <div class="col-md-3">
+                {{ form.alamut_key.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Eventual Alamut Visual Plus API key to be used when opening links to Alamut") }}
```
dnil, line 112:
```suggestion
                {{ form.alamut_key.label(class="control-label",data_toggle="tooltip", data_placement="top", title="Optional Alamut Visual Plus API key to be used when opening links to Alamut") }}
```

dnil (approved):

northwestwitch:
> Just one thing; currently the Alamut button is disabled if there are no bam-files: no doubt a vestige from an attempt to (optionally) load bams to Alamut. I would suggest replacing those with always enabling the link for variants that could be opened. If we later manage to get bam load working we could have a conditional for that load URL, but always use the search/show link anyway.

Sounds good, I'll remove the conditional!

Issue #2725, opened 2021-06-24
Handle Loqus CalledProcessError Exception to avoid variant page crashing


northwestwitch:
This error is causing variants page to crash. I recall that some try/catch block was in place when running queries on LoqusDB cli, but it's not working for the following error:

```subprocess.CalledProcessError: Command '['/home/hiseq.clinical/miniconda/envs/P_loqusdb/bin/loqusdb', '--config', '/home/hiseq.clinical/servers/config/clinical-db.scilifelab.se/loqusdb.yaml', 'variants', '--to-json', '--variant-id', '5_138283171_C_T', '--case-count']' died with <Signals.SIGSEGV: 11>.```

**Comment by mikaell** [comment]
I can have a look at this.

**Comment by mikaell** [comment]
Loqusdb requires a module named *cyvcf2* which seems to be missing in the python environment. This causes Loqusdb to crash on ClinicalDB. However, on Hasta where cyvcf2 is installed, the command works.


```
$ loqusdb --config /home/hiseq.clinical/servers/config/clinical-db.scilifelab.se/loqusdb.yaml variants --to-json --variant-id 5_138283171_C_T --case-count
Traceback (most recent call last):
  File "/home/hiseq.clinical/miniconda/bin/loqusdb", line 7, in <module>
    from loqusdb.__main__ import base_command
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/__main__.py", line 14, in <module>
    from loqusdb.commands import base_command
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/commands/__init__.py", line 4, in <module>
    from .load import load as load_command
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/commands/load.py", line 8, in <module>
    from loqusdb.utils import load_database
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/utils/__init__.py", line 4, in <module>
    from .load import (load_variants, load_family, load_database)
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/utils/load.py", line 8, in <module>
    from loqusdb.vcf_tools import (get_formated_variant, get_vcf)
  File "/home/hiseq.clinical/miniconda/lib/python2.7/site-packages/loqusdb/vcf_tools/__init__.py", line 1, in <module>
    from .vcf import get_vcf
[...]

PR #2764, opened 2021-07-27
Bugfix LoqusDB Crash

mikaell:
This PR adds fixes two bugs.

**How to test**:

a. If the subprocess call to LoqusDB raises an exception, Scout will not crash, but print a stack trace and continue operating.
Reference #2725.

b. Potential bug is fixed which appeared when converting minor versions to floats

```
>>> float('2.5')
2.5
>>> float('2.5.1')
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '2.5.1'
```

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


mikaell:
This branch ran into strange problems. Rebooted in #2765.

PR #2765, opened 2021-07-27
Bugfix loquscrash2 210727 2nd try

### LINKED ISSUE #2725: Handle Loqus CalledProcessError Exception to avoid variant page crashing

mikaell:
*Previous PR #2764ran into strange problems -was it not started from an updated master branch?*

This PR adds fixes two bugs.

**How to test**:

a. If the subprocess call to LoqusDB raises an exception, Scout will not crash, but print a stack trace and continue operating. fix  #2725.

b. Potential bug is fixed which appeared when converting minor versions to floats

```
>>> float('2.5')
2.5
>>> float('2.5.1')
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '2.5.1'
```

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/extensions/loqus_extension.py
```diff
@@ -129,22 +130,21 @@ def get_api_loqus_version(api_url):
         if version is None:
             raise ConfigError(f"LoqusDB API url '{api_url}' did not return a valid response.")
 
-        return float(version)
```
northwestwitch, line 132: 👍🏻

scout/server/extensions/loqus_extension.py
```diff
@@ -129,22 +130,21 @@ def get_api_loqus_version(api_url):
...
         try:
             output = execute_command(call_str)
         except CalledProcessError:
-            LOG.warning("Something went wrong with loqus")
-            return -1.0
+            return "-1.0"
 
         version = output.rstrip().split(" ")[-1]
         LOG.debug("version: {}".format(version))
-        return float(version)
```
northwestwitch, line 147: 👍🏻

scout/server/extensions/loqus_extension.py
```diff
@@ -279,18 +278,13 @@ def get_exec_loqus_variant(self, loqus_instance, variant_info):
...
-            raise err
-
-        res = {}
-        if output:
-            res = json.loads(output)
-
-        return res
+            if output is not None:
+                return json.loads(output)
+        except CalledProcessError:
```
northwestwitch, line 285: I've simulated the error by actually raising the CalledProcessError in a connected local instance of loqusdb (exe):

image


The error is propagated to Scout but the app doesn't crash: 👍🏻 

image

scout/server/extensions/loqus_extension.py
```diff
@@ -99,7 +100,7 @@ def init_app(self, app):
 
     def version_check(self, loqusdb_settings):
         """Check if a compatible version is used otherwise raise an error"""
-        if not loqusdb_settings["version"] >= 2.5:
+        if not loqusdb_settings["version"] >= "2.5":
```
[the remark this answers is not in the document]
northwestwitch: The string comparison should work as expected, as long as loqus returns string-formatted numbers. If by chance it starts returning something like "v1.4" then we would be in trouble, because "v1.4" >= "2.5". But let's leave it like this for now!

northwestwitch (approved):
This is working perfectly on a local instance of Scout connected to a local loqus exe.
Really appreciated also the string to float conversion bug!

I think it's fine to merge as it is!

PR #2778, opened 2021-08-13
Better control over MatchMaker matching nodes and results

### LINKED ISSUE #2773: Matchmaker matching performed on MyGene2 when match is launched on Matchbox
Error found on stage, using this case: https://scout-stage.scilifelab.se/cust101/A1383

northwestwitch:
This current master branch allows matching patient matching against:
- other Scout patients in MatchMaker
- patients from all connected nodes
- patients from a specific selected node:

The latter option has a problem though: even if the user specifies only one external node, the matching is performed again all nodes anyway. 

This PR:
- Fixes this problem: fix #2773
- Removes the option to match against all nodes at once, making the process a bit more active on the user side, increasing its control over requests and results.

**How to test ---> On clinical-db stage**:
1. On stage, master branch:
- [ ]   Grant the role of matchmaker submitter to your user: `scout update user -r mme_submitter -u your_email`
- [ ]   Go to [this patient](https://scout-stage.scilifelab.se/cust101/A1383/mme_matches) and run a match against the MatchBox node
- [ ] You should see that the matching is found on MyGene2 instead (that's because the matching is run on both servers anyway but only MyGene2 has results)
- [ ] Switch to this branch and make sure that
- [ ] The option to run matching against all the external nodes is gone
- [ ] The matching against MatchBox doesn't return results (shouldn't be there at all if it doesn't have results).

**Expected outcome**:

**Review:**
- [x] code approved by DN


scout/server/blueprints/cases/controllers.py
```diff
@@ -1286,6 +1286,8 @@ def matchmaker_match(request, match_type, institute_id, case_name):
             # Against every node
             nodes = [node["id"] for node in matchmaker.connected_nodes]
             for node in nodes:
+                if node != target:
```
northwestwitch, line 1289: This is the actual fix of the PR!

scout/server/blueprints/cases/controllers.py
```diff
@@ -1256,12 +1256,12 @@ def matchmaker_matches(request, institute_id, case_name):
...
 
 
-def matchmaker_match(request, match_type, institute_id, case_name):
+def matchmaker_match(request, target, institute_id, case_name):
     """Initiate a MatchMaker match against either other Scout patients or external nodes
 
     Args:
         request(werkzeug.local.LocalProxy)
-        match_type(str): 'internal' or 'external'
+        target(str): 'internal' for matches against internal patients, or id of a specific node
```
dnil, line 1264: 👍

dnil (approved):

PR #2802, opened 2021-08-23
Update genes command will not trigger an update of database indices any more

northwestwitch:
Removed a line in update genes command that triggers also the update indices command

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/commands/update/genes.py
+++ b/scout/commands/update/genes.py
@@ -181,6 +181,4 @@ def genes(build, downloads_folder, api_key):
 
         load_transcripts(adapter, ensembl_tx_res, genome_build, ensembl_genes_dict)
 
-    adapter.update_indexes()
-
     LOG.info("Genes and transcripts loaded")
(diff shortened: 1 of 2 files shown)


scout/commands/update/genes.py
```diff
@@ -181,6 +181,4 @@ def genes(build, downloads_folder, api_key):
 
         load_transcripts(adapter, ensembl_tx_res, genome_build, ensembl_genes_dict)
 
-    adapter.update_indexes()
```
dnil, line 184: 💯

dnil (approved):
Tentatively approving as discussed! There should be no reason from a production perspective, but it is possible that setup and/or testing is affected, and I have not gone down that road yet.

northwestwitch:
Merging this, as the update_index shouldn't be triggered automatically from anywhere in the code!

PR #2803, opened 2021-08-24
Do not remove update genes command tempdir before resource files are parsed

### LINKED ISSUE #2801: Scout update genes works only if you pre-download update files
If you try to update genes on the fly (`scout update genes --api-key somekey`) you get this error:

image

northwestwitch:
Fix #2801 

**How to test**:
1. Locally, run the command `scout --demo update genes` on master branch. The command should fail
2. Switch to this branch and repeat: the command should update genes

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by mikaell, DN

### DIFF
--- a/scout/commands/update/genes.py
+++ b/scout/commands/update/genes.py
@@ -76,8 +76,6 @@ def fetch_downloaded_resources(resources, downloads_folder, builds):
         builds(list): a list containing both genome builds or one genome build ['37', '38']
 
     """
-    LOG.error(os.listdir(downloads_folder))
-
     for resname, filenames in UPDATE_GENES_RESOURCES.items():
         for filename in filenames:
             resource_path = os.path.join(downloads_folder, filename)
@@ -137,7 +135,6 @@ def genes(build, downloads_folder, api_key):
         downloads_folder = tempfile.TemporaryDirectory(
             dir=os.getcwd()
         )  # Create a temp dir in working dir
-        LOG.error(downloads_folder.name)
         try:
             download_resources(downloads_folder.name, api_key, builds)
         except Exception as ex:
--- a/scout/commands/update/genes.py
+++ b/scout/commands/update/genes.py
@@ -76,6 +76,7 @@ def fetch_downloaded_resources(resources, downloads_folder, builds):
         builds(list): a list containing both genome builds or one genome build ['37', '38']
 
     """
+
     for resname, filenames in UPDATE_GENES_RESOURCES.items():
         for filename in filenames:
             resource_path = os.path.join(downloads_folder, filename)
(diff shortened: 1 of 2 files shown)


northwestwitch:
This is ready for review

scout/commands/update/genes.py
```diff
@@ -133,12 +133,12 @@ def genes(build, downloads_folder, api_key):
...
     # If required resources are missing, download them to a temporary file
     if downloads_folder is None:
-        with tempfile.TemporaryDirectory() as tempdir:
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tempdir:
             try:
                 download_resources(tempdir, api_key, builds)
             except Exception as ex:
                 LOG.error(ex)
-        fetch_downloaded_resources(resources, tempdir, builds)
+            fetch_downloaded_resources(resources, tempdir, builds)
```
northwestwitch, line 141: This is the fix

mikaell:
> This is ready for review

I'll have a look.

scout/commands/update/genes.py
```diff
@@ -133,12 +133,12 @@ def genes(build, downloads_folder, api_key):
 
     # If required resources are missing, download them to a temporary file
     if downloads_folder is None:
-        with tempfile.TemporaryDirectory() as tempdir:
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tempdir:
```
dnil, line 136: Are there any reasons not to honor the ENV TEMP variables? Say by using tempfile.gettempdir() or such?

dnil: Document otherwise; temporary files in CWD is not really expected. 😊

scout/commands/update/genes.py
dnil: Thanks, I never quite got why on earth it should survive outside the with block anyway. Python variable life is a little involved, but this is clearer regardless..

dnil (approved):

dnil:
> > This is ready for review
> 
> I'll have a look.

Ah, missed that; as you like @mikaell - never wrong with more reviewing. We should try to use the "assign" function on PRs to indicate who is looking. But rarely are there two candidates.. 🙄

scout/commands/update/genes.py
```diff
@@ -133,12 +133,12 @@ def genes(build, downloads_folder, api_key):
...
             try:
                 download_resources(tempdir, api_key, builds)
             except Exception as ex:
```
mikaell, line 139: Should execution continue if `download_resources()`? If I read the code correct it will quickly crash.

mikaell:
> > > This is ready for review
> > 
> > 
> > I'll have a look.
> 
> Ah, missed that; as you like @mikaell - never wrong with more reviewing. We should try to use the "assign" function on PRs to indicate who is looking. But rarely are there two candidates.. 🙄

Forgot to assign myself. 

So the code quality will be very good.

scout/commands/update/genes.py
northwestwitch: There's nothing wrong with having the temp dir in ENV TEMP. It's just that in cwd you actually see what is going on during the command execution, and then you see that the temp dir is gone. I can add a comment!

scout/commands/update/genes.py
northwestwitch: I think it's fine. It will just exit with a warning, see line 90 in the code.

northwestwitch: That is basically what happens now before the bugfix

scout/commands/update/genes.py
dnil: I get frustrated with apps that dont listen to any TMPDIR etc. 😆 Perhaps add it to the admin docs and the help text for the command methinks - it is not a common practice.

northwestwitch: Or I just remove that dir=os.getcwd(), I don't want to make it more complicated than it really is. It's a dir that is going to be removed as soon as the command as finished anyway

northwestwitch:
Merging this one in 3, 2, ...

PR #2810, opened 2021-08-24
Avoid case reports triggering the loading spinner

### LINKED ISSUE #2806: Spinner does not stop on download
Confirmed in Chrome and Safari. No document ready event after download completes supposedly. Lets check what is triggered then!

northwestwitch:
Fix #2806

**How to test**:
- Locally, on master branch case page, try to download the PDF delivery report or the PDF general report (RD case) or the Coverage & QC report (cancer case). After they are downloaded the loading spinner won't stop spinning on case page
- Switch to this branch and the loading spinner won't be triggered by the file download

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
+++ b/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
@@ -567,7 +567,7 @@
     </div>
     <div>
       <a href="{{ url_for('cases.coverage_qc_report', institute_id=institute,
-                        case_name=case.display_name, date=date, format='pdf') }}"><i class="fa fa-file-pdf-o"></i></a>
+                        case_name=case.display_name, date=date, format='pdf') }}" target="_blank"><i class="fa fa-file-pdf-o"></i></a>
     </div>
   </div>
 </div>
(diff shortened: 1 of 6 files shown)


northwestwitch:
Ready for review. This should fix the spinner when downloading of 3 PDF reports. Are there any more situations like this?

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -75,7 +75,7 @@
                 <i class="fa fa-link"></i></a>
         </div>
         <div>
-          <a href="{{ url_for('cases.pdf_case_report', institute_id=institute._id, case_name=case.display_name) }}">
+          <a href="{{ url_for('cases.pdf_case_report', institute_id=institute._id, case_name=case.display_name) }}" target="_blank">
```
dnil, line 78: 👍

dnil (approved):
Super! Dont forget the vcf2cytosure one as well!

northwestwitch:
We still have to fix the spinner when the user exports the variant list from variants filters. That one is a bit tricky because it's not a link but a form submit element.

dnil:
Ah, right - good call!

dnil:
How about just setting the form target to "_blank"? Does it break anything or open an annoying page?

northwestwitch:
> How about just setting the form target to `"_blank"`? Does it break anything or open an annoying page?

It opens the annoying page!

PR #2809, opened 2021-08-24
Made Alamut button optional and moved it on the side of igv button

northwestwitch:
Fix #2805 
- Created a config settings option to hide Alamut button (not all centers have this software)
- If SHOW_ALAMUT_LINK is set to True, show it on the side of the igv button.

image

**How to test**:
1. Locally or on stage on both RD and cancer variants.

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #2815, opened 2021-08-24
Reachable alamut button

dnil:

This PR also merges changes from #2809 from @northwestwitch. Sorry about the poor coordination: I'll tentatively blame the GitHub interface, but this user may be slightly at fault as well. 😊 

Screenshot 2021-08-24 at 15 34 32
Screenshot 2021-08-24 at 14 44 20

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/cancer-variant.html
+++ b/scout/server/blueprints/variant/templates/variant/cancer-variant.html
@@ -3,7 +3,7 @@
 {% from "variant/utils.html" import rankscore_panel, overlapping_panel, genes_panel, transcripts_panel, proteins_panel,  pin_button, causative_button, modal_causative %}
 {% from "variant/tx_overview.html" import disease_associated, transcripts_overview %}
 {% from "variant/variant_details.html" import frequencies, gtcall_panel, observations_panel, old_observations, mappability_list, severity_list %}
-{% from "variant/buttons.html" import igv_button, variant_tag_button, variant_tier_button, dismiss_variant_button, mosaic_variant_button, splice_junctions_button %}
+{% from "variant/buttons.html" import variant_tag_button, variant_tier_button, dismiss_variant_button, mosaic_variant_button, splice_junctions_button %}
 {% from "variant/components.html" import alignments, panel_classify, compounds_panel, matching_variants, external_links  %}
 {% from "variant/sanger.html" import sanger_button, modal_sanger, modal_cancel_sanger %}
 {% from "variant/gene_disease_relations.html" import omim_phenotypes %}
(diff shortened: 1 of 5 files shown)


scout/server/blueprints/variant/templates/variant/cancer-variant.html
```diff
@@ -299,35 +300,12 @@ <h6 class="card-header">Variant callers</h6>
...
-        </td>
-      </tr>
-      {% if splice_junctions_tracks %}
-        <tr>
-          <td>{{ splice_junctions_button(institute._id, case.display_name, variant._id) }}</td>
-        </tr>
-      {% endif %}
-    </table>
-  </div>
+    {{ alignments(institute, case, variant, current_user, case_groups, config) }}
```
northwestwitch, line 306: 👍🏻 much better now!

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -29,6 +31,42 @@
...
 {% endmacro %}
 
+{% macro alignments(institute, case, variant, current_user, case_groups, config) %}
+   <ul class="list-group">
+        <li class="list-group-item d-flex justify-content-between">
+          <div>
+            {{ igv_button(case, variant, case_groups) }}
+          </div>
+          <div>
+            {% if variant.alamut_link and config.SHOW_ALAMUT_LINK %}
```
northwestwitch, line 41: I wasn't thinking that a variant might also not have an alamut link. Isn't SHOW_ALAMUT_LINK superfluous then? 🤔 

Otherwise buttons look good, well aligned:
image

northwestwitch (approved):

scout/server/blueprints/variant/templates/variant/components.html
dnil: Ah, I thought you had already tried putting it in links and realised it was looking a bit lonely there. The thing is the link will be made if it can be made, regardless if the user has alamut or not. So the option is still good.

northwestwitch: Right! Then I think it's good to merge!

dnil: Meh, I just realised we do have a config for links: the external primer order. 🙄
Anyway, it is like two mentions. We can move it there later.

dnil: Oh, actually, if we do that: wouldn't it be cleaner to introduce a HIDE_ALAMUT_LINK? Then doing nothing will just produce the link as before. I´ll PR and be done with it while we are still on the same version..

northwestwitch: Agreed!

PR #2818, opened 2021-08-25
Rename hide alamut

dnil:
This PR continues on the last PR #2815, renaming the SHOW_ALAMUT_CONFIG introduced there to HIDE_ALAMUT_CONFIG and moving it to links py instead of the jinja code.

**How to test**:

Screenshot 2021-08-25 at 09 53 53

Screenshot 2021-08-25 at 09 54 02

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/docs/admin-guide/server.md
+++ b/docs/admin-guide/server.md
@@ -41,6 +41,15 @@ MAIL_PASSWORD = 'mySecretPassw0rd'
 
 TICKET_SYSTEM_EMAIL = 'support@sekvens.nu'
 
+## FEATURE FLAGS
+# Allow users to browse variants marked causative from among all their institutes cases
+SHOW_CAUSATIVES = True
+# Optionally show cards with legacy observed variant frequencies on variant pages
+SHOW_OBSERVED_VARIANT_ARCHIVE = True
+# Optionally hide API link to Alamut Visual. If False or non-existing Scout will show Alamut links.
+HIDE_ALAMUT_LINK = False
+
+
 # emails to send error log message to
 ADMINS = ['robin.andeer@scilifelab.se']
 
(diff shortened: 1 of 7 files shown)


scout/server/config.py
```diff
@@ -68,7 +68,7 @@
 # FEATURE FLAGS
 SHOW_CAUSATIVES = True
 SHOW_OBSERVED_VARIANT_ARCHIVE = True
-SHOW_ALAMUT_LINK = True
+HIDE_ALAMUT_LINK = False
```
northwestwitch, line 71: If you don't have this param at all then you see the Alamut button. Good so we don't have to do anything. In a future PR (not this one) we should document the new params added to the config file!

northwestwitch (approved):
👍🏻 Works nicely!

scout/server/config.py
dnil: Good point. I added a few lines to the admin doc. We can revisit that page later, but now there will at least be something there.

docs/admin-guide/server.md
```diff
@@ -41,6 +41,15 @@ MAIL_PASSWORD = 'mySecretPassw0rd'
 
 TICKET_SYSTEM_EMAIL = 'support@sekvens.nu'
 
+## FEATURE FLAGS
+# Allow users to browse variants marked causative from among all their institutes cases
+SHOW_CAUSATIVES = True
+# Optionally show cards with legacy observed variant frequencies on variant pages
+SHOW_OBSERVED_VARIANT_ARCHIVE = True
+# Optionally hide API link to Alamut Visual. If False or non-existing Scout will show Alamut links.
+HIDE_ALAMUT_LINK = False
```
northwestwitch, line 50: 💯

PR #2831, opened 2021-08-31
Fixed MitoMap and HmtVar links for hg38

mhkc:
This fixes that MitoMap and HmtVar links are being displayed when using hg38 genome build. The issue was accidentally introduced in v4.37.

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/utils.html
+++ b/scout/server/blueprints/variant/templates/variant/utils.html
@@ -377,7 +377,7 @@
           <input type="hidden" name="stop" value="{{ initial_view.end }}">
         {% endif %}
 
-        {% if variant.chromosome == "MT" and case.mt_bams %}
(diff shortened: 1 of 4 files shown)


scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -377,7 +377,7 @@ <h6>Phenotype terms (HPO):</h6>
           <input type="hidden" name="stop" value="{{ initial_view.end }}">
         {% endif %}
 
-        {% if variant.chromosome == "MT" and case.mt_bams %}
+        {% if variant.is_mitochondrial and case.mt_bams %}
```
northwestwitch, line 380: 👍🏻 Yes!!!

northwestwitch (approved):
Yes, that was actually a bug, thanks @mhkc!

PR #2832, opened 2021-08-31
Speed up SV variants - gene symbols dont need transcripts

dnil:
This PR refactors the updates to current HGNC gene symbol from `hgnc_id`.

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


dnil:
There is another one for the symbols just after - Ill see if I can't just fix that as well after lunch.

northwestwitch:
Testing it on stage, mmm doesn't look much faster than master. Actually a bit slower even..

northwestwitch:
And actually while I'm loading SV variants using the branch, transcripts are still queried, see the current query:

image

dnil:
its a bit faster, but transcripts are still pulled again for the symbols. Ill fix, but later. The same is true for the case page. Its is some time since we added those checks now, so this is no "smoking gun" for a sudden slow down, more trying to speed up what I can reproducibly find slow. Every time I check what the db is actually doing, it is busy getting transcripts, so removing some of those calls that are not needed seems like a good thing.

dnil:
Tested on scout-stage now: not very impressed by the speed in the end, but I guess it is at least more economical in some sense. Its a lot less bytes shuffled and transcript queries made. Still, they appear rather well indexed and perhaps not what limits the page load time.

scout/commands/view/hgnc.py
```diff
@@ -31,7 +31,7 @@ def hgnc(hgnc_symbol, hgnc_id, build):
         raise click.Abort()
 
     if hgnc_id:
-        result = adapter.hgnc_gene(hgnc_id, build=build)
+        result = adapter.hgnc_gene_caption(hgnc_id, build=build)
```
northwestwitch, line 34: 👍🏻 

image

scout/load/all.py
```diff
@@ -46,12 +46,12 @@ def load_region(adapter, case_id, hgnc_id=None, chrom=None, start=None, end=None
         raise ValueError("Case {} does not exist in database".format(case_id))
 
     if hgnc_id:
-        gene_obj = adapter.hgnc_gene(hgnc_id, case_obj["genome_build"])
-        if not gene_obj:
+        gene_caption = adapter.hgnc_gene_caption(hgnc_id, case_obj["genome_build"])
```
northwestwitch, line 49: 👍🏻

scout/parse/matchmaker.py
```diff
@@ -102,15 +102,15 @@ def genomic_features(store, case_obj, sample_name, genes_only):
...
-        hgnc_genes = var.get("hgnc_ids")
+        hgnc_gene_ids = var.get("hgnc_ids")
         # Looks like MatchMaker Exchange API accepts only variants that hit genes :(
-        if not hgnc_genes:
+        if not hgnc_gene_ids:
             continue
-        for hgnc_id in hgnc_genes:
-            gene_obj = store.hgnc_gene(hgnc_id, case_obj["genome_build"])
-            if not gene_obj:
+        for hgnc_id in hgnc_gene_ids:
```
northwestwitch, line 109: 👍🏻

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -58,11 +58,11 @@ def make_sashimi_tracks(institute_id, case_name, variant_id):
     # Collect locus coordinates. Take into account that variant can hit multiple genes
     variant_genes_ids = [gene["hgnc_id"] for gene in variant_obj.get("genes", [])]
     for gene_id in variant_genes_ids:
-        gene_obj = store.hgnc_gene(hgnc_identifier=gene_id, build=build)
-        if gene_obj is None:
+        gene_caption = store.hgnc_gene_caption(hgnc_identifier=gene_id, build=build)
```
northwestwitch, line 61: 👍🏻

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -666,7 +666,7 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-secondary">Pa
               <span class="float-left">
                 <strong>
                   {% if variant.cadd_score %}
-                    {{ variant.cadd_score|round(1) }}
+                    {{variant.cadd_score|round(1) }}
```
northwestwitch, line 669:
```suggestion
                    {{ variant.cadd_score|round(1) }}
```

northwestwitch: Symmetry!

scout/adapter/mongo/hgnc.py
```diff
@@ -43,6 +43,43 @@ def load_hgnc_bulk(self, gene_objs):
...
         return result
 
+    def hgnc_gene_caption(self, hgnc_identifier, build="37"):
+        """Fetch the current hgnc gene symbol and similar caption info for a gene in a lightweight dict
+        Avoid populating transcripts, exons etc that would be added on a full gene object. Use hgnc_gene() if
+        you need to use those.
+
+        Args:
+            hgnc_identifier(int)
+            build(str)
```
northwestwitch, line 53: What about defaulting to None and including it in the query only if it has a value in ["37","38"]? 
Because I'm looking at your proposed change in the function /scout/server/blueprints/institutes/views.gene_variants and there you don't pass any build (because I guess it could be both builds). But it won't work for build 38 cases because it will default to build 37.

You could perhaps pass nothing and not use build in the query? It will be only one result anyway because you are using find_one.

scout/server/blueprints/institutes/views.py
```diff
@@ -150,11 +150,11 @@ def gene_variants(institute_id):
     if (form.hgnc_symbols.data) and len(form.hgnc_symbols.data) > 0:
         for hgnc_symbol in form.hgnc_symbols.data:
             if hgnc_symbol.isdigit():
-                hgnc_gene = store.hgnc_gene(int(hgnc_symbol))
-                if hgnc_gene is None:
+                hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol))
```
northwestwitch, line 153: This way, don't you get only gene variants for genes in build 37? I've commented also above about this

scout/server/blueprints/variants/controllers.py
```diff
@@ -1126,11 +1126,11 @@ def update_form_hgnc_symbols(store, case_obj, form):
...
         for hgnc_symbol in form.hgnc_symbols.data:
             if hgnc_symbol.isdigit():
-                hgnc_gene = store.hgnc_gene(int(hgnc_symbol), genome_build)
-                if hgnc_gene is None:
+                hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol), genome_build)
+                if hgnc_gene_caption is None:
                     not_found_ids.append(hgnc_symbol)
                 else:
-                    hgnc_symbols.append(hgnc_gene["hgnc_symbol"])
+                    hgnc_symbols.append(hgnc_gene_caption.get("hgnc_symbol", hgnc_symbol))
```
northwestwitch, line 1133: Here also, you will default to build 37, so perhaps use the genome_build param?

northwestwitch (changes requested):
It's good to retrieve only the information one needs without triggering extra queries! 

The changes look safe and effective. I only have this doubt about defaulting to build 37 when the function might also be used for data in build 38, check my comments. That's why I'm requesting changes, but we can discuss about it!

scout/server/blueprints/institutes/views.py
dnil: Yes! It was the same before, but should be fixed.

scout/server/blueprints/variants/controllers.py
dnil: Already do: check L1129. This is just the dict.

scout/server/blueprints/institutes/views.py
dnil: Hm, ah, that is because it iss gene_variants, so variants could be from cases with either.

scout/server/blueprints/variants/controllers.py
northwestwitch: Ah yes, I misread, never mind!

scout/adapter/mongo/hgnc.py
dnil: Yep! It was the same before, and honestly mostly ppl use symbols rather than HGNC-ids in the search field, but you are right. Fixed below by asking for captions from 38 if none was found for digit in 37.

scout/server/blueprints/institutes/views.py
```diff
@@ -150,11 +150,15 @@ def gene_variants(institute_id):
     if (form.hgnc_symbols.data) and len(form.hgnc_symbols.data) > 0:
         for hgnc_symbol in form.hgnc_symbols.data:
             if hgnc_symbol.isdigit():
-                hgnc_gene = store.hgnc_gene(int(hgnc_symbol))
-                if hgnc_gene is None:
+                hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol), build="37")
+
+                if hgnc_gene_caption is None:
+                    hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol), build="38")
```
northwestwitch, line 156: 💯

northwestwitch (approved):
Thanks for fixing that gene_variants thing. I think it's good to merge now!

northwestwitch:
Nooooooo, I fixed it like 5 minutes ago!!! 😭 

image

PR #2830, opened 2021-08-31
Fix multigene cancer variant summary

northwestwitch:
At the moment cancer variant summary shows the info for a single gene even if the variant hits several genes (fix #2826 and partially fixing #2835).

My fix is perhaps naive since I'm removing most of the info obtained in the jinja parsing (see code) to leave only gene ID and gene name. But I was thinking that:
- Most (if not all) of the jinja parsed info is showed later in the page anyway
- That code is a bit hard to read and to maintain. And the choice of the primary transcript or NM transcript to show with associated exon is a bit arbitrary and could be misleading (see also #2835, first problem) . Perhaps a simpler solutions wound't be so bad..

**How to test**:
1. On clinical-db stage go to a cancer variant hitting multiple genes, esample: https://scout-stage.scilifelab.se/cust000/balsamic-validation-gmsmyeloid-tumor-only-2/cancer/ac4fa05770c2e462ed36d23fe673ea4e?cancer=yes
2. Master branch should show just one gene info in variant summary
3. Install this branch and 2 genes should be shown instead

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

Issue #2826, opened 2021-08-30
Problems in cancer scout after the last Scout uppdate


christacosta:
Hi, after the last Scout update a few things are not working as they should in Cancer Scout:

- [x]  Links till “matching causatives from other cases” are not working, e.g.
https://scout.scilifelab.se/cust087/F0041371/cancer/504a020f0068b44557720d5c04c4da0e?cancer=yes

- [x] In the variant page, the gene information is not about the gene in which the mutation was found, e.g. CSDE1 - NRAS
https://scout.scilifelab.se/cust087/F0041371/cancer/91bdff58f63b755cb9ac067040311edb?cancer=yes (in the list with variants it was correctly listed as variant in NRAS)

- [x]  Link to COSMIC is not to the correct variant, e.g.:
Example A: https://scout.scilifelab.se/cust087/F0041371/cancer/504a020f0068b44557720d5c04c4da0e?cancer=yes
There are 3 well-described point mutations at this position, but the link does not include the one that was found in this patient even though it is reported in COSMIC (https://cancer.sanger.ac.uk/cosmic/mutation/overview?id=42635557)

Example B:
https://scout.scilifelab.se/cust087/F0041371/cancer/6a2cb83e3f6bfbda297bb9625b55d58a?cancer=yes
COSMIC link to c.2601C>A instead of C>T --> non-synonymous change which is reported in hematological cases rather than the synonymous aa change that was found in the patient

northwestwitch:
Thanks both for reporting what looks like a bug, or several actually! We'll look into in ASAP!

northwestwitch:
I'll start working on the first problem, in case someone else wants to help! 😃

northwestwitch:
Checking the Cosmic link problem now. At the database level there are 3 cosmic IDs saved for the variant:

![image](https://user-images.githubusercontent.com/28093618/131338936-2fdbbdea-43db-44ff-9456-e8e9b6b8d692.png)

So it's not the VCF parsing, must be the rendering on the web page, I'll fix it!

northwestwitch:
You were right @hassanfa: Scout only showed the link to the first cosmic ID from the list of cosmic IDs associated to the variant. This was not introduced by the latest release, it has been like this all the time. This is now fixed by #2828.

northwestwitch:
This is also fixed, closing!

dnil:
[...]

PR #2840, opened 2021-09-01
Update representative gene - Partly fix #2835 and #2826

### LINKED ISSUE #2835: Scout cancer interface problems (partly duplicated)
I'm copying the text from a customer ticket by Emma Tham:

Dear Scout developers,

I have used the new SCOUT layout for the first time today and have a few issues I would like to give you feedback on.

- [ ] Only one variant is annotated in the summary and it is not always the relevant variant. We are usually interested in the coding variants. See eg pat KI/Brehmer account 091: 37-DS-a, variant [37]1:g.45805919G→C where the non-coding TESK2 is listed, while the MUTYH variant which is the coding variant is not (and I need to scroll down to find it). 

I suggest that if it is not possible to select the coding one, that all variants are listed in the summary.

- [x] The ClinVar database (which is vital for our interpretation) should be clearly visible in the summary with the classification. See eg [37]3:g.178938934G→A in PIK3CA (same patient as above). This variant is a known pathogenic variant and this information should be clearly visible in the summary (it is now only visible on the first page).

- [x]  I would like to be able to dismiss variants from the first page, The functionality seems to be there, but is not active for some reason. This means that I have to open up each variant in a separate window/tab in order to dismiss it even when I can see from the first page that it is not relevant. 

I hope that these suggestions are possible?!

Thank you in advance,

Best regards

Emma

dnil:

This is an alternative to #2830 that instead retains the idea of only showing one gene symbol in summary, using the same logic as in variantS list (most severe SO_TERM consequence). This should be understandable to most users: "Why is this weird research gene with a stopgain in my list? Ah, ok, it is also a synonymous variant in a gene in my panel. That's fine, moving on."

I agree the jinja is getting a bit hard to maintain there. We could consider moving the "primary transcipt" logic into the controller, but I don't do it here.

**How to test**:
1. compare #2830 - note that the same gene as in the variantS list should be selected as representative for the summary

Screenshot 2021-09-02 at 09 44 49

**Expected outcome**:

**Review:**
- [x] code approved by CR

### DIFF
--- a/scout/server/blueprints/variant/utils.py
+++ b/scout/server/blueprints/variant/utils.py
@@ -1,6 +1,6 @@
 import logging
 
-from scout.constants import ACMG_COMPLETE_MAP, CALLERS, CLINSIG_MAP
+from scout.constants import ACMG_COMPLETE_MAP, CALLERS, CLINSIG_MAP, SO_TERMS
 from scout.server.links import add_gene_links, add_tx_links, ensembl
 
 LOG = logging.getLogger(__name__)
@@ -51,6 +51,27 @@ def add_panel_specific_gene_info(panel_info):
 
     return panel_specific
 
+def update_representative_gene(variant_obj, variant_genes):
+    """ Set the gene with most severe consequence as being representative
+    Used for display purposes
+
+    Args:
+        variant_obj(Variant): a variant object
+        variant_genes(list(Genes): a list of genes
+    """
+    if variant_genes:
+        first_rep_gene = min(
+            variant_genes, key=lambda gn: SO_TERMS[gn["functional_annotation"]]["rank"]
+        )
+        # get HGVNp identifier from the canonical transcript
+        hgvsp_identifier = None
+        for tc in first_rep_gene["transcripts"]:
+            if tc["is_canonical"]:
+                hgvsp_identifier = tc.get("protein_sequence_name")
+        first_rep_gene["hgvsp_identifier"] = hgvsp_identifier
+        variant_obj["first_rep_gene"] = first_rep_gene
+    else:
+        variant_obj["first_rep_gene"] = None
 
 def update_transcripts_information(variant_gene, hgnc_gene, variant_obj, genome_build=None):
     """Collect tx info from the hgnc gene and panels and update variant transcripts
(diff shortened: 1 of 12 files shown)


dnil:
Right, the representative could be without NM_ as well. After lunch..

northwestwitch:
Testing this on stage with a multi-gene variant, for instance this one:

https://scout-stage.scilifelab.se/cust000/balsamic-validation-gmsmyeloid-tumor-only-2/cancer/ac4fa05770c2e462ed36d23fe673ea4e?cancer=yes

On my branch (#2830) it would look like this:
image

On this branch:

image


I think it works as it's supposed to, bit I honestly don't know if this is what a user might want to see, can't we ask some real user?

The branch is installed on stage now.  

One thing I would add is tooltips explaining a bit that you are seeing the most representative gene instead of all genes. And also a tooltip on the transcript explaining why that tx.

scout/server/blueprints/variant/utils.py
```diff
@@ -52,6 +52,29 @@ def add_panel_specific_gene_info(panel_info):
     return panel_specific
 
 
+def update_representative_gene(variant_obj, variant_genes):
```
northwestwitch, line 55: Nice with the shared code!

northwestwitch (approved):
It works fine, but I'd add tooltips and ask for feedback before merging!

dnil:
I'm convinced this is what users believe they want to see, though there are definitely edge cases lurking, like say the variant is a known pathogenic regulatory variant in a gene in your panel, but also affects another protein with no disease linked. These are however not abundant, we have tooltips, and all the info is actually in the page after all.

dnil:
Your test variant is actually touching such an edge case for cancer, where the ESR variant, while uninteresting, is the one that the user should evaluate first here, before possibly considering the secondary finding from the SYNE2 change.

dnil:
But note that both users that commented actually want exactly this: ˜"the gene on variantS is right" and ˜"we usually pick the one with the most severe impact". 😸

northwestwitch:
> But note that both users that commented actually want exactly this: ˜"the gene on variantS is right" and ˜"we usually pick the one with the most severe impact". 😸

Ok!

northwestwitch:
@dnil is this ready for merging or do we want to test it more on stage?

dnil:
It's always good to test a bit more. I'll put it on stage.

dnil:
Ok, looks good. I'll merge when the tests finish!

PR #2850, opened 2021-09-03
Reintroduce missing splice junctions button on variant page

### LINKED ISSUE #2847: RNA data not visible in Scout
In a meeting with Anders and Henrik today, it became apparent that we're supposed to be able to view RNA data in Scout. But we cannot. Here are three example families where RNA data exist and apparently should be visible for us:
families 52, 14022, 17172

individual ID for the DNA samples: 52-1-2A, 52-1-4A, 52-2-1U, 52-2-2U; 14022-I-1A, 14022-I-2A, 14022-II-1U, 14022-II-2U; 17172-I-1A
DNA from 17172 has been sequenced twice, from blood and from muscle, but the RNA data comes from blood so I'm guessing it will be linked to family 17172 and not 17172-muscle.

Happy bug hunting ;) Thank you!

northwestwitch:
Fix #2847 

**How to test**:
1. On a local master, note that splice junctions button is not visible on variant page
2. Switch to this branch and the button re-appears

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/components.html
+++ b/scout/server/blueprints/variant/templates/variant/components.html
@@ -62,28 +62,32 @@
   </li>
 {% endmacro %}
 
-{% macro alignments(institute, case, variant, current_user, case_groups, config) %}
+{% macro alignments(institute, case, variant, current_user, case_groups, config, splice_junctions_tracks=False) %}
    <ul class="list-group">
         <li class="list-group-item d-flex justify-content-between">
           <div>
             {{ igv_button(case, variant, case_groups) }}
           </div>
-          <div>
-            {% if variant.alamut_link %}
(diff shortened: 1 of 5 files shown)


scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -62,40 +62,39 @@
   </li>
 {% endmacro %}
 
-{% macro alignments(institute, case, variant, current_user, case_groups, config) %}
+{% macro alignments(institute, case, variant, current_user, case_groups, config, splice_junctions_tracks=False) %}
```
dnil, line 65: 👍

dnil (approved):

PR #2853, opened 2021-09-03
Minor version 4.39

northwestwitch:
### Fixed
- [x] Link to other causative variants on variant page
- [x] Allow multiple COSMIC links for a cancer variant
- [ ] <del>Fixed MitoMap and HmtVar links for hg38 cases --Tested by collaborators-- </del>
- [x] Do not open new browser tabs when downloading files
- [x] Selectable IGV tracks on variant page
- [x] Missing splice junctions button on variant page
- [x] Refactor variantS representative gene selection, and use it also for cancer variant summary
- [x] Missing COSMIC IDs collected from CSQ field named `COSMIC`
### Changed
- [x] Improve Javascript performance for displaying Chromograph images
- [x] Make ClinVar classification more evident in cancer variant page

**Expected outcome**:
The functionalities should be working

**Review:**
- [ ] tests executed by


northwestwitch:
I'll install it on stage!

northwestwitch:
When you download a case report on pdf format it might take quite some time and with the latest changes it looks like nothing is happening when you click on the link. But then, after quite some time, you see the downloaded file. I don't know if this might be perceived as a problem..

dnil:
> When you download a case report in pdf format it might take quite some time and with the latest changes it looks like nothing is happening when you click on the link. But then, after quite some time, you see the downloaded file. I don't know if this might be perceived as a problem..

Well, it is not worse that before really. We could always add something later if that is so. It seems better invested time to make the report faster - perhaps that will just happen with the react version! 😸

dnil:
And to keep track, I was not able to see the RNA junctions track on stage (e.g. https://scout-stage.scilifelab.se/cust003/14022/e4f6bda7d7150a2c455ee12e6a41e2c9 should have it - at least it has the files loaded and readable). Possibly an issue with mangling of booleans to text when passed through macros?

northwestwitch:
Cust003 case 52 should also Have splice junctions data

northwestwitch:
Ok in order to show the splice junctions button, a case individual should have BOTH these 2 keys/values: splice_junctions_bed, rna_coverage_bigwig, see [code](https://github.com/Clinical-Genomics/scout/blob/f0cbeb6ab8f2777fd247328fe83fdc56f6f6dcd9/scout/server/blueprints/variant/controllers.py#L59).

Case 52 doesn't have them (checked in the database), 14002 has them but you are right, bo splice junctions. Something's wrong somewhere!

northwestwitch:
All good, I didn't install the right branch before (😄)  it works!

dnil:
For remembering next week: loading new cancer cases is only locally tested, so we will not update prod loading this time around.

dnil (approved):

PR #2856, opened 2021-09-06
Stop updating database indexes after loading exons via command line

### LINKED ISSUE #2852: Load exons command updates scout indexes
Check here: https://github.com/Clinical-Genomics/scout/blob/af832961880d8531ac899ecbf8d47670ee5ab778/scout/commands/load/exons.py#L57

This should also be removed. We want to trigger it manually and not as part of another command

northwestwitch:
Fix #2852. We don't want indexes being automatically updated after another cli command. We want to have control on when they should be updated.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/commands/load/exons.py
+++ b/scout/commands/load/exons.py
@@ -54,6 +54,4 @@ def exons(build, exons_file):
         LOG.info("Please download a mart dump manually, see instructions in user guide for admins")
         return
 
-    adapter.update_indexes()
-
     LOG.info("Time to load exons: {0}".format(datetime.now() - start))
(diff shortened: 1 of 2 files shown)


scout/commands/load/exons.py
```diff
@@ -54,6 +54,4 @@ def exons(build, exons_file):
         LOG.info("Please download a mart dump manually, see instructions in user guide for admins")
         return
 
-    adapter.update_indexes()
```
dnil, line 57: 👍

dnil (approved):

PR #2864, opened 2021-09-09
Assign OMIM diagnoses to individuals or the entire case

northwestwitch:
- Assign individual OMIM diagnoses:
image

- Improve OMIM terms description in ClinVar submission form:
image

This PR is slightly more complicated than the HPO on individuals one, because current case["diagnosis_phenotypes"] looks like this (a list of integers):

image

To be able to assign OMIM terms to individuals, when a case is open (any case) the format of diagnosis_phenotypes is modified like this (reflects a similar structure of case HPO terms):

image

**How to test**:
1. On master branch, clinical-db stage, assign one or more diagnoses to a case
2. Switch to this branch and make sure the case diagnoses are still visible (and can be removed)
3. Assign one more more diagnoses to one or more individuals of the case
4. Test that the new diagnoses can be removed from the case
5. Pin a variant and add it to a ClinVar submission. Make sure that OMIM diagnoses contain a description and not just the OMIM number in ClinVar submission.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
The conflict in this PR is now fixed. Once this is merged we could start working on assigning individual-level phenotypes using phenotype models.

tests/adapter/conftest.py
```diff
@@ -65,3 +65,17 @@ def parsed_gene():
         "build": "37",
     }
     return gene_info
+
+
+@pytest.fixture
+def omim_term():
```
dnil, line 71: 👍

northwestwitch:
> Great! This will be useful to keep track of disease labels in more complex families. I tested by clicking around a bit; seems to be fully operational. The only thing is that it will not be backwards compatible in its current form, but we seldom revert.

No, unfortunately the current form is structured in a way that doesn't allow much flexibility. Also, it's different from the code we use to handle HPO terms. This PR changes the format of saved OMIM terms in new cases (and old cases when OMIM terms are modified). Searched by OMIM terms return results in both formats.

PR #2876, opened 2021-09-14
Move tx overview logic to backend

### LINKED ISSUE #2862: MANE transcript fix - inconsistent behaviour
In https://github.com/Clinical-Genomics/scout/pull/2683#issue-658334675 , the aim was to fix three things with MANE annotation: MANE Select and Plus Clinical should get green badges, version number should be added to MANE Select, and finally MANE Select should be shown first in the list of transcripts. From my experience with the Lund installation (v4.38), the first two objectives are solved, but the third item is inconsistent. Sometimes the MANE Select transcript is indeed shown first, but in other instances it is further down the list (see attached screenshot). I haven't been able to find a pattern for this inconsistency.
MANE Select

And while I am on the subject, it would be good to add version numbers also to the (few) MANE Plus Clinical transcripts, since that is in line with how MANE is supposed to be.

northwestwitch:
This PR:
- moves a lot of code preparing the transcripts overview panel from jinja to the backend
- Sorts displayed transcripts by "mane_select" and "mane plus" availability (fix #2862)

**How to test**:
1. Install the branch and notice that tx overview on variant page shows the same info as in master branch
2. Check that MANE transcripts are shown on top of tx overview. To do so, open the POT1 variant in mongoDB compass (_id:"4c7d5c70d955875504db72ef8e1abe77"), open its genes and the second transcript (the one that is shown as second tx in the scout variant page) and add a mane_select_transcript value as illustrated in this pic:
image
3. Switch to master and this transcript is shown second in tx overview
4. In this branch it will be shown on top

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Transcript overview on master branch:
image


On this branch:
image

scout/server/blueprints/variant/controllers.py
```diff
@@ -45,6 +45,75 @@
...
+
+            for refseq_id in tx.get("refseq_identifiers", []):
+                decorated_tx = None
+                if ovw_tx["mane"] and ovw_tx["mane"].startswith(refseq_id):
+                    decorated_tx = ovw_tx["mane"]
+                elif ovw_tx["mane_plus"] and ovw_tx["mane_plus"].starstwith(refseq_id):
+                    decorated_tx = ovw_tx["mane_plus"]
+                elif refseq_id.startswith("XM"):
+                    decorated_tx = "".join(
+                        ['<font class="text-muted font-italic">', refseq_id, "</font>"]
```
dnil, line 84: How about flagging it as muted transcripts and defer the styling of it to the template, much like those flagged mane etc?

scout/server/blueprints/variant/controllers.py
```diff
@@ -45,6 +45,75 @@
 LOG = logging.getLogger(__name__)
 
 
+def tx_overview(variant_obj):
+    """Prepares the content of the transcript overview to be shown on variant and general report pages.
```
dnil, line 49: 👍

scout/server/blueprints/variant/templates/variant/tx_overview.html
```diff
@@ -42,91 +42,67 @@
...
-                    {% if transcript.mycancergenome_link %}
-                      <a href="{{ transcript.mycancergenome_link }}" class="badge badge-secondary text-white" target="_blank"
-                        data-toggle="tooltip" title="MyCancerGenome">MCG</a>
-                    {% endif %}
-                  </td> <!-- end of Links col -->
-                </tr>
-              {% endif %} <!-- end of { if "refseq_identifiers" in  transcript } -->
-            {% endfor %} <!-- end of { for transcript in gene.transcripts } loop -->
-          {% endfor %} <!--end of { for gene in variant.genes } loop -->
+          {% for tx in variant.overview_transcripts %}
```
dnil, line 45: 💯

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -10,13 +10,44 @@
...
+            "coding_sequence_name",
+            "protein_sequence_name",
+            "varsome_link",
+            "tp53_link",
+            "cbioportal_link",
+            "mycancergenome_link",
+            "mane",
+            "mane_plus",
+        ]:
+            assert key in tx
```
dnil, line 48: 👍

dnil (approved):
Nice! That was way much logic in the template. Only one suggestion: marking of overview transcripts as uninteresting, rather than produce a html string for them, to keep things systematic.

scout/server/blueprints/variant/controllers.py
northwestwitch: That can certainly be done, good suggestion!

northwestwitch:
Tested after latest changes:
image

Issue #2742, opened 2021-07-01
Update nginx CORS policy


elevu:
Update the nginx CORS policy for scout in **staging** to allow HTTP requests from **localhost** and ** https://scout-stage.web.app**
Also when running Scout in docker, it should be possible to call the endpoints from localhost:3000

elevu:
Note - OPTIONS calls are returning 302 when done from different origin

elevu:
Wait with this task until this is done https://github.com/Clinical-Genomics/scout-react/issues/52

elevu:
inspo: https://stackoverflow.com/questions/22181384/javascript-no-access-control-allow-origin-header-is-present-on-the-requested

northwestwitch:
After brainstorming meeting we understood that this problem is probably caused by the fact that the user is not authenticated in the scout-react demo setup. The easiest was to solve this would be to create a new endpoint in a scout view that would serve the scout-react app page as if it's coming from Scout. Basically in the same way as Scout is serving pages coming from Chanjo-report.

elevu:
https://scout-stage.web.app/cust000/phenotypes this page can be used for testing, it is calling https://scout-stage.scilifelab.se/. To see the error look into the console :)

dnil:
Ok, closing this with #2877. It activates CORS header answers from Flask directly from Scout, instead of adding them in NGINX. We may or may not want to keep it that way in the end, but react development is again enabled.

PR #2877, opened 2021-09-14
Pheno search api - Fix #2749

### LINKED ISSUE #2749: Search endpoint for phenotypes page
The call `GET api/v1/phenotypes` is now returning a list of thousands of phenotypes. It would be better to just have a endpoint like `GET api/v1/phenotypes?search=hemoglobin` that returns a searched term

dnil:

Add an HPO term search endpoint as 
/api/v1/phenotype/search/<QUERY_STRING>

As a bonus, adds CORS header support for js apps via Flask-CORS.

**How to test**:
1. test a search with an hpo term against the api, and notice hpo terms being returned, e.g. `http://localhost:5000/api/v1/phenotypes/search/kidney`
Screenshot 2021-09-14 at 18 04 13
Also note that CORS sensitive parts of Scout are still working as before, e.g. IGV viewer with external tracks.

2. test with a scout-react install and note how CORS errors disappear when this branch is used.

Before:
Screenshot 2021-09-14 at 16 40 22

After:
Screenshot 2021-09-14 at 17 20 20

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN


scout/server/blueprints/phenotypes/views.py
```diff
@@ -28,3 +28,16 @@ def api_hpo_terms():
 
     data = controllers.hpo_terms(store=store)
     return jsonify(data)
+
+
+@hpo_bp.route("/api/v1/phenotypes/search/<search_string>", methods=["GET"])
```
northwestwitch, line 33: Where do you see in the react page that the address will become like this?

scout/server/blueprints/phenotypes/views.py
```diff
@@ -28,3 +28,16 @@ def api_hpo_terms():
...
+@public_endpoint
+def api_hpo_term_search(search_string):
+    """Public API for HPO phenotype term: search for a HPO terms matching a string
+
+    Returns:
+        data(dict): result dict with key "phenotypes" set to an array of all matching phenotype terms
+    """
+
+    data = controllers.hpo_terms(store=store, query=search_string)
```
northwestwitch, line 42: The search doesn't work but the CORS error in the html disappears:

image

scout/server/app.py
```diff
@@ -48,6 +49,8 @@
 def create_app(config_file=None, config=None):
     """Flask app factory function."""
     app = Flask(__name__)
+    CORS(app)
```
northwestwitch, line 52: 👍🏻

northwestwitch (commented):
Given that this branch is now installed on clinical-db, there is some error in the page and no phenotypes are returned (therefore I am not approving right away). The CORS fix should be working but it would be nice to see a working example. Where does the route "/api/v1/phenotypes/search/<search_string>" come from?

scout/server/blueprints/phenotypes/views.py
northwestwitch: I mean, shouldn't this be working with @elevu's example phenotype page?

dnil: I dont think so yet. It wasn't there to call before, and I used endpoint request syntax instead of a get url '?' argument.

dnil: The route is new, and can be seen working by typing it into a browser (compare picture above). 😊

dnil: I assumed the page was a work in progress, but can have a look there as well. There were som DOM container errors as well that should not be related to a search returning results or not.

northwestwitch: Yes I noticed the DOM errors. I am asking because I don't know if @elevu and @moedarrah have a specific idea in mind on how to handle the phenotype search in their phenotypes page. Might also be that they retrieve all phenotypes and then filter on the fly using some js code instead of this type of search?

dnil: Nah, then there would be no need for a new endpoint. The old one would have been sufficient if the js did some matching. There is also evident use if a search url, but I dont quite yet understand how it is supposed to switch between mock and live stage data. There are some hardcoded server urls there that I suppose are for testing:
https://github.com/Clinical-Genomics/scout-react/blob/2209e9800813c00d38a3fab72d80a55147f74d2d/src/services/ScoutApi.ts#L73

northwestwitch: But I guess providing a query option is always good to have, I am not really concerned about the query. But it would be nice to see that the CORS fix actually work with the react app before merging..

dnil: Good point. I'll see what I can do: if I read that correctly, the setPhenotypes() function is still missing? 
The DOM errors are by the way unrelated - they appear on any page with the top menu, likely just a missing div id tag somewhere.

dnil: For the record, it was there, from a little js hook `useState`, but the search function had been temporarily repurposed for other testing.

elevu: we asked for the Search endpoint because GET /phenotypes returns thousands of objects, it is too slow even on the current Scout.
This new endpoint works perfectly on the React app, thank you!

elevu: https://scout-stage.web.app/cust000/phenotypes some change is needed on the UI but the table with phenotypes works!

northwestwitch: Nice, we can merge this PR then!

northwestwitch (approved):

northwestwitch:
Fix #2742. You were right from the beginning @elevu , it was the CORS, sorry we didn't get sooner!

dnil:
Indeed. Here is a branch based off @elevu's CORS-fix on Scout-react where the phenotype search is working, when used against this branch.
https://github.com/Clinical-Genomics/scout-react/tree/fix-cors-use-search-table
Screenshot 2021-09-15 at 11 32 00

PR #2752, opened 2021-07-07
Scout-chanjo development docker-compose integration -- WORK IN PROGRESS --

northwestwitch:
A docker-compose integration containing images of chanio, chanio-report and scout

**How to test**:
1. Move the docker-compose file under scout/
2. Edit the Scout config file as described in docker-compose file
3. Launch the docker-compose with: `docker-compose -f docker-compose-chanjo-report.yml up`
 
**Expected outcome**:
The functionality should be working -- **IT'S NOT, BUT I COULD USE SOME HELP TO FIX IT**! --
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #2878, opened 2021-09-15
Development docker-compose file with scout/chanjo integration

northwestwitch:
This PR adds a development docker-compose file that integrates Scout and Chanjo-report

**How to test**:
- [ ] Switch to this branch on your local computer
- [ ] Follow the instructions at the top of the new docker-compose file
- [ ] Go to the browser: localhost:5000 and demo case
- [ ] On the side bar there should be the "coverage" report available, click on it and it should open a window with a report not containing samples. That's because the demo samples loaded with chanjo (sample1, sample2 and sample3) have different names from those in scout (ADM1059A1, ADM1059A2, ADM1059A3). This link actually works: http://localhost:5000/reports/report?sample_id=sample1&sample_id=sample2&sample_id=sample3&level=10&panel_name=Test+panel+%281.0%29. I think demo files should ideally be fixed in chanjo and not in scout. This PR is just for showing that the setup works.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Report when clicking from the case page:

image


Report after adjusting sample names:
image

northwestwitch:
It took me some a while to understand why Scout installed via conda requires chanjo-report installed manually on the side, while the docker-compose from this PR just runs it (notice I'm just using chanjo to populate the MariaDB database in the docker-compose, but I haven't included or installed a chanjo-report service/package in there). Then it occurred to me:

chanjo-report is installed in the extras_require in the [scout setup](https://github.com/Clinical-Genomics/scout/blob/6ac4790fe81e9e0e5c321070950cbd2c77ec187f/setup.py#L72), which means it is installed if some other package requires it. This doesn't happen when you install scout in a conda env.

Scout Dockerfile has been recently (#2752) modified with this line https://github.com/Clinical-Genomics/scout/blob/6ac4790fe81e9e0e5c321070950cbd2c77ec187f/Dockerfile#L29 and it's installing chanjo-report directly when scout is installed.

While this is convenient because with Docker you skip installing chanjo and chanjo report, these 2 packages are not strictly part of Scout so it would make sense to keep them separate. Otherwise why don't we just add them to the scout requirements.txt since they're going to be installed always anyway?  

Any suggestions @dnil @Mropat @henrikstranneheim ?

northwestwitch:
Basically my previous comment was to day that we don't need to transfer chanhio-report to the virtual farm if we use the current Scout Dockerfile

Mropat:
Docker-compose will not work with podman on CGVS. However if chanjo-report can be installed directly as a requirement of scout, or even moved into scout codebase, this could work and make things easier!

northwestwitch:
> Docker-compose will not work with podman on CGVS

Yes I know, this is just a demo. Let's keep the Dockerfile as it is then!

northwestwitch:
> Should this be merged into master? Since it works

Yes, could you approve the PR?

Mropat (approved):
:star2:

PR #2869, opened 2021-09-10
Feature configparse redux

northwestwitch:
This PR contains the same code as #2496, which I've merged to master by mistake (and then reverted) today. Sorry @mikaell !!  Fix #2496

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by CR
- [x] tests executed by CR, ML

PR #2880, opened 2021-09-16
Images in variant page

mhkc:
This PR adds the ability to add custom images to STR variant page by associating an image with the HGNC id.

image

**How to test**:
1. Load the demo case 643594 into the database.
2. Verify that STRs with repeat locus AFF2 has a new panel containing the custom images

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


mhkc:
I cant debug the tests on my computer for some reason because  i am getting an unrelated error when mocking the mongo database.

`pymongo.errors.OperationFailure: Index already exists with a different name: build_hgnc_symbol`

Could you please help me fix the failing test?

scout/build/case.py
```diff
@@ -255,7 +255,7 @@ def build_case(case_data, adapter):
     case_obj["chromograph_image_files"] = case_data.get("chromograph_image_files")
     case_obj["chromograph_prefixes"] = case_data.get("chromograph_prefixes")
 
-    case_obj["custom_images"] = case_data.get("custom_images", {})
+    case_obj["custom_images"] = case_data.get("custom_images", {'case': {},})['case']
```
dnil, line 258: 👍

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -171,6 +178,38 @@
 {% endblock %}
 
 
+{% macro panel_custom_images(images) %}
+  <div id="attached-img-panel" class="card mt-3 panel-default card">
+    <div class="panel-heading">
+      <div
+        class="collapsed"
```
dnil, line 185: 👍

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -378,6 +417,15 @@
   <script src="//cdn.datatables.net/1.10.20/js/jquery.dataTables.min.js"></script>
 
   <script>
+    const flipArrowIcon = (element) => {
```
dnil, line 420: 😸 👍

scout/server/blueprints/variant/controllers.py
```diff
@@ -346,6 +347,11 @@ def variant(
             **CANCER_SPECIFIC_VARIANT_DISMISS_OPTIONS,
         }
 
+    # re-encode images as base64
+    if variant_obj.get("custom_images"):
+        for img in variant_obj["custom_images"]:
+            img["data"] = b64encode(img["data"]).decode("utf-8")
```
dnil, line 353: 👍

dnil (approved):
Ill approve since this seems to be working fine, but before merging, think once more about the hgnc_symbols. I think what you mean is REPID, and that is also what you match against on the variant. If you indeed mean HGNC, I would strongly suggest using HGNCId instead, to avoid any later issues with changing hgnc symbols.

dnil:
> I cant debug the tests on my computer for some reason because i am getting an unrelated error when mocking the mongo database.
> 
> `pymongo.errors.OperationFailure: Index already exists with a different name: build_hgnc_symbol`
> 
> Could you please help me fix the failing test?

Hehe, I missed that it is not passing. 🙄 Ill check!

dnil:
Right, I think that was the last test, but lets see - I've been wrong before. 😸

Sorry this took some time: its both a bit too complex functions. When we have a calm couple of days, refactoring loading, setup and test setup to work identically would be a blessing. I took a couple of liberties when fixing the tests. 

For one I changed the style of the previous button in the str variant page image accordion to look more like the other card titles. The text wasn't really visible as it was before. I also keep both the decoded case and str images a little longer, partly because it was practical for the tests, partly because Im lazy and it was easier to think about it if the dict looked the same a little longer.

Still, do check that repid vs hgnc_symbol above. I'm sure it would not fit our setup here, and I suspect the same will be true for you with variants in genes with more than one significant locus, like say ARX or HOXA13.

dnil:
And for the record: tested locally.
Screenshot 2021-09-17 at 16 54 49

scout/adapter/mongo/case.py
```diff
@@ -631,18 +631,28 @@ def load_case(self, config_data, update=False, keep_actions=True):
 
                 variant_type = vcf_file["variant_type"]
                 category = vcf_file["category"]
+                # remove old variants if performing update
```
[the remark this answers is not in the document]
dnil: Like so - I just wanted somthing to retrigger the tests.

scout/adapter/mongo/variant_loader.py
```diff
@@ -474,6 +475,15 @@ def _load_variants(
                     # We need to have a max size of the bulk
                     if len(bulk) > 10000:
                         load = True
+                # Associate variant with image
+                if custom_images:
+                    images = [
+                        img
+                        for img in custom_images
+                        if img["hgnc_symbol"] == variant_obj["str_repid"]
```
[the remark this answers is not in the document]
mhkc: good catch. Ill change it to str_repid

dnil:
Ping @mhkc - what about those extra like bed files? The black stuff is fine, but would be slightly more elegant not to include changes to files you didn't really intend to change. But don't worry about it or overwork it - it happens once in a while. I had a broken black on one env since shortly after the last release as well; had to reinstall one of its deps to get it working again. 🙄

mikaell:
I'm adapting this change to work with Pydantic config parsing. It shall work when we merge, but try to avoid editing the config parsing before #2869 is merged 🙂

PR #2893, opened 2021-09-23
Removed padding after punctuation in comments

### LINKED ISSUE #2891: Comment field inserts spaces after punctuation marks
The comment field inserts a space after a punctuation marks. This causes issues when writing gene nomenclature and decimal numbers which can cause the user to misinterpret the results.

Eg: `foo!doo?moo.crow,flow` => `foo! doo? moo. crow, flow`

The padding seems to be applied when rendering the template as the comment is not padded in the database.

mhkc:
This removes the automatic insertion of a space after a punctuation mark in comments. (fix #2891)

**How to test**:
1.  write the following comment "foo!doo?moo.crow,flow"

**Expected outcome**:
The displayed comment should be "foo!doo?moo.crow,flow"

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -197,11 +197,6 @@ def register_filters(app):
         return cosmicId
 
     @app.template_filter()
-    def fix_punctuation(text):
-        """Adds a white space after puntuation"""
-        return re.sub(r"(?<=[.,:;?!])(?=[^\s])", r" ", text)
-
-    @app.template_filter()
     def count_cursor(pymongo_cursor):
         """Count numer of returned documents (deprecated pymongo.cursor.count())"""
         # Perform operations on a copy of the cursor so original does not move
--- a/scout/server/blueprints/cases/templates/cases/case_report.html
+++ b/scout/server/blueprints/cases/templates/cases/case_report.html
@@ -1159,7 +1159,7 @@
           {% for comment in variant.comments %}
             <tr>
               <td style="width:50%;">{{comment.created_at.strftime('%Y-%m-%d')}} {{comment.user_name}}</td>
-              <td class="text-break"><strong>{{ comment.content | fix_punctuation}}</strong></td>
+              <td class="text-break"><strong>{{ comment.content }}</strong></td>
             </tr>
           {% endfor %}
         </tbody>
@@ -1327,7 +1327,7 @@
           {% for comment in variant.comments %}
             <tr>
               <td style="width:50%;">{{comment.created_at.strftime('%Y-%m-%d')}} {{comment.user_name}}</td>
-              <td class="text-break"><strong>{{ comment.content | fix_punctuation}}</strong></td>
+              <td class="text-break"><strong>{{ comment.content }}</strong></td>
             </tr>
           {% endfor %}
         </tbody>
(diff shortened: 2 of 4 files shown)


northwestwitch:
Mmm I remember fixing it like this some time ago, see my PR: https://github.com/Clinical-Genomics/scout/pull/2455

The problem was that some very long strings of comment with no space were breaking the views

scout/server/templates/utils.html
```diff
@@ -42,7 +42,7 @@ <h6 class="comments-title ml-3 mt-2"></i><i class="far fa-comments"></i> Comment
     <style>.odd { background-color: #d5e8f0 }</style>
     {% for comment in comments %}
       <div class="row ml-1 alert {% if loop.index0 % 2 %}even{% else %}odd{% endif %}">
-          <div class="col-7 text-break">{{ comment.content|fix_punctuation }}</div>
```
northwestwitch, line 45: That function was introduced to prevent [a bug](https://github.com/Clinical-Genomics/scout/issues/2451). It inserts a space after .,:;?!
Perhaps we could remove the space after the dot instead?

northwestwitch (commented):
See my comment. We have several cases or very long comments with no spaces that break either case page, variant page or case report page. For this reason we need to break these comments somehow. But I understand this might be a problem when the punctuation with no space is needed. One could avoid the spaces after numbers or dots in general?

scout/server/templates/utils.html
northwestwitch: Bah perhaps the idea was that with spaces the comments would be more readable instead of breaking every 50 or whatever chars

mhkc:
I have tried the example comment from the merged PR and I dont see the issue.

image

image

image

northwestwitch:
That's right! I guess we fixed also the behavior of the div to split the comment into new lines. Reproduced also by my side:

image


image

image


I guess then it's fine to merge!

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -199,7 +199,7 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 	                    {% for comment in comments %}
 	                      <tr>
 	                        <td style="width:50%;">{{comment.created_at.strftime('%Y-%m-%d')}} {{comment.user_name}}</td>
-	                        <td class="text-break"><strong>{{ comment.content | fix_punctuation }}</strong></td>
+	                        <td class="text-break"><strong>{{ comment.content }}</strong></td>
```
northwestwitch, line 202: 👍🏻

northwestwitch (changes requested):
Looks like fix_punctuation is not really needed and the class="text-break" handles these long strings nicely! 👍🏻 

I did a quick check and looks like there's a couple more `{{ comment.content | fix_punctuation }}` in the case report template.

And since it will be no longer used we could also remove the original fix_punctuation function in the app.py file?

mhkc:
I have removed the function. Feel free to merge when you have the time.

scout/server/app.py
```diff
@@ -196,11 +196,6 @@ def cosmic_prefix(cosmicId):
             return "COSM" + str(cosmicId)
         return cosmicId
 
-    @app.template_filter()
```
northwestwitch, line 199: 👍🏻

northwestwitch (approved):
👍🏻 Good job!

PR #2909, opened 2021-10-05
Use official MongoDB 4.4 in docker-compose

### LINKED ISSUE #2907: Change the Docker-compose and actions files to use a MongoDB version 4.4
To mirror the database that soon will be used in production

northwestwitch:
Fix #2907 

**How to test**:
1. Run docker-compose up and make sure the app can be launched and contains demo data

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ A simple demo instance of Scout requires the installation of Docker and can be l
 The repository includes a Makefile with common shortcuts to simplify setting up and working with Scout. To see a full list and description of these shortcuts run: `make help`.
 
 This demo is consisting of 3 containers:
-- a lightweight mongodb instance
+- a MongoDB instance running on port 27013
 - scout-cli --> the Scout command line, connected to the database. Populates the database with demo data
 - scout-web --> the Scout web app, that serves the app on localhost, port 5000.
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,14 +5,14 @@ version: '3'
 services:
 
   mongodb:
-    image: mvertes/alpine-mongo
+    image: mongo:4.4.9
     container_name: mongodb
-    ports:
-      - '27017:27017'
-    expose:
-      - '27017'
     networks:
       - scout-net
+    ports:
+      - '27013:27013'
+    expose:
+      - '27013'
 
   scout-cli:
     build: .
@@ -47,7 +47,3 @@ services:
 networks:
   scout-net:
     driver: bridge
-    ipam:
-      driver: default
-      config:
-        - subnet: 172.21.0.0/24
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -5,14 +5,14 @@ version: '3'
...
-    ports:
-      - '27017:27017'
-    expose:
-      - '27017'
     networks:
       - scout-net
+    ports:
+      - '27013:27013'
+    expose:
+      - '27013'
```
northwestwitch, line 15: I changed this port so it doesn't mess up with an eventual database hosted on the user machine

northwestwitch:
Container are created: 🆗 

image

App works as it should: 🆗 
image

docker-compose.yml
dnil, line 15:
```suggestion
      - '27017'
```

dnil: Nice idea, but you dont really change the port that the mongod answers to? That would be possible with options to the startup, or a custom mongod.conf for the container. But can I suggest just to leave it on 27017 in the container, remap it to the host 27013 for any debugging, and expose the container 27017 to the compose network.

dnil: Like so with these changes: 
Screenshot 2021-10-08 at 11 07 02
Screenshot 2021-10-08 at 11 07 12

northwestwitch: Good suggestion!! 👍🏻

northwestwitch: It was actually what I wanted to do but I did it wrong in the PR :)

docker-compose.yml
```diff
@@ -38,7 +37,6 @@ services:
     command: scout --host mongodb --demo serve --host 0.0.0.0
     volumes:
       - ./scout:/home/worker/app/scout
-      - ./volumes/scout/data:/home/worker/data
```
dnil, line 41: 👍 We are not really using it there as it is now. Devs should likely map their own volumes for both really, one for a live scout dir, one for a live persistent database. We could make an example of that in containers/ at some point, but not this PR.

README.md
```diff
@@ -34,7 +34,7 @@ A simple demo instance of Scout requires the installation of Docker and can be l
 The repository includes a Makefile with common shortcuts to simplify setting up and working with Scout. To see a full list and description of these shortcuts run: `make help`.
 
 This demo is consisting of 3 containers:
-- a lightweight mongodb instance
+- a MongoDB instance running on port 27013
```
dnil, line 37:
```suggestion
- a MongoDB instance, on the default port 27017 in the container, mapped to host port 27013
```

Issue #2906, opened 2021-10-05
Scout cli should be able to connect to a replica set MongoDB with connection params from a yaml file


northwestwitch:
Test the situation where a yaml file contains a database connection string containing all the urls of the replica set

northwestwitch:
I'll try setting up a 3 nodes replica set in docker-compose to test that cli can load demo data in the 3 databases. I can do it! 💪🏻

northwestwitch:
Let's split up this task in 3 smaller tasks:
- [ ] Create a docker-compose file with a MongoDB replicaset composed of 3 nodes. Set up the Databases so it will be possible to connect from localhost (not from Scout yet)
- [ ] Fix Scout connectivity via .py config file so that connects to the replicaset
- [ ] Fix Scout connectivity via .yaml config file so that connects to the replicaset

PR #2912, opened 2021-10-06
Connect to a MongoDB replicaset using config.py params - WORKS!!!

northwestwitch:
Fixes partially #2906 

**How to test --locally--**:
1. Switch off mongod instances you might have running locally and start a Dockerized MongoDB replica set with the command: 
```
docker run --name mongo-replica -d -p 27017:27017 -p 27018:27018 -p 27019:27019 northwestwitch/mongo4.4-replica-set
```
2. In the config.py uncomment the following line:
```
MONGO_URI = "mongodb://localhost:27017,localhost:27018,localhost:27019/?replicaSet=rs0&readPreference=primary&ssl=false" 
```
3. **Wait half a minute**
3. Load demo data in demo database with the command:  `scout setup demo `
4. Then run the server with `scout --demo serve`. 
5. Stop the container with: `docker stop mongo-replica`
5. Make sure Scout still connects to localhost:27017 if you comment the MONGO_URI and start your local mongod.

**Expected outcome**:

**Review:**
- [x] code approved by DNIL
- [x] tests executed by CR

### DIFF
--- a/scout/commands/serve.py
+++ b/scout/commands/serve.py
@@ -21,7 +21,8 @@ def serve(host, port, debug, livereload, test):
     """Start the web server."""
     if test:
         if current_app.config.get("MONGO_DATABASE"):
-            return "Connection could be established"
+            LOG.info("Connection could be established")
+            return
 
     if livereload:
         server = Server(current_app.wsgi_app)


northwestwitch:
This is finally working as it should now. Ready for review!

scout/adapter/client.py
```diff
@@ -54,11 +55,12 @@ def get_connection(
...
 
     LOG.info("Try to connect to %s" % log_uri)
+
+    client = MongoClient(uri, serverSelectionTimeoutMS=timeout)
     try:
-        client = MongoClient(uri, serverSelectionTimeoutMS=timeout)
-    except ServerSelectionTimeoutError as err:
-        LOG.warning("Connection Refused")
-        raise ConnectionFailure
+        client.server_info()
```
northwestwitch, line 61: This is the actual step that checks if the connection is valid. If it crashes it's not a valid connection (check [here](https://stackoverflow.com/questions/30539183/how-do-you-check-if-the-client-for-a-mongodb-instance-is-valid))

scout/commands/serve.py
```diff
@@ -27,26 +25,9 @@ def serve(host, port, debug, livereload, test):
         MONGO_DBNAME=current_app.config.get("MONGO_DBNAME", "scout"),
         MONGO_USERNAME=current_app.config.get("MONGO_USERNAME", None),
         MONGO_PASSWORD=current_app.config.get("MONGO_PASSWORD", None),
+        MONGO_URI=current_app.config.get("MONGO_URI", None),
     )
 
-    valid_connection = check_connection(
```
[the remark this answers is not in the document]
dnil: 👍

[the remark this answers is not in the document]
dnil: Agreed: that also makes the whole pymongo_config dict redundant, doesn't it - so perhaps just drop it?
And move the slightly more informative warning about starting mongod over to scout/adapter/client?

scout/adapter/client.py
```diff
@@ -44,6 +43,8 @@ def get_connection(
 
     """
     authdb = authdb or mongodb
+    log_uri = uri
```
dnil, line 46: I guess what he was trying to do is keep passwords out of the log. To me that makes a bit of sense. Perhaps just say connecting to `"mongod using config URI string"`, and reserve the full URI output for debug? That way, production logs would not contain passwords.

scout/adapter/client.py
```diff
@@ -54,11 +55,12 @@ def get_connection(
...
     LOG.info("Try to connect to %s" % log_uri)
+
+    client = MongoClient(uri, serverSelectionTimeoutMS=timeout)
     try:
-        client = MongoClient(uri, serverSelectionTimeoutMS=timeout)
-    except ServerSelectionTimeoutError as err:
-        LOG.warning("Connection Refused")
-        raise ConnectionFailure
+        client.server_info()
+    except (ServerSelectionTimeoutError, OperationFailure, ConnectionFailure) as err:
```
dnil, line 62: 👍

scout/adapter/client.py
dnil: NIce!

scout/adapter/utils.py
```diff
@@ -8,19 +8,20 @@
...
 from pymongo import MongoClient
-from pymongo.errors import OperationFailure, ServerSelectionTimeoutError
+from pymongo.errors import ConnectionFailure, OperationFailure, ServerSelectionTimeoutError
 
 LOG = logging.getLogger(__name__)
 
 
 def check_connection(
-    host="localhost", port=27017, username=None, password=None, authdb=None, max_delay=1
+    host="localhost", port=27017, uri=None, username=None, password=None, authdb=None, max_delay=1
```
dnil, line 17: 👍 
I guess we really wont use it any more, except for in some old scripts. Good to fix if its around, but we should probably deprecate it as well.

scout/adapter/client.py
northwestwitch: I think I'll just remove it, it's not something that we see anyway. And if there are problems with the connection we would see in the `LOG.error(f"Database connection error:{err}")` line

PR #2964, opened 2021-11-08
Display and download HPO gene panels' genes in italics

### LINKED ISSUE #2960: HPO gene list in italics?
Just improvement of life...

I wonder if it would be possible to get the gene names in _italics (kursivt)_ in the pdf we get when we download the HPO gene panel (bifogas som bilaga med våra svar). 

image

image

TACK på förhand!

northwestwitch:
Fix #2960.
Display HPO genes in italics on case page and in the downloaded PDFs

**How to test**:
2. Pick a case and assign one or more HPO phenotypes
3. Create a gene list

**Expected outcome**:
- [x] Make sure the HPO genes displayed on case page are in italics
- [x] When downloading the gene list using these buttons, the documents should contain genes in italics.
image

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/gene_panel.html
+++ b/scout/server/blueprints/cases/templates/cases/gene_panel.html
@@ -96,13 +96,15 @@
 
     <!-- scrollable div containing dynamic gene list -->
     <div class="mt-3 ml-3" style="max-height:200px; overflow-y: scroll; min-height: 100px;">
-      <i>
-      {% for hpo_gene in case.dynamic_gene_list %}
-        {{ hpo_gene.hgnc_symbol }}&nbsp;&nbsp;
+      {% if case.dynamic_gene_list %}
+        <i>
+        {% for hpo_gene in case.dynamic_gene_list %}
+          {{ hpo_gene.hgnc_symbol }}&nbsp;&nbsp;
+        {% endfor %}
+        </i>
       {% else %}
         <span class="text-muted">No gene list generated</span>
-      {% endfor %}
-      </i>
+      {% endif %}
     </div>
 
     {% if case.dynamic_gene_list %}


northwestwitch:
Before:

image


After:

image

scout/server/blueprints/cases/views.py
```diff
@@ -907,7 +907,7 @@ def download_hpo_genes(institute_id, case_name, category):
     phenotype_terms_with_genes = controllers.phenotypes_genes(store, case_obj, is_clinical)
     html_content = ""
     for term_id, term in phenotype_terms_with_genes.items():
-        html_content += f"<hr><strong>{term_id} - {term.get('description')}</strong><br><br>{term.get('genes')}<br>"
+        html_content += f"<hr><strong>{term_id} - {term.get('description')}</strong><br><br><i>{term.get('genes')}</i><br>"
```
dnil, line 910: 👍

scout/server/blueprints/cases/templates/cases/gene_panel.html
```diff
@@ -96,11 +96,13 @@ <h6 class="ml-3 mt-3"><span class="fas fa-clipboard-list"></span>&nbsp;HPO gene
 
     <!-- scrollable div containing dynamic gene list -->
     <div class="mt-3 ml-3" style="max-height:200px; overflow-y: scroll; min-height: 100px;">
+      <i>
       {% for hpo_gene in case.dynamic_gene_list %}
         {{ hpo_gene.hgnc_symbol }}&nbsp;&nbsp;
       {% else %}
         <span class="text-muted">No gene list generated</span>
```
dnil, line 103: Optionally leave this non-italic, eg by adding in a counter non-italic block. Or using an if instead of the for loop else.
```suggestion
       </i><span class="text-muted">No gene list generated</span><i>
```

dnil (approved):

scout/server/blueprints/cases/templates/cases/gene_panel.html
northwestwitch: Ok I'll change it 😄

northwestwitch: Changed as you suggested by introducing an if condition. Tested locally and when there are no genes it looks like this:

image

Issue #2959, opened 2021-11-04
local observations count occasionally missing


ielvers:
Hello!

We've noticed in a few cases that the local observations count is missing for occasional variants. Here is a recent occurrence:

Case 18172-2, rerun 2021-10-21. Variant https://scout.scilifelab.se/cust003/18172-2/88aa0ffdeccea4155eee94a48d0fa672. Two local occurrences reported, one is case 18172 which consists of this individual and her sister (but it wasn't on the clinical list when they were first run).

Case 18172-1, rerun 2021-11-03. This is the sister of 18172-2. The same variant is present (https://scout.scilifelab.se/cust003/18172-1/077a24ba8989348e84c928b62b5ddb83) but the local observation count is N/A.

It's obviously not a big problem, I'm just reporting it.

ielvers:
Umm, edit. I opened the whole list of variants and then went back to the variant of interest and now local observations count is present. I wish I had taken a screen shot.

dnil:
Thanks, we'll try keep an eye on it. We wanted to change the loqus integration anyway. If it's ok perhaps not until next week though, but do shout if it becomes a big issue!

ielvers:
I came across this again, so I'll paste the screenshots here, but this is really not a problem.
It's case 21286. I added some HPO terms and opened the clinical list. I noticed that the local observation count was missing in one variant. And then when I open the variant page again, the local observation count is loaded. 

I screenshotted the whole thing to prove that it's the same variant but the local observations count is the point :)

The first time I opened a variant:
![image](https://user-images.githubusercontent.com/57791644/141423727-706a7f95-3f14-4b0f-9f71-f76ad4ba853a.png)

Then opening the same variant again:
![image](https://user-images.githubusercontent.com/57791644/141422035-57f1b8cd-5410-43f2-ac11-e92a9f0d78d8.png)

Also, in all variants the archive local observation count (Nr obs. and Nr homo.) seems to be reported as N/A when it's 0.

Again, it's not a problem, just an observation.

dnil:
Thank you @ielvers! It is good timing as well, as we are in the process of moving the loqusdb backend over to the new server farm. We will be switching to using loqusdbapi, which has a nice REST API, rather than the CLI-app, which is rather easily overwhelmed by multiple requests within a short time on the current old server. We'll have a look at the 0 vs N/A as well, and see if we can't do something about it when we are at it!

dnil:
So, two things here:
[...]

PR #2975, opened 2021-11-12
Fix old counts N/A

dnil:

Attempt to fix the part of #2959 where available old obs 0 counts show as N/A. Missing keys (total counts, homozygotes) should still display as N/A.

Also allow admins to see case ids for loqus matches, even if their user institute is not a case collaborator.

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -213,9 +213,9 @@
         </thead>
         <tbody>
           <tr>
-            <td>{{ variant.local_obs_old or 'N/A' }}</td>
-            <td>{{ variant.local_obs_hom_old or 'N/A' }}</td>
-            <td>{{ variant.local_obs_total_old or 'N/A' }}</td>
+            <td>{{ variant.local_obs_old|default('N/A') }}</td>
+            <td>{{ variant.local_obs_hom_old|default('N/A') }}</td>
```
dnil, line 217: And to be frank, oftentimes the `Obs` and `Hom` keys would simply be unset, which this won't fix. We could try to resolve this by adding some convention with MIP crew. Perhaps the new addition of local_obs_total_count from next update will suffice?

dnil:
🤷‍♂️ Pretty ok, but again, if keys are not set in the vcf, we will show them as N/A. I think that might be the best we can currently do on the display side, without a careful agreement with the supplier of that obs annotation, like if a certain key is available in the header, missing keys on variants should be counted as zeros on load.

Screenshot 2021-11-12 at 09 56 22
Screenshot 2021-11-12 at 09 57 10

scout/server/blueprints/variant/templates/variant/variant_details.html
northwestwitch: Perhaps!

northwestwitch: Nice fix! Reading in the [docs](https://jinja.palletsprojects.com/en/3.0.x/templates/?highlight=default#jinja-filters.default), if local_obs_hom_old exists and it's empty, for instance then nothing will be printed in the page. So what about adding `true`, as a second param to the function, to default to N/A in case the key exists and it's blank?

northwestwitch (approved):
Nice. I have a suggestion, but it should work also like it is!

Issue #2961, opened 2021-11-08
Deprecated wtforms.TextField and problem with flask-ldap3-login after WTForms upgrade


northwestwitch:
TextField was removed in wtforms v.3.0.0 (released yesterday Nov 7th). Now all the imports of this element are broken and should be replaced by wtforms.StringField. https://wtforms.readthedocs.io/en/3.0.x/whats_new/?highlight=textfield

northwestwitch:
The fixed looked quite straightforward but here comes a problem: we are using a library named flask_ldap3_login to handle LDAP logins. flask_ldap3_login is also using the deprecated TextField and it's bugged at the moment. I'll try to create a PR in the other repo to fix the issue!


```
(py38) chiararasi@ChiaraRMBP:~/Documents/work/GITs/scout$ scout --demo serve
Traceback (most recent call last):
  File "/Users/chiararasi/miniconda3/envs/py38/bin/scout", line 33, in <module>
    sys.exit(load_entry_point('scout-browser', 'console_scripts', 'scout')())
  File "/Users/chiararasi/miniconda3/envs/py38/bin/scout", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/Users/chiararasi/miniconda3/envs/py38/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/Users/chiararasi/miniconda3/envs/py38/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/chiararasi/Documents/work/GITs/scout/scout/commands/__init__.py", line 1, in <module>
    from .base import cli
  File "/Users/chiararasi/Documents/work/GITs/scout/scout/commands/base.py", line 25, in <module>
    from scout.server.app import create_app
  File "/Users/chiararasi/Documents/work/GITs/scout/scout/server/app.py", line 13, in <module>
    from .blueprints import (
[...]

PR #2980, opened 2021-11-17
Update requirements to use flask-ldap3-login>=0.9.17 instead of freezing WTForm

northwestwitch:
flask-ldap3-login was just released to fix [this dependency issue](https://github.com/nickw444/flask-ldap3-login/issues/75#issuecomment-970715102) caused by the upgrade of WTForm to v3.0. This is a better fix than what we have for [the issue in Scout](https://github.com/Clinical-Genomics/scout/issues/2961) I think!

**How to test**:
1. Tested here: https://github.com/nickw444/flask-ldap3-login/issues/75#issuecomment-970385810

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,6 @@ Flask-Bootstrap
 Flask-CORS
 path.py
 Flask-Markdown
-WTForms<3 # Remove this dependency after flask_ldap3_login issue #75 is fixed
 Flask-WTF
 Flask-Mail
 Flask-WeasyPrint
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -26,7 +25,7 @@ cryptography<3.4
 
 # webapp login
 authlib
-flask_ldap3_login
+flask_ldap3_login>=0.9.17
```
dnil, line 28: 💯 That is super! Much better!

dnil (approved):
🆗

PR #2349, opened 2021-02-10
Refactor LoqusDB settings parsing and introducing Loqus queries via REST API

northwestwitch:
This PR adds support to queries to loqus using REST APIs. Fix #2295

Note that the connection to Loqusdb at the moment is a really messy story and I think it needs fixing as a Scout major release fix, see a suggestion in #2357

In this PR (**that will not make it necessary to release a major**), I'm introducing a warning message if a user provides LoqusDB settings that I think should be **deprecated**, for instance:

```
LOQUSDB_SETTINGS = {
  "binary_path":"/Users/username/miniconda3/envs/loqus/bin/loqusdb",
  "config_path":"/Users/username/Documents/work/GITs/loqusdb/config.yaml"
}
```

or:

```
LOQUSDB_SETTINGS = [ 
  {
    "binary_path":"/Users/username/miniconda3/envs/loqus/bin/loqusdb",
    "config_path":"/Users/username/Documents/work/GITs/loqusdb/config.yaml"
  },
  {
    "id": "cancer_loqus",
    "binary_path":"/Users/username/miniconda3/envs/loqus/bin/loqusdb2",
    "config_path":"/Users/username/Documents/work/GITs/loqusdb/config2.yaml"
  }
]
```
**Note that these settings will still work. I just want to scare users into using this type of config:**

```
LOQUSDB_SETTINGS = {
  "default": {
       "binary_path":"/Users/username/miniconda3/envs/loqus/bin/loqusdb",
       "config_path":"/Users/username/Documents/work/GITs/loqusdb/config.yaml"
   },
   "custom_other_loqus" : {
       "binary_path":"/Users/username/miniconda3/envs/loqus/bin/loqusdb2",
       "config_path":"/Users/username/Documents/work/GITs/loqusdb/config2.yaml"
    },
    "custom_loqus_api": {
       "api_url" : ""http://127.0.0.1:9000"
    }
}
```

Note that the config above has 3 different loqus instances. **This PR introduces support for the latter type (REST APIs).**

**Prepare for testing**
1. clone and install in a dedicated environment [loqusDB executable](https://github.com/moonso/loqusdb), without setting up its demo database.
1. clone (but don't install) the [loqusdbapi](https://github.com/Clinical-Genomics/loqusdbapi)
1. ` cd loqusdbapi`
1. Restore the loqusdbapi database in a local mongo instance (localhost, port 27017) so that it can be used by loqusdb executable:
```mongorestore --host localhost:27017 -d loqusdb dump/```
[...]

PR #2989, opened 2021-11-22
Test and fix loqusdbapi sv call 

dnil:

**How to test**:
1. check loqus sv freq on a variant - note it generally missing, even if the case in question is loaded
2. apply fix and see the variant reappear if loaded

**Expected outcome**:

Before:
Screenshot 2021-11-22 at 18 10 00

After:
Screenshot 2021-11-22 at 18 09 46

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/extensions/loqus_extension.py
+++ b/scout/server/extensions/loqus_extension.py
@@ -236,7 +236,7 @@ class LoqusDB:
 
             sv_type = variant_info["variant_type"]
             search_url = f"{search_url}/svs/?chrom={chrom}&end_chrom={end_chrom}&pos={pos}&end={end}&sv_type={sv_type}"
-            LOG.info("Loqus search: %s", search_url)
+            LOG.debug("Loqus search: %s", search_url)
         search_resp = api_get(search_url)
         if search_resp.get("status_code") != 200:
             LOG.info(search_resp.get("detail"))
--- a/scout/server/extensions/loqus_extension.py
+++ b/scout/server/extensions/loqus_extension.py
@@ -236,7 +236,8 @@ class LoqusDB:
 
             sv_type = variant_info["variant_type"]
             search_url = f"{search_url}/svs/?chrom={chrom}&end_chrom={end_chrom}&pos={pos}&end={end}&sv_type={sv_type}"
-            LOG.debug("Loqus search: %s", search_url)
+
+        LOG.debug("Loqus search: %s", search_url)
         search_resp = api_get(search_url)
         if search_resp.get("status_code") != 200:
             LOG.info(search_resp.get("detail"))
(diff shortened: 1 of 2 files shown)


dnil:
Ahem, but on my local the total counts also disappeared. Maybe some naming issue with the returned total counts on svs as well? Checking.

dnil:
> Ahem, but on my local the total counts also disappeared. Maybe some naming issue with the returned total counts on svs as well? Checking.

Ah, ok, that was just on local. I have an old Docker somewhere, pre the first v1.0 where "total" is still "nr_cases". Works fine on stage:
Screenshot 2021-11-22 at 18 38 24

scout/server/extensions/loqus_extension.py
```diff
@@ -235,7 +235,7 @@ def get_api_loqus_variant(api_url, variant_info):
             end = variant_info["end"]
 
             sv_type = variant_info["variant_type"]
-            search_url = f"{search_url}/svs/?chrom={chrom}&end_chrom={end_chrom}&pos={pos}&end={end}&sv_type={sv_type}"
+            search_url = f"{search_url}/?chrom={chrom}&end_chrom={end_chrom}&pos={pos}&end={end}&sv_type={sv_type}"
```
northwestwitch, line 238: This is interesting. Obviously an error I've introduced when I've created the support for the loqus API. But how come that this works if the loqusdbapi endpoint is shaped like this: https://github.com/Clinical-Genomics/loqusdbapi/blob/3fb1830e3b8f3d04a6e638f12d454e95475e8030/loqusdbapi/main.py#L98 ??

🤔

northwestwitch (commented):
Didn't we test how to retrieve SVs in https://github.com/Clinical-Genomics/scout/pull/2349?

Look at the comments with the validations.. 🤔

scout/server/extensions/loqus_extension.py
dnil: The search url already has the category here:
https://github.com/Clinical-Genomics/scout/blob/3995a367b4a5e56dd23e9ee172992cc74b69e6e0/scout/server/extensions/loqus_extension.py#L227

dnil: So it was added twice, basically, once explicitly once implicitly..

northwestwitch: Aha! I see now!

dnil: > Didn't we test how to retrieve SVs in #2349?
> 
> Look at the comments with the validations.. 🤔

Yes, and I also tested it in https://github.com/Clinical-Genomics/loqusdbapi/issues/19, where it failed first, then seemed to work. I think I am missing something about the connection session longevity: I guess both extensions must exist at startup, but perhaps we choose a little too early on a page which one to use?

northwestwitch, line 238: 🚀

northwestwitch (approved):
Of course, I see the error now. Good job! 💯

PR #2992, opened 2021-11-24
Cancer SV AD filter

dnil:

Add filters for depth, af, alt count; similar to the cancer SNV view.
Screenshot 2021-11-24 at 15 42 42

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/variant/controllers.py
```diff
@@ -434,6 +434,8 @@ def observations(store, loqusdb, case_obj, variant_obj):
 
     if not obs_data:
         obs_data["total"] = loqusdb.case_count(variant_category=category, loqusdb_id=loqusdb_id)
+        if obs_data["total"]:
+            obs_data["observations"] = 0
```
northwestwitch, line 438: 👍🏻

scout/server/blueprints/variants/forms.py
```diff
@@ -164,3 +164,16 @@ class SvFiltersForm(VariantFiltersForm):
     clingen_ngi = IntegerField("ClinGen NGI obs")
     swegen = IntegerField("SweGen obs")
     clinical_filter = SubmitField(label="Clinical filter")
+
+
+class CancerSvFiltersForm(SvFiltersForm):
+    """Extends SvFiltersForm for cancer structural variants"""
+
+    depth = IntegerField("Depth >", validators=[validators.Optional()])
```
northwestwitch, line 172: What is depth, AD? Write also (alt. AD) or add a tooltip perhaps?

northwestwitch:
I took a liberty to modify a test to increase the coverage. The test was already there, but with the wrong case type (RD instead of cancer)

northwestwitch (approved):
I have a few suggestions but the code works already!

northwestwitch:
Nice new functionality! 🥇

scout/server/blueprints/variants/controllers.py
```diff
@@ -888,6 +895,15 @@ def populate_filters_form(store, institute_obj, case_obj, user_obj, category, re
             }
         )
         clinical_filter = MultiDict(clinical_filter_dict)
+    elif category == "cancer_sv":
```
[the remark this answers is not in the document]
dnil: Exactly: no way around that I fear since this will have both somatic stuff and sv stuff. Multiple inheritance might simplify it slightly, at least if this was c++.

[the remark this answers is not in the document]
dnil: Yes, the solution for this pattern is really to have member methods on the classes, and use polymorphism. I'll have a look again.

scout/server/blueprints/variant/controllers.py
dnil: Something of an afterthought, but it feels rather safe and in line with the earlier fix.

scout/server/blueprints/variants/forms.py
```diff
@@ -164,3 +173,28 @@ class SvFiltersForm(VariantFiltersForm):
...
+    alt_count = IntegerField("Min alt count", validators=[validators.Optional()])
+    control_frequency = BetterDecimalField(
+        "Normal alt AF <", places=2, validators=[validators.Optional()]
+    )
+    tumor_frequency = BetterDecimalField(
+        "Tumor alt AF >", places=2, validators=[validators.Optional()]
+    )
+
+
+FILTERSFORMCLASS = {
```
dnil, line 194: It should perhaps be a clever dynamic constructor instead, but at least it is better than before.

PR #3007, opened 2021-11-30
Allow filtering for several chromosomes

### LINKED ISSUE #2971: SVFilters
Hi,

When using SVFilters to select SV variants would it be possible to:
-add the possibility to mark several chromosomes at the same time? Now one can only select one chromosome 
-add the posibility to filter by the AD: allele depth of the alternate allele (or by allele frequency), both for the normal and for the tumor samples? (filtering for the AD of the reference allele both in the normal and tummor sample can also be useful)

This would be VERY useful 

Thanks

Teresita

dnil:

Fix #2971 part 1 (part 2 was solved in another PR). 

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Ok, it now has the base functionalities:
- filtering for multiple chr works
- selectable coords are populated from all selected chr (though in honesty we can't really use that yet, so perhaps it is just bloat..)
I think I will call it a day here, but there are some open questions to consider, like how to treat the "All" option?

dnil:
This is more or less ready for review, but no rush. I might fix the complex function on a better time of day.. 😬

scout/server/blueprints/variants/forms.py
```diff
@@ -113,7 +113,7 @@ class VariantFiltersForm(FlaskForm):
         render_kw={"placeholder": "<chr>:<start pos>-<end pos>[optional +/-<span>]"},
     )
 
-    chrom = NonValidatingSelectField("Chromosome", choices=[], default="")
+    chrom = NonValidatingSelectMultipleField("Chromosome", choices=[], default="")
```
northwestwitch, line 116: 👍🏻

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,69 +1,107 @@
...
+			}
+			console.log("chrom is array: add multiple chr cytobands. first " + i + ":" + chrom[i] + "("+ typeof (chrom[i])+")")
+			if (i !== 0) {
+				first = false
+			}
+			populateCytobandsSingleChr(cytobands, chrom[i], first)
+		}
+		return
+	}
+	populateCytobandsSingleChr(cytobands, chromosome, false)
```
northwestwitch, line 59: Here you changed the populateCytoband by adding possible cytoband options for every chromosome that is preselected in the chromosome select. 

For instance if chromosome 1 and 2 are selected, then all cytobands for chr 1 and 2 are selectable. I understand how you reasoned but perhaps here codefactor has a point: aren't we complicating things too much?

Why not simply inactivating the cytoband select instead? Because your solution is not making it possible to select more than one cytoband at the time anyway, so even if you have 2 chromosome available, your cytoband selection is de facto restricting the search to only one chromosome..

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,69 +1,107 @@
...
-    var el = document.createElement("option");
-    el.textContent = optionText;
-    el.value = opt["stop"];
-    if(endElem.value === el.value){
-      el.selected = true;
-    }
-    cytoEnd.appendChild(el);
-  }
+	for (elem of [cytoStart, cytoEnd]) {
+		if (first && elem.options.length > 0) {
```
northwestwitch, line 69: All this code is quite complicated, so yes, if somebody is interested in more than chromosome, I'd just deactivate the cytoband search, easy peasy!

scout/adapter/mongo/query.py
```diff
@@ -247,9 +246,19 @@ def build_query(
                 continue
 
             if criterion == "chrom" and query.get("chrom"):  # filter by coordinates
+                query_chrom = query.get("chrom")
```
northwestwitch, line 249: What about unselecting "All" if any specific chromosome is selected in the multiselect? Because the current code allows you to do this:

image


Which is not returning interesting results, if the user meant to just select chromosome 1 and 2 for instance

northwestwitch (commented):
I have a few comments, pre-code review! :)

scout/adapter/mongo/query.py
dnil: Sure! I gave it a shot, but the logic was so complex I decided to unwrap the functions a bit, using the event handler a bit more. Prototyped on the sv page now. I still haven't found a quick way to update the select "summary" when any option status changes. I guess something that triggers on input and summarises...

dnil: Its sort of ok now methinks. The only thing is the immediacy of the values changing behind the scenes when All is selected. If one has the dropdown open already, it does not change to reflect the current values, although e.g. the summary on dropdown control does.

dnil:
No rush, but this should be ready for re-review. I cleaned up the js to decrease complexity, and moved most to a macro, so while it looks like a lot of changes, this is mostly due to the moving of nearly identical scripts from all variantS pages to one macro.

northwestwitch:
I'll re-review now!

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,69 +1,107 @@
...
+		}
+		return
+	}
+}
+
+function getSelectedChromosomes() {
+	var chrom_select = document.forms["filters_form"].elements["chrom"];
+
+	var chrom = [];
+  var options = chrom_select && chrom_select.options;
```
northwestwitch, line 37: You could simplify a lot the code from this line. Something like this perhaps?

```
if (chrom_select.selectedIndex === 0) {
		return chrom;
}

for (var i=0; i < chrom_select.selectedOptions.length; i++) {
		chrom.push(chrom_select.selectedOptions[i].value);
}
return chrom;
```

scout/server/blueprints/variants/static/form_scripts.js
```diff
@@ -1,69 +1,107 @@
...
-  if (typeof document.forms["filters_form"].elements["chrom_pos"] !== "undefined") {
-  	chromPos=document.forms["filters_form"].elements["chrom_pos"].value;
+/* exported sanitizeChromSelOptions() */
+function sanitizeChromSelOptions() {
+	var chrom = [];
+	var chrom_select = document.forms["filters_form"].elements["chrom"];
+  var options = chrom_select && chrom_select.options;
+  var opt;
+
+  // clear other options if All is selected
```
northwestwitch, line 8: Later in the code you have created a function that collects the selected chromosomes (getSelectedChromosomes). perhaps you can reuse it in here? There is some duplicated code that you could remove!

scout/adapter/mongo/query.py
northwestwitch: I'm testing the latest version. The simplest situation is when you want to select a single chromosome, for instance chr3. If you select it, you find yourself in this situation and you have to manually remove "All" otherwise the research will return all chroms:

image

This behavior is a bit cumbersome since I think most searches would include one chromosome only..

I understand the difficulty of the task, because what you want is:
- to deselect "All" when any single chromosome is selected
- to deselect all any single chromosomes when "All" is the last selected item of the select.

I think jquery is more useful than javascript to achieve the above. Check for instance the snippet provided here: 
https://stackoverflow.com/questions/39329332/bootstrap-multiselect-how-to-uncheck-and-deselect-multiselect-options-using-jqu

northwestwitch (changes requested):
The functionality is improving but there are a couple of things to fix I think. The first one is described in my comment (having to manually deselect All when you select single chromosomes). This is not something that doesn't work, but an annoying thing that is not going to be appreciated I'm afraid.

One thing that doesn't work is populating the cytobands once you select single chromosomes, have you checked.

Consider using a jquery solution, you could avoid many lines of code and potential errors..

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -2,6 +2,38 @@
 {% from "utils.html" import comments_table %}
 {% from "variants/indicators.html" import pin_indicator, causative_badge, comments_badge, evaluations_badge, research_assessments_badge, clinical_assessments_badge, group_assessments_badge %}
 
+{% macro filter_script_main(cytobands) %}
```
northwestwitch, line 5: Nice with the reusable macro! 🥇

scout/server/blueprints/variants/static/form_scripts.js
dnil: The complication with that is that I would also have to add a check for the empty value `All` option: since it can be selected or not, I can't just assume I know its position in the selectedOptions.

scout/server/blueprints/variants/static/form_scripts.js
dnil: Sure! I was just saving some runtime, but it is such a short list it makes no sense.

scout/server/blueprints/variants/templates/variants/utils.html
dnil: But apparently it broke the populateCytobands rerun for some reason. It tested ok before. lll have a look!

scout/server/blueprints/variants/static/form_scripts.js
northwestwitch: But you know it's position actually, it's always the first if it's selected, no?

northwestwitch: If you have: 
```
(chrom_select.selectedIndex === 0) {
		return chrom;
}
```

before that loop, whenever the first option (All) is selected, you will not even reach the loop after. Check the docs: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

dnil: That is correct, although it may not be super evident that the following requires it. But that can be said for a lot. Using  the `selectedOptions` collection and going from 0 does make that loop a little easier to read - let's go with that.

scout/server/blueprints/variants/templates/variants/utils.html
dnil: Yeah, looks like there is something odd with the argument passing via the macro. Perhaps that is why it wasn't before. :grin: Ill give it a few more minutes..

scout/server/blueprints/variants/static/form_scripts.js
northwestwitch: > may not be super evident

You could add a comment perhaps?

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: Feel free to open a PR to switch to b5 🤣

northwestwitch: There is one also for b4: https://mdbootstrap.com/docs/b4/jquery/forms/multiselect/

Check the example under "Multiselect with Bootstrap styles"

dnil: Interesting, though that particular example is bugged (note two Select All - the deselected and the selectable).

dnil: On the macro part, that was actually ok; I was not paying attention to the priority order when passing a function with an argument to addEventListener. The function actually got executed on load and the return value passed to addEventListener. 🙄

scout/demo/madeline.xml
```diff
@@ -192,7 +192,7 @@ text{
 }
 
 ]]></style>
- <script type="text/javascript" xlink:href="javascript/madeline.js"></script>
+ <script type="text/javascript" xlink:href="/cases/static/madeline.js"></script>
```
dnil, line 195: I just wanted to keep the console error free by default.. and this should be correct currently, though Im not so sure it is what cg does..

dnil:
I think you will like this one! I swapped from the custom stuff to using the select picker databox select all.
Screenshot 2022-02-02 at 18 21 10
Let me just layout that on all the pages first.. 😅

scout/adapter/mongo/query.py
```diff
@@ -248,9 +247,19 @@ def build_query(
...
+                if isinstance(query_chrom, list):
```
northwestwitch, line 251: Wouldn't be easier to just pass a list here, also when there is only one chromosome?

scout/demo/madeline.xml
northwestwitch: I'm not sure what that line does 🤔

scout/server/blueprints/variants/controllers.py
```diff
@@ -47,10 +47,10 @@
 
 
 def populate_chrom_choices(form, case_obj):
-    """Populate the option of the chromosome select accordig to the case genome build"""
+    """Populate the option of the chromosome select according to the case genome build"""
```
northwestwitch, line 50: 👍🏻

scout/server/blueprints/variants/controllers.py
```diff
@@ -47,10 +47,10 @@
...
     # Populate chromosome choices
     chromosomes = CHROMOSOMES if "37" in str(case_obj.get("genome_build")) else CHROMOSOMES_38
-    form.chrom.choices = [("", "All")] + [(chrom, chrom) for chrom in chromosomes]
```
northwestwitch, line 53: That's the easiest solution 👍🏻

northwestwitch (approved):
Great improvement! Now it works perfectly and it's user-friendly as well!

I have a comment regarding what you pass to the mongo query function, that code-wise would be simpler to just pass a list, even when it contains only one chrom, but if you like it this way it'f fine by me!

Nice new functionality! 🥇

scout/demo/madeline.xml
dnil: It imports some scripts that the svg uses: init() is the one that gets called and reported missing on the console. We serve it as a static from scout, but I think the path might have changed in cg without that being reflected in scout. Regardless, this works, and removes an error when running demo data. The other part would need some investigation on a live case and an issue to cg.

scout/adapter/mongo/query.py
dnil: It would. The only thing is Im reluctant to put in the work to find all places it is called. Maybe I could use like pydantic to recast it. 😁

northwestwitch: > Maybe I could use like pydantic to recast it.

You could ask @mikaell that is very familiar with the library 🤣

dnil: If you don't mind I will leave it for another time. I reread the criterion, and it would only shorten that very slightly, as there still is a need for different coordinate handling for SNV and SV on one hand, and for single and multiple chromosomes, with the latter not consistent with start-end coordinates. Looking at it again, I realised we were missing "cancer" in there, so they were always going as SV-overlaps. Should improve performance a little!

dnil: To justify the laziness, the places to look for changes is not trivial. It's not just the forms, its all variants queries e.g. https://github.com/Clinical-Genomics/scout/blob/92753c7a07b0d2c5602a327bbf7d8541c00e358c/scout/server/blueprints/cases/controllers.py#L484). Also stored filters would be affected and would probably need resetting or script updating.

northwestwitch: I don't mind! 😆

PR #3016, opened 2021-12-03
Do not drop HPO terms collection when updating HPO terms via the cli

northwestwitch:
Fix #3015 

**How to test**:
1. Locally, set up a demo instance
2. Check that the hpo terms collection has several indexes
3. On main branch, run `scout --demo update hpo`
4. Make sure indexes are gone (except for _id)
5. Repeat from one from this branch
6. After updating the hpo terms, indexes should be retained

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

PR #3017, opened 2021-12-03
Release 4.43.1

dnil:
This PR marks a new Scout release. We apply semantic versioning. This is a patch release since it patches an issue without changing the intended functionality or any interfaces.

### Added
 - [x] Added database index for OMIM disease term genes
 ### Changed
 ### Fixed
~Do not drop HPO terms collection when updating HPO terms via the command line~
- [x] Do not drop disease (OMIM) terms collection when updating diseases via the command line

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/commands/update/hpo.py
```diff
@@ -43,8 +43,9 @@ def hpo(hpoterms, hpo_to_genes):
     adapter = store
 
     LOG.info("Dropping HPO terms")
-    adapter.hpo_term_collection.drop()
+    adapter.hpo_term_collection.delete_many({})
```
northwestwitch, line 46: You have incorporated also this change from my previous PR. But it should be fine when we merge #3016

northwestwitch (approved):
👍🏻

scout/commands/update/hpo.py
dnil: Yes, that was intentional! All good!

dnil:
This is now on stage for testing!

dnil:
Adding index 🆗 
Screenshot 2021-12-03 at 09 53 10

dnil:
Disease terms: 🆗 
Screenshot 2021-12-03 at 10 26 34

Screenshot 2021-12-03 at 10 27 33

As for HPO ahem, right, this is clever:
Screenshot 2021-12-03 at 10 21 28
And indexes are fine ofcourse:
Screenshot 2021-12-03 at 10 27 33

Non-dropping of indices was tested locally though in the respective PRs so I think we are fine.

PR #3021, opened 2021-12-03
Fix #3020 - add index

### LINKED ISSUE #3020: Add index for disease_term genes
Most fetching of disease_term is done via `_id` but we also have an operation searching arrays, which could be indexed.

https://github.com/Clinical-Genomics/scout/blob/c89cc3c0025ea918d1430dec07ccbcd92326ca0d/scout/adapter/mongo/omim.py#L113

dnil:
This PR adds a functionality - index on disease term genes

**How to test**:
Screenshot 2021-12-03 at 09 39 19

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/constants/indexes.py
+++ b/scout/constants/indexes.py
@@ -1,6 +1,12 @@
 from pymongo import ASCENDING, DESCENDING, TEXT, IndexModel
 
 INDEXES = {
+    "disease_term": [
+        IndexModel(
+            [("genes", ASCENDING)],
+            name="genes",
+        )
+    ],
     "hgnc_gene": [
         IndexModel(
             [("build", ASCENDING), ("chromosome", ASCENDING)],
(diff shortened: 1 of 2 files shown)


scout/constants/indexes.py
```diff
@@ -1,6 +1,12 @@
 from pymongo import ASCENDING, DESCENDING, TEXT, IndexModel
 
 INDEXES = {
+    "disease_term": [
```
northwestwitch, line 4: It should work fine!

northwestwitch (approved):

dnil:
Yes, taking hits very quickly! I like only started the local and klicked one variant.

Screenshot 2021-12-03 at 09 53 10

dnil:
This is actually used when uploading cases. Might give a small improvement in upload speed as well.

PR #3023, opened 2021-12-03
Fix #3022 - add melter toggle

### LINKED ISSUE #3022: Allow users to toggle cases for rerun monitoring (melter)
As in https://github.com/Clinical-Genomics/melter/issues/5, add a toggle to the case collapsible action bar, triggering new events for melter monitoring of a case.

dnil:

Screenshot 2021-12-03 at 16 26 26

Screenshot 2021-12-03 at 17 14 09

Screenshot 2021-12-03 at 17 31 18

Screenshot 2021-12-03 at 17 39 10

**How to test**:
1. toggle rerun monitoring on and off for cases
2. note that this should be logged as events for the case
3. when rerun monitoring is enabled for a case, it should be exported ok using `scout export cases --rerun-monitor`

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
This is working, but I will add a couple of tests to get coverage back up again.

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -275,6 +275,11 @@
   {% else %}
     <button type="button" class="btn btn-warning form-control btn-sm" data-toggle="modal" data-target="#rerun-modal">Request rerun</button>
   {% endif %}
+  <p>
+    <form action="{{ url_for('cases.monitor', institute_id=institute._id, case_name=case.display_name) }}" method="POST">
+      <span><input type="checkbox" id="rerun_monitoring" onChange="this.form.submit()" name="rerun_monitoring" value="monitor" {% if case.rerun_monitoring %}checked{% endif %}></span><span class="text-white" data-toggle="popover" data-content="Check box to enable rerun monitoring with melter."> Request rerun monitoring</span>
```
henningonsbring, line 280:
```suggestion
      <span><input type="checkbox" id="rerun_monitoring" onChange="this.form.submit()" name="rerun_monitoring" value="monitor" {% if case.rerun_monitoring %}checked{% endif %}></span><span class="text-white" data-toggle="popover" data-content="A new analysis will start when relevant parts of the workflow is updated"> Activate monitoring</span>
```

scout/commands/export/case.py
```diff
@@ -21,6 +21,7 @@
 @click.option("-f", "--finished", is_flag=True, help="archived or solved")
 @click.option("--causatives", is_flag=True, help="Has causative variants")
 @click.option("--research-requested", is_flag=True, help="If research is requested")
+@click.option("--rerun-monitor", is_flag=True, help="If case rerun monitoring is requested")
```
henningonsbring, line 24: "If case rerun monitoring is requested", from just reading this text I do not really understand what it does, can it be more clear? Now it sounds like you ask if a specific case is monitored, as I understand it you will get all monitored cases returned together with some metadata.

dnil: Yeah, I just followed the style we have for that endpoint, but I agree; it's not easy to parse without examples. I've added a few words; hopefully it's a little better.

dnil:
Ok, fixed conflict. I have now added a config option to show the melter interaction in the web interface, so that we could merge the code without having to think too much about how the users will receive it when the melter backend is not yet up in production.

There is one thing that will require a little interaction testing with cg: the switch to our own `json_converter` instead of the `bson_handler`. I think it is the way to do it. Using the `bson_handler` is actually bugged on the current `main`. Export case with PNGs stored, like for the demo, with the `--json` flag crashes. It might still be that some cg component is surprised - e.g. if it expects the broken 'datetime.datetime()`-strings in the output. Probably not, but better safe than sorry.

.github/workflows/vulture.yml
```diff
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 --ignore-names cls,args ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls,args,kwargs ${{steps.files.outputs.all}}
```
northwestwitch, line 22: 👍🏻

CHANGELOG.md
```diff
@@ -80,6 +80,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Add utils for dictionary type
 - Filter institute cases using multiple HPO terms
 - Vulture GitHub action to identify and remove unused variables and imports
+- Add a toggle for melter rerun monitoring of cases
```
northwestwitch, line 83: It has been a while, so these 2 should be moved up in the changelog

scout/adapter/mongo/case_events.py
```diff
@@ -413,11 +413,73 @@ def request_rerun(self, institute, case, user, link):
...
+            institute=institute,
+            case=case,
+            user=user,
+            link=link,
+            category="case",
+            verb="monitor",
+            subject=case["display_name"],
+        )
+
+        updated_case = self.case_collection.find_one_and_update(
```
northwestwitch, line 439: This works. But one thing that I noticed while working on a recent PR is that in general all case updates go through the [update_case](https://github.com/Clinical-Genomics/scout/blob/c1630a12b53509b779abf7a06fb8666b36842674/scout/adapter/mongo/case.py#L755).

I think the code would be easier to maintain if you update the case in the same way

scout/adapter/mongo/case_events.py
```diff
@@ -413,11 +413,73 @@ def request_rerun(self, institute, case, user, link):
...
+            institute=institute,
+            case=case,
+            user=user,
+            link=link,
+            category="case",
+            verb="unmonitor",
+            subject=case["display_name"],
+        )
+
+        updated_case = self.case_collection.find_one_and_update(
```
northwestwitch, line 470: Same here, invoke `self.update_case` instead?

scout/commands/export/case.py
```diff
@@ -7,6 +7,7 @@
 
 from scout.constants import CASE_STATUSES
 from scout.server.extensions import store
+from scout.server.utils import jsonconverter
 
 from .export_handler import bson_handler
```
northwestwitch, line 12: I think Vulture is right here, it's not used any more

scout/commands/export/case.py
```diff
@@ -21,6 +22,11 @@
 @click.option("-f", "--finished", is_flag=True, help="archived or solved")
 @click.option("--causatives", is_flag=True, help="Has causative variants")
 @click.option("--research-requested", is_flag=True, help="If research is requested")
+@click.option(
```
northwestwitch, line 25: Useful! 👍🏻

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -276,16 +276,33 @@
...
     <button type="button" class="btn btn-warning btn-sm form-control" data-toggle="modal" data-target="#rerun-modal">Request rerun</button>
   </div>
 {% endif %}
 {% endmacro %}
 
+
+{% macro melter_interact(institute, case) %}
+  <div class="ml-3 mt-1">
+    <form action="{{ url_for('cases.monitor', institute_id=institute._id, case_name=case.display_name) }}" method="POST">
+      <span><input type="checkbox" id="rerun_monitoring" onChange="this.form.submit()" name="rerun_monitoring" value="monitor" {% if case.rerun_monitoring %}checked{% endif %}></span><span class="text-white" data-toggle="tooltip" title="A new analysis will start when relevant parts of the workflow is updated"> Activate monitoring</span>
```
northwestwitch, line 301:
```suggestion
      <span><input type="checkbox" id="rerun_monitoring" onChange="this.form.submit()" name="rerun_monitoring" value="monitor" {% if case.rerun_monitoring %}checked{% endif %}></span><span class="text-white" data-toggle="tooltip" title="A new analysis will start when relevant parts of the workflow are updated"> Activate monitoring</span>
```

scout/commands/export/case.py
```diff
@@ -70,7 +80,7 @@ def cases(
             LOG.info("No cases could be found")
 
     if json:
-        click.echo(json_lib.dumps(models, default=bson_handler))
+        click.echo(json_lib.dumps(models, default=jsonconverter))
```
northwestwitch, line 83: 👍🏻

scout/constants/case_tags.py
```diff
@@ -88,6 +88,8 @@
     "share": "shared case with",
     "unshare": "revoked access for",
     "rerun": "requested rerun of",
+    "monitor": "requested rerun monitoring for",
```
northwestwitch, line 91:
```suggestion
    "rerun_monitor": "requested rerun monitoring for",
```

scout/constants/case_tags.py
```diff
@@ -88,6 +88,8 @@
...
+    "unmonitor": "disabled rerun monitoring for",
```
northwestwitch, line 92:
```suggestion
    "rerun_unmonitor": "disabled rerun monitoring for",
```

scout/adapter/mongo/case_events.py
```diff
@@ -413,11 +413,73 @@ def request_rerun(self, institute, case, user, link):
...
+             updated_case
+        """
+
+        self.create_event(
+            institute=institute,
+            case=case,
+            user=user,
+            link=link,
+            category="case",
+            verb="monitor",
```
northwestwitch, line 435:
```suggestion
            verb="rerun_monitor",
```

scout/adapter/mongo/case_events.py
```diff
@@ -413,11 +413,73 @@ def request_rerun(self, institute, case, user, link):
...
+          updated_case
+        """
+
+        self.create_event(
+            institute=institute,
+            case=case,
+            user=user,
+            link=link,
+            category="case",
+            verb="unmonitor",
```
northwestwitch, line 466:
```suggestion
            verb="rerun_unmonitor",
```

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -276,16 +276,33 @@
 </a>
 <!-- Submenu content -->
 {% if case.rerun_requested %}
+  {% if config.SHOW_MELTER %}
```
northwestwitch, line 279: SHOW_MELTER is a bit tricky as a name for a config param. It's not self-explanatory and it has a meaning just for CG Solna. Rename it to something like `RERUN_MONITORING` ?

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -276,16 +276,33 @@
...
+    <div id='submenu3' class="collapse">
+      {{ melter_interact(institute, case) }}
```
northwestwitch, line 281: Also here, perhaps rename to something that can be understood and used also by other centers?

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -276,16 +276,33 @@
...
+    </div>
+  {% endif %}
   <span class="d-flex align-items-center">
     <i class="text-white mt-1 ml-3">Rerun pending</i>
   </span>
 {% else %}
   <div id='submenu3' class="collapse">
+    {% if config.SHOW_MELTER %}
+      {{ melter_interact(institute, case) }}
```
northwestwitch, line 290: Here as well

scout/server/blueprints/cases/views.py
```diff
@@ -828,6 +828,21 @@ def rerun(institute_id, case_name):
     return redirect(request.referrer)
 
 
+@cases_bp.route("/<institute_id>/<case_name>/monitor", methods=["POST"])
+def monitor(institute_id, case_name):
```
northwestwitch, line 832:
```suggestion
def rerun_monitor(institute_id, case_name):
```

scout/server/config.py
```diff
@@ -97,6 +97,7 @@
 SHOW_CAUSATIVES = True
 SHOW_OBSERVED_VARIANT_ARCHIVE = True
 HIDE_ALAMUT_LINK = True
+SHOW_MELTER = True
```
northwestwitch, line 100: RERUN_MONITORING ?

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -91,6 +89,48 @@ def test_reanalysis(app, institute_obj, case_obj, mocker, mock_redirect):
         assert resp.status_code == 302
 
 
+def test_monitor(app, institute_obj, mocker, mock_redirect):
```
northwestwitch, line 92:
```suggestion
def test_rerun_monitor(app, institute_obj, mocker, mock_redirect):
```

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -91,6 +89,48 @@ def test_reanalysis(app, institute_obj, case_obj, mocker, mock_redirect):
...
+            data=form_data,
+        )
+        assert resp.status_code == 302
+        updated_case = store.case_collection.find_one({"_id": a_case["_id"]})
+        # THEN the case should be updated
+        assert updated_case["rerun_monitoring"] is True
+
+        # AND an unmonitor event should be created
+        rerun_event = store.event_collection.find_one()
+        assert rerun_event.get("verb") == "monitor"
```
northwestwitch, line 131:
```suggestion
        assert rerun_event.get("verb") == "rerun_monitor"
```

northwestwitch (commented):
Nice! I have a few suggestions, but nothing so complicated!

What happens when the case is rerun? Then the rerun monitor should be set to False for the case, right?

dnil:
> Nice! I have a few suggestions, but nothing so complicated!

Thanks a bunch! I'll check.

> What happens when the case is rerun? Then the rerun monitor should be set to False for the case, right?

No, that doesn't really affect it yet. Maybe the case was solved by the first rerun, maybe it wasn't. Possibly it should be auto reset if the case is marked solved, but then there is partially solved, so I think we can let the user decide for starters.

At some point we will need to present results from specifically the melter reruns. It will likely entail a little bit of tagging of variants of interest to avoid complete re-evaluation of the case. That will be a later PR...

scout/commands/export/case.py
dnil: Yeah, that likely was the last place.

CHANGELOG.md
dnil: Indeed! I thought I did that in a merge, but time flies.

scout/adapter/mongo/case_events.py
dnil: Hm, this is still in the case adapter.. But, sure, the fewer places we touch the db the better.

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
dnil: He, sure, it's just a macro name, so not very exposed. But it doesn't matter, so why not!

scout/server/config.py
```diff
@@ -97,6 +97,8 @@
 SHOW_CAUSATIVES = True
 SHOW_OBSERVED_VARIANT_ARCHIVE = True
 HIDE_ALAMUT_LINK = True
+# Display case rerun monitoring toggle
+RERUN_MONITOR = True
```
northwestwitch, line 101:
```suggestion
# RERUN_MONITOR = True
```

northwestwitch (approved):
Tested again after the changes and works perfectly! 💯

scout/server/config.py
northwestwitch: Just an idea: since other centers are not likely going to have a use for the functionality just now, we could switch if off in the demo settings?

dnil: Hm, this is the demo config. I would suggest keeping it enabled as possible, when other components are not really invoked. I mean, a test user could run the cli export command for monitored cases. But it doesn't matter very much either way.

dnil: Ah, but that reminds me - what happens if the key does not exist! Will fix!

northwestwitch: Ok, see, my comment has been useful anyway! :)

dnil: > Ok, see, my comment has been useful anyway! :)

Essentially always! 😸

PR #3026, opened 2021-12-07
Fix variants searching by symbol and alias

### LINKED ISSUE #3024: Genes not included in gene panel versions in use when loading this case (clinical list):['TAP1', 'TAP2']
The **express** case ([421701400259fam](https://scout.scilifelab.se/cust004/421701400259fam)) was analysed with [PID](https://scout.scilifelab.se/panels/603f92c9d05a2282ecd74c50) (Primary Immune deficiency) as default panel. This panel contains the genes TAP1 and TAP2, but if you try to filter for variants in these 2 genes it complains:

image

WHY?

northwestwitch:
Fix #3024. This should fix the bug, but perhaps the functionality might be improved to return all variants that have TAP1 as gene symbol or TAP1 as an alias.

**How to test**:
1. Load on scout stage and look for variants in TAP1 or TAP2 in case that contains these variants

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by DN

### DIFF
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -316,13 +316,21 @@ class GeneHandler(object):
         LOG.info("All genes fetched")
         return hgnc_dict
 
+    def gene_by_symbol_or_aliases(self, symbol, build="37"):
+        """Return an iterable with only one gene is found with a given symbol or a list of genes
+           with the provided symbol contained in the aliases list.
+        Args:
+            symbol(str)
+            build(str)
+
+        Returns:
+            res(dict or pymongo.Cursor(dict)): return a gene dictionary or a cursor with gene dictionaries
+        """
+        res = self.hgnc_collection.find_one({"hgnc_symbol": symbol, "build": str(build)})
+        return res or self.gene_by_alias(symbol, build="37")
+
     def gene_by_alias(self, symbol, build="37"):
-        """Return an iterable with hgnc_genes.
-
-        If the gene symbol is listed as primary the iterable will only have
-        one result. If not the iterable will include all hgnc genes that have
-        the symbol as an alias.
-
+        """Return an iterable with hgnc_genes which have the provided symbol in the gene aliases
         Args:
             symbol(str)
             build(str)
@@ -332,12 +340,7 @@ class GeneHandler(object):
         """
         LOG.debug("Fetch gene by symbol if possible: {}".format(symbol))
 
-        res = self.hgnc_collection.find({"hgnc_symbol": symbol, "build": str(build)})
-
-        if self.hgnc_collection.find_one({"aliases": symbol, "build": str(build)}) is None:
-            LOG.debug("No gene with symbol {} was found. Attempting an alias.".format(symbol))
-            res = self.hgnc_collection.find({"aliases": symbol, "build": str(build)})
-
+        res = self.hgnc_collection.find({"aliases": symbol, "build": str(build)})
         return res
 
     def genes_by_alias(self, build="37", genes=None):
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -324,10 +324,13 @@ class GeneHandler(object):
             build(str)
 
         Returns:
-            res(dict or pymongo.Cursor(dict)): return a gene dictionary or a cursor with gene dictionaries
+            res(list or pymongo.Cursor(dict)): return a list with one gene or a cursor with several gene dictionaries
         """
         res = self.hgnc_collection.find_one({"hgnc_symbol": symbol, "build": str(build)})
-        return res or self.gene_by_alias(symbol, build="37")
+        if res:
+            return [res]
+        else:
+            return self.gene_by_alias(symbol, build="37")
 
     def gene_by_alias(self, symbol, build="37"):
         """Return an iterable with hgnc_genes which have the provided symbol in the gene aliases
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -317,8 +317,8 @@ class GeneHandler(object):
         return hgnc_dict
 
     def gene_by_symbol_or_aliases(self, symbol, build="37"):
-        """Return an iterable with only one gene is found with a given symbol or a list of genes
-           with the provided symbol contained in the aliases list.
+        """Return an iterable with only one gene when gene with a given symbol if found
+           or a cursor with genes where the provided symbol is among the aliases.
         Args:
             symbol(str)
             build(str)
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -173,8 +173,7 @@ class GeneHandler(object):
         return self.hgnc_collection.find({"aliases": hgnc_symbol, **build_query})
 
     def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
-        """Find one hgnc genes that match a hgnc symbol. Replaces deprecated
-        pymongo.cursor.count()
+        """Find one hgnc genes that match a hgnc symbol.
 
         Check both hgnc_symbol and aliases
 
@@ -186,9 +185,7 @@ class GeneHandler(object):
         """
 
         LOG.debug("Find one genes with symbol %s" % hgnc_symbol)
-        return self.hgnc_collection.find_one(
-            filter={"build": build, "hgnc_symbol": hgnc_symbol, "aliases": hgnc_symbol}
-        )
+        return self.hgnc_collection.find_one(filter={"build": build, "aliases": hgnc_symbol})
 
     def get_query_alias_or_id(self, hgnc_symbol, build):
         """Return query to search for hgnc-symbol or aliases"""
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -330,9 +330,9 @@ class GeneHandler(object):
         if res:
             return [res]
 
-        return self.gene_by_alias(symbol, build="37")
+        return self.gene_aliases(symbol, build="37")
 
-    def gene_by_alias(self, symbol, build="37"):
+    def gene_aliases(self, symbol, build="37"):
         """Return an iterable with hgnc_genes which have the provided symbol in the gene aliases
         Args:
             symbol(str)
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -341,8 +341,6 @@ class GeneHandler(object):
         Returns:
             res(pymongo.Cursor(dict))
         """
-        LOG.debug("Fetch gene by symbol if possible: {}".format(symbol))
-
         res = self.hgnc_collection.find({"aliases": symbol, "build": str(build)})
         return res
 
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -329,8 +329,8 @@ class GeneHandler(object):
         res = self.hgnc_collection.find_one({"hgnc_symbol": symbol, "build": str(build)})
         if res:
             return [res]
-        else:
-            return self.gene_by_alias(symbol, build="37")
+
+        return self.gene_by_alias(symbol, build="37")
 
     def gene_by_alias(self, symbol, build="37"):
         """Return an iterable with hgnc_genes which have the provided symbol in the gene aliases
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -173,7 +173,7 @@ class GeneHandler(object):
         return self.hgnc_collection.find({"aliases": hgnc_symbol, **build_query})
 
     def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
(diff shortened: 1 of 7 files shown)


scout/adapter/mongo/hgnc.py
```diff
@@ -186,7 +186,9 @@ def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
         """
 
         LOG.debug("Find one genes with symbol %s" % hgnc_symbol)
-        return self.hgnc_collection.find_one(filter={"build": build, "aliases": hgnc_symbol})
+        return self.hgnc_collection.find_one(
+            filter={"build": build, "hgnc_symbol": hgnc_symbol, "aliases": hgnc_symbol}
```
dnil, line 190: Hm, aliases should always contain hgnc_symbol. We should probably fix this where it is called. Probably that place should use hgnc_genes instead of hgnc_genes_find_one?

northwestwitch:
Here is a case with a variant in those genes on stage, to test the fix: https://scout-stage.scilifelab.se/cust002/F0013649/0aee14cfad75c537d7a6568bcf3d3c32

scout/server/blueprints/variants/controllers.py
```diff
@@ -1072,20 +1072,21 @@ def check_form_gene_symbols(
...
-            if hgnc_symbol not in clinical_symbols:
-                # clinical symbols from gene panels might not be up to date with latest gene names
-                # but their HGNC id would still match
-                outdated_symbols.append(hgnc_symbol)
-        else:  # clinical variants
-            non_clinical_symbols.append(hgnc_symbol)
+        hgnc_genes = store.hgnc_genes(hgnc_symbol=hgnc_symbol, build=genome_build)
+
+        for hgnc_gene in hgnc_genes:
+            if hgnc_gene is None:
```
northwestwitch, line 1078: This doesn't make sense, does it?

scout/adapter/mongo/hgnc.py
```diff
@@ -317,13 +316,24 @@ def hgncsymbol_to_gene(self, build="37", genes=None):
...
+        res = self.hgnc_collection.find_one({"hgnc_symbol": symbol, "build": str(build)})
+        if res:
+            return [res]
 
-        If the gene symbol is listed as primary the iterable will only have
-        one result. If not the iterable will include all hgnc genes that have
-        the symbol as an alias.
+        return self.gene_by_alias(symbol, build="37")
 
+    def gene_by_alias(self, symbol, build="37"):
```
dnil, line 335: Hm, isn't the convention that we always find_one with the gene_ but allow multiple with the geneS_?

dnil: It should have been genes_by_alias before really.. 😁

northwestwitch: I think it would be better to return them all in this case, because you don't know which alias the right one is

northwestwitch: there is another function, down, that is named geneS_by_alias.. 😆

northwestwitch: We have to rename the other function in case

northwestwitch: But I agree, should be geneS

scout/adapter/mongo/hgnc.py
```diff
@@ -317,28 +316,32 @@ def hgncsymbol_to_gene(self, build="37", genes=None):
...
         """
-        LOG.debug("Fetch gene by symbol if possible: {}".format(symbol))
+        res = self.hgnc_collection.find_one({"hgnc_symbol": symbol, "build": str(build)})
+        if res:
+            return [res]
 
-        res = self.hgnc_collection.find({"hgnc_symbol": symbol, "build": str(build)})
+        return self.gene_aliases(symbol, build="37")
 
-        if self.hgnc_collection.find_one({"aliases": symbol, "build": str(build)}) is None:
```
dnil, line 338: 👍

scout/commands/load/variants.py
```diff
@@ -108,7 +108,7 @@ def variants(
         if hgnc_id:
             gene_obj = adapter.hgnc_gene(hgnc_id, case_obj["genome_build"])
         if hgnc_symbol:
-            for res in adapter.gene_by_alias(hgnc_symbol, case_obj["genome_build"]):
+            for res in adapter.gene_aliases(hgnc_symbol, case_obj["genome_build"]):
```
dnil, line 111: 👍

scout/server/blueprints/variants/controllers.py
```diff
@@ -1072,20 +1072,22 @@ def check_form_gene_symbols(
     clinical_symbols = store.clinical_symbols(case_obj)
 
     for hgnc_symbol in hgnc_symbols:
-        hgnc_gene = store.hgnc_genes_find_one(hgnc_symbol, genome_build)
-
-        if hgnc_gene is None:
+        # Retrieve a gene with "hgnc_symbol" as hgnc symbol or a list of genes where hgnc_symbol is among the aliases
+        hgnc_genes = store.gene_by_symbol_or_aliases(symbol=hgnc_symbol, build=genome_build)
+        if not hgnc_genes:
```
dnil, line 1077: 👍

scout/server/blueprints/variants/controllers.py
```diff
@@ -1072,20 +1072,22 @@ def check_form_gene_symbols(
     clinical_symbols = store.clinical_symbols(case_obj)
 
     for hgnc_symbol in hgnc_symbols:
-        hgnc_gene = store.hgnc_genes_find_one(hgnc_symbol, genome_build)
```
dnil, line 1075: The new version of this is already so much better!

dnil (changes requested):
Great start! Works on the error scenario, and fixes a weirdly implemented/incorrect old function.

I still find a couple of issues with it though:
- does not actually bring in aliases in search, even if the gene is in the clinical list (try e.g. RING4 or RING11)
- when reading up I started worrying about cases sensitivity, as is fixed by e.g. hgnc_genes(). Case can be expected to be correct when input on like a panel, but when freetexted in a box.. It wasn't working for some time, so lets not overdo it though.
- Warnings for entirely missing genes appear to have gone away. Sth with None returns? Like [ None ] == True?

northwestwitch:
Regarding the first comment of your review: `does not actually bring in aliases in search, even if the gene is in the clinical list (try e.g. RING4 or RING11)`. I think it's up to the user to search for genes using an updated symbol no? Because the alternative is returning all genes that have either symbol= symbol or symbol in aliases. And if we do that, every time a user looks for a gene then they will get A TON of warnings, since it will be a list of genes returned all the time, and probably not just the gene they intended to change

northwestwitch:
> * Warnings for entirely missing genes appear to have gone away. Sth with None returns? Like [ None ] == True?

I missed this! I'll fix it, thanks!

dnil:
> > * Warnings for entirely missing genes appear to have gone away. Sth with None returns? Like [ None ] == True?
> 
> I missed this! I'll fix it, thanks!

With that working again, both the other disputable issues should become more evident and less confusing!

dnil:
And we already have a caption for the query field saying "case sensitive". My bad - let's keep it like that! 😊

northwestwitch:
> With that working again, both the other disputable issues should become more evident and less confusing!

This should be fixed now:

image

scout/server/blueprints/variants/controllers.py
```diff
@@ -1072,20 +1072,23 @@ def check_form_gene_symbols(
...
-                # but their HGNC id would still match
-                outdated_symbols.append(hgnc_symbol)
-        else:  # clinical variants
-            non_clinical_symbols.append(hgnc_symbol)
+
+        for hgnc_gene in hgnc_genes:
+            if is_clinical is False:  # research variants
+                updated_hgnc_symbols.append(hgnc_symbol)
+            elif hgnc_gene["hgnc_id"] in clinical_hgnc_ids:  # clinical variants
+                updated_hgnc_symbols.append(hgnc_symbol)
```
dnil, line 1085:
```suggestion
                updated_hgnc_symbols.append(hgnc_gene["hgnc_symbol"])
```
That way the updated symbol goes in.

scout/server/blueprints/variants/controllers.py
```diff
@@ -1072,20 +1072,23 @@ def check_form_gene_symbols(
...
-            if hgnc_symbol not in clinical_symbols:
-                # clinical symbols from gene panels might not be up to date with latest gene names
-                # but their HGNC id would still match
-                outdated_symbols.append(hgnc_symbol)
-        else:  # clinical variants
-            non_clinical_symbols.append(hgnc_symbol)
+
+        for hgnc_gene in hgnc_genes:
+            if is_clinical is False:  # research variants
+                updated_hgnc_symbols.append(hgnc_symbol)
```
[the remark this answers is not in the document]
northwestwitch: YES!

dnil:
One more thing: on the latest few I get double appends to the outdated_symbols for non-clinical genes on my local?
Screenshot 2021-12-07 at 16 35 54

northwestwitch:
> One more thing: on the latest few I get double appends to the non_clinical_symbols for non-clinical genes on my local? Screenshot 2021-12-07 at 16 35 54

Ah good, let's use sets perhaps?

dnil:
> Ah good, let's use sets perhaps?
I was thinking the same before!

dnil:
Bugs can be painful, but this looks so much cleaner already! Lets check tomorrow again with fresh brains, but well done already!

dnil:
Tested on stage. Works well with both alias and current symbol.
Screenshot 2021-12-08 at 10 45 31

dnil:
Just to document, the duplication problem
Screenshot 2021-12-08 at 10 49 51
 on local is fixed, and missing gene warning works again. 
Screenshot 2021-12-08 at 10 49 31

tests/adapter/mongo/test_gene_handler.py
```diff
@@ -420,6 +420,24 @@ def test_hgnc_symbol_to_gene(adapter):
...
+
+    # THEN using the function to search the gene by symbol should return a list with the gene
+    result = adapter.gene_by_symbol_or_aliases(symbol=parsed_gene["hgnc_symbol"])
+    assert type(result) == list
+    assert len(result) == 1
+
+    # GIVEN a gene symbol not found in "hgnc_symbol" field of any gene
+    # The function should return a pymongo iterable (results of searching genes by alias)
+    result = adapter.gene_by_symbol_or_aliases(symbol="FOO")
+    assert type(result) != list
```
dnil, line 438: Sure. But in this case also probably NoneType? Still, works out fine either way really so keep it like this, or possibly add None to the comment above. 😆

dnil (approved):
All good. I'd still suggest moving the warning for alias used up to cover also single gene hits. That would probably be more common. Or just removing it. There should still be a warning for outdated symbols used anyway.

tests/adapter/mongo/test_gene_handler.py
northwestwitch: No it's an empty mongomock cursor

scout/server/blueprints/variants/controllers.py
```diff
@@ -1063,29 +1063,45 @@ def check_form_gene_symbols(
...
-                # clinical symbols from gene panels might not be up to date with latest gene names
-                # but their HGNC id would still match
-                outdated_symbols.append(hgnc_symbol)
-        else:  # clinical variants
-            non_clinical_symbols.append(hgnc_symbol)
+        if not hgnc_genes:
+            not_found_symbols.add(hgnc_symbol)
+            continue
+
+        if len(hgnc_genes) > 1:  # Gene was not found using provided symbol, aliases were returned
```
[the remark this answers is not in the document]
northwestwitch: Right! We could check before the conversion! Great suggestion!

tests/adapter/mongo/test_gene_handler.py
dnil: Ah, right, it is not real_adapter. And evenso it would still be an empty cursor actually, so all good!

scout/server/blueprints/variants/controllers.py
```diff
@@ -1063,29 +1063,48 @@ def check_form_gene_symbols(
...
-        elif hgnc_gene["hgnc_id"] in clinical_hgnc_ids:  # clinical variants
-            updated_hgnc_symbols.append(hgnc_symbol)
-            if hgnc_symbol not in clinical_symbols:
-                # clinical symbols from gene panels might not be up to date with latest gene names
-                # but their HGNC id would still match
-                outdated_symbols.append(hgnc_symbol)
-        else:  # clinical variants
-            non_clinical_symbols.append(hgnc_symbol)
+        if (
+            isinstance(hgnc_genes, list) is False
```
dnil, line 1079: 👍

tests/adapter/mongo/test_gene_handler.py
northwestwitch: I wrote it like this because it's a pain to check if the type is class.Mongomock.cursor, this was easier!

northwestwitch:
> All good. I'd still suggest moving the warning for alias used up to cover also single gene hits. That would probably be more common. Or just removing it. There should still be a warning for outdated symbols used anyway.

I introduced such warning just in case users search for a gene symbol and get results in other genes, so they know!

PR #3031, opened 2021-12-08
Display genes beneath each splice junction track

### LINKED ISSUE #2990: RNA-seq IGV track improvements
Hi,

We have analysed our cases that already have the option to see the "RNA splicing" in IGV and we would like to ask you for some possible additions.

Is it possible to add to our menu options in IGV of RNA-seq data the possibility to set the minimum junction coverage that will be displayed? 
Is it possible to include the gene track more than just once? We know that we can move the track around but sometimes in families it would be easier to have more gene tracks to compare.

Thank you for all your help.
Sandrina Correia

northwestwitch:
Fix #2990 
Display genes track beneath each splice junctions track in igv.js browser instead of just one at the bottom of the page.

**How to test**:
1. Setup a demo database with the modified case config file
2. Go the whichever snv and check the RNA tracks
3. Each one should have a gene track beneath

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by DN

### DIFF
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
@@ -80,7 +80,7 @@
                                 }
                             ]
                           }, // end of sashimi track with data
-                          {
+                          { // genes track
                             name: geneTrack.name,
                             type: geneTrack.type,
                             format: geneTrack.forma,
@@ -93,7 +93,7 @@
                             searchable: true,
                             order: {{counter.loop}},
                             infoURL: "https://www.ncbi.nlm.nih.gov/gene/?term=$$"
-                          },
+                          }, // end of genes track
                         {% endfor %}
                     ] // end of tracks
             };
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
@@ -83,7 +83,7 @@
                           { // genes track
                             name: geneTrack.name,
                             type: geneTrack.type,
-                            format: geneTrack.forma,
+                            format: geneTrack.format,
                             sourceType: geneTrack.sourceType,
                             url: geneTrack.url,
                             indexURL: geneTrack.indexURL,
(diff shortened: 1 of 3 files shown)


northwestwitch:
Before:

image

After:

image

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -45,28 +54,15 @@ <h3>{{case}}: Splice junctions</h3>
...
-                            indexURL: "{{ custTrack.indexURL|replace('%2F','/') }}",
-                            displayMode: "{{ custTrack.displayMode }}",
-                            visibilityWindow: 300000000,
-                            order: Number.MAX_VALUE,
-                            searchable: true
-                          },
-                        {% endif %}
-                      {% endfor %}
                         {% for track in tracks %} // sample tracks
+                        {% set counter.loop = counter.loop + 1 %}
```
northwestwitch, line 58: This is what does the trick!

scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
```diff
@@ -84,6 +80,20 @@ <h3>{{case}}: Splice junctions</h3>
                                 }
                             ]
                           }, // end of sashimi track with data
+                          { // genes track
+                            name: geneTrack.name,
+                            type: geneTrack.type,
+                            format: geneTrack.forma,
```
dnil, line 86:
```suggestion
                            format: geneTrack.format,
```

northwestwitch: Thanks! Weird that the tracks were displayed anyway. The param is not probably required, but let's have it, and correct :)

dnil:
Super! Good call to put one gene track on each junction track - much easier to read. The old gene track was also very narrow, but with the very inclusive UCSC transcript sets shown, I think I will find myself increasing track size of the gene track pretty often. What do you think about having one of them default to a larger track height? Its already improved, so approving as is as well: the other could be a future issue.

dnil (approved):

northwestwitch:
Yes, I also noticed that the gene track is way too narrow and one has to scroll down often. Perhaps let's have it on another PR where we fix the gene track height also in the normal (dna aligns) browser?

PR #3032, opened 2021-12-09
New minor version 4.44

northwestwitch:
### Changed
- [x] Display Gene track beneath each sample track when displaying splice junctions in igv browser
### Fixed
- [x] Fix filter variants by genes
- [x] Genes API endpoint to return a json formatted error when request is malformed
- [x] Typo in ACMG BP6 tooltip

**Review:**
- [x] code approved by DN
- [x] tests executed by DN. CR


northwestwitch:
Ok this is a bit weird: 
image

But looking for a real gene works:
image

And also returning error:
image

But I guess we are good to deploy with that error, nobody is using the api anyway..

dnil:
Sorry, did you paste the wrong image for the first dummy case? 😊 You pasted one where you forgot to escape the `&`, hence the command goes bg and one can't see the output?

dnil:
It works if one escapes the `&`:
Screenshot 2021-12-09 at 11 49 23

northwestwitch:
Yes I can reproduce if I use a dockerized version of Mongo 4.4. I'll fix!

dnil (approved):
All good, latest tested on stage.

PR #3036, opened 2021-12-09
Cancer gene symbols check - fix #3035

### LINKED ISSUE #3035: On current main, cancer variants etc do not check aliases etc
Also, when fixing, let's make sure the warnings are fine also for someone adding 1400 genes to a filter...

dnil:
This PR adds a functionality or fixes a bug (fix #3035).

Screenshot 2021-12-09 at 12 44 06

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -528,7 +528,7 @@ class PanelHandler:
 
     def clinical_symbols(self, case_obj):
         """Return all the clinical gene symbols for a case."""
-        panel_ids = [panel["panel_id"] for panel in case_obj["panels"]]
+        panel_ids = [panel["panel_id"] for panel in case_obj.get("panels", [])]
         query = self.panel_collection.aggregate(
             [
                 {"$match": {"_id": {"$in": panel_ids}}},
@@ -540,7 +540,7 @@ class PanelHandler:
 
     def clinical_hgnc_ids(self, case_obj):
         """Return all the clinical gene hgnc IDs for a case."""
-        panel_ids = [panel["panel_id"] for panel in case_obj["panels"]]
+        panel_ids = [panel["panel_id"] for panel in case_obj.get("panels", [])]
         query = self.panel_collection.aggregate(
             [
                 {"$match": {"_id": {"$in": panel_ids}}},
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -1066,6 +1066,7 @@ def check_form_gene_symbols(
     non_clinical_symbols = set()
     not_found_symbols = set()
     outdated_symbols = set()
+    aliased_symbols = set()
     updated_hgnc_symbols = set()
 
     clinical_hgnc_ids = store.clinical_hgnc_ids(case_obj)
@@ -1083,7 +1084,7 @@ def check_form_gene_symbols(
             isinstance(hgnc_genes, list) is False
         ):  # Gene was not found using provided symbol, aliases were returned
             hgnc_genes = list(hgnc_genes)
-            outdated_symbols.add(hgnc_symbol)
+            aliased_symbols.add(hgnc_symbol)
 
         if not hgnc_genes:
             not_found_symbols.add(hgnc_symbol)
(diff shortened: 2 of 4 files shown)


scout/adapter/mongo/panel.py
```diff
@@ -540,7 +540,7 @@ def clinical_symbols(self, case_obj):
 
     def clinical_hgnc_ids(self, case_obj):
         """Return all the clinical gene hgnc IDs for a case."""
-        panel_ids = [panel["panel_id"] for panel in case_obj["panels"]]
+        panel_ids = [panel["panel_id"] for panel in case_obj.get("panels", [])]
```
northwestwitch, line 543: This way is also safer code!

scout/server/blueprints/variants/controllers.py
```diff
@@ -1071,6 +1071,10 @@ def check_form_gene_symbols(
     clinical_hgnc_ids = store.clinical_hgnc_ids(case_obj)
     clinical_symbols = store.clinical_symbols(case_obj)
 
+    # if no clinical symobols / panels were found loaded, warnings are treated as with research
+    if len(clinical_hgnc_ids == 0) and len(clinical_symbols == 0):
```
northwestwitch: You could have also checked if the case had gene panels, but this works as well! 👍🏻

scout/server/blueprints/variants/controllers.py
```diff
@@ -1079,10 +1083,7 @@ def check_form_gene_symbols(
             isinstance(hgnc_genes, list) is False
         ):  # Gene was not found using provided symbol, aliases were returned
             hgnc_genes = list(hgnc_genes)
-            flash(
-                f"Could not find a gene with symbol '{hgnc_symbol}'. Search was extended to genes using symbol as an alias.",
-                "warning",
-            )
+            outdated_symbols.add(hgnc_symbol)
```
northwestwitch, line 1086: This is already done a few lines down no? In that case could also be the gene panel that contains an outdates symbol.. Do we want to add another key/flash message value to the warnings dictionary to distinguish the 2 cases?

northwestwitch (approved):
Much better! And finally also cancer variants filter will contain the fixes in the search and the warnings. Check my comment regarding the outdated_symbols set!

scout/server/blueprints/variants/controllers.py
dnil: I was considering that: it would be clearers - the only issue is that it would be longer if there are both.

dnil: Maybe its good. Then we can sharpen the warning messages a bit as well - they are intentionally vague as is. 😁

scout/server/blueprints/variants/controllers.py
```diff
@@ -1124,10 +1125,15 @@ def check_form_gene_symbols(
             "label": "warning",
         },
         "outdated_symbols": {
-            "message": "Outdated gene symbols either provided in search or found in the panels used for the analysis.",
+            "message": "Outdated gene symbols found in the clinical panel loaded for the analysis.",
             "gene_list": outdated_symbols,
             "label": "info",
         },
+        "aliased_symbols": {
```
northwestwitch, line 1133: 💯

northwestwitch (approved):

northwestwitch:
This in the end has the potential to generate more types of warnings, but also more precise messages!

scout/server/blueprints/variants/controllers.py
```diff
@@ -1123,10 +1126,15 @@ def check_form_gene_symbols(
             "label": "warning",
         },
         "outdated_symbols": {
-            "message": "Outdated gene symbols either provided in search or found in the panels used for the analysis.",
+            "message": "Outdated gene symbols found in the clinical panel loaded for the analysis.",
```
northwestwitch, line 1129:
```suggestion
            "message": "Outdated gene symbols found in the clinical panels loaded for the analysis.",
```

northwestwitch (approved):

dnil:
> This in the end has the potential to generate more types of warnings, but also more precise messages!

Yes, I like it better this way so far. Should relieve some of the confusion from overlapping warnings, methinks!

scout/server/blueprints/variants/controllers.py
dnil: Yeah, and that could then be done before the other two checks. But then one would have to preset them to like empty lists aso.. CodeFactor is not screaming about the complexity yet. 😬

PR #3037, opened 2021-12-09
Fix gene api queries input check

northwestwitch:
Fix an error when running a gene query using the API with illegal chars query param. Example:

`curl 'localhost:5000/api/v1/genes?query=lalala\&build=sas'`

## How to test locally ##
- Launch a demo app connected to a MongoDB version 4.4
- On main, the query should make the app crash
- Switch to this branch and it should return an empty list instead

**Expected outcome**:
The query should return an empty list

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Check outdated gene symbols and update with aliases also for cancer
 ### Fixed
 - Genes API endpoint to return a json formatted error when request is malformed
+- Gene api queries input check
 
 ## [4.43.1]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 ### Changed
 ### Fixed
+- Genes API endpoint to return a json formatted error when request is malformed
 
 ## [4.43.1]
 ### Added
--- a/scout/server/blueprints/genes/views.py
+++ b/scout/server/blueprints/genes/views.py
@@ -50,6 +50,11 @@ def gene(hgnc_id=None, hgnc_symbol=None):
 def api_genes():
     """Return JSON data about genes."""
     query = request.args.get("query")
-    build = request.args.get("build", "all")
-    json_out = controllers.genes_to_json(store, query, build)
-    return jsonify(json_out)
+    if query is None:
+        return jsonify({"code": 400, "message": "missing 'query' param in request"})
+    build = request.args.get("build", "37")
+    try:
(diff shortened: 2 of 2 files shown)


scout/server/blueprints/genes/views.py
```diff
@@ -50,11 +50,9 @@ def gene(hgnc_id=None, hgnc_symbol=None):
 def api_genes():
     """Return JSON data about genes."""
     query = request.args.get("query")
-    if query is None:
-        return jsonify({"code": 400, "message": "missing 'query' param in request"})
+    if query is None or query.isalnum() is False:
```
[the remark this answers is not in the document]
dnil: Good point. I guess one could add a few extra chars that are allowed, but perhaps even better to check what chars mongo doesn't like?

northwestwitch:
I've been reading the documentation here: https://www.genenames.org/about/guidelines/

And looks like it's fine to have letters numbers and dashes

northwestwitch:
tested locally with Mongodb 4.4: 

image

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Display Gene track beneath each sample track when displaying splice junctions in igv browser
 - Check outdated gene symbols and update with aliases also for cancer
 ### Fixed
-- Fix filter variants by genes
```
dnil, line 13: Ok! Perhaps add like a comma to the line above?

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Display Gene track beneath each sample track when displaying splice junctions in igv browser
 - Check outdated gene symbols and update with aliases also for cancer
```
dnil, line 11:
```suggestion
- Check outdated gene symbols and update with aliases for both RD and cancer variantS
```

scout/server/blueprints/genes/views.py
```diff
@@ -50,11 +50,9 @@ def gene(hgnc_id=None, hgnc_symbol=None):
 def api_genes():
     """Return JSON data about genes."""
     query = request.args.get("query")
-    if query is None:
-        return jsonify({"code": 400, "message": "missing 'query' param in request"})
+    if query is None or query.replace("-", "").isalnum() is False:
```
dnil, line 53: Ok, yes, that should work. I can't think of any gene with other chars in the symbol, and the docs seem good.

dnil (approved):

CHANGELOG.md
dnil: And feel free to edit that in the v4.4 changelog instead - that would be at least as appropriate. 😁

PR #3049, opened 2021-12-10
Decouple scout from broken loqusdbapi instances

### LINKED ISSUE #3048: Avoid crashing if configured loqusdb instance is not running
Loqusdbapi might be down for a number of reasons, especially the staging one. Scout should be able to run anyway.

```
"/home/hiseq.clinical/SERVER/apps/scout-stage/git/scout-stage/scout/server/extensions/loqus_extension.py", line 131, in get_api_loqus_version
  raise ConfigError(f"LoqusDB API url '{api_url}' did not return a valid response.")
scout.exceptions.config.ConfigError: LoqusDB API url 'http://127.0.0.1:28005' did not return a valid response.
```

northwestwitch:
fix #3048

Start Scout also if a loqusdb instance is not reachable. But show a warning on variants pages:

image

**How to test**:
1. Install on clinical-db stage and make sure scout runs and shows the warning described above all all variants pages

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -401,6 +401,15 @@ def observations(store, loqusdb, case_obj, variant_obj):
     Returns:
         obs_data(dict)
     """
+    institute_id = variant_obj["institute"]
+    institute_obj = store.institute(institute_id)
+    loqusdb_id = institute_obj.get("loqusdb_id") or "default"
+    if loqusdb.loqusdb_settings[loqusdb_id]["version"] == None:
+        flash("Could not connect to the preselected loqusdb instance", "danger")
+        return {
+            "total": "N/A",
+        }
+
     chrom = variant_obj["chromosome"]
     end_chrom = variant_obj.get("end_chrom", chrom)
     pos = variant_obj["position"]
@@ -426,13 +435,6 @@ def observations(store, loqusdb, case_obj, variant_obj):
         "variant_type": variant_obj.get("sub_category", "").upper(),
         "category": category,
     }
-
-    institute_id = variant_obj["institute"]
-    institute_obj = store.institute(institute_id)
-    loqusdb_id = institute_obj.get("loqusdb_id") or "default"
-    if loqusdb.loqusdb_settings[loqusdb_id]["version"] == None:
-        flash("Could not connect to the preselected loqusdb instance", "danger")
-
     obs_data = loqusdb.get_variant(variant_query, loqusdb_id=loqusdb_id)
 
     if not obs_data:
@@ -472,7 +474,6 @@ def observations(store, loqusdb, case_obj, variant_obj):
 
         obs_data["cases"].append(dict(case=other_case, variant=other_variant))
 
-    flash(obs_data)
     return obs_data
 
 
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -404,7 +404,7 @@ def observations(store, loqusdb, case_obj, variant_obj):
     institute_id = variant_obj["institute"]
     institute_obj = store.institute(institute_id)
     loqusdb_id = institute_obj.get("loqusdb_id") or "default"
-    if loqusdb.loqusdb_settings[loqusdb_id]["version"] == None:
+    if loqusdb.loqusdb_settings[loqusdb_id]["version"] is None:
         flash("Could not connect to the preselected loqusdb instance", "danger")
         return {
             "total": "N/A",
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -2,7 +2,7 @@ import logging
 from base64 import b64encode
 from datetime import date
 
-from flask import url_for
+from flask import flash, url_for
 from flask_login import current_user
 
 from scout.constants import (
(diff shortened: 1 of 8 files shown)


scout/server/extensions/loqus_extension.py
```diff
@@ -93,13 +93,19 @@ def init_app(self, app):
...
-
-        self.loqus_ids = self.loqusdb_settings.keys()
-        LOG.debug(f"LoqusDB setup: {self.__repr__()}")
+            try:
+                self.version_check(setting)
+                self.loqus_ids = self.loqusdb_settings.keys()
+                LOG.debug(f"LoqusDB setup: {self.__repr__()}")
+            except EnvironmentError as env_ex:
+                LOG.warning(env_ex)
+                app.config["LOQUSDB_SETTINGS"] = "Fail"
```
dnil, line 102: 👍

scout/server/extensions/loqus_extension.py
```diff
@@ -93,13 +93,19 @@ def init_app(self, app):
...
 
     def version_check(self, loqusdb_settings):
         """Check if a compatible version is used otherwise raise an error"""
+        if loqusdb_settings["version"] is None:
+            raise EnvironmentError("LoqusDB instance not available")
```
dnil, line 107: 👍

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -169,9 +169,13 @@
         </thead>
         <tbody>
           <tr>
+            {% if config.LOQUSDB_SETTINGS == 'Fail' %}
+              <td style="background-color: #ff9999;" colspan=3>Failed to contact loqusdb server</td>
+            {% else %}
```
dnil, line 174: Indent?

dnil (approved):

scout/server/blueprints/variant/templates/variant/variant_details.html
northwestwitch: On the left?

dnil: Its html, so personal preference! I would at least do the same with the two if-else blocks, like either move the above left, or move the tds below right.. 😸

northwestwitch:
I've changed the error message to this:

image

Because could be that your institute is using another loqusdb instance that is working instead?

northwestwitch:
I'm still not entirely satisfied with this PR, because if there are present different loqusdb instances and a customer is using one that runs, then there is no point in showing the error message. I'll try to fix it so it shows only when it's needed

dnil:
Another thing that you could include if you feel like fiddling with it is an error for if the CLI or API did not return a result on the query for some reason. We have N/A but a verbose one would be nice.

northwestwitch:
> Another thing that you could include if you feel like fiddling with it is an error for if the CLI or API did not return a result on the query for some reason. We have N/A but a verbose one would be nice.

This is something we could do, but I'd rather no do it in this PR. This was more something meant to be able to test stage even if can't use loqusdbapi at the moment..

dnil:
> > Another thing that you could include if you feel like fiddling with it is an error for if the CLI or API did not return a result on the query for some reason. We have N/A but a verbose one would be nice.
> 
> This is something we could do, but I'd rather no do it in this PR. This was more something meant to be able to test stage even if can't use loqusdbapi at the moment..

All good: just though about it due to the placement of the warning.

northwestwitch:
The branch is installed on clinical-db stage now.
Changing the loqus instance in use for cust000 from "prod-cli" to "default" triggers a warning every time the user is on a variant page:

image

scout/server/blueprints/variant/controllers.py
```diff
@@ -401,6 +401,15 @@ def observations(store, loqusdb, case_obj, variant_obj):
     Returns:
         obs_data(dict)
     """
+    institute_id = variant_obj["institute"]
+    institute_obj = store.institute(institute_id)
+    loqusdb_id = institute_obj.get("loqusdb_id") or "default"
+    if loqusdb.loqusdb_settings[loqusdb_id]["version"] == None:
+        flash("Could not connect to the preselected loqusdb instance", "danger")
+        return {
+            "total": "N/A",
```
dnil, line 410: 💯

tests/server/conftest.py
```diff
@@ -20,6 +20,14 @@ class LoqusdbMock:
     def __init__(self):
         self.nr_cases = 130
         self.variants = {"1_169898014_T_C": {"families": ["vitalmouse"]}}
+        self.loqusdb_settings = {
+            "default": {
+                "api_url": "http://127.0.0.1:9000",
+                "instance_type": "api",
+                "id": "default",
+                "version": "2.5",
```
dnil, line 28: And this version-of-whatever-the-api-talks-to is kind of silly, but it is not for this PR.

scout/server/extensions/loqus_extension.py
```diff
@@ -93,13 +93,18 @@ def init_app(self, app):
                 setting["version"] = "2.5"
             else:
                 setting["version"] = self.get_instance_version(setting)
-            self.version_check(setting)
-
-        self.loqus_ids = self.loqusdb_settings.keys()
-        LOG.debug(f"LoqusDB setup: {self.__repr__()}")
+            try:
+                self.version_check(setting)
+            except EnvironmentError as env_ex:
```
dnil, line 98: 👍

dnil (approved):
All of it kind of violates the fail-early idea, but it is for good cause. 😸

tests/server/conftest.py
northwestwitch: It's just an example for the test!

PR #3051, opened 2021-12-13
Bump version 4.45

dnil:
This PR marks a new Scout release. We apply semantic versioning. This is a minor release for reasons, since it fixes issues while slightly changing the functionality.

### Changed
- [x] Start Scout also when loqusdbapi is not reachable
- [x] Clearer definition of manual standard and custom inheritance models in gene panels
### Fixed
- [x] Gene panel crashing on edit action

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR


northwestwitch (approved):
👍🏻

PR #3053, opened 2021-12-14
Create a Dockerfile for serving the app via gunicorn - with GitHub action

northwestwitch:
- Created a Dockefile to be used to serve the dockerized app in production
- Modified the code to collect database params specified as env vars
- Created a GitHub action that pushes the Dockerfile-server image to scout-server-stage every time a PR is opened

**How to test**:
1. Run docker-compose up and make sure the demo app is served on **localhost:8000**

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [ ] tests executed by

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,13 +28,21 @@ services:
       - mongodb
 
   scout-web:
-    build: .
-    container_name: scout-web
+    build:
+      context: .
+      dockerfile: Dockerfile-server
     expose:
-      - '5000'
+      - '8000'
     ports:
-      - '5000:5000'
-    command: scout --host mongodb --demo serve --host 0.0.0.0
+      - '8000:8000'
+    environment:
+      SCOUT_CONFIG: /home/worker/app/scout/server/config.py
+      MONGO_HOST: mongodb
+      MONGO_DBNAME: scout-demo
+      GUNICORN_WORKERS: 1
+      GUNICORN_TREADS: 1
+      GUNICORN_BIND: 0.0.0.0:8000
+      GUNICORN_TIMEOUT: 400
     volumes:
       - ./scout:/home/worker/app/scout
     networks:
--- a/docs/admin-guide/server.md
+++ b/docs/admin-guide/server.md
@@ -4,78 +4,111 @@ This document will explain how we like to run Scout in both development and prod
 
 ## Config
 
-The server has it's own config file, separate from the command line tool. It's written in Python for full expressivity. These are the settings you can use:
+In order to run the web server, it is usually necessary to provide a number of parameters. All available parameters should be specified in a python-formatted configuration file. This config file is passed as a parameter when the application is launched. The following example of a config file is is used for running the development server (read further down):
 
 ```python
-# flask.conf.py
+SECRET_KEY = "this is not secret..."
+REMEMBER_COOKIE_NAME = "scout_remember_me"
 
-# to encrypt cookie data
-SECRET_KEY = 'makeThisSomethingNonGuessable'  # required
+# MONGO_URI = "mongodb://127.0.0.1:27011,127.0.0.1:27012,127.0.0.1:27013/?replicaSet=rs0&readPreference=primary"
+MONGO_DBNAME = "scout"
 
-# connection details for MongoDB
-MONGO_DBNAME = 'scout'                        # required
-MONGO_PORT = 27017
-MONGO_USERNAME = 'mongoUser'
-MONGO_PASSWORD = 'mongoUserPassword'
+BOOTSTRAP_SERVE_LOCAL = True
+TEMPLATES_AUTO_RELOAD = True
+
+DEBUG_TB_INTERCEPT_REDIRECTS = False
+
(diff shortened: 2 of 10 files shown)


docker-compose.yml
```diff
@@ -28,13 +28,21 @@ services:
...
+      dockerfile: Dockerfile-server
     expose:
-      - '5000'
+      - '8000'
     ports:
-      - '5000:5000'
-    command: scout --host mongodb --demo serve --host 0.0.0.0
+      - '8000:8000'
+    environment:
+      SCOUT_CONFIG: /home/worker/app/scout/server/config.py
```
northwestwitch, line 39: This is the env var that should point tp the Scout config file

dnil: Still need the yaml as well for the CLI I guess, though not really the info in it anymore. But that's for laterz.

scout/server/extensions/mongo_extension.py
```diff
@@ -9,14 +11,14 @@ class MongoDB:
     def init_app(app):
         """Initialize from flask"""
 
-        db_name = app.config.get("MONGO_DBNAME", "scout")
+        db_name = os.environ.get("MONGO_DBNAME") or app.config.get("MONGO_DBNAME", "scout")
```
dnil, line 14: Ah, ok, you don't want to pass the yaml? Or why do you want to set these from ENV? 😊

northwestwitch: I wasnt't thinking about the yaml file. In prod settings (or even development settings) you don't need to set up these env vars since you are passing the path to the config file. But you might need them in the eventuality you can't change the demo config file but still want to customize the database you are connecting to (like in out demo docker-compose setup)

northwestwitch: Because otherwise in the demo config file the database host is localhost (not mongodb), port 27017 (which is fine), and database name is scout (not scout-stage)

dnil: Well, yes, it is a demo so all good. For a prod solution, we would of course really just like to skip the yaml. And as a fallback option pass it. But that could be done with like mounting over the default file location when firing up the instance, so not really a big issue, right?

dnil (approved):
Partially tested with https://github.com/Clinical-Genomics/servers/pull/714

scout/server/extensions/mongo_extension.py
northwestwitch: I don't think we should keep using the yaml file, especially in prod. 

When launching the cli, we could fix it that we pass database uri, database name and OMIM api key as env vars instead. As an alternative we might add the OMIM api key as one of the params of the .py config file (of course we should make sure that the app reads it and uses it.

So we can deprecate the use of the yaml file for good! 😀

northwestwitch:
I've updated the admin docs with the newest config file and included the missing parameters in the demo config file as well. If it's fine with you @dnil I'd merge this one!

docs/admin-guide/server.md
```diff
@@ -123,6 +156,8 @@ gunicorn \
     scout.server.auto:app
 ```
 
+Note that while it might still be necessary to provide a `SCOUT_CONFIG` environment variable pointing to python config file to run a full web server, some parameters like those used to connect to MongoDB (`MONGO_HOST`, `MONGO_DBNAME`, `MONGO_PORT`, `MONGO_USERNAME`, `MONGO_PASSWORD`, `MONGO_URI`) can be also be passed as environment variable. In that case they will have precedence over those provided in the config file.
```
dnil, line 159:
```suggestion
Note that while it might still be necessary to provide a `SCOUT_CONFIG` environment variable pointing to python config file to run a full web server, some parameters like those used to connect to MongoDB (`MONGO_HOST`, `MONGO_DBNAME`, `MONGO_PORT`, `MONGO_USERNAME`, `MONGO_PASSWORD`, `MONGO_URI`) can be also be passed as environment variables. In that case they will have precedence over those provided in the config file.
```

docs/admin-guide/server.md
```diff
@@ -106,7 +139,7 @@ FLASK_DEBUG=1 \
 SCOUT_CONFIG="/full/path/to/flask.conf.py" \
 flask run
 ```
-The server can also be started from the scout cli with the command `scout serve`. There are multiple options here. To use a alternative flask config pass the path with option `-f/ --flask_config` like `scout -f path/to/flask.conf.py serve`.
+The server can also be started from the scout cli with the command `scout serve`. The default config file that will be used when no other configuration option is provided will be the one present under scout/server/config.py. To use an alternative flask config, you can use the `-f/ --flask_config` option. Example: `scout -f path/to/flask.conf.py serve`.
```
dnil, line 142: 👍

docs/admin-guide/server.md
northwestwitch, line 159: Added this part to document the use of the environment variables for the MongoDB connection params

northwestwitch: Ah yes, thanks!

PR #3060, opened 2021-12-16
Managed variant csv file verification

dnil:
This PR adds a functionality and updates documentation. 

**How to test**:
1. how to test it, possibly with real cases/data:
- try loading a managed variants file with a faulty line, e.g.
```
## Managed variants from some reliable source
#chromosome;position;end;reference;alternative;category;sub_category;description
14;76548781;76548781;CTGGACC;G;snv;indel;IFT43 indel test
17;48696925;48696925;G;T;snv;snv;CACNA1G intronic test
;;;T;C;snv;snv;silly
7;124491972;124491972;C;A;snv;snv;POT1 test snv
```
- ensure uploading a normal example still works

**Expected outcome**:
Screenshot 2021-12-17 at 09 15 48

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/managed_variants/controllers.py
```diff
@@ -141,6 +148,32 @@ def upload_managed_variants(store, lines, institutes, current_user_id):
     return new_managed_variants, total_variant_lines
 
 
+def validate_managed_variant(managed_variant_info):
```
northwestwitch, line 151: This could also be easily done in the HTML code by adding the "required" key in the input fields. At that level you could even specify that position should be a number. This way the form won't be submitted until you add all the required fields. So you can skip the form validation after, and also adding the tests to the form validation..

northwestwitch (commented):
Check out my suggestion!

scout/server/blueprints/managed_variants/controllers.py
northwestwitch: https://www.w3schools.com/tags/att_input_required.asp

docs/user-guide/variants.md
```diff
@@ -124,6 +124,36 @@ The out put files of expansion hunter if preferably annotated with [stranger][st
 ### Compounds (top 20)
 Only interesting when the compound inheritance pattern is required, the list can be very long but is by default cropped to the top 20 highest ranked ones as shown in the heading.
 
+## Managed variants
+
+### Managed variants upload file format
```
northwestwitch, line 129: 🏅

scout/server/blueprints/managed_variants/controllers.py
dnil: Im sorry, I dont quite follow? How would you do that for the file parsing? We could (and if we dont should) do that for the manual single variant add form, but the file parser would still need a validator?

dnil: Ah, ok, looks like I skipped making them required in the other form in favour of easier inheritance. I can fix that. It wasn't really in the scope of the PR, but I agree. It is nicer if the user can't even click submit, though there should of course be validation later in order not to trust this to the browser.

northwestwitch: Ah yes, your are right, I didn't even test the submitted file with missing fields (topic of THIS PR), I've just try to click submit when the field is empty 😅

docs/user-guide/variants.md
```diff
@@ -124,6 +124,36 @@ The out put files of expansion hunter if preferably annotated with [stranger][st
...
+The managed variants csv file is a tab or semicolon (;)-separated text file with a mandatory header that describes the columns and one line for each gene entry. You must use the same delimiter for the whole file. Ideally do not use the delimiter characters in other places in the file. Consult with an admin if you need to use the delimiter characters in other fields for help with escaping them or ensuring a higher priority separator is used on a previous line.
```
northwestwitch, line 130: If it's tab or semicolon-separated perhaps let's not name it csv, otherwise users might think they have to use commas as separators?

northwestwitch, line 130:
```suggestion
The managed variants text file is a tab or semicolon (;)-separated text file with a mandatory header that describes the columns and one line for each gene entry. You must use the same delimiter for the whole file. Ideally do not use the delimiter characters in other places in the file. Consult with an admin if you need to use the delimiter characters in other fields for help with escaping them or ensuring a higher priority separator is used on a previous line.
```

docs/user-guide/variants.md
```diff
@@ -124,6 +124,36 @@ The out put files of expansion hunter if preferably annotated with [stranger][st
...
+- **sub_category(str)** Sub category of variant. For "snv" and "cancer" this is either "snv" or "indel". For "sv" and "cancer_sv" this is one of "ins", "del", "dup", "cnv", "inv" or "bnd". *Mandatory*
+- **build(str)** Defaults to "37". *Optional*
+- **description(str)** Free text description of the variant. Classification status, evidence and references are appropriate, and if applicable affected gene(s), protein consequences and alternative variant names. *Optional*
+- **maintainer(str)** Placeholder, replaced by the current uploading user. *Optional placeholder*
+- **institutes(list(str))** Placeholder, replaced with the current uploading user institutes. *Optional placeholder*
+
+Example:
+
+```csv
+## Managed variants from some reliable source
```
northwestwitch, line 150: 👍🏻

northwestwitch (approved):
Works perfectly! 🥇 

image

I have a few suggestions:

- Since we are touching the functionality: improving the looks of this form by placing all  elements on one line, with the right spacing and vertical padding.

image


- In the form above replace `CSV file` with `text file`, since it's either tab or semicolon-separated text file

- Add `required` to the required fields in the other form, when you add a single variant, since it's crashing as well whenever the right fields are not found
image

docs/user-guide/variants.md
dnil: Sure, this is mostly MS-legacy terminology.

tests/server/blueprints/managed_variants/test_managed_variants_controllers.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+from scout.server.blueprints.managed_variants import controllers
+
+
+def test_upload_managed_variants(adapter, managed_variants_lines):
```
northwestwitch, line 6: Nice!

dnil:
Thank you for the review!
I fixed the upload button layout, added the same file-name-updater as for gene panels, and made mandatory fields `required` on both add and edit forms.

dnil:
Meh, but now the edit is partially broken. I'll check, but perhaps after dinner. 😅 

Screenshot 2021-12-20 at 16 53 19

dnil:
Ok, it was just a small thing, broken with an earlier update..

PR #3063, opened 2021-12-17
Gene panels api endpoint - fix #3001

### LINKED ISSUE #3001: Communicate the date when a gene panel was updated last time to melter
[melter](https://github.com/Clinical-Genomics/melter) keeps track of cases flagged for perpetual monitoring in scout and starts new analysis of such cases when appropriate. One criteria for redoing an old analysis is if the gene panel has been updated. To make this feature work scout could communicate which date the gene panel was updated last time (and that could then be compared by melter to when the analysis was done the last time).

This issue can be closed when:
- [ ] scout can communicate to melter the date a gene panel was updated last time
or
- [ ] melter can identify if a gene panel has been updated since the last analysis (with another strategy than mentioned above) - please give suggestions if we should close the issue via this option

dnil:
This PR adds a functionality: gene panels JSON REST API. Will return all versions of a named panel, including update dates as key `dates`. This date can be compared to the date (and if desired version and later actual gene changes) on the monitored case `case.panels[i].updated_at` for matching `panel_name` in `case.panels[i].panel_name`. 

**How to test**:
1. Attempt to retrieve a demo panel via the public API 
```
$ curl http://localhost:5000/api/v1/panels/panel1 
[{"_id": "5ef44f61d020103d9c9ad74e", "panel_name": "panel1", "institute": "cust000", "version": 1.0, "date": "2016-12-09T00:00:00", "maintainer": [], "display_name": "Test panel", "description": null, "genes": [{"hgnc_id": 7481, "symbol": "MT-TF"},
```

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/panels/controllers.py
+++ b/scout/server/blueprints/panels/controllers.py
@@ -267,8 +267,8 @@ def panel_export(store, panel_obj):
     return dict(panel=panel_obj)
 
 
-def panels_to_json(store, panel_name):
-    """Fetch matching gene panels and convert to JSON."""
+def get_panels(store, panel_name):
+    """Fetch matching gene panels and return a list."""
     gene_panels = list(store.gene_panels(panel_id=panel_name, include_hidden=True))
 
     return gene_panels
--- a/scout/server/blueprints/panels/controllers.py
+++ b/scout/server/blueprints/panels/controllers.py
@@ -269,6 +269,8 @@ def panel_export(store, panel_obj):
 
 def panels_to_json(store, panel_name):
     """Fetch matching gene panels and convert to JSON."""
-    gene_panels = store.gene_panels(panel_id=panel_name, include_hidden=True)
+    gene_panels = list(store.gene_panels(panel_id=panel_name, include_hidden=True))
 
-    return list(gene_panels)
+    log.debug("gene panels %s", gene_panels)
+
+    return gene_panels
--- a/scout/server/blueprints/panels/controllers.py
+++ b/scout/server/blueprints/panels/controllers.py
@@ -265,3 +265,10 @@ def panel_export(store, panel_obj):
     panel_obj["name_and_version"] = full_name
 
     return dict(panel=panel_obj)
+
+
+def panels_to_json(store, panel_name):
+    """Fetch matching gene panels and convert to JSON."""
+    gene_panels = store.gene_panels(panel_id=panel_name, include_hidden=True)
+
+    return list(gene_panels)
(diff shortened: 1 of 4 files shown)


scout/server/blueprints/panels/controllers.py
```diff
@@ -265,3 +265,10 @@ def panel_export(store, panel_obj):
     panel_obj["name_and_version"] = full_name
 
     return dict(panel=panel_obj)
+
+
+def panels_to_json(store, panel_name):
```
northwestwitch, line 270: This function is named panels_to_json but it's not really what it does. The json conversion happens in the view instead. Perhaps you don't even need it..

scout/server/blueprints/panels/views.py
```diff
@@ -16,6 +17,17 @@
 panels_bp = Blueprint("panels", __name__, template_folder="templates")
 
 
+@panels_bp.route("/api/v1/panels/<panel_name>", methods=["GET", "POST"])
+@public_endpoint
+def api_panels(panel_name):
+    """Return JSON data about panels with a given panel name.
+    Returns all versions.
+    """
+    json_out = controllers.panels_to_json(store, panel_name)
```
northwestwitch, line 26: Why not just calling the store.gene_panels from here? I am a big fun of using controllers to avoid too much code in the views, but in this case you are replacing a line with another line..

northwestwitch (changes requested):
The endpoint works perfectly. I'm not approving at once because the new controller function has a misleading name "panels_to_json", see my comment. I would in fact just remove that function and invoke the adapter from the view, but if you like it this way it's fine. Just perhaps rename it to get_panel or something?

scout/server/blueprints/panels/views.py
dnil: Same here, so I'd like to keep it like that. Compare the main view for panels, where there are now multiple calls to the store directly. It gets complex fast, so one extra context switch could be worth it.

scout/server/blueprints/panels/controllers.py
dnil: It's the same as the other API calls, but I do agree!

scout/server/blueprints/panels/controllers.py
```diff
@@ -265,3 +265,10 @@ def panel_export(store, panel_obj):
     panel_obj["name_and_version"] = full_name
 
     return dict(panel=panel_obj)
+
+
+def get_panels(store, panel_name):
```
northwestwitch, line 270: 👍🏻

northwestwitch (approved):

henningonsbring:
"including update dates as key dates" @dnil can you clarify?

Is the key `date` or `dates`?

henningonsbring:
will the key date/dates point on a datetime object? would be great if it would for consistency with the date here: https://github.com/Clinical-Genomics/scout/pull/3023

PR #3064, opened 2021-12-20
Speed-up CI tests further by randomizing test groups contents

northwestwitch:
CI testing now is a bit faster, but splitting tests by groups makes it so that there are some groups (for instance group 5) that contain slower tests than other groups. Randomizing group contents should re-distribute slower tests into all groups, making them a but more balanced. This way the entire process should be faster.

**How to test**:
1. Check the automatic tests in the GitHub actions

**Expected outcome**:
- [x] Tests executed in test groups should be in random order
- [x] CI tests execution time should be faster than in the last merged PR

**Review:**
- [x] code approved by DN
- [x] tests executed by GitHub action

### DIFF
--- a/.github/workflows/tests_and_cov.yml
+++ b/.github/workflows/tests_and_cov.yml
@@ -61,7 +61,7 @@ jobs:
           pip install --upgrade --upgrade-strategy eager -r requirements-dev.txt -e .
 
       - name: Run pytest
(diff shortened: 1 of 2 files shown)


northwestwitch:
Result is promising!

image

dnil (approved):

PR #3067, opened 2021-12-20
Simplistic CDN alternative - take three

dnil:
This PR adds a functionality.

A student had a need for a local scout for use inside a highly secure facility (no direct internet). This solves the issue. Just run it on a machine with web access. It will dl all css and js used in the code, and edit accordingly to make the pages use the local copy, like a very simplistic CDN service.

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Now the command is working: 

image


Tested it by:
- Going offline
- Removing cached images and files

Start the demo server: 😨 

image

dnil:
Ok, that should be it again. I didn't include the debug maps for css, so you will probably still see some issues with resource loading, but I think all styles and scripts should be covered. 😊

northwestwitch:
Much better, but some libs are still missing, for instance font-awesome, that is used in many pages:

image


image

dnil:
Ah, ok the fa one we have (now?) includes some stuff from paths relative to its server, like `format('embedded-opentype'),url('../fonts/fontawesome-webfont.woff?v=4.2.0')`. Ok, lets see..

dnil:
Ok, I think this served its purpose. If needed later, I won't remove the branch, but for a complete version one would have to do a deeper pull, saving also the files that are fetched by the js that is explicitly imported.

PR #3069, opened 2021-12-21
Make phenotype non-mandatory when assigning partial causatives

### LINKED ISSUE #2562: partial causative function
image

I wish to use partially causative

What can i do if there are not OMIM or HPO that are appropriate?

can we have a free text option? Or the possibility to select the partially causative without being forced to write an explanation?

northwestwitch:
Fix #2562 --> Specify or not an HPO or OMIM phenotype when marking a variant as partial causative

**How to test -- locally or on clinical-db stage**:
- [x] Try to mark a variant [snv, sv, cancer] as partial causative without adding a diagnosis or phenotype
- [x] Mark another one as partial causative but specify phenotype or diagnosis  
- [x] Check the case report and see that it works and displays the right info for both variants.

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by Mikael


mikaell (approved):
Looks good! 99% removal of code 😎

northwestwitch:
> Looks good! 99% removal of code

Removing stuff is always nice, thanks for the review!

PR #3070, opened 2021-12-21
Remove hgnc_genes_find_one function, fix code that used it and bug in variants search form

### LINKED ISSUE #3030: Get rid of 'hgnc_genes_find_one' function and fix code where it is used
This function returns a random gene which has a given alias: https://github.com/Clinical-Genomics/scout/blob/7023c99abdd4f7dcfa67d26c571e41a8ecd3cbc1/scout/adapter/mongo/hgnc.py#L175. Results are unpredictable and can lead to errors.

Once #3026 is merged, the function above will only be used while loading gene panels [from the variants page (variants filters?)](https://github.com/Clinical-Genomics/scout/blob/c89cc3c0025ea918d1430dec07ccbcd92326ca0d/scout/server/blueprints/variants/controllers.py#L802). Call another function that returns a more appropriate result instead!

northwestwitch:
Fix #3030

This function was still used in 2 places:
- Variants filter, searching for variants matching a gene symbol list provided in an uploaded file. This functionality is broken at the moment
image

- Institute pages, gene variants search (Search SNVs and INDELs). In this case I'm proposing a change to the code that will **simplify** and speed-up the variants search

**How to test**:
1. From master branch, create a simple txt file containing a list of gene symbols, some of which don't exist in the database: example:
POT1
ELN
HELLO

Provide this file for filtering variants in a demo instance and make sure that the functionality is not working as intended.

2. Switch to this branch and make sure that the filtering works

3. Go to institute page, gene variants functionality and search for the same gene using the form:
image

4. Make sure results are returned

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -172,21 +172,6 @@ class GeneHandler(object):
             )
         return self.hgnc_collection.find({"aliases": hgnc_symbol, **build_query})
 
-    def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
-        """Find one hgnc genes that match a hgnc symbol.
-
-        Check both hgnc_symbol and aliases
-
-        Args:
-            hgnc_symbol(str)
-            build(str): The build in which to search
-        Returns:
-           One single hgnc document
-        """
-
-        LOG.debug("Find one genes with symbol %s" % hgnc_symbol)
-        return self.hgnc_collection.find_one(filter={"build": build, "aliases": hgnc_symbol})
-
     def get_query_alias_or_id(self, hgnc_symbol, build):
         """Return query to search for hgnc-symbol or aliases"""
         query = {
--- a/scout/adapter/mongo/hgnc.py
+++ b/scout/adapter/mongo/hgnc.py
@@ -330,7 +330,7 @@ class GeneHandler(object):
         if res:
             return [res]
 
-        return self.gene_aliases(symbol, build="37")
+        return self.gene_aliases(symbol, build=build)
 
     def gene_aliases(self, symbol, build="37"):
         """Return an iterable with hgnc_genes which have the provided symbol in the gene aliases
--- a/scout/server/blueprints/institutes/views.py
+++ b/scout/server/blueprints/institutes/views.py
@@ -9,6 +9,7 @@ from flask_login import current_user
 from werkzeug.datastructures import Headers
 
 from scout.constants import ACMG_COMPLETE_MAP, ACMG_MAP, CASEDATA_HEADER, CLINVAR_HEADER
+from scout.server.blueprints.variants.controllers import update_form_hgnc_symbols
 from scout.server.extensions import loqusdb, store
 from scout.server.utils import institute_and_case, jsonconverter, templated, user_institutes
 
@@ -144,6 +145,10 @@ def gene_variants(institute_id):
         form.variant_type.data = ["clinical"]
 
     variant_type = form.data.get("variant_type")
+
+    # check if supplied gene symbols exist
+    if form.hgnc_symbols.data:
+        update_form_hgnc_symbols(store=store, case_obj=None, form=form)
 
     variants_query = store.gene_variants(
         query=form.data,
--- a/scout/server/blueprints/institutes/views.py
+++ b/scout/server/blueprints/institutes/views.py
@@ -145,58 +145,21 @@ def gene_variants(institute_id):
 
     variant_type = form.data.get("variant_type")
 
-    # check if supplied gene symbols exist
-    hgnc_symbols = []
-    not_found_symbols = []
-    not_found_ids = []
-    data = {}
-    if (form.hgnc_symbols.data) and len(form.hgnc_symbols.data) > 0:
-        for hgnc_symbol in form.hgnc_symbols.data:
-            if hgnc_symbol.isdigit():
-                hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol), build="37")
+    variants_query = store.gene_variants(
+        query=form.data,
+        institute_id=institute_id,
+        category="snv",
+        variant_type=variant_type,
+    )
 
-                if hgnc_gene_caption is None:
-                    hgnc_gene_caption = store.hgnc_gene_caption(int(hgnc_symbol), build="38")
+    result_size = store.count_gene_variants(
+        query=form.data,
+        institute_id=institute_id,
+        category="snv",
+        variant_type=variant_type,
+    )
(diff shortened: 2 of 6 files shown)


northwestwitch:
Test fixing the variants filtering by gene file (containing genes: POT1, ELN, HELLO, AJEJE)

Master branch is broken: ⛔ 
image


This branch works: 🆗 
image

<hr>

Test searching gene variants with the same genes above: 🆗 

image

scout/adapter/mongo/hgnc.py
```diff
@@ -330,7 +315,7 @@ def gene_by_symbol_or_aliases(self, symbol, build="37"):
         if res:
             return [res]
 
-        return self.gene_aliases(symbol, build="37")
+        return self.gene_aliases(symbol, build=build)
```
northwestwitch, line 318: This was a potential bug introduced not so long ago, when I modified this code

scout/server/blueprints/institutes/views.py
```diff
@@ -145,58 +145,21 @@ def gene_variants(institute_id):
 
     variant_type = form.data.get("variant_type")
 
-    # check if supplied gene symbols exist
```
dnil, line 148: Good; the update part of this is duplicated in the controller now. But, there is no user feedback from there. Perhaps add flashes with warnings there, or maybe it's possible to reuse one of the regular variant routines for that even? I'm uneasy about skipping warnings altogether - its so easy to typo a gene symbol.

scout/adapter/mongo/hgnc.py
```diff
@@ -172,21 +172,6 @@ def hgnc_genes(self, hgnc_symbol, build="37", search=False):
             )
         return self.hgnc_collection.find({"aliases": hgnc_symbol, **build_query})
 
-    def hgnc_genes_find_one(self, hgnc_symbol, build="37"):
```
dnil, line 175: Good riddance! 🥇

scout/server/blueprints/variants/controllers.py
```diff
@@ -805,13 +805,17 @@ def upload_panel(store, institute_id, case_name, stream):
     raw_symbols = [
         line.strip().split("\t")[0] for line in stream if line and not line.startswith("#")
     ]
-    # check if supplied gene symbols exist
-    hgnc_symbols = []
+    # check if supplied gene symbols exist, search genes by official symbol or alias
+    hgnc_symbols = set()
```
dnil, line 809: 👍

tests/server/blueprints/institutes/test_institute_views.py
```diff
@@ -733,33 +733,6 @@ def test_gene_variants_filter(app, institute_obj, case_obj):
         assert "POT1" in str(resp.data)
 
 
-def test_gene_variants_no_valid_gene(app, institute_obj, case_obj, mocker, mock_redirect):
```
dnil, line 736: I would suggest to keep this functionality; doesn't have to be very complicated.

dnil (changes requested):
This works well, fixes a bug, and removes a function that easily gets abused. The only problem with the current state of the fix is it drops a warning that was kind of useful - and you had to drop a test for the warning to get that through... 😉

scout/server/blueprints/institutes/views.py
northwestwitch: Good suggestion, I can try to re-use code already present in variantS controller, that does exactly the same thing!

northwestwitch: Reintroduced genes checks:

image

scout/server/blueprints/institutes/views.py
```diff
@@ -146,57 +147,24 @@ def gene_variants(institute_id):
...
-            )
-
-        if hgnc_symbols == []:
-            # If there are not genes to search, return to previous page with a warning
-            flash("No valid gene provided for variant search.", "warning")
-            return redirect(request.referrer)
-
-        form.hgnc_symbols.data = hgnc_symbols
+    if form.hgnc_symbols.data:
+        update_form_hgnc_symbols(store=store, case_obj=None, form=form)
```
northwestwitch, line 151: This is calling functions in variants.controllers that check gene validity and print eventual errors

scout/server/blueprints/variants/controllers.py
```diff
@@ -1045,6 +1049,54 @@ def populate_sv_filters_form(store, institute_obj, case_obj, category, request_o
     return form
 
 
+def _flash_gene_symbol_errors(
```
northwestwitch, line 1052: Code here is just moved from another function to decrease its complexity

scout/server/blueprints/variants/controllers.py
```diff
@@ -1069,10 +1121,10 @@ def check_form_gene_symbols(
     aliased_symbols = set()
     updated_hgnc_symbols = set()
 
-    clinical_hgnc_ids = store.clinical_hgnc_ids(case_obj)
-    clinical_symbols = store.clinical_symbols(case_obj)
+    clinical_hgnc_ids = store.clinical_hgnc_ids(case_obj) if case_obj else []
+    clinical_symbols = store.clinical_symbols(case_obj) if case_obj else []
 
-    # if no clinical symobols / panels were found loaded, warnings are treated as with research
+    # if no clinical symbols / panels were found loaded, warnings are treated as with research
```
dnil, line 1127: 💯

scout/server/blueprints/variants/controllers.py
```diff
@@ -1109,39 +1161,9 @@ def check_form_gene_symbols(
...
-            "label": "info",
-        },
-    }
-
-    # warn user if gene symbols are corresponding to any current gene,
-    for error in errors.values():
-        if not error["gene_list"]:
-            continue
-        flash(f'{error["message"]}:{error["gene_list"]}', error["label"])
+    _flash_gene_symbol_errors(
```
dnil, line 1164: 👍

tests/server/blueprints/institutes/test_institute_views.py
dnil: I guess you could still keep the test now, possibly changing it to check for the error message instead? But, that is optional.

dnil (approved):
Super!

tests/server/blueprints/institutes/test_institute_views.py
northwestwitch: The function that is called here is already covered by tests, check [here](https://app.codecov.io/gh/Clinical-Genomics/scout/blob/main/scout/server/blueprints/variants/controllers.py), line 1048. I think it's fine to just remove this function here!

Issue #2897, opened 2021-09-28
login issue


ielvers:
Hi there!
I was using Scout all morning with no issue but about half an hour ago I needed to download a coverage report and got a server network error. When that happens I normally just log out of Scout and then log in again and then downloading works just fine. However now I can't log back in. I can log out and in of my google clinicalgenomics account in the web browser but when I try to log into Scout by clicking here
![image](https://user-images.githubusercontent.com/57791644/135079997-9e1b8580-e414-442d-8ce1-54db75bf3c48.png)
all I get is this:
![image](https://user-images.githubusercontent.com/57791644/135080070-81ec8e94-606e-42f4-9de2-e225a2eff3c3.png)

This happens regardless of whether I'm logged in or logged out of my clinicalgenomics account. I've tried this for 30 minutes now.

Any ideas why this is happening? None of my colleagues dare log out to see if they'll be able to log in again. They're such chickens.

dnil:
Thank you for the report! We have been seeing mysterious errors in the log, especially during the last hour and were just about to investigate who was having trouble! 😸 We'll have a look!

dnil:
```
2021-09-28 13:43:35,899 - scout.server.app - ERROR: Exception on /authorized [GET] [in /home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py:1892]
Traceback (most recent call last):
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
   response = self.full_dispatch_request()
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
   rv = self.handle_user_exception(e)
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
   reraise(exc_type, exc_value, tb)
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
   raise value
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
   rv = self.dispatch_request()
 File "/home/hiseq.clinical/miniconda/envs/prod/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
[...]

PR #3083, opened 2022-01-11
Use a FlaskForm with a CSRF token when passing form data to chanio-report

northwestwitch:
This is an attempt at fixing #2589 and #2897

I'm not sure how to reproduce the situation of these bugs and what I'd do is just giving this alternative code a try. If we don't get the error from the bugs above any more then the bug is fixed! 😄 

**How to test**:
1. Install on Scout stage and try to download the coverage report for a case (the case should have coverage in chanjo db), for instance [this one](https://scout-stage.scilifelab.se/cust000/NZ-LCH-lucigentest) or [this one](https://scout-stage.scilifelab.se/cust000/15035-miptest)

**Expected outcome**:
The functionality should be working and should return the same coverage for these samples as the coverage report produced by "main" branch

**Review:**
- [ ] code approved by
- [x] tests executed by CR


northwestwitch:
Coverage report produced by "main" branch:

image


Coverage report produced by this branch: 🆗 

image

northwestwitch:
I am not sure this PR will fix the mentioned issues but I think it's still worth reviewing and merging because refactors the code to prepare requests to chanjo-report and makes the process more clear.

PR #3100, opened 2022-01-20
Add export fields

dnil:
This PR adds a functionality.

The GERP conservation boolean we store for variant page display is not so informative. The user will at some point benefit more from actually using the likes of `bcftools` to query an annotated original VCF, so adding all possible annotations is hardly mandatory. Versioning of databases and tools becomes an issue with exported sheets.

**How to test**:
1. pick a case, download variantS and note the indicated columns are present.
2. try both single and multiple sample cases as genotype is for each individual

Screenshot 2022-01-20 at 12 39 12

**Expected outcome**:

**Review:**
- [x] code approved by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
+##
+### Added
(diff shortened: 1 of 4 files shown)


northwestwitch:
Installed on cg-vm1 now. I'm going to test it there!

northwestwitch (approved):
Works perfectly! 🥇

PR #3106, opened 2022-01-25
Reintroduce links to dbSNP, Beacon and SweGen in RD variant summary panel

### LINKED ISSUE #3105: Beacon query button is gone in RD variant page
Must have disappeared when we restyled the variant page

northwestwitch:
Fix #3105 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Install on cg-vm1 or locally and make sure these links are visible while using this branch

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variant/templates/variant/buttons.html
+++ b/scout/server/blueprints/variant/templates/variant/buttons.html
@@ -1,5 +1,5 @@
 {% macro database_buttons(variant) %}
-<div>
+<div class="mt-3">
   <a class="btn btn-sm btn-info text-white mr-1 mb-1" role="button" href="{{ variant.swegen_link }}" target="_blank">SweGen</a>
   <a class="btn btn-sm btn-info text-white mr-1 mb-1" role="button" href="{{ variant.beacon_link }}" target="_blank">Beacon</a>
   {% if variant.cosmic_links %} <!-- This is a list of tuples : [(id1, link1), (id2, link2), ..] -->
--- a/scout/server/blueprints/variant/templates/variant/buttons.html
+++ b/scout/server/blueprints/variant/templates/variant/buttons.html
@@ -1,5 +1,4 @@
 {% macro database_buttons(variant) %}
-<div class="h6 mt-1">Databases</div>
 <div>
   <a class="btn btn-sm btn-info text-white mr-1 mb-1" role="button" href="{{ variant.swegen_link }}" target="_blank">SweGen</a>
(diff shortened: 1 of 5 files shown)


scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -138,17 +138,6 @@
         </tbody>
       </table>
     </div>
-    {% if variant.track == "rd" %}
```
northwestwitch, line 141: This was not used and not working anyway

northwestwitch:
One variant on master branch:

image

On this branch:

image

northwestwitch:
Ready for review now!

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -1,3 +1,38 @@
+{% macro database_buttons(variant) %}
```
dnil, line 1: 👍

scout/server/blueprints/variant/templates/variant/variant_details.html
dnil: As for the placement, what made it weigh over for the Summary card? I would say thematically it fits the "current" Frequencies better. Layout wise, the ClinVar table in Summary can grow, but on most variants it would be short so that Summary and the up to three cards on freq end up roughly the same size. Touch and go there. Without more arguments, I weigh towards keeping these links in Frequencies.

dnil (approved):
This fixes a bug, and improves code complexity and redundancy! As stated, I dont see the reason for changing the layout. If you have good reason, go with it, but otherwise moving this is now very easy thanks to your macro!

northwestwitch:
> I dont see the reason for changing the layout

I made it the same as in cancer variant, check whichever panel case and you'll see that these buttons (and some more) are exactly there. We could have these buttons in frequency as well, but then we should move them also for cancer variant!

dnil:
> > I dont see the reason for changing the layout
> 
> 
> 
> I made it the same as in cancer variant, check whichever panel case (also on prod scout) and you'll see that these buttons (and some more) are exactly there. We could have these buttons in frequencies as well, but then we should move them also for cancer variant!

Ok, good. I very much agree with keeping the only active placement then.😁 We could easily change later; thematically it does fit the freq card.

PR #3127, opened 2022-02-02
Show Swedac logo in portrait-oriented delivery report

### LINKED ISSUE #3115: Fix export of the delivery report logo (PDF version)
Adding it here otherwise we forget: the logo is only visible in the HTML version of the report, and since it looks ugly it's not exported in the PDF one.

Come up with a solution, perhaps with an example directly in the demo delivery report?

### LINKED ISSUE #3126: Text is missing in the pdf version of the delivery report
## Description

When downloading the pdf version of the delivery report all information doesn't fit, the Swedac logo and "Svarsdatum" is not visible.

From the pdf:
Screenshot 2022-01-25 at 10 05 46

From the web link:
Screenshot 2022-01-25 at 10 07 09

## Suggested solution

*Write here if there is a suggested way to solve the issue*

## This can be closed when

All information is visible in the pdf version of the delivery report.

northwestwitch:
- Fix missing text from the delivery report, while keeping in in portrait orientation (fix #3126)
- Fix Swedac logo missing from delivery report (fix #3115)

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Install this branch on cg-vm1
2. Pick a case with delivery report, for instance this one: https://scout-dev.scilifelab.se/cust002/F0042222
3. Make sure that the exported PDF delivery report using this branch shows the swedac logo and the page looks fine

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/demo/delivery_report.html
+++ b/scout/demo/delivery_report.html
@@ -17,7 +17,7 @@
 
 <style>
     @page {
-      size: landscape;
+      size: A4;
       margin-left: 0rem;
       margin-right: 0rem;
     }
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -3,7 +3,6 @@ import datetime
 import json
 import logging
 import os.path
-import re
 import shutil
 from operator import itemgetter
 
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -741,33 +741,43 @@ def gene_fusion_report(institute_id, case_name, report_type):
 def coverage_qc_report(institute_id, case_name):
     """Display coverage and qc report."""
     _, case_obj = institute_and_case(store, institute_id, case_name)
-    data = controllers.multiqc(store, institute_id, case_name)
-    if data["case"].get("coverage_qc_report") is None:
+    coverage_qc_report = case_obj.get("coverage_qc_report")
+
+    if coverage_qc_report is None:
         return abort(404)
 
-    coverage_qc_report = data["case"]["coverage_qc_report"]
     report_format = request.args.get("format", "html")
+
+    out_dir = os.path.abspath(os.path.dirname(coverage_qc_report))
+    filename = os.path.basename(coverage_qc_report)
+
     if report_format == "pdf":
-        try:  # file could not be available
-            html_file = open(coverage_qc_report, "r")
-            source_code = html_file.read()
-            return render_pdf(
-                HTML(string=source_code),
-                download_filename=case_obj["display_name"]
-                + "_"
-                + datetime.datetime.now().strftime("%Y-%m-%d")
-                + "_coverage_qc_report.pdf",
-            )
+        try:
+            with open(os.path.abspath(coverage_qc_report), "r") as html_file:
+                source_code = html_file.read()
+                bytes_file = html_to_pdf_file(source_code, "landscape", 300)
+                file_name = "_".join(
+                    [
+                        case_obj["display_name"],
+                        datetime.datetime.now().strftime("%Y-%m-%d"),
+                        "coverage_qc_report.pdf",
+                    ]
+                )
+                return send_file(
+                    bytes_file,
+                    attachment_filename=file_name,
+                    mimetype="application/pdf",
+                    as_attachment=True,
+                )
         except Exception as ex:
             flash(
-                "An error occurred while downloading delivery report {} -- {}".format(
+                "An error occurred while converting report to PDF: {} -- {}".format(
                     coverage_qc_report, ex
                 ),
                 "warning",
             )
-
-    out_dir = os.path.dirname(coverage_qc_report)
-    filename = os.path.basename(coverage_qc_report)
+            LOG.error(ex)
+            return redirect(request.referrer)
 
     return send_from_directory(out_dir, filename)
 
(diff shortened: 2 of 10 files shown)


northwestwitch:
[This case](https://scout-dev.scilifelab.se/cust002/F0042222) on stage Scout

Main branch:

image


This branch: 

image

scout/demo/delivery_report.html
```diff
@@ -17,7 +17,7 @@
 
 <style>
     @page {
-      size: landscape;
+      size: A4;
```
northwestwitch, line 20: Same as in current prod

scout/server/blueprints/cases/views.py
```diff
@@ -692,16 +691,20 @@ def delivery_report(institute_id, case_name):
         try:  # file could not be available
             html_file = open(delivery_report, "r")
             source_code = html_file.read()
-            # remove image, since it is problematic to render it in the PDF version
```
dnil, line 695: 🥳

scout/server/blueprints/cases/views.py
```diff
@@ -692,16 +691,20 @@ def delivery_report(institute_id, case_name):
...
-            source_code = re.sub(
-                '<img class=.*?alt="SWEDAC logo">', "", source_code, flags=re.DOTALL
+
+            bytes_file = html_to_pdf_file(source_code, "portrait", 300)
+            file_name = "_".join(
```
dnil, line 696: ✅

scout/server/blueprints/cases/views.py
```diff
@@ -710,6 +713,10 @@ def delivery_report(institute_id, case_name):
                 ),
                 "warning",
             )
+            LOG.error(
+                f"An error occurred while downloading delivery report {delivery_report} -- {ex}"
+            )
+            return redirect(request.referrer)
```
dnil, line 719: Hm, not quite; this dls the html page for the pdf endpoint, as the A HREF is marked DOWNLOAD in the sidebar!
Screenshot 2022-02-03 at 09 04 51

dnil (changes requested):
Looks super! The error handling was a nice touch, but needs a change!

scout/server/blueprints/cases/views.py
northwestwitch: Ah, so it looks like it downloads anyway. I'll check!

northwestwitch: This problem is caused by the download attribute in this line ---> https://github.com/Clinical-Genomics/scout/blob/11c8be6d40128d29d524945abe5a2f42ec892f34/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html#L521

We put it there some time ago because otherwise the page spinner continued to spin even after the download of the file. I'll try to see if there is a simple way to fix it!

northwestwitch: One solution would be, while populating the case data to be shown on the case page, check which resource files (delivery reporr, cancer qc report, etx) do actually exist and if they don't exist do not show the at all in the sidebar. This would be the cleanest solution I think. But perhaps in another PR?

dnil: Yes, that would be nice, and agree on other PR! But it could also change dynamically, as when we e.g. have a mount issue or case update, so we kind of need another way. Perhaps we should just like `abort(404)` instead? I want to avoid a situation where a user thinks they have downloaded the report, uploading it to the LIMS without having checked it. They should perhaps react to the change of file suffix, but not all win users show those.

northwestwitch: Unless you remove that "download" property from the link tag, it will always download a failed page, no matter how you handle the error, try. Now you are seeing this problem a lot because most cases in scout stage, unlike scout prod,  don't have a delivery report

northwestwitch: I honestly don't know how to fix it other than removing the "download" but also removing the loading spinner

dnil: Right, a downloaded 404 is useless. I can't remember why it wasn't possible to just stop the spinner on completion/error with a download script on the case sidebar?

northwestwitch: https://github.com/Clinical-Genomics/scout/issues/2806

northwestwitch: I think we should merge this one and focus on the missing files problem somewhere else. Because you could perhaps even check if a file is existing in jinja (before printing "download") but that would be an ugly solution, since the whole logic should be in the view.. And this PR is not actually introducing the error, because it's already like this in prod

dnil: Would unbinding beforeunload work here as well? A small script on the sidebar would probably as well. But, fine, there is a flash and the file type changes after download. Lets just add an issue for this and merge?

dnil (approved):

scout/server/blueprints/cases/views.py
northwestwitch: > Lets just add an issue for this and merge?

I've done it, we can fix this problem before the next release, so the problem before won't show!

dnil: Hihi, but as you very correctly identify it was a preexisting problem! You only changed it slightly, but a similar situation existed before.

PR #3129, opened 2022-02-03
Attempt at fixing file system not mounted IGV crash

### LINKED ISSUE #3128: Fix crashes and tighten down file access for file endpoint.
When container mounts to expected fs go down, this can happen:
```
2022-02-02 14:42:09,150 - scout.server.app - ERROR: Exception on /remote/static [GET] [in /venv/lib/python3.8/site-packages/flask/app.py:1457]
 File "/venv/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
 File "/venv/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
 File "/venv/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
   rv = self.dispatch_request()
 File "/venv/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
 File "/home/worker/app/scout/server/blueprints/alignviewers/views.py", line 72, in remote_static
   if not range_header and (file_path.endswith(".bam") or file_path.endswith(".cram")):
AttributeError: 'NoneType' object has no attribute 'endswith'
```
It wouldn't hurt adding some checks for if the user is asking for files belonging to cases they should have access to. It would take quite some guessing of case names etc to access, but it might be possible to combine with some other situation where we feel that info is trivial to show. Also worth glitch checking. 😬

northwestwitch:
Tentative fix #3128, only the crash part.

I couldn't reproduce locally, so any idea is welcome! My idea is to pass an empty path to the function that servers the file, so it will return "file not found" instead. Basically it will except here: https://github.com/Clinical-Genomics/scout/blob/679bcd1098cdf95010a75b7c56cec9e458e97adc/scout/server/blueprints/alignviewers/partial.py#L54

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Cast MNVs to SNV for test
 - Export verified variants from all institutes when user is admin
 - Cancer coverage and QC report not found for demo cancer case
+- IGV browser crashing when file system with alignment files is not mounted
 
 ## [4.47]
 ### Added
--- a/scout/server/blueprints/alignviewers/partial.py
+++ b/scout/server/blueprints/alignviewers/partial.py
@@ -28,7 +28,10 @@ def parse_byte_range(byte_range):
 def send_file_partial(path):
     range_header = request.headers.get("Range", None)
     if not range_header:
-        return send_file(path)
+        try:
+            return send_file(path)
+        except IOError:
+            return abort(404, "File not found")
 
     try:
         byte_range = parse_byte_range(request.headers["Range"])
--- a/scout/server/blueprints/alignviewers/views.py
+++ b/scout/server/blueprints/alignviewers/views.py
@@ -3,16 +3,7 @@ import logging
 import os.path
 
 import requests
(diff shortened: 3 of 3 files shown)


dnil:
Good start! I took the liberty of adding a try-except, to remove a crash on missing non-ranged files as well.
We should really think a little about if we don't want to add a layer of security so users can't just snatch each others files, but that can be a separate issue. As long as it doesn't happen by mistake, it is not immediately a bug. We have a reasonable amount of trust in the users.

northwestwitch:
> if we don't want to add a layer of security so users can't just snatch each others files,

I'll think about something. Shouldn't be that hard, no?

northwestwitch:
> if we don't want to add a layer of security so users can't just snatch each others files,

To do it properly, this needs a lot fo refactoring, so perhaps let's have it in another PR?

dnil (approved):
Right!

PR #3146, opened 2022-02-07
add ranges to rank score results on variant pages

### LINKED ISSUE #791: Add ranking ranges to Rank score results

fc-io:
resolve #791

Add ranges to the rank score results. The range is pulled from the corresponding `rank_model_-<version>-.ini` file in the `Clinical-Genomics/reference-files` repo and stored in the database. Storing it so it does not have to fetch every time. If github is down the first time the ini file is fetched things will not work. I'd build a better fallback if appropriate (?).

New dep (`configobj`) for parsing the ini file. So one need to run pip (`pip install --editable .`) to make magic happen.

**old version**
image

**new version full width**
image

**new version not full width**
image


scout/server/blueprints/variant/controllers.py
```diff
@@ -21,10 +21,11 @@
 )
 from scout.server.blueprints.variant.utils import update_representative_gene
 from scout.server.extensions import cloud_tracks, gens
-from scout.server.links import ensembl, get_variant_links
```
fc-io, line 24: ensembl was unused, could have been removed in a separate commit, sorry for the confusion

scout/server/utils.py
```diff
@@ -106,6 +106,49 @@ def institute_and_case(store, institute_id, case_name=None):
     return institute_obj
 
 
+def rank_score(variant, rank_model):
+    """Calculate rank ranges for the variant and its corresponding rank model."""
+
+    def get_scores(field):
+        return [
+            int(v.get("score"))
```
fc-io, line 114: score are always ints, or? :D

dnil: Or else the lovely ValueError exception crash!

scout/server/utils.py
```diff
@@ -106,6 +106,49 @@ def institute_and_case(store, institute_id, case_name=None):
...
+            if isinstance(v, dict) and isinstance(int(v.get("score")), (int))
+        ]
+
+    def get_rank(model, info_fields, variant, category):
+        fields = [f for f in info_fields if f.get("category") == category]
+        scores = [get_scores(field) for field in fields]
+        flattend_scores = [n for v in scores for n in v]
+        # finds the already pre-calculated score
+        rank_score = next(
+            (r for r in variant.get("rank_score_results") if r.get("category") == category), None
```
dnil, line 125: This fails if no result is available in the category. next(None) does not work. If you really, really want to use the list comprehension anyway, perhaps cast it in a containing list(). Otw expand to a loop with some conditionals to handle non existing values?

dnil (changes requested):
Great start! Super clean, clever code with good practices! This will be a very appreciated feature. Just to know for another time, before we start looking at PRs they should normally:
- have a CHANGELOG.md entry
- pass tests; the one that is currently failing illustrates an issue!
- introduce new tests if the PR adds a new feature, or otherwise more than a non-negligible number of lines of moving parts
- ideally name the branches without '/' as the branch names get reused for Docker image names, and it can be a pain to get Podman to pull the right image with certain special characters in the name.

There is one major issue with the logic still. It does not consider score categories with a `category_aggregation = sum` properly. These need sums of max and min values for all predictors within the category. Consider `protein_prediction` in the demo case: there two protein predictors in the 1.20 model, Sift and Polyphen. Each of those can give one point to the score, so e.g. http://localhost:5000/cust000/643594/4c7d5c70d955875504db72ef8e1abe77 gets score 2 but the range is shown as 0<x<1.

We could always finish this off for you if you don't have the time during the semester - just let us know!

scout/adapter/mongo/rank_model.py
```diff
@@ -0,0 +1,49 @@
...
+        Args:
+            rank_model_url(string): A string with the url to the rank model ini file to fetch.
+
+        Returns:
+            rank_model(dict): a copy of what was inserted
+        """
+        LOG.info("Adding rank model to the database")
+
+        response = urllib.request.urlopen(rank_model_url, timeout=20)
+        config = ConfigObj(response).dict()
```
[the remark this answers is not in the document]
fc-io: I think it was that it could not handle nested hierarchies. configobj was the thing used in genmod as well. Why do python devs hate on JSON? ;)

scout/server/blueprints/variant/static/rank_score_results.css
```diff
@@ -0,0 +1,45 @@
+.rank_score_results_heading {
```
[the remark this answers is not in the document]
fc-io: You have a point. :)

scout/adapter/mongo/rank_model.py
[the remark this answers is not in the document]
dnil: I know! Would be so much easier if they were just json directly!

scout/server/utils.py
```diff
@@ -106,6 +106,49 @@ def institute_and_case(store, institute_id, case_name=None):
...
+            "min": min(flattend_scores),
+            "max": max(flattend_scores),
+        }
+
+    def get_rank_score_results(model):
+        fields = [v for v in model.values() if isinstance(v, dict) and v.get("field") == "INFO"]
+
+        return [get_rank(model, fields, variant, c) for c in model["Categories"].keys()]
+
+    return sorted(get_rank_score_results(rank_model), key=lambda k: k["category"].casefold())
```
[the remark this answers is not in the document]
fc-io: :D

scout/adapter/mongo/rank_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Args:
+            rank_model_url(string): A string with the url to the rank model ini file to fetch.
+
+        Returns:
+            rank_model(dict): a copy of what was inserted
+        """
+        LOG.info("Adding rank model to the database")
+
+        response = urllib.request.urlopen(rank_model_url, timeout=20)
```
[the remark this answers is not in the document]
fc-io: coolio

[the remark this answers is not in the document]
fc-io: will fix

scout/server/utils.py
fc-io: good catch! but I'm a bit confused on the explanation.

you mean it crashes 💥 on the later rank_score.get? or crashes when `category` is None? or it fails if the variant does not have a score? or something else? :D

dnil: Probably me too, but either I guess. If the inner expression in the next() evaluates to none, you should get like a "NoneType is not an iterable", right? I'm no longer at the computer so won't check how likely either of the inner ones is to be missing - or if they are well checked for existence beforehand, but at the very least you saw fit to add a fallback None, so I'm guessing it can happen. 😅

scout/server/utils.py
fc-io: I wrapped the whole thing. :D Seems lazy but nice. Also added an error (warning) message to the rank score results view if it fails for some reason. I mean it should not, hence I think it would be overkill to make it fail in a more granular way (i.e. have it show results but not ranges).

image

fc-io: the same thing would shown if the `.ini` file could not be fetched/parsed

northwestwitch:
Can I help fixing this one, so we can merge it @fc-io @dnil ?

fc-io:
Most det, totally fine by me 😄

dnil:
Ok, closing the original branch as well as this is now merged via another PR! Will be super nice to see this deployed! 😸

PR #3147, opened 2022-02-08
Fixed some Pydantic case parsing model types and default values

northwestwitch:
Some model types in pydantic parsing are incorrect or could have better (or none) values

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Load a demo case with and without the modified params in the config file and check that the created case document is correct

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/.github/workflows/vulture.yml
+++ b/.github/workflows/vulture.yml
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 --ignore-names cls ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls,args ${{steps.files.outputs.all}}
         continue-on-error: true
--- a/.github/workflows/vulture.yml
+++ b/.github/workflows/vulture.yml
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls ${{steps.files.outputs.all}}
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


scout/parse/models.py
```diff
@@ -281,15 +281,15 @@ class ScoutLoadConfig(BaseModel):
...
+    madeline_info: Optional[str] = Field(None, alias="madeline")  #!
     multiqc: Optional[str] = None
     owner: str = None
     peddy_ped: Optional[str] = None
-    peddy_ped_check: Optional[str] = Field("", alias="peddy_check")
-    peddy_sex_check: Optional[str] = Field("", alias="peddy_sex")
+    peddy_ped_check: Optional[str] = Field(None, alias="peddy_check")
+    peddy_sex_check: Optional[str] = Field(None, alias="peddy_sex")
     phenotype_terms: Optional[List[str]] = None
     rank_model_version: Optional[str] = ""
```
dnil, line 294: What about this and the sv_rank_model_version? None seems more prudent, but I haven't checked in the templates/controllers.

.github/workflows/vulture.yml
```diff
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls,args ${{steps.files.outputs.all}}
```
dnil, line 22: Sure, that saves a lot of #noqa: F841!

scout/parse/models.py
northwestwitch: I was thinking the same, I'll check!

northwestwitch: After setting them to None, and trying to load demo case, the values are still correctly parsed:

image

northwestwitch: If case config.yaml doesn't contain any rank_model_version or sv_rank_model_version, then the created document case doesn't contain these keys, which I think is fine.

PR #3155, opened 2022-02-10
Remove null non-mandatory key/values when updating a case

### LINKED ISSUE #3153: Re-uploaded RD cases contain document keys that are specific for cancer cases
For instance: cnv_report and coverage_qc_report.

northwestwitch:
Fix #3153 --> Avoid saving empty keys (also non specific to that type of case) when updating a case 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Run scout setup demo and check in MongoDB compass that case contains the expected key/values.
2. While on master branch, re-upload the case: `scout --demo load case scout/demo/643594.config.yaml -u` 
3. Make sure that a bunch of non-necessary keys with empty values get saved in the updated case:
 - cnv_report
 - coverage_qc_report
 - gene_fusion_report_research
 - mme_submission
 - multiqc
4. Repeat both steps from this branch and check that these extra key/values don't get saved in case document

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/adapter/mongo/case.py
```diff
@@ -827,30 +828,29 @@ def update_case(self, case_obj, keep_date=False):
...
                     "rerun_requested": case_obj.get("rerun_requested", False),
                     "research_requested": case_obj.get("research_requested", False),
-                    "smn_tsv": case_obj.get("smn_tsv"),
+                    "smn_tsv": case_obj["smn_tsv"] if case_obj.get("smn_tsv"),
                     "status": case_obj.get("status"),
-                    "sv_rank_model_version": case_obj.get("sv_rank_model_version"),
+                    "sv_rank_model_version": case_obj["sv_rank_model_version"] if case_obj.get("sv_rank_model_version"),
                     "track": case_obj.get("track", "rare"),
                     "updated_at": updated_at,
-                    "variants_stats": case_obj.get("variants_stats"),
```
northwestwitch, line 853: These numbers are better being recalculated after a case is re-uploaded, because they may vary from before

scout/adapter/mongo/case.py
```diff
@@ -850,13 +849,38 @@ def update_case(self, case_obj, keep_date=False):
...
+            "gene_fusion_report_research",
+            "mme_submission",
+            "multiqc",
+            "rank_model_version",
+            "smn_tsv",
+            "sv_rank_model_version",
+        ]:
+            if updated_case.get(key):  # Do not remove key if it has a value
+                continue
+            unset_keys[key] = 1
```
dnil, line 874: I might have picked dict[key]="true", or just added the key to a set/list, but this works!

tests/adapter/mongo/test_matchmaker.py
```diff
@@ -36,4 +36,4 @@ def test_case_mme_delete(adapter, case_obj, user_obj, institute_obj, mme_patient
     updated_case = adapter.case_mme_delete(submitted_case, user_obj)
 
     # Case should not have associated MME submission data any more
-    assert updated_case["mme_submission"] is None
+    assert updated_case.get("mme_submission") is None
```
dnil, line 39: 👍

dnil (approved):
Good fix for existing ones! One might avoid setting them in the first place as well, like in `add_case()`, but maybe that works already after the previous pydantic PR?

scout/adapter/mongo/case.py
northwestwitch: I think you can also use "" or whatever you like apparently, as the value shouldn't affect the operation: https://docs.mongodb.com/manual/reference/operator/update/unset/

northwestwitch:
> Good fix for existing ones! One might avoid setting them in the first place as well, like in `add_case()`, but maybe that works already after the previous pydantic PR?

The problem is only when you update, and it's not specifically related to the pydantic parsing (this issue was there from before). In the beginning I was thinking the same: let's avoid avoid setting those keys without values, but sometimes you find yourself in this situation, for instance

old_case[mme_submission] = {some stuff}

You want to remove the mme submission for the case, so you have new_case with no "mme_submission" key

you have two options here, you either replace old case with new case (but you have to be 100% sure that all the other keys of new case are exactly as you want them to be) or you first set old_case[mme_submission] = None and then you remove the key (as in this PR). 

I think the second way is less risky and more conservative

scout/adapter/mongo/case.py
northwestwitch: I've changed to the value from 1 to "" now, so it's the same as in mongodb docs!

northwestwitch: > I've changed to the value from 1 to "" now, so it's the same as in mongodb docs!

The updated command works: 

image

PR #3175, opened 2022-02-21
Fix #3174 - silence chr-coord if hgvs in clinvar export

### LINKED ISSUE #3174: ClinVar submission file export - chrom-start-end and hgvs exclusive
It seems ClinVar no longer treats chrom-start-end as optional if a complete hgvs exists. Suggest we silence the genome coordinates if we have a hgvs for the variant.

dnil:

ClinVar presubmission checks complain if both HGVS and chr-start-end exist. Give only one, with prio for HGVS.
Screenshot 2022-02-21 at 15 30 36

With test case, in CSV on current `main`:

Screenshot 2022-02-21 at 15 33 52

Looks :ok: in CSV file after applying this patch:

Screenshot 2022-02-21 at 15 32 02

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/parse/clinvar.py
```diff
@@ -162,10 +162,17 @@ def clinvar_submission_lines(submission_objs, submission_header):
             header_key,
             header_value,
         ) in submission_header.items():  # header_keys are the same keys as in submission_objs
+
             if (
                 header_key in submission_obj
             ):  # The field is filled in for this variant/casedata object
-                csv_line.append('"' + submission_obj.get(header_key) + '"')
+                if (
+                    header_key in CLINVAR_SILENCE_IF_EXISTS.keys()
```
northwestwitch, line 170: CLINVAR_SILENCE_IF_EXISTS is not a very self-explanatory variable name. But I guess you wanted to have something general in case something else should be skipped in the presence of other available values? This works, but I wonder if it wound't be more simple and readable if you wrote the explicit code, perhaps with a comment? 

```
# Leave chromosome, start and stop columns empty if HGVS is available
if header_key in ['chromosome', 'start', 'stop'] and 'hgvs' in submission_obj:
```

dnil: Yes, I have a feeling that might happen. The function was nicely general so I didn't like the idea of adding a specialised case there. The possible keys are defined in another constant around there, so I would prefer to keep these on the same level. But I can add a comment at least!

scout/parse/clinvar.py
```diff
@@ -162,10 +162,18 @@ def clinvar_submission_lines(submission_objs, submission_header):
...
+                    and CLINVAR_SILENCE_IF_EXISTS[header_key] in submission_obj
+                ):  # ignore certain fields if other prioritised fields are also filled in
+                    # e.g. columns [chrom, start, stop] should not be used if the [hgvs] column is filled in
```
northwestwitch, line 173: 👍🏻

northwestwitch (approved):
👍🏻

PR #3176, opened 2022-02-21
Fix #3173 - ClinVar export prefer OMIM condition

### LINKED ISSUE #3173: ClinVar submission file export - Condition ID type
Condition ID type values currently are written once per each DB term. As it is right now, ClinVar only accept one value for that column, although multiple ids of the same type can then be given in the Condition ID value column.
For multiple, we have to pick one source. I would advocate using OMIM if it exists, otherwise HPO.

dnil:

ClinVar pre-submission errors found on variants with conditions:
Screenshot 2022-02-21 at 16 43 20

After fix, when adding multiple conditions of the same type, only one type caption is written to the variant.

Screenshot 2022-02-21 at 16 27 43

 Also, when adding both OMIM and HPO conditions, OMIM is preferred.
Screenshot 2022-02-21 at 16 29 38

✅ 
Screenshot 2022-02-21 at 16 36 26

In summary:
- Only one ClinVar submission variant contition type per variant should be used 
- ClinVar submission variant condition prefers OMIM over HPO if available

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


CHANGELOG.md
```diff
@@ -17,7 +17,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Removed unused `add_compounds` param from variant controllers function
 - Changed default hg19 genome for IGV.js to legacy hg19_1kg_decoy to fix a few problematic loci
 - Reduce code complexity (parse/ensembl.py)
-
+- Only one ClinVar submission variant contition type per variant is used
```
northwestwitch, line 20:
```suggestion
- Only one phenotype condition (type OMIM or HPO) per variant is used in ClinVar submissions
```

scout/server/blueprints/variant/views.py
```diff
@@ -303,19 +303,29 @@ def clinvar(institute_id, case_name, variant_id):
...
                 variant_id = key.split("@")[1]
-                cond_types = []
-                cond_values = []
+                possible_cond_types = []
+                possible_cond_values = []
 
                 for condition in conditions:
-                    cond_types.append(condition.split("_")[0])  # 'HPO' or 'OMIM'
-                    cond_values.append(condition.split("_")[1])  # HPO id or OMIM ID
+                    possible_cond_types.append(condition.split("_")[0])  # 'HPO' or 'OMIM'
```
northwestwitch, line 310: If you wanted to avoid some lines of code below, you could define a dictionary of possible conditions like this:

condition_types:{
"OMIM":[],
"HPO": []
}

and add terms to this dictionary here

northwestwitch (approved):
I have a couple of suggestions for improvement but the code already works! 👍🏻 

There is also some warning from Vulture for an unused variant that could be removed!

scout/server/blueprints/variant/views.py
dnil: Agree; that is what I would have chosen from scratch. I would however have to rewrite the stuff ds of this as well to fit that data structure instead. It would be simpler in the end, but not super simple (since OMIM should still have prio). I think I will pass on the opportunity, but feel free! 😁

dnil:
> There is also some warning from Vulture for an unused variant that could be removed!

Yep, thanks - I didn't notice! I made a naive fix now, but didn't have smtp configured on the local. Let's test on stage before merging.

scout/server/blueprints/variant/views.py
northwestwitch: Do you mind if I try this approach? We could also move a big chunk of this code to the controllers!

dnil: Moving it to the controller would make a lot of sense! Feel very free!

scout/server/blueprints/variant/templates/variant/clinvar.html
```diff
@@ -262,7 +262,7 @@ <h5>
                       <td style="width:30%">Condition HPO term(s):<br>(associated to the variant)</td>
                       <td>
                         {% for hpo in pheno_hpos %}
-                          <input type="checkbox" id="conditions@{{pinned._id}}" value="HPO_{{hpo}}" name="conditions@{{pinned._id}}"><a href="http://hpo.jax.org/app/browse/term/{{hpo}}"
+                          <input type="checkbox" id="hpo_terms@{{pinned._id}}" value="{{hpo}}" name="hpo_terms@{{pinned._id}}"><a href="http://hpo.jax.org/app/browse/term/{{hpo}}"
```
northwestwitch, line 265: To simplify the downstream parsing I've modified the original form fields to distinguish between OMIM and HPO terms.

northwestwitch:
Tested locally on a case containing 2 HPO terms and 2 OMIM terms. If you select all HPO terms (in variant and case forms) + the 2 OMIM terms:

image


On main branch:

image


On this branch:

image

northwestwitch:
I think this one is fine now, take a look when you have time @dnil!

scout/server/blueprints/variant/views.py
```diff
@@ -292,58 +292,23 @@ def acmg():
 @templated("variant/clinvar.html")
 def clinvar(institute_id, case_name, variant_id):
     """Build a clinVar submission form for a variant."""
-    data = clinvar_export(store, institute_id, case_name, variant_id)
+
     if request.method == "GET":
+        data = clinvar_export(store, institute_id, case_name, variant_id)
```
dnil, line 297: It now becomes rather apparent that this should be two different endpoints, one for GET, one for POST. But, with all request processing as well in the controller it is already so easy to read it hardly matters. Laters if at all.

scout/server/blueprints/variant/templates/variant/clinvar.html
```diff
@@ -273,7 +273,7 @@ <h5>
                       <td style="width:30%">OMIM diagnoses:<br>(associated to the variant)</td>
                       <td>
                         {% for omim in pheno_omims %}
-                          <input type="checkbox" id="conditions@{{pinned._id}}" value="OMIM_{{omim.disease_nr}}" name="conditions@{{pinned._id}}"><a href="http://omim.org/entry/{{omim.disease_nr}}"
+                          <input type="checkbox" id="omim_terms@{{pinned._id}}" value="{{omim.disease_nr}}" name="omim_terms@{{pinned._id}}"><a href="http://omim.org/entry/{{omim.disease_nr}}"
```
dnil, line 276: 👍

dnil:
Yes, looking all good - tested locally, and for verification on stage. 👍

PR #3201, opened 2022-03-02
Filter out variants that are found only in unaffected individuals

### LINKED ISSUE #3008: Filtering unaffected-only variants in families
Much like for matching causatives, it is of disputable use currently to show variants only present in unaffected family members, perhaps especially ClinVar pathogenic variants. This is likely mostly a vestige from sequencing with limited coverage.

Compare e.g. matching causatives where we no longer report matches unaffected family members: https://github.com/Clinical-Genomics/scout/pull/1434.

northwestwitch:
Fix #3008 --> Option to filter out variants that are only in unaffected individuals

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Install on cg-vm1 and test filtering out vars

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
Ok, This one is ready for review

scout/adapter/mongo/query.py
```diff
@@ -339,6 +342,37 @@ def build_query(
...
+                "$or": []
+            }  # At least one of the affected individuals should harbor the variant
+        }
+        for ind in case_inds:
+            if ind["phenotype"] in [1, "unaffected"]:  # 1=unaffected, 2=affected
+                continue
+            affected_query["$elemMatch"]["$or"].append(
+                {
+                    "sample_id": ind["individual_id"],
+                    "genotype_call": {"$nin": ["0/0", "./.", "./0", "0/."]},
```
dnil, line 368: 👍

scout/server/blueprints/variants/forms.py
```diff
@@ -123,6 +123,8 @@ class VariantFiltersForm(FlaskForm):
     filter_variants = SubmitField(label="Filter variants")
     export = SubmitField(label="Filter and export")
 
+    hide_unaffected = BooleanField("Hide variants present only in unaffected", default=False)
```
dnil, line 126: I think it would make sense to leave this on by default.

```suggestion
    hide_unaffected = BooleanField("Hide variants present only in unaffected", default=True)
```
One could restate in the positive, like `Show variants only present in unaffected`, but maybe it looks odd together with 'Hide dismissed'?

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -33,6 +33,12 @@
       var the_form = document.forms['filters_form'];
       the_form.submit();
     };
+
+    document.getElementById('hide_unaffected').onchange = function() {
+      document.getElementById('expand_search').value = "False";
+      var the_form = document.forms['filters_form'];
+      the_form.submit();
```
dnil, line 40: 👍

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -49,16 +55,22 @@
...
+        {{ form.hide_unaffected(class="form-check-input") }}
+        {{ form.hide_unaffected.label(class="form-check-label") }}
+      </div>
     </div>
-    <div class="col-3 offset-3 d-flex justify-content-start">
+    <div class="col-3 d-flex justify-content-end">
       Filter returns {{result_size}} / {{ total_variants }} variants.
     </div>
-    <div class="col-3 offset-1 d-flex justify-content-end">
+    <div class="col-3 d-flex justify-content-start">
```
dnil, line 73: Hm, this looks odd on my local: 
Screenshot 2022-03-03 at 12 22 59
Try the other way around with start-end again?

scout/server/blueprints/variants/templates/variants/utils.html
dnil: Slick with the insta-reload and spinner!

tests/adapter/mongo/test_query.py
```diff
@@ -47,8 +47,31 @@ def test_build_query(adapter):
     assert query["variant_type"] == "clinical"
 
 
+def test_build_query_hide_not_in_affected(adapter, case_obj):
```
dnil, line 50: 👍

dnil (approved):
🏅 Just a small layout thing. And it makes sense to have it on by default: the idea is to avoid seeing incidental stuff, so seeing more should be what requires an additional active choice.

scout/server/blueprints/variants/forms.py
northwestwitch: I agree that having it on by default is the way to go. Yesterday I've tried exactly the same code in your suggestion and didn't work, let's try today!

northwestwitch: Nope, doesn't work. I'll try to fix it in another way

scout/adapter/mongo/query.py
```diff
@@ -353,6 +353,8 @@ def affected_inds_query(self, mongo_query, case_id):
             "$elemMatch": {"$or": []}  # Any of the affected individuals should exibit the variant
         }
         case_obj = self.case(case_id=case_id)
+        if len(case_obj.get("individuals"), []) == 1:  # No point in adding this filter
```
dnil: 💯 But probably hide the checkbox then as well, right!

scout/server/blueprints/variants/forms.py
northwestwitch: I think I'll just change the label of the checkbox into "Show variants present only in unaffected" and change the query from $nin to $in, much easier this way. Also because when you land into the variantS page (any type= with the GET method) then the query for filtering out these vars doesn't contain hide_unaffected=True and to add it for each type of var is cumbersome..

scout/server/blueprints/variants/forms.py
```diff
@@ -123,6 +123,8 @@ class VariantFiltersForm(FlaskForm):
     filter_variants = SubmitField(label="Filter variants")
     export = SubmitField(label="Filter and export")
 
+    hide_unaffected = BooleanField("Hide variants present only in unaffected", default="checked")
```
northwestwitch, line 126: Ok I found out that if I use render_kw={'checked': True} instead of default="checked", then it works.

Other people with the same problem: https://github.com/wtforms/flask-wtf/issues/362

northwestwitch: But then if you uncheck and submit the form it doesn't stay unchecked when the page reloads :expressionless:

dnil: How about the js suggestion they had on the page you linked, like setting the checked if the value is true?

northwestwitch: I'm trying with jinja

northwestwitch: Didn't work, tried also with jquery. So I made it like this: 😆 

image

scout/adapter/mongo/query.py
northwestwitch: I thought about it, but I haven't removed it from the template for 3 reasons:

1. When they are in a case with only one individual, people might wonder if they are seeing all variants or not (of course they are seeing them all, but you know..)
2. To keep the form consistent in all cases, which is also part of point 1
3. Because then you have to pass the case object to this macro, and this macro is invoked from different forms: the SNVs form, the SVs form, the cancer vars form, ??. So it's a lot of additional code that I wasn't really sure was worth to add. Also because I'm lazy..

So unless you really want to hide I would leave it like that.. ?

northwestwitch: image

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -49,16 +55,22 @@
...
-      {{ form.hide_dismissed }}
-      {{ form.hide_dismissed.label(class="ml-3") }}
+    <div class="col-4">
+      <div class="form-check d-flex justify-content-start">
+        {{ form.hide_dismissed(class="form-check-input") }}
+        {{ form.hide_dismissed.label(class="form-check-label") }}
+      </div>
+      <div class="form-check d-flex justify-content-start">
+        {{ form.show_unaffected(class="form-check-input") }}
+        {{ form.show_unaffected.label(class="form-check-label", **{'data-toggle': 'tooltip', 'title':'This filter applies only to cases with more than one sample'}) }}
```
dnil, line 67: Perhaps explain also the general function? E.g. 
```suggestion
        {{ form.show_unaffected.label(class="form-check-label", **{'data-toggle': 'tooltip', 'title':'This filter applies only to cases with more than one sample. When this is unchecked, only variants that are predicted by callers to be present in an affected individual genotype are show.'}) }}
```

northwestwitch:
Test fixed! The problem was that the variant object used in the failing tests had no sample-specific genotype calls, so when the query filtered by individuals the variant got excluded from the results. I think I'm ready with this one now!

PR #3212, opened 2022-03-04
Shallow link two STR loci

### LINKED ISSUE #3209: Multilocus STR gene deep linking does not work
I guess that is ARX and HOXA13. The loci on Stripy are to "ARX_1" and "ARX_2" instead of the normal gene symbol. Same in GnomAD. GnomAD is also broken in the sense that if you go to the ARX gene overview, you will find links only to the first of the two loci. It thus seems likely this model will change with time.

dnil:

Fix #3209.

ARX and HOXA13 get shallow links to the respective db loci lists, since the GnomAD and STRipy pages have their own nomenclature for those atm. I would not be surprised to see them make landing/joint pages for those in the future, esp if we find more such genes, but this will solve the issue for now.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/scout/server/links.py
+++ b/scout/server/links.py
@@ -3,6 +3,8 @@ from flask import current_app
 from scout.constants import SPIDEX_HUMAN
 from scout.utils.convert import amino_acid_residue_change_3_to_1
 
+SHALLOW_REFERENCE_STR_LOCI = ["ARX", "HOXA13"]
+
 
 def add_gene_links(gene_obj, build=37):
     """Update a gene object with links
@@ -70,6 +72,9 @@ def stripy_gene(hgnc_symbol):
 
     if not hgnc_symbol:
         return None
+
+    if hgnc_symbol in SHALLOW_REFERENCE_STR_LOCI:
+        return "https://stripy.org/database"
     return link.format(hgnc_symbol)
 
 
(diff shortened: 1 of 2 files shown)


dnil:
Tested on cg-vm1:
Screenshot 2022-03-04 at 10 41 06

scout/server/links.py
```diff
@@ -3,6 +3,8 @@
 from scout.constants import SPIDEX_HUMAN
 from scout.utils.convert import amino_acid_residue_change_3_to_1
 
+SHALLOW_REFERENCE_STR_LOCI = ["ARX", "HOXA13"]
```
northwestwitch, line 6: Could this list be moved to constants? Makes sense, also in case we need to add some more?

northwestwitch (approved):
Looks good and test works fine. See my suggestion! 👍🏻

scout/server/links.py
dnil: It could definately! I kept it as its currently only used in one file, so made more sense to have it at hand. Im ok with either!

dnil: If all else is equal, laziness wins? 😉

PR #3218, opened 2022-03-04
Default SV callers to Manta (WES). Fix #3206

### LINKED ISSUE #3206: WES SVs default to "all callers" - MIP tag interaction issue?
All randomly drawn WES SV vars look like they were called by all callers. I would assume they are rather not tagged properly, and this is a default?

Screenshot 2022-03-03 at 16 48 49

dnil:
This PR adds a functionality or fixes a bug. See Clinical-Genomics/MIP#1977.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. load a wes MIP 10 case
2. note SV call filter status: should now be manta:PASS rather than all callers PASS.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/parse/variant/callers.py
+++ b/scout/parse/variant/callers.py
@@ -52,6 +52,6 @@ def parse_callers(variant, category="snv"):
         filter_status = "Pass"
         if variant.FILTER is not None:
             filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
-        callers["wes"] = filter_status
+        callers["manta"] = filter_status
 
     return callers
--- a/scout/parse/variant/callers.py
+++ b/scout/parse/variant/callers.py
@@ -48,4 +48,10 @@ def parse_callers(variant, category="snv"):
             filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
         callers["gatk"] = filter_status
 
+    if category == "sv" and not raw_info or other_info:
+        filter_status = "Pass"
+        if variant.FILTER is not None:
+            filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
+        callers["wes"] = filter_status
+
     return callers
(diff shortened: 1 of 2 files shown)


scout/parse/variant/callers.py
```diff
@@ -48,4 +48,10 @@ def parse_callers(variant, category="snv"):
             filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
         callers["gatk"] = filter_status
 
+    if category == "sv" and not raw_info or other_info:
+        filter_status = "Pass"
+        if variant.FILTER is not None:
+            filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
+        callers["manta"] = filter_status
```
northwestwitch, line 55: That's because if the variant is called then it means that has passed in manta, right?

northwestwitch: It still feels a bit arbitrary to me, especially thinking about other facilities that might have different pipelines..

dnil: It is very arbitrary. I am not fond of it, but it follows from the same pattern we have for the SNVs with at least MIP<11. Lund use the more sensible caller tags so they should be fine. And for others, I don’t think seeing manta:pass is any worse than all callers:pass…

scout/parse/variant/callers.py
```diff
@@ -48,4 +48,10 @@ def parse_callers(variant, category="snv"):
             filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
         callers["gatk"] = filter_status
 
+    if category == "sv" and not raw_info or other_info:
+        filter_status = "Pass"
+        if variant.FILTER is not None:
+            filter_status = "Filtered - {}".format(filter_status.replace(";", " - "))
```
northwestwitch, line 54: 👍🏻

northwestwitch (approved):
👍🏻 

Perhaps let's load a test case before merging?

scout/parse/variant/callers.py
dnil: Agreed!

dnil:
Meh, that's silly. I should have checked the VCFs first. Even the one-caller wes sv vcfs have 'set=Intersection' so in scout terms they will have `raw_info` `call` as `Intersection`. I'll close this.

PR #3233, opened 2022-03-09
Test compound filter - Partially Fix #374

### LINKED ISSUE #374: compound partner level filter
Hi, 

sitting with the clinical genetics folks discussing scout and returned to an old issue - couldn't find it here though. We would very much like to be able to filter (or adjust rank) for strong compound pairs. We would like to be able to screen for variants where *both* fulfill criteria such as rare, nonsense or known pathogenic etc. See also https://github.com/moonso/genmod/issues/41
Either solving it pre-loading or post-loading is fine. Ability to filter on aggregate compound score might be enough? Then one could include the AR_compounds for a suspected AR-trio without being swamped with potential carrier variants..

Cheers!

// Daniel

dnil:
This PR adds a functionality.

Screenshot 2022-03-09 at 16 33 18

This is the oldest still open issue, and according to the last user meeting the most wanted feature. While implementing it properly, with real filtering on all variants and the same for all their compounds, ideally only when inheritance models that involve compound recessive inheritance are on the table either in search or on the variants, is very hard, the present PR might help a little when scanning long lists.

The idea is to allow hiding compound buttons / shading the compounds below a given rank score on request.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/variants/views.py
```diff
@@ -123,7 +123,9 @@ def variants(institute_id, case_name):
     if request.form.get("export"):
         return controllers.download_variants(store, case_obj, variants_query)
 
-    data = controllers.variants(store, institute_obj, case_obj, variants_query, result_size, page)
+    data = controllers.variants(
+        store, institute_obj, case_obj, variants_query, result_size, page, query_form=form.data
```
northwestwitch, line 127: The entire form dictionary is passed to controllers even though only one key/value is used afterwards. I'm wondering if it wouldn't be easier just to pass the key/value in question (either a number or don't pass it at all?)

scout/server/blueprints/variants/controllers.py
```diff
@@ -366,6 +378,13 @@ def parse_variant(
             variant_obj["compounds"], key=lambda compound: -compound["combined_score"]
         )
 
+        # check compound against current query rank score
+        for compound in variant_obj["compounds"]:
+            if (query_form and query_form.get("compound_rank_score") is not None) and (
```
northwestwitch, line 383: If you pass only the variant you could avoid the if (query_form and query_form.get("compound_rank_score") at least

scout/server/blueprints/variants/controllers.py
```diff
@@ -53,7 +53,16 @@ def populate_chrom_choices(form, case_obj):
...
-def variants(store, institute_obj, case_obj, variants_query, variant_count, page=1, per_page=50):
+def variants(
+    store,
+    institute_obj,
+    case_obj,
+    variants_query,
+    variant_count,
+    page=1,
+    per_page=50,
+    query_form=None,
```
northwestwitch, line 64: Passing only the compound_filter value (an int) or None?

scout/server/blueprints/variants/controllers.py
```diff
@@ -366,6 +378,13 @@ def parse_variant(
             variant_obj["compounds"], key=lambda compound: -compound["combined_score"]
         )
 
+        # check compound against current query rank score
+        for compound in variant_obj["compounds"]:
+            if (query_form and query_form.get("compound_rank_score") is not None) and (
+                compound.get("rank_score") < query_form.get("compound_rank_score")
```
northwestwitch, line 384: I think this line is causing a potential bug. Tested on stage typing a number (3, 8066?) as value of compound_rank_score:

```
2022-03-14 08:28:28,055 - [scout.server.app](http://scout.server.app/) - ERROR: Exception on /cust000/F0006358-mip7/variants [POST] [in /venv/lib/python3.8/site-packages/flask/app.py:1457]
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
  File "/home/worker/app/scout/server/utils.py", line 71, in decorated_function
    context = func(*args, **kwargs)
  File "/home/worker/app/scout/server/blueprints/variants/views.py", line 126, in variants
    data = controllers.variants(
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 114, in variants
    parse_variant(
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 384, in parse_variant
    compound.get("rank_score") < query_form.get("compound_rank_score")
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -280,7 +280,19 @@
...
     </div>
+    <div class="col-2 offset-3">
+      <div id="hidecompounds" class="collapse">
+        {{ form.compound_rank_score.label(class="control-label", data_toggle="tooltip", data_placement="top", title="Shade compounds below this score as if dismissed") }}
+        {{ form.compound_rank_score(class="ml-1 form-control", onchange="this.form.submit()") }}
+      </div>
+    </div>
+    <div class="col-1">
+      <a class="btn btn-secondary btn-sm form-control" data-toggle="collapse" href="#hidecompounds" aria-expanded="false" aria-controls="hidecompounds">
+        Hide compounds
```
northwestwitch, line 291: Are the compounds hidden or shaded?

dnil:
😄 Sorry, thanks for testing - should have marked it draft!

northwestwitch (changes requested):
Nice long overdue feature, and smart way of achieving something that is indeed very difficult to implement! 🥇 

I've been trying it but looks like there is a bug affecting the search (check the email and one of my comments).

I would simplify the code by passing to the controllers only the variable in question and not the entire search form, unless you want to make the PR even more epic by moving a lot of code from the view to the controllers? 😈 

The looks of the 2 new elements in the variants form could also be improved, check my comments.

northwestwitch:
> 😄 Sorry, thanks for testing - should have marked it draft!

NP! :)

scout/server/blueprints/variants/controllers.py
dnil: Yes, as it is now that would make sense. I am thinking we may sooner or later want to filter on e.g. CADD or mirror the whole variant filter query onto the compounds in which case it makes more sense to pass it all. I could make it a two stage thing I guess, where the query form is first broken down into like a condensed query object and then passed on to the controller.

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -280,7 +280,19 @@
...
       </div>
     </div>
+    <div class="col-2 offset-3">
+      <div id="hidecompounds" class="collapse">
+        {{ form.compound_rank_score.label(class="control-label", data_toggle="tooltip", data_placement="top", title="Shade compounds below this score as if dismissed") }}
+        {{ form.compound_rank_score(class="ml-1 form-control", onchange="this.form.submit()") }}
+      </div>
+    </div>
+    <div class="col-1">
+      <a class="btn btn-secondary btn-sm form-control" data-toggle="collapse" href="#hidecompounds" aria-expanded="false" aria-controls="hidecompounds">
```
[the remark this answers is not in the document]
dnil: Ah, ok! Looks good on Safari; will test!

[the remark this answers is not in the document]
dnil: It wasn't so much the browser as the window size and zoom level. I could recreate it on Safari by making the window smaller. Made it partially inline now. The string coords kind of have to follow the pattern and keeping that input large enough for its placeholder is the real limiting factor. One could keep fiddling, but other things in the form starts looking really awkward when making the page too narrow as well.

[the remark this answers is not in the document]
dnil: Screenshot 2022-03-14 at 12 17 05

Screenshot 2022-03-14 at 12 23 40

scout/server/blueprints/variants/templates/variants/utils.html
dnil: 😸 That depends! If the last compound on the variant compound badge list is shaded, the whole badge is hidden. The button could say either, the tooltip has both.

scout/server/blueprints/variants/views.py
dnil: Yeah, as above, it would certainly be easier now, though more complicated later. I can change if you like, and we just change it back later; either is fine by me!

scout/server/blueprints/variants/controllers.py
dnil: Yeah, I didn't think about what if the cases were unranked. It should be good now!

scout/server/blueprints/variants/views.py
northwestwitch: I would keep it easier now and change it later. I also think that when we change it later we should try to move much of the code present in the view to the controllers, so the code would probably be quite different from now!

dnil: Right: it sure has some weirdness like direct store calls - and quite a bit of passing on the form to the controller. Landing more on the controller side would also make it easier to split the One View with multiple jobs into several with a single task.

scout/server/blueprints/variants/controllers.py
dnil: Actually it seems like more fun to try for a more complete filter. I've added a sketch, but the compounds need to be a bit more populated to allow filtering. I'll look into it any day now! 😁

northwestwitch:
image


Codecov has started taking liberties since I've been messing with the codecov app. I'll fix!

dnil:
Still, codecov is right in as much as that this needs a couple of additional tests. I was saving them until I knew how this would work. I think I'll add a couple of more filter mirroring things, then tests and ask for a re-review to get some more input! I showed this to a few ppl at cust002 yesterday who seemed positive.

northwestwitch:
> and ask for a re-review to get some more input!

Sounds good!

dnil:
OK, this is actually essentially ready for some more input now. I'll add some tests tomorrow!

scout/constants/variant_tags.py
```diff
@@ -207,7 +207,7 @@
 SPIDEX_HUMAN = {
     "low": {"neg": [-1, 0], "pos": [0, 1]},
     "medium": {"neg": [-2, -1], "pos": [1, 2]},
-    "high": {"neg": [-2, -float("inf")], "pos": [2, float("inf")]},
```
northwestwitch, line 210: Yes!

scout/server/blueprints/variants/controllers.py
```diff
@@ -328,33 +339,19 @@ def compounds_need_updating(compounds, dismissed):
...
-    variant_obj,
-    update=False,
-    genome_build="37",
-    get_compounds=True,
-    case_dismissed_vars=[],
-):
-    """Parse information about variants.
-    - Adds information about compounds
-    - Updates the information about compounds if necessary and 'update=True'
+def update_compounds(store, variant_obj, case_dismissed_vars):
```
northwestwitch, line 342: So nice with extracting the code fro parse_variant and creating new smaller functions! 🥇

scout/server/blueprints/variants/forms.py
```diff
@@ -94,6 +94,8 @@ class VariantFiltersForm(FlaskForm):
     genetic_models = SelectMultipleField(choices=GENETIC_MODELS)
 
     cadd_score = BetterDecimalField("CADD", places=2, validators=[validators.Optional()])
+    compound_rank_score = IntegerField("Compound rank score")
+    compound_follow_filter = BooleanField("Follow filter")
```
northwestwitch, line 98: If I understand it correctly, the same filter settings chosen by the users are also applied to the compound vars?

At least  the settings chosen for coordinates (lt, gt), frequency (gnomad, local obs, clingen ngi and swegen), in (region_annotations, functional_annotations, clinsig, svtype, genetic_models) and spidex


image

I think this becomes a very powerful tool and the central concept should be super clear.

This form item with the tooltip don't look very self explanatory to me:

image

For this reason I would rename "follow filter" to something like "apply same filters to variant's compounds"

The tooltip is better, but perhaps "Shade compounds as if dismissed if they are excluded by selected filters"

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -271,16 +271,30 @@
...
     <div class="col-2">
-      <a class="btn btn-secondary btn-sm" data-toggle="collapse" href="#stringcoords" aria-expanded="false" aria-controls="stringcoords">
+      <a class="btn btn-secondary btn-sm form-control" data-toggle="collapse" href="#stringcoords" aria-expanded="false" aria-controls="stringcoords">
         String coordinates
       </a>
     </div>
-    <div class="col-3 offset-1">
-      <div id="stringcoords" class="collapse">
-        {{ wtf.form_field(form.chrom_pos) }}
+    <div class="col-4">
```
northwestwitch, line 278: image

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -271,16 +271,30 @@
...
+      <div id="stringcoords" class="collapse form-group row">
```
northwestwitch, line 279: Would be nice if the div stays uncollapsed after a search that contains compound rank score and/or follow filter, if it's not super complicated, otherwise it's fine!

tests/adapter/mongo/test_query.py
```diff
@@ -489,8 +489,8 @@ def test_build_spidex_high(adapter):
                     "$or": [
                         {
                             "$and": [
-                                {"spidex": {"$gt": -2}},
-                                {"spidex": {"$lt": -float("inf")}},
+                                {"spidex": {"$gt": -float("inf")}},
```
northwestwitch, line 492: 👍🏻

tests/server/blueprints/variants/test_variants_controllers.py
```diff
@@ -52,6 +52,222 @@ def test_compounds_need_updating():
...
+    query_form = {"cadd_score": test_cadd + 1, "compound_follow_filter": True}
+
+    # WHEN calling the function
+    hide_compounds_query(store, variant_obj, query_form)
+
+    # WHEN checking its compounds
+    compounds = variant_obj.get("compounds")
+
+    # THEN the only compound now appears dismissed
+    assert compounds[0]["is_dismissed"]
```
northwestwitch, line 94: 👍🏻

northwestwitch (commented):
Nice with refactoring on the controller and the variant's filter following the compounds!

I have a few suggestions

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: This part is a bit crammed, perhaps give more space to Local obs by reducing the space occupied by Chromosome?

dnil: You are too zoomed in! The local obs box already looks weird! 😄 Perhaps I can fix it as well!

northwestwitch: I'm zoomed at 100% on safari and chrome, but perhaps's it's because I'm using my laptop and not a monitor?

dnil: Probably! Its going to be the same for many - I've tweaked it a little in the last push! See if you like it! I made the String button the same size as the Hide to retain color symmetry on the page. It would look a little lopsided if they are different.

scout/server/blueprints/variants/forms.py
dnil: Right, but that is a veeeery long name on a button. I think the users will learn to identify what it does, but if you have another succinct suggestion that captures the functionality (I considered the likes of "emulate" or "replicate") I'm all ears! 

 I much agree on the tool tip - updated!

scout/server/blueprints/variants/templates/variants/utils.html
dnil: Right, I'll give it a shot!

scout/server/blueprints/variants/forms.py
northwestwitch: > Right, but that is a veeeery long name on a button. I think the users will learn to identify what it does, but if you have another succinct suggestion that captures the functionality (I considered the likes of "emulate" or "replicate") I'm all ears!
> 
> I much agree on the tool tip - updated!

Apply variant's filter?

dnil: I considered ˜Apply filter˜ but it has two downsides from my perspective: 1) apply is often used as a name for update-form-with-what-I-have; I would have interpreted that as use-the-rank-score-next-to-it 2) not so important, but "apply" to me has very little wiggle room for not actually applying all of the filter (such as genes and chromosomes that really makes little sense).

northwestwitch: `Extend variant's filter`? To me "follow filter" is not really helpful as a name..

dnil: Nah, it doesn't extend it, rather does roughly the same only not quite the same way. So it follows, emulates, perhaps replicates or clones. It is also has context, right between the "compound rank score" and "hide compounds" buttons, so doesn't have to contain all of that. More suggestions please! 😄 

Btw I saw that it has issues with multiple in-items again, likely after the previous test fix... I'll fix and add another test.

northwestwitch: Fine, name it "follow filter"! :roll_eyes:

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: Looking good now:

image

scout/server/blueprints/variants/forms.py
dnil: 😄 How about "Parallel filter"? One more word should fit as well. Say "According to filter"? How about not relying on context, like "Compounds follow filter" or "Shade with filter"?

scout/server/blueprints/variants/forms.py
```diff
@@ -95,7 +95,7 @@ class VariantFiltersForm(FlaskForm):
 
     cadd_score = BetterDecimalField("CADD", places=2, validators=[validators.Optional()])
     compound_rank_score = IntegerField("Compound rank score")
-    compound_follow_filter = BooleanField("Follow filter")
+    compound_follow_filter = BooleanField("Compounds follow filter")
```
northwestwitch, line 98: 😄

tests/server/blueprints/variants/test_variants_controllers.py
```diff
@@ -258,6 +258,97 @@ def test_hide_compounds_follow_filter_region(app, variant_obj):
     compound_variant_dict["is_dismissed"] = False
 
 
+def test_hide_compounds_follow_filter_in(app, variant_obj):
```
northwestwitch, line 261: Nice with all the tests! 🥇

northwestwitch (approved):
Code-wise it looks good to me, so I'm approving.
I saw a couple of errors on Friday but I'm not sure about the situations that created them.

The errors:

  ```
File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 728, in hide_compounds_query
    compound_follow_filter(compound, compound_var_obj, query_form)
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 692, in compound_follow_filter
    if _compound_follow_filter_clnsig(compound, compound_var_obj, query_form):
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 621, in _compound_follow_filter_clnsig
    if clnsig_value in query_rank or str_re.match(clnsig_value):
TypeError: expected string or bytes-like object
```

and 

```
 File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 639, in hide_compounds_query
    compound_follow_filter(compound, compound_var_obj, query_form)
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 607, in compound_follow_filter
    if _compound_follow_filter_in(compound, compound_var_obj, query_form):
  File "/home/worker/app/scout/server/blueprints/variants/controllers.py", line 539, in _compound_follow_filter_in
    compound_items = compound_var_obj.get(item)
AttributeError: 'NoneType' object has no attribute 'get'
```
Are they fixed in the latest code? I think so, right?

dnil:
> Are they fixed in the latest code? I think so, right?

Yes, should be! Both had some then clear explanations, but I can't quite recall. One was the different ways of storing clinsig on new and old cases. I'll deploy once more to stage and check around, but should be fine. I also remembered I wrote that the always_clinsig flag doesn't affect the search, but that would look kind of silly on the clinical query. I've fixed that one as well and updated the tooltip.

dnil:
~7 years discussion of full feature. 14 days implementation and testing of Good Enough partial fix that covers the bulk of it. 😅To be fair several previous refactors and features partly aimed this way (Cheers @moonso, @northwestwitch and @moedarrah). Do we learn anything from that? Nah. Grit is good. The sun is shining, and time is all we have! 🌞

PR #3238, opened 2022-03-14
Fix Marrvel link for variants in build 38

### LINKED ISSUE #3236: Fix MARRVEL link
The current MARRVEL link on the SNV/INDEL variant page doesn't work on genome build GRCh38, since it sends genomic coordinates that MARRVEL interprets as GRCh37. It seems a fix could be to instead send cDNA coordinates, e.g., http://marrvel.org/human/variant/NM_012160.3:c.541A>G

northwestwitch:
Fix Marrvel link for variants in build 38 by attempting to do a liftover to build 37 (fix #3236)
If the liftover is not possible, do not display marrvel link

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [ ] tests executed by


scout/server/links.py
```diff
@@ -412,10 +413,36 @@ def get_variant_links(institute_obj, variant_obj, build=None):
...
+    """Create link to Marrvel (http://marrvel.org/) for a variant"""
+
+    url_template = "http://marrvel.org/human/variant/{}:{} {}>{}"
+    chrom = variant_obj["chromosome"]
+    start = variant_obj["position"]
+    ref = variant_obj["reference"]
+    alt = variant_obj["alternative"]
+
+    if build == 38:  # liftover is necessary before returning link
+        client = EnsemblRestApiClient()
```
dnil, line 431: Ooh, do we have a timeout on that, and know anything about their rate limits? This can end up a little bit of traffic, at least when someone opens a list of pages to background. But if the are happy receiving loads of requests I suppose its fine. What bothers me more is that it will slow the page load down with as much as this request takes.

Transcript linkout coordinates would be one option, or delaying the liftover until the user clicks the link, like first pointing to an endpoint on Scout for the final redirect - cf the align viewer that doesn't lift over until one clicks RNA-seq. One could also consider like an async request so that we let the rest of the page finish first, and have like a js that pulls this into the document when it is ready?

tests/server/test_links.py
```diff
@@ -1,8 +1,69 @@
...
+    mycancergenome,
+    snp_links,
+)
+
+
+def test_marrvel_link_build_37(variant_obj):
+    """Test creating a MARRVEL link for a variant in build 37"""
+    # GIVEN a variant object of genome build 37:
+    m_link = marrvel_link(variant_obj, 37)
+    # THEN the Marrvel link should contain the original variant coordinates
```
dnil, line 20: 💯

tests/server/test_links.py
```diff
@@ -1,8 +1,69 @@
...
+    assert (
+        m_link
+        == f"http://marrvel.org/human/variant/{variant_obj['chromosome']}:{variant_obj['position']} {variant_obj['reference']}>{variant_obj['alternative']}"
+    )
+
+
+@responses.activate
+def test_marrvel_link_build_38():
```
dnil, line 28: ☀️ Super nice with tests!

dnil (commented):
Super nice! I just realised it will slow down the variant page load. What do you think? Wouldn't it be worth a workaround to keep at least the rest of the page loading independent of Ensembl?

scout/server/links.py
northwestwitch: It's a good point. From what I've seen so far, they don't care about the number of request they get from a DNS, as long as they are not bombed with them. I don't think it's our case. I don't think it would slow down the variant page too much as well (variant links are not generated in the variantS page right? That would be a problem otherwise), but I agree that it's best to do the liftover ONLY if the user clicks on the button. I'll change the code to do as you suggested, use a scout entrypoint to try to do the liftover and redirect the user to marrvel (if liftover works)!

northwestwitch:
I rewrote it entirely as you suggested @dnil!

dnil:
> I rewrote it entirely as you suggested @dnil!

Ooh, super! I'll have a look in the afternoon!

dnil (approved):
🥇 Super!

dnil:
Ahem, I only had 37 cases loaded. Did you test with 38 already perhaps? 😁

northwestwitch:
> Ahem, I only had 37 cases loaded. Did you test with 38 already perhaps?

Of course not! 😆

PR #3274, opened 2022-03-25
Sort panels alphabetically according to name (panelS) and id (variant)

### LINKED ISSUE #3257: List gene panels alphabetically

dnil:
Fix #3257

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Nice! Installing it on stage server!

dnil:
Lets remember to check a case without any panels; it could backfire.

scout/adapter/mongo/panel.py
```diff
@@ -522,9 +522,11 @@ def apply_pending(self, panel_obj, version):
 
     def latest_panels(self, institute_id, include_hidden=False):
         """Return the latest version of each panel."""
-        panel_names = self.gene_panels(
-            institute_id=institute_id, include_hidden=include_hidden
-        ).distinct("panel_name")
+        panel_names = sorted(
```
northwestwitch, line 525: Mmm shouldn't show in panel page?

I see this:

image

dnil:
Actually, it should be fine (empty panels that is); it seems to work on local.

scout/adapter/mongo/panel.py
dnil: Should and does locally! Let me check..

northwestwitch: It's on stage

dnil: Looks good to me:
Screenshot 2022-03-25 at 12 04 34
But yeah, maybe there is something odd with the PanelApp panels!

ehre: Great job, thanks! Will this alphabetical order also be reflected in the drop-down list on the SNV/INDELs variants page?

dnil: Maybe it would also be more obvious if I sort on display name instead!

dnil: > Great job, thanks! Will this alphabetical order also be reflected in the drop-down list on the SNV/INDELs variants page?

Hi, np! No, not immediately; for us that is a very short list, and kind of sectioned into the ones from the case, and the institute own additions. But, maybe I could add like a separator between those and sort each. I can give it a quick look.

ehre: Oh. Well, for us it would certainly be very helpful to get the panels alphabetically also there.

dnil: Ok, like this for the panelS page: now with display name instead of panel name as well, which should make it better.
Screenshot 2022-03-25 at 12 32 55

northwestwitch: > Looks good to me:

Perhaps I was still on the old branch? Yes it works!

northwestwitch:
image

If you are wondering why actions get stuck!

northwestwitch (approved):
Code wise looks good, let's just give it a final check on stage, but I think it's fine to merge! 👍🏻

dnil:
Thanks! Looks good on stage now as well:

Screenshot 2022-04-26 at 15 00 04

Still no separators on the dropdown, but let's leave that for BS5 dropdown update...

Screenshot 2022-04-26 at 14 59 43

PR #2861, opened 2021-09-08
Refactored and filtered general report data

northwestwitch:
Fix #2813.

- Filtered case key/values
- Filtered case individuals key/values
- Filtered institute key/values
- Filtered variants key/values (there is still a lot of information but it's required in the report)
- Modified variants list and arranged in this object:
variants = {
 "causatives_detailed" : [],
 "partial_causatives_detailed" : [],
 "suspects_detailed": [],
 "classified_detailed": [],
 "tagged_detailed": [],
 "tier_detailed": [],
 "dismissed_detailed": []
} 

**How to test**:
1. Install on stage and compare with branch master. General report produced with this branch should be exactly the same as the one produced by master branch.
2. Use the api to retrieve report data. It should contain all the required json info to produce the report

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #3286, opened 2022-03-30
Fix  mismatch between SV n.genes in case report and SV variants page

### LINKED ISSUE #3285: Different number of genes in SV variantS page and general report
Take for instance the first SV variant of this case, an inversion that contains 197 genes: https://scout-stage.scilifelab.se/cust002/F0042222/sv/variants?variant_type=clinical

I marked it as causative, so it shows on the general report

But when you see the report it says that it has 32 genes!

northwestwitch:
Tentative fix #3285 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -31,8 +31,8 @@ VARIANT_REPORT_VARIANT_FEATURES = [
     "cadd_score",
     "genetic_models",
     "acmg_classification",
-    "genes",
-    "hgnc_symbols",
+    "genes",  # This is a list of dictionaries with gene info
+    "hgnc_symbols",  # This is a list of gene symbols (strings)
     "comments",
     "category",
     "dismiss_variant",
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -32,6 +32,7 @@ VARIANT_REPORT_VARIANT_FEATURES = [
     "genetic_models",
     "acmg_classification",
     "genes",
+    "hgnc_symbols",
     "comments",
     "category",
     "dismiss_variant",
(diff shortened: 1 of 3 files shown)


northwestwitch:
Tested with same variant as in the open issue

Case report using this branch: 🆗 

image


Same variant, dismissed: 🆗 

image

scout/constants/variant_tags.py
```diff
@@ -32,6 +32,7 @@
     "genetic_models",
     "acmg_classification",
     "genes",
+    "hgnc_symbols",
```
northwestwitch, line 35: I hate this change we introduced some time ago (https://github.com/Clinical-Genomics/scout/pull/2861). Every time something is missing from general report but should be there it takes me forever to figure out why! I wish we could revert 😠

dnil: Heh, yes, I guess the react version of the report is still on hold, so perhaps the use case isn't really there?

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -791,7 +791,7 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-warning">Part
 {% macro dismissed_gene_list(variant) %}
   {% if variant.genes %}
     {% if variant.genes|length > 3%}
-      {{ variant.genes|length }}&nbsp;genes
+      {{ variant.hgnc_symbols|length }}&nbsp;genes
```
dnil, line 794: Perphaps also add a small note in e.g. the controller or view docstring describing the difference so it's not confusing if one needs to change it later?

dnil (approved):
Well found!

scout/server/blueprints/cases/templates/cases/case_report.html
northwestwitch: Good idea, I'll fix!

PR #3292, opened 2022-04-01
General report frequencies and callers...

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
I kind of suspect this will not be enough, but I didn't have a good test case loaded on local, so testing on stage..

dnil:
And the callers...
Screenshot 2022-04-01 at 11 36 51

 The callers construct is only present after the decoration. One could either add all callers individually to the variant filter constant, but that would sort of defeat the purpose of having the `CALLERS` constant. I opted for repopulating via CALLERS before decorating. Its a little complex as CodeFactor says. I can break it out into a function.

northwestwitch:
> And the callers... Screenshot 2022-04-01 at 11 36 51
> 
> The callers construct is only present after the decoration. One could either add all callers individually to the variant filter constant, but that would sort of defeat the purpose of having the `CALLERS` constant. I opted for repopulating via CALLERS before decorating. Its a little complex as CodeFactor says. I can break it out into a function.

Nice! Perhaps a function that can be shared with variant page?

dnil:
> Nice! Perhaps a function that can be shared with variant page? 

Yes, perhaps at least with the variant export functionality where it is even more similar. I'll check.

> Could also be a jinja macro

Mm, well, yes, that would be nice. As it is, we have macros for the variant types. They tend to be internally repetitive as well, but I dont think this is the time to touch that. But it would be nice to do it...

northwestwitch:
> Mm, well, yes, that would be nice. As it is, we have macros for the variant types. They tend to be internally repetitive as well, but I dont think this is the time to touch that. But it would be nice to do it...

We can fix it later as well, yes!

dnil:
Looking tentatively good. For a large var:
Screenshot 2022-04-01 at 12 00 22
And a smaller one with fewer panels:
Screenshot 2022-04-01 at 12 00 46

northwestwitch:
Is there something missing here as well? 

image

dnil:
>Is there something missing here as well? 
Indeed!

scout/constants/variant_tags.py
```diff
@@ -37,6 +39,9 @@
     "category",
     "dismiss_variant",
     "variant_rank",
+    "sub_category",
+    "end_chrom",
```
northwestwitch, line 43: 👏🏻

dnil: Screenshot 2022-04-01 at 12 26 08

northwestwitch: Well done, let's merge before we see something else?

PR #3293, opened 2022-04-01
Bootstrap 5 - Dark

dnil:
This PR adds a functionality.

- Bootstrap 5
- Dark or light mode auto select according to browser media preference

Screenshot 2022-04-01 at 19 01 03

Very, very obviously just a draft. Not super bad, but still many little details. The major change is moving away from jQuery as a strict dependency. It still plays with it of course, but all internal functions are now native js. They now use more explicit constructors, rather than a first jsQuery call to a function. Also few new naming schemes for things, like most (all?) `data-` prefix things are now `data-bs-`.

To get colors to play with dark mode, I opted to first reset essentially all our custom colors just to get stuff visible. Then one can bring them back one by one and select an appropriate light and a dark mode color alternative for each.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
The PR I've just merged caused a conflict in scout/server/blueprints/institutes/templates/overview/causatives.html .. sorry! 😞

dnil:
Thanks, that is expected - I'll try to find a day or so to wrap this up in the near term so as to not do this too much. 

So far so good; many moving parts with js and bootstrap scripted components need a little tweak, but then work just fine. Feel very free to have look and suggest e.g. theme/style/colors or things from the new components that are must-have from the get-go!

dnil:
> @dnil do you want help fixing the sonarcloud issues in this branch?

I can fix the ones caused directly, but the old ones touched I'll see about. There will be a lot anyway. Some security related, like the integrity for the CDNs and rel-noopener make sense, and the accessibility compliance is something we have talked about doing before. But a lot is just noise.

northwestwitch:
#### Sonarcloud's approval becomes even more important than your mother's.. 


image

dnil:
Ah, I just noticed the dropdown conflicts a bit with the sticky-top table headers. Lets see if one can lift the dropdown, otherwise it would likely be fine as before without the sticky on the pages where the dropdown covers it. I'll call it a day now, but this would be ready for comments. See you Monday!

dnil:
Ah, I just noticed the dropdown conflicts a bit with the sticky-top table headers. Lets see if one can lift the dropdown, otherwise it would likely be fine as before without the sticky on the pages where the dropdown covers it. I'll call it a day now, but this would be ready for comments. See you Monday!

northwestwitch (approved):
Haven't checked the code (changes in 70+ pages 😆 ) but looks gorgeous! 🥇 

I think we are ready to merge 

Some small things:

- The tumor purity thing that you also have noticed:

- STRs pages in light mode (too much color, especially blue I think):
image

SVs pages, transcripts panel:

image

Add also a bit of space from the panel above:

image

northwestwitch:
I've noticed one more thing: on stage, if you go to cust000 and then on one cancer case, if you go back to all cases of cust000, then you only see cancer cases.

dnil:
Thanks, interesting! I can't recreate it locally off the bat. Will check!

dnil:
To summarise, it was the same with another branch on stage, so we decided it's ok to merge this.

PR #3295, opened 2022-04-04
Move beacon code to a dedicated extension. Fix beacon remove vars

### LINKED ISSUE #3277: Error while submitting to beacon VCF files filtered by a gene panel
Example, [this case](https://scout-stage.scilifelab.se/cust002/F0038815) from cust002 in stage:

Filter variants using 2 gene panels:

image

Error:

image

northwestwitch:
- Fix #3277 
- Remove most of the code from cases views and controllers and put it into an app extension

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Tested on stage and delete variants functionality now works:

image


Since I've tested with a case copied from prod and the panels for that case don't exist on stage (or don't exist in case dict?), I get an error when trying to create the submission data with the panel's genes:

image


Perhaps I can fix the code to fail a bit more gracefully..

northwestwitch:
I think one of the problems is that the add variants endpoint times out when you provide one or more gene panels with many genes (probably the request is not even sent to the beacon). I'm investigating a solution (an async endpoint, using a thread?), but so far is still not working..

northwestwitch:
Aha! Perhaps I don't need to complicate the code with async requests, there's a permission issue on the dockerized beacon app on stage: https://github.com/Clinical-Genomics/cgbeacon2/issues/227

northwestwitch:
Another problem when you send requests with data containing thousands of genes:

image


I'm not sure this would work

dnil:
> Another problem when you send requests with data containing thousands... 
> I'm not sure this would work

No, a say POST request with general data would be preferable to using the rather limited session object. I haven't had any time to read the code yet, so probably you already did! 😊 Let's check tomorrow!

northwestwitch:
> > Another problem when you send requests with data containing thousands...
> > I'm not sure this would work
> 
> No, a say POST request with general data would be preferable to using the rather limited session object. I haven't had any time to read the code yet, so probably you already did! 😊 Let's check tomorrow!

Wait, it's not ready yet!

northwestwitch:
> No, a say POST request with general data would be preferable to using the rather limited session object.

Agreed! We already have problems of this type with chanjo! I've been thinking of ways to "compress" a list of number in something that is smaller. Looks like Google has an algorithm that does exactly the same: https://developers.google.com/maps/documentation/utilities/polylinealgorithm

Check out this python implementation: https://github.com/amit1rrr/numcompress

northwestwitch:
I think this is ready for review now!

scout/server/extensions/beacon_extension.py
```diff
@@ -0,0 +1,192 @@
...
+            form(ImmutableMultiDict): request form submitted by user. Example:
+                [('case', 'internal_id'), ('samples', 'affected'), ('vcf_files', 'vcf_snv'), ('vcf_files', 'vcf_snv_research'), ('panels', '6246b25121d86882e127710c')]
+
+        """
+
+        # Check if user has rights to submit case to beacon
+        user_obj = store.user(current_user.email)
+
+        if "beacon_submitter" not in user_obj.get("roles", []):
+            flash("You don't have permission to use the Beacon tool", "warning")
```
dnil, line 102: Also something like please write to support and ask your institute primary contact to approve you as a Beacon submitter?

scout/server/extensions/beacon_extension.py
```diff
@@ -0,0 +1,192 @@
...
+        gene_ids = set()
+        for panel in form.getlist("panels"):
+            gene_ids.update(store.panel_to_genes(panel_id=panel, gene_format="hgnc_id"))
+
+        if gene_ids:
+            data["genes"] = {"ids": list(gene_ids), "id_type": "HGNC"}
+
+        return data
+
+    def add_variants(self, store, case_obj, form):
```
dnil, line 87: Passing the form all the way to the extension does connect things quite a bit. But the code is currently clean enough that I think changing it to parse the form in the view would make it more complex. All good for now!

tests/server/extensions/test_beacon_extension.py
```diff
@@ -0,0 +1,150 @@
...
+        # WHEN user submits variants from a case to the Beacon
+        beacon.add_variants(store, case_obj, None)
+
+        # THEN case will not be updated with submission data
+        updated_case = store.case_collection.find_one(case_obj)
+        assert "beacon" not in updated_case
+
+
+@responses.activate
+def test_add_variants_wrong_dataset(app, user_obj, case_obj):
```
dnil, line 55: 👍

dnil (approved):
Looks good to me, but lets test at stage as well.

scout/server/extensions/beacon_extension.py
northwestwitch: Yes, good idea!

scout/server/extensions/beacon_extension.py
northwestwitch: Did you mean that it is fine to pass the raw request of it should be parsed in the view first? Sorry i didn't get it

dnil: I think it is in principle more correct if the view parses requests - or possibly a dedicated function from the controller. But we don’t hold to that in general, and this code is still very clean - the forms are not touched until they reach the extension, so I do not mind an exception!

northwestwitch: Ah OK, thanks!

northwestwitch:
Test trying to submit a case to a dataset that doesn't exist in beacon: 🆗 

image

Test add variants to dataset that exists (cust002 GRCH37) 🆗 

image

Test removing vars for the same case: 🆗 

image

northwestwitch:
Merging this one as well!

PR #3252, opened 2022-03-22
Refactoring of igv tracks building

northwestwitch:
Fix #3139 and fix #3250

- Build case IGV tracks starting from case and variant objects, and not passing all params from an HTML form
- Check that user has access to a case track before redirecting to IGV view (using [institute_and_case](https://github.com/Clinical-Genomics/scout/blob/4c177c838639cb648bbe03c4be6f710f3dcc4d2c/scout/server/utils.py#L89) function to retrieve case info and check if user is allowed to see resources)
- Check that user has access to RNA splice junctions before redirecting to IGV sashimi plot (also using institute_and_case function)

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
</details>

**How to test**:
1. Install on stage (cg-vm1) and open alignments:
- [ ] At the case level (from case sidebar)
- [ ] For a SNV variant
- [ ] For a SV variant
- [ ] For STRs
- [ ] For cancer variants
- [ ] For a case in a group with other cases

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #3303, opened 2022-04-19
Additional check for user logged in status and requested resource before serving files using remote static

### LINKED ISSUE #3302: Refactor endpoint so that /remote/static/unindexed can't be used without being authenticated
Check the use case from this report: https://huntr.dev/bounties/7acac778-5ba4-4f02-99e2-e4e17a81e600/

Reproduce: `docker-compose up`
Then `curl --path-as-is 'http://127.0.0.1:8000/remote/static/unindexed?file=../../../../../../../etc/passwd'

northwestwitch:
Fix #3302 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test - locally**:
1. On main branch, serve demo app (`scout --demo serve`)
2. Try to get a resource as a non-logged user via curl: `curl http://localhost:5000/remote/static?file=config.py`
3. The response should return the resource
4. Get the resource in the browser as a logged in user by using this URL: `http://localhost:5000/remote/static?file=config.py`
5. The response should return the resource

6. Switch to this branch and repeat 2 and 4
7. You should get a 403 code (forbidden) in both cases

**Expected outcome**:

**Review:**
- [x] code approved by mikaell
- [x] tests executed by mikaell


northwestwitch:
I'll write a test!

northwestwitch:
Ok, now I'm really done with this one!

mikaell:
Fun bug. I'll have a look.

mikaell (approved):
Code looks good and the tests pass! 💪

northwestwitch:
@dnil I've modified this PR following your feedback of yesterday --> I've introduced sessions to check which files can be served by the remote static endpoint. Feel free to give it a check when you have time

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -14,6 +14,22 @@
...
+
+    Args:
+        display_obj(dict): A display object containing case name, list of genes, lucus and tracks
+    """
+    session_tracks = list(display_obj.get("reference_track", {}).values())
+    for key, track_items in display_obj.items():
+        if key not in ["tracks", "custom_tracks", "sample_tracks"]:
+            continue
+        for track_item in track_items:
+            session_tracks += list(track_item.values())
```
dnil, line 28: Hm, one would need a plan to purge old session tracks as well, otherwise we will rather soon have users unable to log in due to overfull session stores.

scout/server/blueprints/alignviewers/views.py
```diff
@@ -62,7 +60,15 @@ def remote_cors(remote_url):
 @alignviewers_bp.route("/remote/static", methods=["OPTIONS", "GET"])
 def remote_static():
     """Stream *large* static files with special requirements."""
-    file_path = request.args.get("file") or ""
+    file_path = request.args.get("file") or "."
```
dnil, line 63: Would it be really difficult to just replace that file arg with case-individual-filetype/configkey, much as in #3252? That way, the the actual path could be kept internal to the app instead of passing via the users browser. I can have a look if you are tired of this issue by now. 😊

northwestwitch: It's quite a lot of code because it's not just case key/value. I see some issues a

- Sometimes the track is at the case level, sometimes at the individual level
- The tracks have an index that should also be passed by this function, and the index file suffix is extrapolated by the main track suffix 
- The tracks need to be built anyway before (in the controllers) because already in the template you need to specify which type of track igv should expect (wig, alignment, bed, etc. see [here](https://github.com/Clinical-Genomics/scout/blob/1200f4a0f947f35e1fbae8f1403d4e59ca0f8fd2/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html#L120)) , so you'd end up having almost the same code twice

northwestwitch: But it if you want to try, please do! Perhaps in a separate branch, because I'll instroduce the code to purge the session in this one

scout/server/blueprints/alignviewers/controllers.py
northwestwitch: I don't think overfull session stores would happen anyway, since that session object doesn't contain much info, try to printout the content! And the session["igv_tracks"] gets overwritten every time you get to the alignments of a new case.

The object is a list of strings with the tracks for that case, not a dict with track data for each case you visit

northwestwitch: I've been investigating a solution and found this:: https://stackoverflow.com/questions/48994440/execute-a-function-after-flask-returns-response

I've pushed some code to this PR. Basically after the request is returned from igv or sashimi_igv, the session["ivg_tracks"] key/value gets removed

scout/server/blueprints/alignviewers/views.py
```diff
@@ -91,7 +97,18 @@ def sashimi_igv(institute_id, case_name, variant_id):
     )  # This function takes care of checking if user is authorized to see resource
 
     display_obj = controllers.make_sashimi_tracks(case_obj, variant_id)
-    return render_template("alignviewers/igv_sashimi_viewer.html", **display_obj)
+    controllers.set_session_tracks(display_obj)
+
+    response = Response(render_template("alignviewers/igv_sashimi_viewer.html", **display_obj))
+
+    @response.call_on_close
```
northwestwitch, line 104: This executes when the response is closed removes the session item

tests/server/blueprints/alignviewers/test_alignviewers_views.py
```diff
@@ -72,3 +128,6 @@ def test_igv_authorized(app, user_obj, case_obj, variant_obj):
 
         # THEN the response should be a valid HTML page
         assert resp.status_code == 200
+        # AND when the reponse is closed case IGV tracks should be remove from session
```
northwestwitch, line 131: Added a test for the new code

scout/server/blueprints/alignviewers/views.py
```diff
@@ -91,7 +94,16 @@ def sashimi_igv(institute_id, case_name, variant_id):
...
+    @copy_current_request_context
+    def clear_session_traks():
```
dnil, line 103:
```suggestion
    def clear_session_tracks():
```

scout/server/blueprints/alignviewers/views.py
dnil: Cool thing!

scout/server/blueprints/alignviewers/views.py
```diff
@@ -120,4 +132,13 @@ def igv(institute_id, case_name, variant_id=None, chrom=None, start=None, stop=N
...
 
     display_obj = controllers.make_igv_tracks(case_obj, variant_id, chrom, start, stop)
-    return render_template("alignviewers/igv_viewer.html", **display_obj)
+    controllers.set_session_tracks(display_obj)
+
+    response = Response(render_template("alignviewers/igv_viewer.html", **display_obj))
+
+    @response.call_on_close
+    @copy_current_request_context
+    def clear_session_traks():
```
dnil, line 141:
```suggestion
    def clear_session_tracks():
```

northwestwitch: Ah yes! 😆

scout/server/blueprints/alignviewers/views.py
dnil: Yeah, I think 1 and 2 above should be rather possible. It's harder to get around 3 without adding some complexity. One could perhaps have a common helper for the two, producing the same track info and use just the file name for this one, but I haven't tried so no idea yet if there is anything that is difficult to get around. 

How about we focus on merging this, which a lot safer than before, and we can refactor later as time allows?

northwestwitch: yes, I agree!

PR #3304, opened 2022-04-19
Script to fix OMIM format for old cases

### LINKED ISSUE #3301: Standardize OMIM info in case documents before case export
Some oldish cases contain OMIM diagnoses as a list of integers, some newer cases save OMIM terms as a list of dictionaries. Some downstream applications (CG - https://github.com/Clinical-Genomics/cg/pull/1464) might be strict on parsing the output of the case export functionality, so it would be best to first modify the case and save the OMIM diagnoses in the new format before exporting cases

northwestwitch:
I think the easiest way to fix #3301 would be with a scripts that loops over the cases with the old OMIM format and fixes them one by one. To be first tested on stage

**How to test**:
1. On hasta, create a file with the content of the script.
2. "us" to use scout stage env (python and other libs are required)
3. Run a simulation to check which cases contain wrong diagnosis format: `python convert_case_omim_format.py --db_uri stage_db_uri`
- [x] Make sure that there are cases with old diagnosis format (a list of integers). These cases won't be modified by the script
4. Run the script to fix the stage cases with old  diagnosis format: `python convert_case_omim_format.py --db_uri stage_db_uri --fix`
- [x] These cases should be fixed. You can make sure by running the command `python convert_case_omim_format.py --db_uri stage_db_uri`, that will no longer show any cases with old dia now.

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
Test on scout stage (hasta): 🆗 

image

When re-running the simulation there are no more case with old diagnosis format: 🆗 
image

dnil (approved):
👍

northwestwitch:
Nice, later today we can run the script to fix prod cases on cg-prod-services!

dnil:
> Nice, later today we can run the script to fix prod cases on cg-prod-services!

Perhaps safest to take a short downtime, right? There would be a very slight risk to loose edits to cases. Shouldn't take too long since its "just" all cases. We could send an email preparing it, and if really prudent find a small slot with prod as well?

northwestwitch:
> > Nice, later today we can run the script to fix prod cases on cg-prod-services!
> 
> Perhaps safest to take a short downtime, right? There would be a very slight risk to loose edits to cases. Shouldn't take too long since its "just" all cases. We could send an email preparing it, and if really prudent find a small slot with prod as well?

Shouldn't affect prod since it's just a bunch of old cases which have diagnosis in our database. Let's find out with the dry run. I think it's good to stop the browser and have a downtime for the users, but I think it's safe for production

dnil:
> > > Nice, later today we can run the script to fix prod cases on cg-prod-services!
> > 
> > 
> > Perhaps safest to take a short downtime, right? There would be a very slight risk to loose edits to cases. Shouldn't take too long since its "just" all cases. We could send an email preparing it, and if really prudent find a small slot with prod as well?
> 
> Shouldn't affect prod since it's just a bunch of old cases which have diagnosis in our database. Let's find out with the dry run. I think it's good to stop the browser and have a downtime for the users, but I think it's safe for production

The only cases that could come into question are reruns, and they would have to go in with a really awkward timing as well.

northwestwitch:
I've just runned the script on prod now. 144 cases were fixed. I have a local copy of the fixes executed, if we need it later!

Issue #3041, opened 2021-12-09
Personal event history


dnil:
Show (recent) event activity for a user, making it easy to get back to a case/analysis recently done, or remembering that case with that pinned variant in whatever gene.

mikaell:
How would this feature differ from the Activity panel on the case page?

Is it showing all cases activities gathered on the first page?

dnil:
1. It would be a new feature show all activity for any case and variant but only for the active logged in user.
2. That is a very good start. Additional events will appear in due time, but there is no request for new events here.

dnil:
It could be a new view e.g. under the institutes/overview blueprint collection, and be accessed/linked from the user dropdown on the `templates/layout_bs4` `top_nav_right` menu. 

Certain users have a large number of events, so the view would be paginated.

dnil:
The individual user info now displayed on the Users list page would make a nice addition to such a view:

![Screenshot 2022-02-24 at 12 28 42](https://user-images.githubusercontent.com/758570/155515952-cb547056-2101-43b5-95cf-f39654cdb962.png)

mikaell:
I was thinking in the lines of updating the first page the user arrives to. This would contain a table of 15-25 most recent events. Every table entry would be a link to the case. Full user history would be reachable from the more specific user menu.

A quick and dirty html-hack. 
![image](https://user-images.githubusercontent.com/1150065/155521987-6c152570-272f-4864-976b-18ff864c9f02.png)

northwestwitch:
Just an idea, you might consider to include not every type of recent event in this list. For instance I don't think it would be so informative that we show that the user has dismissed 20 variants of a single case. But if we show that the user has assigned a causative for a case, dismissed a bunch of variants for another (as a single event), commented on another case it would be more helpful I think!

mikaell:
This is what I done so far. Is it moving in the right direction?

- [ ] Don't display events of minor importance
- [ ] Nicer presentation (whole row link, no underline?)
- [ ] Things I've forgotten

![image](https://user-images.githubusercontent.com/1150065/157442472-dfa99ceb-a349-4f19-9aab-081c51492b1e.png)

northwestwitch:
Let's analyze the whole list of events that can be diaplayed:

```
VERBS_MAP = {
    "assign": "was assigned to",
    "unassign": "was unassigned from",
    "status": "updated the status for",
    "comment": "commented on";
    "comment_update": "updated a comment for",
[...]

PR #3310, opened 2022-04-20
Add Eventlog to Index.html

mikaell:
**Requires Python ≥ 3.7** Why? Because dicts will maintain order without being OrderedDicts().

This PR adds a functionality. Now a summary of a user's recent and important events are shown on index.html. Each entry will be link to the case in question. Reference issue #3041.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Load one or several cases2. 
2. Perform one or more actions. Such as mark a variant causative, add a comment, etc.3. 
3. Go to `index.html` most often `localhost:5000`4. 

**Expected outcome**:
Functionality should be working.
1. Most important events are placed first2. (*)
2. Clicking the case-id will link to the case-page

image

**Review:**
- [ ] code approved by
- [ ] tests executed by

 (*) The order of important events is determined by the macro `VERBS_MAP`  in `scout/constants/case_tags.py`. I think there is room for improvement and I would appreciate feedback on the ordering.


northwestwitch:
In order to test it we should install the branch on the stage vm but I see that the action that is pushing the image on Docker Hub (the same image that is deployed on our stage VM) is failing:

image

I think it doesn't like the / in the branch name?

scout/adapter/mongo/event.py
```diff
@@ -184,9 +184,9 @@ def case_events_by_verb(self, category, institute, case, verb):
         }
         return self.event_collection.find(query)
 
-    def user_events(self, user_obj=None):
+    def user_events(self, user_obj=None, case=None):
```
dnil, line 187: Hm, but querying for` case=None` is probably not what you want? Rather perhaps different queries if the case param is given.

mikaell:
Replaced by #3314 to fix issue with Docker incompatible branch name.

PR #3311, opened 2022-04-21
Add option for admins to create beacon datasets from institute settings

### LINKED ISSUE #3276: Missing dataset error when user tries to load case data to a beacon
The current setup of the beacon app, allows only to submit variants to pre-existing datasets.

So if for instance a user from cust002 wants to submit a case that is sequenced in build 37, they can do it as long as in the beacon database exists a dataset named cust002_GRCh37:

But what about a user of cust003, or a user from cust002 that wants to submit a case sequenced in build 38?

On one hand it would be easier if we create only 2 datasets (cases build 37 and cases in build 38), and all users submit to one of the other according to the genome build found in the case documents. But it makes also sense to separate by institute, since one institute might contain cancer cases, another one cases affected by one pathology instead of another and so on..

the best option would probably be allowing the user to choose which dataset submit to, but this still relies on admins to create as many datasets as the number of institutes with cases to submit to the beacon.. 🤔

### LINKED ISSUE #3299: Admins should be able to create missing beacon datasets for a given customer and genome build
Create this functionality from institute settings page if Scout is connected to a Beacon (connection params in Scout settings) So from that page it (from instance cust001) should be only possible to create 2 datasets for the time being

cust001_GRCh37 (public dataset)
cust001_GRCh38 (public dataset)

In the future we could also introduce the option to create registered and controlled datasets.

The institute datasets above should be created only of they don't already exist

northwestwitch:
It is not possible to submit a case to the beacon is the institute doesn't already have a dataset in beacon. Dataset IDs are something like this `custXXX_GRCh37` or `custXXX_GRCh38`, so it's a controlled dictionary.

Fix #3299 and fix #3276. Scout admins will be able to create a new dataset for ai institute (with controlled dictionary) from the institute settings page:

image

The option will not be visible for non-admin users

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**Before testing**
Your user should have beacon_submitter role in scout-stage database. If it doesn't:
1. ssh to hasta
2. "us"
3. `scout update user -u <your.email@scilifelab.se> -r beacon_submitter`
4. If you are not an admin: `scout update user -u <your.email@scilifelab.se> -r admin`

**How to test**:
1. Install on scout-stage (cg-vm1)
5. Go to a case from cust000 (doesn't have any dataset registered in beacon)
6. Try to add to beacon a recent case from this institute, you should get an error message.
7. Go to cust004 institute settings and create a new beacon dataset id for data in genome build 37
8. Try again to add variants of a cust000 case to the beacon and now it should work.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
This is now ready for review!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -131,6 +132,45 @@ def institute(store, institute_id):
...
+    """Populate the form select of scout.server.blueprints.institutes.forms.BeaconDatasetForm with data.
+    Available data is a controlled dictionary of dataset names depending of the datasets already existing in the Beacon server
+
+    Args:
+        institute_obj(dict) An institute object
+    """
+    beacon_form = BeaconDatasetForm()
+
+    if current_user.is_admin is False:
+        LOG.warning("User must be an admin to create a beacon dataset")
```
dnil, line 145: That will be many warnings; most users are not admin. Better at the actual action endpoint!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -131,6 +132,45 @@ def institute(store, institute_id):
...
+        return beacon_form
+
+    if not all(
+        [current_app.config.get("BEACON_URL") and current_app.config.get("BEACON_TOKEN")]
+    ):  # Send a request to the beacon to collect the existing datasets
+        LOG.warning(
+            "Both BEACON_URL and BEACON_TOKEN params are required in Scout configuration file to create a new Beacon dataset"
```
dnil, line 152: A bit same here; maybe warn if one is set but not the other?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -131,6 +132,45 @@ def institute(store, institute_id):
...
+        )
+        return beacon_form
+
+    # Collect all dataset IDs names present on the Beacon
+    beacon_dsets = beacon.get_datasets()
+
+    dset_options = []  # List of tuples containing dataset select options
+    for build in ["GRCh37", "GRCh38"]:
```
dnil, line 160: Are these already in use? If so, all good, they change slowly. Otherwise perhaps use and share a constant like BUILDS or a new one!

scout/server/blueprints/institutes/forms.py
```diff
@@ -33,7 +33,7 @@ def pre_validate(self, form):
 
 
 class InstituteForm(FlaskForm):
-    """Instutute-specif settings"""
+    """Instutute-specific settings"""
```
dnil, line 36:
```suggestion
    """Institute-specific settings"""
```
😆

scout/server/blueprints/institutes/templates/overview/institute_settings.html
```diff
@@ -31,11 +31,12 @@
   <div class="row" id="body-row"> <!--sidebar and main container are on the same row-->
     {{ institute_actionbar(institute) }} <!-- This is the sidebar -->
   <div class="col">
-    {{ institute_settings(form, institute, current_user, loqus_instances) }}
+    <div>
```
dnil, line 34: Not that it matters, but why the extra div?

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -123,8 +124,32 @@ <h5 class="header">General Institute Settings</h5>
       </div>
     </div>
   </div> <!-- end of div class="row" -->
-</div> <!-- end of container div -->
+
+</form>
+
+<!-- Display form to add a Beacon dataset if user is an andmin and if institute lacks one or more Beaocn datasets -->
+{% if current_user.is_admin and config.BEACON_URL and config.BEACON_TOKEN %}
+<form action="{{ url_for('overview.add_beacon_dataset', institute_id=institute._id) }}" class="form-inline" method="post">
```
dnil, line 132: `form-inline` has been dropped in bs5. Use grid or utils instead! Which you already do, so just skip! 😁

scout/server/blueprints/institutes/views.py
```diff
@@ -144,10 +173,12 @@ def institute_settings(institute_id):
...
 
     institute_obj = store.institute(institute_id)
-    form = InstituteForm(request.form)
+    institute_form = InstituteForm(request.form)
+
+    beacon_form = controllers.populate_beacon_form(institute_obj)
 
     # if institute is to be updated
-    if request.method == "POST" and form.validate_on_submit():
+    if request.method == "POST" and beacon_form.validate_on_submit():
```
dnil, line 181: I'm not sure I follow this; why replace the form.validate - didn't you rather want to add the beacon_form.validate?

scout/server/blueprints/institutes/views.py
```diff
@@ -144,10 +173,12 @@ def institute_settings(institute_id):
         return redirect(request.referrer)
 
     institute_obj = store.institute(institute_id)
-    form = InstituteForm(request.form)
+    institute_form = InstituteForm(request.form)
```
dnil, line 176: 👍

scout/server/extensions/beacon_extension.py
```diff
@@ -84,6 +84,33 @@ def base_submission_data(self, store, case_obj, form):
 
         return data
 
+    def add_dataset(self, institute_obj, dataset_id):
+        """Add a missing dataset for an institute
+
+        Args:
+            institute_obj(dict): scout.models.Institute
+            dataset_it(str): a string like this "<institute_id>_<genome_build>". Example: cust000_GRCh38
```
dnil, line 92:
```suggestion
            dataset_id(str): a string like this "<institute_id>_<genome_build>". Example: cust000_GRCh38
```

tests/server/extensions/test_beacon_extension.py
```diff
@@ -27,6 +27,24 @@ def test_beacon_create_app():
         assert beacon.token
 
 
+@responses.activate
+def test_add_dataset(app, institute_obj):
+    """Test the function that sends a POST request to the Beacon with the info to create a new dataset"""
+
+    # GIVEN a mocked Beacon server dataset add endpoint
+    url = "http://localhost:6000/apiv1.0/add_dataset"
```
dnil, line 35: A fixture (or a constant) for at least the base URL would make it easier to read/maintain!

dnil (approved):
Very useful; sorry for the late review! Only minor comments.

scout/server/blueprints/institutes/controllers.py
northwestwitch: This warning would never show if the params in the config file are set correctly. But it's useful to see a warning if there is a misconfiguration, so one fixes it. I'm going to remove the warning in the code above, but I'd like to leave this one as it is.

scout/server/blueprints/institutes/controllers.py
northwestwitch: These naming is used by the beacon app, not by Scout. In Scout we have often ['37','38']. Was this your question?

northwestwitch: https://github.com/ga4gh-beacon/specification/blob/b151e38e16644d9ae0bff048b859d03d96a34b7c/beacon.yaml#L147

scout/server/blueprints/institutes/templates/overview/institute_settings.html
northwestwitch: Naa, they're just extra, I've removed them, thanks!

scout/server/blueprints/institutes/controllers.py
dnil: Sure; we should never see it in prod anyway. Consider my suggestion, but its fine.

scout/server/blueprints/institutes/controllers.py
dnil: Thanks, it would not hurt to have a const with the official build ids, but the downside I guess could be that someone might use them for case build or such if they are too prominently global. As you wish - this will change very slowly so no biggie.

tests/server/extensions/test_beacon_extension.py
northwestwitch: Good idea, Added it in a constant!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -131,6 +132,44 @@ def institute(store, institute_id):
...
+
+    Args:
+        institute_obj(dict) An institute object
+    """
+    beacon_form = BeaconDatasetForm()
+
+    if current_user.is_admin is False:
+        return beacon_form
+
+    if not all(
```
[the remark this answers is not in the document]
northwestwitch: Sorry I don't understand this suggestion. In order to work properly, the functionality needs 2 parameters in config file: BEACON_URL **and** BEACON_TOKEN. What is this code suggestion checking?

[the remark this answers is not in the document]
dnil: The intention is to warn if one but not both are set. A user can set up without beacon (eg locally) so would not need the warning if both unset.

scout/server/blueprints/institutes/controllers.py
northwestwitch: It can be a constant in the beacon extension I guess, but I wouldn't add it in the scout constants because they are not used in other places in scout other than here

scout/server/blueprints/institutes/controllers.py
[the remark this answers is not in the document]
northwestwitch: Ah OK, you mean if somebody is using scout without beacon. Yes I understand now!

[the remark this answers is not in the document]
northwestwitch: Good point!

scout/server/blueprints/institutes/controllers.py
northwestwitch: MM I don't think the code suggestion works as intended, If I remove both params from the conf file it executes the lines below the if anyway. I'll fix!

scout/server/blueprints/institutes/views.py
northwestwitch: Good catch, this is a typo. I meant to write institute_form.validate_on_submit():

northwestwitch: When you click on submit on the beacon form you don't POST a request to this page, but to another endpoint in the views (`overview.add_beacon_datase`t)

northwestwitch:
@dnil I think I've covered all your comments now. Thanks for the review, good suggestions that prevented potential bugs btw. Do you want to give it another check before I merge?

scout/server/blueprints/institutes/controllers.py
dnil: Perfect - then they also won't be confused.

scout/server/blueprints/institutes/templates/overview/utils.html
dnil, line 132:
```suggestion
<form action="{{ url_for('overview.add_beacon_dataset', institute_id=institute._id) }}" method="post">
```
Or `class="row-cols-*"` but I dont think you need it.

northwestwitch: Ah there was a form-inline. I think you meant that there was no need because I didn't have a form-inline before, that's why I've marked it as resolved, sorry!

northwestwitch: Yes, no need for it, thanks!

image

PR #3312, opened 2022-04-21
Add ranges to rank score results on variant pages - 2

### LINKED ISSUE #791: Add ranking ranges to Rank score results

northwestwitch:
PR based on #3146, with some fixes and the branch renamed to be compliant with the GitHub action to deploys on cg-vm1

fix #791

Add ranges to the rank score results. The range is pulled from the corresponding `rank_model_-<version>-.ini` file in the `Clinical-Genomics/reference-files` repo and stored in the database. Storing it so it does not have to fetch every time. If github is down the first time the ini file is fetched things will not work. I'd build a better fallback if appropriate (?).

New dep (`configobj`) for parsing the ini file. So one need to run pip (`pip install --editable .`) to make magic happen.

**old version**
image

**new version**
image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
All tests pass now. I've addressed also the codefactor problems.

- There are still thing to fix, for instance the rank panel on the SV page doesn't work:

image

- The custom panel styles present on scout/server/blueprints/variant/static/rank_score_results.css should be probably removed and replaced with something more standard

- Tests should be written for the code introduced in this PR

northwestwitch:
Another thing I noticed: given a variant which has rank scores, if case config file doesn't contain `rank_model_version` or `sv_rank_model_version` or the respective model is not found using a request, then no score value is not shown at all.

For instance, take demo case and remove 
```
rank_model_version: '1.20'
sv_rank_model_version: '1.5'
```
from the config file. If you add the case using main branch, the variant will look like this:

image

Using this branch instead, you won't see any rank value at all:

image


Less info is better than no info, so this must be fixed!

northwestwitch:
I've **created tests** and fixed the issue that the link to retrieve the rank model .ini file (only the SNV model code was retrieved - not the SV one) was hardcoded in the adapter. The links to get **both SNV and SV .ini files are now parsed from the app config files**.

Another thing I've changes is that I'm not parsing numbers from the model items (for instance spllicing from -5 to 12). The reason why is that the parsing was bugged and the values not very informative, I think. I've decided instead to **display all the model values in a popover that describes what the eventual values might represent**, like this:

image

image

northwestwitch:
This is ready for review now!

dnil:
> I've **created tests** and fixed the issue that the link to retrieve the rank model's .ini file (only the SNV model code was retrieved - not the SV one) was hardcoded in the adapter. The links to get **both SNV and SV .ini files are now parsed from the app config file**.

Super with tests and the sv .ini file parsing! 

> Another thing I've changed is that I'm not parsing numbers from the model items (for instance spllicing from -100 to 100). The reason is that the parsing was bugged and the values not very informative, I think. I've decided instead to **display all the model values in a popover that describes what the eventual values might represent**, like this:
> 
> image
> 
> image

Hm, but that also does not display the mode of aggregaition - like sum, min, max so the user would be about as informed as with the faulty parsing. If you wish, I could give the parsing a shot?

dnil (commented):
The aggregation mode for category needs to be recorded at least for display, and when that is done, the range should be rather computable again, right?

northwestwitch:
> The aggregation mode for category needs to be recorded at least for display, and when that is done, the range should be rather computable again, right?

Agreed! Shouldn't be hard, I can fix after lunch!

northwestwitch:
> The aggregation mode for category needs to be recorded at least for display, and when that is done, the range should be rather computable again, right?

image

Aggregation rule is displayed now. The ranges are computable. Shall we include them or you think it is sufficient to show the model info like this? Because it's a bit more of parsing and perhaps it's not necessary?

dnil:
Super nice now: both the storing and parsing (thanks @fc-io), display of full item (@northwestwitch 🦸) and the actual ranges with category aggregation model accounted for that I've been going on and on about.
Screenshot 2022-04-26 at 18 26 32

dnil:
Ok, I think I'm done now - check and see what you think (tomorrow obviously)! 😁

northwestwitch:
Testing the branch on stage: there is still a problem in the code. I'll fix!

```
2022-04-27 07:44:19,246 - [scout.server.app](http://scout.server.app/) - ERROR: Exception on /cust002/F0006270-2/208a09ede2020bc3d1f7b99e03ff88a1 [GET] [in /venv/lib/python3.8/site-packages/flask/app.py:1455]
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 2077, in wsgi_app
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1525, in full_dispatch_request
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1523, in full_dispatch_request
    rv = self.dispatch_request()
  File "/venv/lib/python3.8/site-packages/flask/app.py", line 1509, in dispatch_request
  File "/home/worker/app/scout/server/utils.py", line 71, in decorated_function
    context = func(*args, **kwargs)
  File "/home/worker/app/scout/server/blueprints/variant/views.py", line 49, in variant
    data = variant_controller(store, institute_id, case_name, variant_id=variant_id)
  File "/home/worker/app/scout/server/blueprints/variant/controllers.py", line 375, in variant
    "rank_score_results": variant_rank_scores(store, case_obj, variant_obj),
  File "/home/worker/app/scout/server/blueprints/variant/controllers.py", line 418, in variant_rank_scores
    score["model_ranges"] = store.get_ranges_info(rank_model, category)
  File "/home/worker/app/scout/adapter/mongo/rank_model.py", line 101, in get_ranges_info
    if category in rank_model_categories:
TypeError: argument of type 'NoneType' is not iterable
```

northwestwitch:
This PR has be be accompanied by another PR in servers to fix the URs in the scout config file to retrieve the .ini files, I'll create it!

northwestwitch:
PR created on servers --> https://github.com/Clinical-Genomics/servers/pull/805

scout/server/blueprints/variant/controllers.py
```diff
@@ -413,12 +413,11 @@ def variant_rank_scores(store, case_obj, variant_obj):
             rm_link_prefix, rank_model_version, rm_file_extension
         )
         # Loop over each rank score category and collect model explanation to display on variant page
-        for score in rank_score_results:
-            category = score.get("category")  # examples: Splicing, Consequence, Deleteriousness
-            if rank_model is None:
-                continue
-            score["model_ranges"] = store.get_ranges_info(rank_model, category)
-            (score["min"], score["max"]) = store.range_span(score["model_ranges"])
+        if rank_model:
```
dnil, line 416: 👍

northwestwitch:
After deploying the change on the scout config file on stage server it works! 🥳 

image

dnil:
FC also had a nice touch up filter for capitalising the Categories on table display. One moment, I'll dig it up..

dnil:
I don't think we need category abbrevs, do we? Let's keep them in mind if someone complains about the page width.

northwestwitch:
> I don't think we need category abbrevs, do we? Let's keep them in mind if someone complains about the page width.

They were hidden using css in the original PR, also because many abbreviation were exactly the same as the extended categories

dnil:
Oki, I have a good feeling about this one now! Anything else we should cover? Otherwise we merge?!

dnil (approved):
I like! Thanks both!

northwestwitch:
> I like! Thanks both!

Thank you! Let's merge!

PR #3314, opened 2022-04-22
Add Eventlog to Index.html 

mikaell:
**Requires Python ≥ 3.7** Why? Because dicts will maintain order without being OrderedDicts().

**Replaces PR #3310 which had a branch name incompatible with Docker**

This PR adds a functionality. Now a summary of a user's recent and important events are shown on index.html. Each entry will be link to the case in question. Reference issue #3041.

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
</details>

**How to test**:
1. Load one or several cases2. 
2. Perform one or more actions. Such as mark a variant causative, add a comment, etc.3. 
3. Go to `index.html` most often `localhost:5000`4. 

**Expected outcome**:
Functionality should be working.
1. Most important events are placed first2. (*)
2. Clicking the case-id will link to the case-page

image

**Review:**
- [ ] code approved by
- [ ] tests executed by

 (*) The order of important events is determined by the macro `VERBS_MAP`  in `scout/constants/case_tags.py`. I think there is room for improvement and I would appreciate feedback on the ordering.

PR #3318, opened 2022-04-27
Configure an optional session timeout - now you get logged out after a month of inactivity!

northwestwitch:
Might be safer to log out a user after an amount of minutes of inactivity. **Users get logged out after a month (!) of inactivity at the moment**. That's because a month is the default value when you don't set any session value.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Locally, uncomment and modify the `SESSION_TIMEOUT_MINUTES` config file to 1
2. Run demo scout and log in.
3. Make sure that you are logged out after a minute

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/.github/workflows/vulture.yml
+++ b/.github/workflows/vulture.yml
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 --ignore-names cls,args,kwargs ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls,args,kwargs,real_variant_database ${{steps.files.outputs.all}}
         continue-on-error: true
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -95,7 +95,6 @@ def create_app(config_file=None, config=None):
                 # combine visited URL (convert byte string query string to unicode!)
                 next_url = "{}?{}".format(request.path, request.query_string.decode())
                 login_url = url_for("public.index", next=next_url)
-                LOG.error("HERE BITCHES")
                 return redirect(login_url)
 
     return app
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -4,7 +4,7 @@ import os
 from datetime import timedelta
 
 import coloredlogs
-from flask import Flask, current_app, redirect, request, session, url_for
+from flask import Flask, current_app, redirect, request, url_for
 from flask_babel import Babel
 from flask_cors import CORS
 from flask_login import current_user
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -1,9 +1,10 @@
 """Code for flask app"""
 import logging
 import os
+from datetime import timedelta
 
 import coloredlogs
-from flask import Flask, current_app, redirect, request, url_for
+from flask import Flask, current_app, redirect, request, session, url_for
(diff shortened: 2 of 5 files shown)


.github/workflows/vulture.yml
```diff
@@ -19,5 +19,5 @@ jobs:
         uses: anaynayak/python-vulture-action@v1.0
         id: vulture
         with:
-          vulture-args: --min-confidence 80 --ignore-names cls,args,kwargs ${{steps.files.outputs.all}}
+          vulture-args: --min-confidence 80 --ignore-names cls,args,kwargs,real_variant_database ${{steps.files.outputs.all}}
```
dnil, line 22: One could just save it to something local as well, or disable quality checking for those lines, but sure, why not!

scout/server/app.py
```diff
@@ -60,6 +62,12 @@ def create_app(config_file=None, config=None):
     if config_file:  # Params from an optional .py config file provided by the user
         app.config.from_pyfile(config_file)
 
+    session_timeout_minutes = app.config.get("SESSION_TIMEOUT_MINUTES")
+    if session_timeout_minutes:
+        session_duration = timedelta(minutes=session_timeout_minutes)
+        app.config["PERMANENT_SESSION_LIFETIME"] = session_duration
```
dnil, line 68: Its time..

scout/server/config.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 SECRET_KEY = "this is not secret..."
-REMEMBER_COOKIE_NAME = "scout_remember_me"
+REMEMBER_COOKIE_NAME = "scout_remember_me"  # Prevent session timeout when user closes browser
+# SESSION_TIMEOUT_MINUTES = 60  # Minutes of inactivity before session times out
```
dnil, line 4: 👍

dnil (approved):
Looks good!

scout/server/config.py
northwestwitch: Should we fix our config files you think? And how much time do we set?

.github/workflows/vulture.yml
northwestwitch: This looked easier 😆

scout/server/config.py
dnil: Yes. But possibly wait to deploy that until the personal event log PR is merged?

On the order of 100h should more or less not be noticed. Also anything on the order of 10h should be rather hard to argue against. Setting it to anything below 1h will have an impact on how one can work with open tabs, returning to tasks or showing others partial results. 

How about 10h for starters then? It may force a kind of wrap-up for the day in the evening. But having a recent log to bounce back to would lessen the need for open tabs and the like?

northwestwitch: > Yes. But possibly wait to deploy that until the personal event log PR is merged?

I think it's more the other way around, that's why I wrote this PR. Checking #3314 I've realized that if you are logged out and you log in you don't get to see the events panel, but you are immediately redirected to the instituteS page. I've commented on that PR that I think the events panel should be moved to some other place. Perhaps you could take also a look at that PR to see what I mean?

northwestwitch: > How about 10h for starters then?

10 h sounds good. It will mean login every day

PR #3322, opened 2022-05-03
Minor version 4.52

northwestwitch:
### Added
- [x] Demo cancer case gets loaded together with demo RD case in demo instance
- [x] Parse REVEL_score alongside REVEL_rankscore from csq field and display it on SNV variant page **(_automatic tests only_)**
- [x] Rank score results now show the ranking range
- [x] cDNA and protein changes displayed on institute causatives pages
- [x] Optional SESSION_TIMEOUT_MINUTES configuration in app config files
- [x] Script to convert old OMIM case format (list of integers) to new format (list of dictionaries)
- [x] Additional check for user logged in status before serving alignment files
- [x] Download .cgh files from cancer samples table on cancer case page
- [x] Number of documents and date of last update on genes page
### Changed
- [x] Verify user before redirecting to IGV alignments and sashimi plots
- [x] Build case IGV tracks starting from case and variant objects instead of passing all params in a form
- [x] Unfreeze Werkzeug lib since Flask_login v.0.6 with bugfix has been released
- [x] Sort gene panels by name (panelS and variant page)
- [x] Removed unused `server.blueprints.alignviewers.unindexed_remote_static` endpoint
- [x] User sessions to check files served by `server.blueprints.alignviewers.remote_static` endpoint
- [x] Moved Beacon-related functions to a dedicated app extension
- [x] Audit Filter now also loads filter displaying the variants for it
### Fixed
- [x] Handle `attachment_filename` parameter renamed to `download_name` when Flask 2.2 will be released
- [x] Removed cursor timeout param in cases find adapter function to avoid many code warnings
- [x] Removed stream argument deprecation warning in tests
- [x] Handle `no intervals found` warning in load_region test
- [x] Beacon remove variants
- [x] Protect remote_cors function in alignviewers view from Server-Side Request Forgery (SSRF)
- [x] Check creation date of last document in gene collection to display when genes collection was updated last

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN


northwestwitch:
Let's try to wrap up the work we made so far so we can have a new release soon!

northwestwitch:
Deployed on stage now

northwestwitch:
I've just deployed the [changes of the scout config]( https://github.com/Clinical-Genomics/servers/pull/806) to have the user's session time out after 10h. I'll restart the scout-stage server so we can make sure it works!

dnil:
Super for testing; I still think we should wait with changing that for prod until we have the history PR merged as well.

dnil:
It´s easy to loose track of what you were doing when you have to log in again, if you are not very careful about what you are doing. Will be very nice to have a log.

northwestwitch:
Regarding the tests for `Handle attachment_filename parameter renamed to download_name when Flask 2.2 will be released`:

- download_name is used in cases views, `pdf_case_report`
- PDF gene panel exports
- MT report export
- Download verified vars from dashboard

They all work!

dnil:
But, let's see, the session timeout didn't, did it? Maybe the stage-config wasn't updated yet? We could test locally with a few minutes I guess!

northwestwitch:
> But, let's see, the session timeout didn't, did it? Maybe the stage-config wasn't updated yet? We could test locally with a few minutes I guess!

Didn't work for me, but I've been offline for less than 10 hours perhaps?

northwestwitch:
> Didn't work for me, but I've been offline for less than 10 hours perhaps?

But no, I definitely wasn't on Scout last night..

dnil:
The session logout works, but it will take some time before it makes a difference, as most active users will still have cookie tokens valid for a month. They should be the last long-term ones though. We should be able to force a new login from all by drawing a new secret key, but again, I think we should wait with that until we have the event history merged.

northwestwitch:
> The session logout works, but it will take some time before it makes a difference, as most active users will still have cookie tokens valid for a month.

Ah that why?!

northwestwitch:
I've re-deployed the updated branch on stage now

PR #3328, opened 2022-05-06
Gene collection update time corresponds to last gene creation date

northwestwitch:
Genes in build 37 and 38 can be updated separately, so instead of collecting a random gene to see when gene collection was updated, use date of last inserted gene.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/genes/controllers.py
+++ b/scout/server/blueprints/genes/controllers.py
@@ -26,7 +26,7 @@ def genes(store, query):
     nr_genes_37 = store.nr_genes(build="37")
     nr_genes_38 = store.nr_genes(build="38")
     genes_subset = list(store.all_genes(add_transcripts=False, limit=20))
-    last_inserted_gene = store.hgnc_collection.find({}).sort([("timestamp", -1)]).limit(1)
+    last_inserted_gene = store.hgnc_collection.find({}).sort("_id", -1).limit(1)
     last_updated = document_generated(last_inserted_gene[0]["_id"] if last_inserted_gene else None)
     return dict(genes=genes_subset, last_updated=last_updated, nr_genes=(nr_genes_37, nr_genes_38))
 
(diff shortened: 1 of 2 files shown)


northwestwitch:
Tested locally: 

image

scout/server/blueprints/genes/controllers.py
```diff
@@ -26,7 +26,8 @@ def genes(store, query):
     nr_genes_37 = store.nr_genes(build="37")
     nr_genes_38 = store.nr_genes(build="38")
     genes_subset = list(store.all_genes(add_transcripts=False, limit=20))
-    last_updated = document_generated(genes_subset[0]["_id"] if genes_subset else None)
+    last_inserted_gene = store.hgnc_collection.find({}).sort([("timestamp", -1)]).limit(1)
```
dnil, line 29: :+1: that should work!

dnil: (I never realised the ObjectIds themselves were part creation time stamp! Nothing preventing the rest of the document from changing I suppose, but I think we drop the relevant documents when updating, right?)

dnil (approved):

scout/server/blueprints/genes/controllers.py
dnil: > Tested locally:
> 
> image

Very few hg38 genes you have there! 😄

northwestwitch: I don't think my previous version worked? 

It selected a document created at 15 something.

If you instead use `.sort('_id',-1).limit(1)` it returns a document created after (16 something)

scout/server/blueprints/genes/controllers.py
```diff
@@ -26,7 +26,7 @@ def genes(store, query):
     nr_genes_37 = store.nr_genes(build="37")
     nr_genes_38 = store.nr_genes(build="38")
     genes_subset = list(store.all_genes(add_transcripts=False, limit=20))
-    last_inserted_gene = store.hgnc_collection.find({}).sort([("timestamp", -1)]).limit(1)
+    last_inserted_gene = store.hgnc_collection.find({}).sort("_id", -1).limit(1)
```
dnil, line 29: Yes, its basically the same, as the four most significant bytes of the _id is the timestamp!

northwestwitch: Nope, update time didn't change:

image


Perhaps the server timezone?

PR #3334, opened 2022-05-09
Fix #3331 - parse dbSNP CSQ

### LINKED ISSUE #3331: Variant IDs in MIP11 no longer rsID
Scout assumes the ID column of VCF contain dbSNP rsID. As of MIP11 this appears to be a short variant ID similar to the Scout variant display ID instead (`chr_pos_ref_alt`). This breaks linkout to dbSNP, and looks a little funny:
Screenshot 2022-05-06 at 16 38 54

Presumably the intention is to populate the `INFO.dbSNP` field with the rsID, so we should likely switch parsing the dbSNP id to that as an alternative if the ID column does not match `/rs.*/` (cosmic not considered). From MIP11 VCF:
```
##INFO=<ID=dbSNP,Number=A,Type=String,Description="dbSNP ID of the variant">
```
but this field does not seem populated. Likely VEP needs another switch as well, @jemten?

dnil:

Attempt parsing a MIP file, ensure the situation in #3331 with variant ID-like strings in dbSNP button does not occur. If a recent patch of MIP11, VEP CSQ fields should instead be parsed, and dbSNP ids be visible again.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN on stage and GitHub Actions


scout/parse/variant/variant.py
```diff
@@ -282,8 +283,9 @@ def parse_variant(
     cosmic_ids = set()
     for parsed_transcript in parse_transcripts(raw_transcripts):
         parsed_transcripts.append(parsed_transcript)
-        for dbsnp in parsed_transcript.get("dbsnp", []):
-            dbsnp_ids.add(dbsnp)
+        for dbsnpkey in ["dbsnp", "rs_dbSNP150", "rs_dbSNP"]:
+            for dbsnp in parsed_transcript.get(dbsnpkey, []):
+                dbsnp_ids.add(dbsnp)
```
northwestwitch, line 288: What about duplicates? What if rs_dbSNP150 is rs12345 and rs_dbSNP is also rs12345?

dnil: Right, I guess that's why its a set, though it is not super likely that someone would have both?

northwestwitch: ah yes, of course, it's using add and not append. Sorry, lack of sleep!

northwestwitch (approved):
Looks good, but of course let's load a MIP11 case to test this pR and the GATK one before deploying!

dnil:
Yes, there should be a patch version for `justhusky` now! I can check.

scout/parse/variant/transcript.py
```diff
@@ -252,13 +252,15 @@ def get_sift_prediction(entry):
...
-    variant_ids = entry.get("EXISTING_VARIATION")
-
-    if variant_ids:
-        for variant_id in variant_ids.split("&"):
-            if variant_id.startswith("rs"):
-                dbsnp_list.append(variant_id)
+    dbsnp_list = set()
+    dbsnp_keys = ["EXISTING_VARIATION", "RS_DBSNP150", "RS_DBSNP"]
+
+    for key in dbsnp_keys:
```
northwestwitch, line 258: Wow looks more complicated now! Perhaps expand also docstring with an example of a complicated entry?

dnil: Yes, and probably another parse test for a current VEP; the VEP97 plus VEP OldSkool+ ones we carry are not necessarily current.

dnil:
Ok, it works on stage, but I'll add that test for good measure before we merge. 

Before:
Screenshot 2022-05-10 at 14 14 25

After:
Screenshot 2022-05-10 at 16 40 16

PR #3342, opened 2022-05-16
Select pinned variants to export to ClinVar directly from case page

### LINKED ISSUE #3188: Move ClinVar submission button from variant pages to case page (pinned vars panel)
This way the feature would be more evident

northwestwitch:
The current situation is that if the variant is pinned, **there is a button in variant page** (classify panel) to export it to ClinVar.
Alternatively one can export ALL pinned variants to ClinVar if the selected checkbox of ClinVar export page is selected.

But **it is not possible to specifically select which variants to include in ClinVar submissions**

This PR adds a functionality or fixes a bug:
- Moves button to create ClinVar submissions from variant page to case page (fix #3188)
- Allows user to specifically select the variants to include

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Pin one or more variants (SNVs and SNVs) for a case and make sure that creation of excel files to be used in ClinVar submissions work

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


scout/server/blueprints/variant/controllers.py
```diff
@@ -644,36 +644,32 @@ def build_clinvar_submission(store, request, institute_id, case_name, variant_id
...
-    ]
-    variant_obj = store.variant(variant_id)
+    variant_objs = [store.variant(variant_id) or variant_id for variant_id in variants]
 
     # gather missing transcript info from entrez (refseq id version)
-    for pinned_var in pinned:
+    for var_obj in variant_objs:
         # Exclude variants that aren't loaded
-        if isinstance(pinned_var, str):
+        if isinstance(var_obj, str):
```
mikaell, line 670: Is `var_obj` a string (as tested and expected) or something else?

scout/server/blueprints/variant/controllers.py
```diff
@@ -644,36 +644,32 @@ def build_clinvar_submission(store, request, institute_id, case_name, variant_id
...
     case_diagnoses = case_obj.get("diagnosis_phenotypes", [])
     if case_diagnoses and isinstance(case_diagnoses[0], int):
         case_obj = store.convert_diagnoses_format(case_obj)
-    pinned = [
-        store.variant(variant_id) or variant_id for variant_id in case_obj.get("suspects", [])
-    ]
-    variant_obj = store.variant(variant_id)
+    variant_objs = [store.variant(variant_id) or variant_id for variant_id in variants]
 
     # gather missing transcript info from entrez (refseq id version)
```
mikaell, line 667: Typo: entrez

scout/server/blueprints/variant/controllers.py
```diff
@@ -644,36 +644,32 @@ def build_clinvar_submission(store, request, institute_id, case_name, variant_id
...
     """Gather the required data for creating the clinvar submission form
 
     Args:
         store(scout.adapter.MongoAdapter)
-        institute_id(str): Institute ID
-        case_name(str): case ID
-        variant_id(str): variant._id
+        institute_obj(str): Institute ID
+        case_obj(str): case ID
+        variants(list): list of variant._id strings
```
mikaell, line 654: Are these variables strings or instances of a class? I find the code hard to read. In previous version they were named for example `case_name` and naturally interpreted  as a string. Now named `case_obj` but is it a string as the docstring says or an instance of a class, as the name indicates?

mikaell (commented):
See comments. Most regard naming and documentation.

scout/server/blueprints/variant/views.py
```diff
@@ -299,17 +299,28 @@ def acmg():
     return jsonify(dict(classification=classification))
 
 
-@variant_bp.route("/<institute_id>/<case_name>/<variant_id>/clinvar", methods=["POST", "GET"])
+@variant_bp.route("/<institute_id>/<case_name>/clinvar", methods=["POST"])
 @templated("variant/clinvar.html")
-def clinvar(institute_id, case_name, variant_id):
-    """Build a clinVar submission form for a variant."""
+def clinvar_create(institute_id, case_name):
+    """Create a ClinVar submission document in database for one or more variants from a case."""
```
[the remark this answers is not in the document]
northwestwitch: In general we don't include docstrings in views functions, only in controllers and functions of any other module

northwestwitch:
Thanks for the review! I have now fixed according to your comments:
- Fixed adding new vars to the submission (good catch!)
- FIxed confusing docstring in variant.controllers function

I haven't added detailed docstring to the create_submission view since all views get data from requests and not from other functions. So the params are always request.args (GET requests) or request.form (POST requests). We don't have detailed docstrings in any other view that is a request endpoint.

scout/server/blueprints/variant/controllers.py
northwestwitch: Fixed

scout/server/blueprints/variant/controllers.py
northwestwitch: Removed

scout/server/blueprints/variant/controllers.py
northwestwitch: I'll add a comment. basically when a case has a variant is an object, but if a case gets re-uploaded without a variant, this becomes a string

northwestwitch: Ah there is already a comment

northwestwitch:
I've just realized that I've modified only the SNV variant page with the code to create a ClinVar submission object. i need to do the same on SV variant page and cancer page as well otherwise it won't work. Converting this PR to draft now!

northwestwitch:
The old links were present only in SV variant pages, I've removed them so it should be OK now!

mikaell (approved):
Code looks good. I'll try it out once more before merging though.

mikaell:
Functionality seems to work. 

Feel free to merge 🎊🎉

PR #3382, opened 2022-06-08
Fix variants' pagination after filter

### LINKED ISSUE #3381: Appearance that no variants returned after filter
Hi all, thanks for the great software!

We've noticed a behaviour in scout whereby if you are a few pages in when looking at variants (ie. not on the first page of variants), and then apply a filter that reduces the number of variants to fewer than is able to fill to the current page, on face value it can appear that no variants were returned.

This behaviour should be reproducible with the demo database provided by scout. For case 643594, looking at "Clincial SNV and INDELs", navigate to page 2 of variants, then apply a filter for CADD score of 20. This returns a page that says "No matching variants" while also saying that "Filter returns 6 / 680 variants". This is remedied by clicking the "First page" button to take you back to the first page of variants where the 6 variants can be viewed. I interpret this to mean I'm still on page 2 after the filter is applied, but there are not enough variants after my filter to fill up 2 pages, hence my page is empty.

no_variants_returned

Although this is workaroundable by simply clicking to the first page, it caused some confusion with our collaborators at the hospital and I'd argue it would be more logical to return to the first page when a new filter is applied. What are your thoughts?

I'm using [scout version 4.53](https://github.com/Clinical-Genomics/scout/tree/v4.53)

Let me know if you require any further information or if I can help with testing :)

Leah

northwestwitch:
Fix #3381 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Locally or on stage, go to variants list, page 2 for a case.
2. Apply some filter and make sure that first page of vars is shown
3. Repeat for cancer variants and structural variants

**Expected outcome**:

**Review:**
- [x] code approved by DN


dnil (approved):
That should do it! Don't forget the cancer and cancer-sv filters (in utils).

scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
```diff
@@ -175,8 +175,8 @@
         </a>
       </div>
       <div class="col-10 d-flex">
-        {{ form.filter_variants(class_="btn btn-primary form-control") }}
-        {{ form.clinical_filter(class_="btn btn-info form-control") }}
+        {{ form.filter_variants(class_="btn btn-primary form-control", onclick="resetPage()") }}
```
dnil, line 178: Oooh, this is an unused macro, isn't it? The one that is called is as the others in utils. Perhaps a good time to merge them? Or to make an issue, its hardly a rush.

northwestwitch:
> That should do it! Don't forget the cancer and cancer-sv filters (in utils).

Those should be already accounted for, since also those pages 

> That should do it! Don't forget the cancer and cancer-sv filters (in utils).

Cancer SV filters should be fine because it's using the same macro as SV filters. now I've fixed cancer filters. I'm deploying on stage so we can test with real cases!

northwestwitch:
> > That should do it! Don't forget the cancer and cancer-sv filters (in utils).
> 
> Those should be already accounted for, since also those pages
> 
> > That should do it! Don't forget the cancer and cancer-sv filters (in utils).
> 
> Cancer SV filters should be fine because it's using the same macro as SV filters. now I've fixed cancer filters. I'm deploying on stage so we can test with real cases!

Ah no, cancer SV filter is not OK yet!

scout/server/blueprints/variants/templates/variants/cancer-sv-variants.html
northwestwitch: Let's just remove it, no?

northwestwitch:
Ok, cancer vars look good now on stage. I'm going to merge

PR #3404, opened 2022-06-13
Fix logo bg, update logo

dnil:

Update KUH logo, fix dark mode backgrounds so it doesn't look like we gave the logo a new shape - that would violate the graphics profile.

Screenshot 2022-06-13 at 13 13 50

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Nice! Would it be possible to add a white background also to the GMS logo, so it will look similar to the other ones?

dnil:
The thought crossed my mind. I can check if they have a positive logo! That one is negative, so the text would not be visible on white. Also it might be a large white area in the end if we keep the separation and don't want to add borders around the GMS logo either that might look like it has a new shape.

northwestwitch:
> The thought crossed my mind. I can check if they have a positive logo! That one is negative, so the text would not be visible on white. Also it might be a large white area in the end if we keep the separation and don't want to add borders around the GMS logo either that might look like it has a new shape.

Check how it looks like, otherwise this PR is already fine I think!

scout/server/blueprints/public/templates/public/index.html
```diff
@@ -63,12 +63,12 @@ <h1 class="display-4">Scout</h1>
     {% endif %}
 
     <p class="pb-1 mb-0">Developed by:</p>
-    <div class="row justify-content-center align-items-center">
+    <div class="row justify-content-center align-items-center bg-white rounded-3">
```
northwestwitch, line 66: 👍🏻

northwestwitch (approved):
Nice!

dnil:
> > The thought crossed my mind. I can check if they have a positive logo! That one is negative, so the text would not be visible on white. Also it might be a large white area in the end if we keep the separation and don't want to add borders around the GMS logo either that might look like it has a new shape.
> 
> Check how it looks like, otherwise this PR is already fine I think!

If I switch to light mode - not that anyone would like to do that knowingly, mind you - but assuming that some might not have changed their settings yet, the GMS logo text is not visible on light mode. I'll try giving it a row of darker bg, or ask for a positive logo. They do have it in some presentation, but no material online that I found.

dnil:
I've asked for a positive color GMS logo, and should probably ask for a negative KUH-one, but in the meantime this should comply with their profiles, if not being the height of elegance.

PR #3453, opened 2022-06-17
display timeline

### LINKED ISSUE #3041: Personal event history

northwestwitch:
A timeline of events page for a user. ---> https://scout-stage.scilifelab.se/overview/timeline

fix #3041

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Deploy on stage and check the page!

**Expected outcome**:

**Review:**
- [x] code approved by DN


northwestwitch:
I don't really know how to fix the 2 bugs found by sonarcloud since they are not really bug.

It complains that I have `<li>stuff</li>` but the <ul> is there, just not on the very html page, but on the template that gets imported in the page

northwestwitch:
I think this is ready for review now! I could definitely use some feedback on the functionality

dnil:
> I don't really know how to fix the 2 bugs found by sonarcloud since they are not really bugs.
> 
> It complains that I have `<li>stuff</li>` that are not enclosed by an `<ul>` but the `<ul>` is there, just not on the very html page, but on the template that gets imported in the page 😠

I report those as false positives. It’s just sonarcloud not being able to parse jinja macro.

northwestwitch:
Yesterday I've made some more improvements that are pushed now!

scout/server/blueprints/institutes/templates/overview/timeline.html
```diff
@@ -0,0 +1,98 @@
...
+
+    <!-- Improve different cases visibility by introducing a space between the block of events of each different case -->
+    {% if eventg._id.case_id !=  ns.last_case %}
+      {% set ns.space_top = true %}
+    {% else %}
+      {% set ns.space_top = false %}
+    {% endif %}
+
+    <!-- Set icon to display and according to verb type-->
+    {% if eventg._id.verb in ["assign", "unassign"] %}
```
dnil, line 48: This looks like a config const dict, right?

scout/server/blueprints/institutes/views.py
```diff
@@ -37,6 +36,14 @@
 )
 
 
+@blueprint.route("/overview/timeline", methods=["GET"])
+@templated("overview/timeline.html")
+def timeline():
+    data = {"events": controllers.get_timeline_data(request.args.get("limit") or "100")}
+    data["verbs_map"] = VERBS_MAP
```
dnil, line 43: 👍

scout/server/templates/layout.html
```diff
@@ -50,6 +50,17 @@
...
             {% block top_nav_right %}
               {% if current_user.is_authenticated %}
+                <li class="nav-item dropdown">
+                  <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown">Timeline<span class="caret"></span></a>
+                  <ul class="dropdown-menu">
+                    <li>
+                      <a class="dropdown-item" href="{{ url_for('overview.timeline', limit=100) }}">Last events</a>
+                    </li>
+                    <li>
+                      <a class="dropdown-item" href="{{ url_for('overview.timeline', limit=-1) }}">All events</a>
```
dnil, line 60: Heh, that is a very long list for some users. One could paginate, filter on time, or allow users to set a count. But, as always, improvements can happen later - having a start is good.

scout/server/blueprints/institutes/controllers.py
```diff
@@ -31,6 +31,29 @@
...
+        limit(str): "100" to display latest events or "-1" to display all events
+    """
+    timeline_results = []
+    results = store.user_timeline(current_user.email, int(limit))
+    for eventg in results:  # Add links to cases pages
+        case_obj = store.case(case_id=eventg["_id"]["case_id"])
+        if case_obj is None:
+            continue
+        eventg["_id"]["case_name"] = case_obj.get("display_name")
+        eventg["_id"]["link"] = url_for(
```
dnil, line 48: Consider grouping consecutive events for the same case. It easily gets a bit redundant with many (different) events for the same case. Those could be in running text, or on separate lines with the link button only once.
Screenshot 2022-06-18 at 16 20 22

dnil: It would be super sweet to know a little about the gene, variant affected by pinning or causation, or the old/new status of status changes. But, that easily gets extensive and different for different variant categories etc. We can always add later.

dnil (approved):
I think this is a very useful and pretty addition! There are a few things that could do with polishing, especially more grouping. I have a feeling you have mostly looked at users with rather few events per case, not really at users with "normal" RD use, right? But it works, and seems a solid basis for further improvement - and is a decided improvement already.

scout/server/blueprints/institutes/controllers.py
northwestwitch: Ohi look like you are still seeing the old version! I haven't deployed the newest on stage. Newest version is a bit less of a eyesore and has case grouping as well!

northwestwitch: I moved the icons to the constants and did. better case grouping. Now it looks like this:

image

northwestwitch: I've deployed the latest on stage now!

scout/server/templates/layout.html
northwestwitch: In the lastest version of this PR I've included this: 
image

I've tested on stage with my user and it's fast enough to retrieve all events for my user since 2018. I'm wondering how it looks like for you (you should have many more events in the database than me). If it gets stuck I can change it to show less events instead!

dnil:
Nice! Events that give both case and variant entries are still doubled up (like causative, pin, Sanger, ...). Since we are merging there won't be any links for the variants, so we may as well go with only one set of those.

Screenshot 2022-06-20 at 08 20 48


On the nitpicking side the back-to-top can end up overlapping with an event icon:
Screenshot 2022-06-20 at 08 25 52

northwestwitch:
> Nice! Events that give both case and variant entries are still doubled up (like causative, pin, Sanger, ...)

Right, I'll fix it!

scout/server/templates/layout.html
dnil: Yes, again, good start - that was actually there, I updated stage with what was on the repo when I reviewed. It loads, and rather quickly - then our highest ranking officers can have an order of magnitude more. I dont know what the best is, but on the order of hundreds sounds like a starting point. Perhaps make the lowest level like a few hundred instead? The goal should be to cover what one might keep open in tabs for later perusal over a few days.

northwestwitch: I've changed it to display 50 and 500 events now:

image


The upper level should give just the latest stuff. 500 events should go back in time a bit more (a month or so perhaps) for the most active users. Does it look OK?

PR #3470, opened 2022-06-22
Local archive date and count

dnil:

Thanks to MIP11, values for count and date of archive (or fall back to description with filename) is available.

See https://github.com/Clinical-Genomics/MIP/pull/1763.

Screenshot 2022-06-22 at 17 48 20

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR


scout/adapter/mongo/variant_loader.py
```diff
@@ -379,6 +380,8 @@ def _load_variants(
             category(str): ['snv','sv','cancer','str']
             sample_info(dict): A dictionary with info about samples.
                                Strictly for cancer to tell which is tumor
+           custom_images(dict): A dict with custom images for a case.
```
northwestwitch, line 383: 💯

scout/parse/variant/headers.py
```diff
@@ -23,6 +23,42 @@ def parse_rank_results_header(vcf_obj):
     return rank_results_header
 
 
+def parse_local_archive_header(vcf_obj):
+    """Return a dict with local archive data for the case.
+
+    Check if the rank result is in the vcf header. If it exists return a dict
```
northwestwitch, line 29: Isn't this the loqusdb stats and not the rank results?

scout/parse/variant/headers.py
```diff
@@ -23,6 +23,42 @@ def parse_rank_results_header(vcf_obj):
...
+    for header_line in vcf_obj.raw_header.split("\n"):
+        if "Software" in header_line and "loqusdb" in header_line:
+            software_entry = header_line.split("<")[1]
+            entries = software_entry.split(",")
+            for entry in entries:
+                assignment = entry.split("=")
+                the_key = assignment[0]
+                if the_key == "Date":
+                    the_value = assignment[1]
+                    local_archive_header["Date"] = the_value
```
northwestwitch, line 57: 👍🏻

northwestwitch (approved):
Good job, we can finally get rid of the old issue of the hard coded (wrong date) on the old obs panel! 🥇 🥳 

Works nicely (tested on the demo instance), We could perhaps try to load one real MIP11 case later as well.

I've noticed that in the code you interchange the `local_archive_info` and `local_archive_header` terms. I know it's the same thing, since the info is captured from the header (it works slightly different from rank_results_header for instance, where the header is further used to capture values from the variant lines). But might be confusing, Why don't you add a comment in the code saying just that?

You could also add a test. Should be easy now that the variants VCF contains a MIP11-like header no?

scout/parse/variant/headers.py
dnil: Indeed! 😆

northwestwitch:
Tested on Hasta server by uploading  a recent case [lovinggoat](https://scout-stage.scilifelab.se/cust000/NA24143-200ng/variants?variant_type=clinical).

The local observation doesn't change from the same case loaded on prod server. Isn't this case a MIP 11 case?

dnil:
> Tested on Hasta server by uploading a recent case [lovinggoat](https://scout-stage.scilifelab.se/cust000/NA24143-200ng/variants?variant_type=clinical).
> 
> The local observations date doesn't change from the same case loaded on prod server. Isn't this case a MIP 11 case?
> 
> Mmm I can't find the "Software" and "loqusdb" strings in the SNVs header..

lovinggoat now on stage:

Screenshot 2022-06-23 at 12 20 10

The locusdb Software tag is indeed missing, ping @jemten.

dnil:
Oki, merging with what we have. The Software loqus tag parsing works on local and test.

PR #3495, opened 2022-07-04
Remove verified variants stats from dashboard and add them to institute's dedicated page

### LINKED ISSUE #3489: Confusing "Validation ordered Cases" field in dashboard
For instance in demo case if I pin a variant, then:
- mark it as validated (true positive)
-  mark it as validated (false positive)
- mark it as Not validated

I get this, where it says that I have requested validation for 3 cases!

image

northwestwitch:
- Remove verified vars stats from dashboard and add it to a dedicated page that can be accessed from the institute sidebar
- fix #3489

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test -> Install the branch on stage and compare the following things against current prod server**:

1. At the moment the stats on the validated variants (sanger) can be seen and downloaded from the dashboard (third tab, named "Variant statistics". The page doesn't work so well and requires a query (that can take forever) before showing stats. It look like this:

image

1. I'm removing entirely this tab (that doesn't show the variants unless you download them) and I'm creating a new dedicated page that can be reached from instituteS page, sidebar, "Verified"

image

1. This page looks like the "Causatives" page, that is also accessible from that sidebar

1. This way all variants can be displayed and eventually downloaded in the same way as the the causatives, using the export buttons on top.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Tested on stage and looks like it's not crashing. Ready for review I'd say!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -75,6 +75,50 @@ def get_timeline_data(limit):
...
+
+    Returns:
+        verified(list): list of variant objects (True positive or False Positive)
+    """
+    verified = []
+    for variant_obj in store.verified(institute_id=institute_id):
+        if variant_obj["category"] in ["snv", "cancer"]:
+            update_representative_gene(
+                variant_obj, variant_obj.get("genes", [])
+            )  # required to display cDNA and protein change
```
dnil, line 92: 👍

scout/server/blueprints/dashboard/controllers.py
```diff
@@ -144,98 +144,8 @@ def get_dashboard_info(adapter, data={}, institute_id=None, slice_query=None):
         },
     ]
 
-    # Data from "Variant statistics tab" is not filtered by slice_query and numbers will
```
dnil, line 147: 💯 It was always a little strange that it didn't conform to the overall query.

scout/server/blueprints/institutes/controllers.py
```diff
@@ -75,6 +75,50 @@ def get_timeline_data(limit):
...
+    true_pos = 0
+    false_pos = 0
+    for var in verified_vars:
+        if var.get("validation") == "True positive":
+            true_pos += 1
+        else:
+            false_pos += 1
+
+    n_validations_ordered = len(list(store.validations_ordered(institute_id)))
+    return true_pos, false_pos, n_validations_ordered - (true_pos + false_pos)
```
dnil, line 119: 👍

dnil (changes requested):
This has a few issues on testing. When comparing, the causatives view (now) has the same, perhaps after some library update. I would suggest we solve them both here, but I'm also very open to merging this and making an issue to fix both.

northwestwitch:
> This is really nice! Unfortunately I noticed a few things that are broken both here and in the Causatives export when opening the xslx in Excel. Presumably they were ok once, but have broken over successive updates.
> 
> All variants except where many ref nucleotides changed have an overflow `N">` partial tag in hgvs c: Screenshot 2022-08-15 at 14 11 32 Screenshot 2022-08-15 at 14 12 24
> 
> SVs get a similar overflow with the variant type (`_clinical">`) in the position column. Screenshot 2022-08-15 at 14 11 40
> 
> Screenshot 2022-08-15 at 14 11 12
> 
> The sticky-top/sticky-offset is a little off, as seen when the main menu drops. We could just disable it for this kind of page for now.
> 
> Screenshot 2022-08-15 at 14 06 36
> 
> Screenshot 2022-08-15 at 14 06 22

Thanks for the review, I'll try to fix in this PR!

northwestwitch:
I've solved the second problem (dropdown menu hidden by datatables header) now. I think the export problem is due to the presence of a tooltip element in the cells that cause problems. I'll try to find a solution!

northwestwitch:
@dnil both errors should be fixed now. I've deployed the latest on hasta stage to test!

dnil:
> both errors should be fixed now. I've deployed the latest on hasta stage to test!

@northwestwitch Much better now! Still a few left though (on STRs and SVs in particular): 

Screenshot 2022-08-16 at 10 46 58

Screenshot 2022-08-15 at 14 12 24

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -174,3 +174,124 @@ <h5 class="header">General Institute Settings</h5>
...
+            </td>
+            <td>
+              <a href="{{ url_for('variant.variant',
+                  institute_id=institute._id,
+                  case_name=case.display_name,
+                  variant_id=variant._id) }}" data-bs-toggle='tooltip' data-bs-container='body' title="{{ variant.display_name }}">{{ variant.display_name|truncate(15, True) }}
+              </a>
+            </td>
+            <td>
+              <span class="text-muted" style="display: inline-block; margin-left: 10px;" data-bs-toggle='tooltip' data-bs-container='body' title="{{ variant.reference|truncate(30, True) }} → {{ variant.alternative|truncate(30, True) }}">
```
dnil, line 215: There's a mix of quoting styles here, " and '. Might that be part of what is confusing the excel export script?

northwestwitch: I think it might be the ">" character, check this out: https://datatables.net/forums/discussion/30014/problem-with-jquery-datatables-buttons-excel-export-using-special-character-like-or

I've spent many hours trying the suggested solutions and some more stuff found on google, but didn't work. I think the easiest would be removing the tooltips for now, so it will work. We can investigate further later if we thing tooltips are necessary there.

northwestwitch:
I've simplified the dashboard page by removing the two distinct panels and put all together, now that there is space:

image

See if you like it on stage!

dnil:
Looks great, thanks for all the fixes! I compared the exported xslx to the old ones. Most is there, but we do miss AD for ref and alt, and info on the callers that found the variant. We also miss "functional annotations" and CADD score, but I really doubt anyone would care about them in this context: that is more for the causatives view.

northwestwitch:
> Looks great, thanks for all the fixes! I compared the exported xslx to the old ones. Most is there, but we do miss AD for ref and alt, and info on the callers that found the variant. We also miss "functional annotations" and CADD score, but I really doubt anyone would care about them in this context: that is more for the causatives view.

Yeah basically you see the same info available for the causatives: my idea was to standardize the views. Then if something is missing we can add to the them all altogether!

northwestwitch:
> we do miss AD for ref and alt

The info is available in a tooltip when you hover on the "Zygosity". Since it is not exported I'll try to render it better

northwestwitch:
I've added also AD info (exportable), functional annotation and cadd score now. I've left causatives page as it is, we can fix it later when causatives share the code with validated

dnil:
That leaves only caller then, right?

northwestwitch:
> That leaves only caller then, right?

That's a lot of info:

image

Perhaps include only callers with a value?

dnil:
> > That leaves only caller then, right?
> 
> That's a lot of info:
> 
> image
> 
> Perhaps include only callers with a value?

Yes, or like one col with the ones that are actually there! For most SNVs it would only be like "deepvariant" nowadays. SVs and cancer are a little different still.

northwestwitch:
Added also callers now. I had to make it scrollable horizontally to display all the info, otherwise it doesn't fit the page. Deployed the latest on stage

northwestwitch:
> Added also callers now. I had to make it scrollable horizontally to display all the info, otherwise it doesn't fit the page. Deployed the latest on stage

Oh no, large SV now are ugly:

image

I'll fix

northwestwitch:
Ok, I think I'm done with this one now. Feel free to check it again on stage (it's deployed) when you have time @dnil

tests/server/blueprints/institutes/test_institute_views.py
```diff
@@ -686,6 +686,17 @@ def test_cases_similar_phenotype_query(app, case_obj, institute_obj, test_hpo_te
         assert case_obj["display_name"] in str(resp.data)
 
 
+def test_verified(app, institute_obj):
+    """Test the endpoint that returns the list of verified variants for an institute"""
+    # GIVEN an initialized app
+    with app.test_client() as client:
```
northwestwitch, line 692: I'm not really satisfied with this test, tomorrow I could improve it to include a case where you mark a variant as validated and then you search it in the validated page

dnil:
I can't tell, not visible on dark mode! 😉 

Screenshot 2022-08-16 at 16 17 22

northwestwitch:
> can't tell, not visible on dark mode! 😉

OMG I completely forgot about the dark mode :roll_eyes:

northwestwitch:
> I think this works well now! Test and a dark mode will be nice.

Fixed both test and dark mode. Those 2 cols not visible look like this now:

image

Issue #3510, opened 2022-07-28
Flask-Login 0.6.2 Will Break Tests


mikaell:
Latest version of Flask-login, release on June 25th (0.6.2) will break our test suite. 0.6.1 works.

Test
```
pytest tests/server/blueprints/alignviewers/test_alignviewers_views.py::test_igv_not_authorized
```

Will not return unauthorised `403`, as expected, but OK `200`. This will cause the test to fail
```
FAILED tests/server/blueprints/alignviewers/test_alignviewers_views.py::test_igv_not_authorized - assert 200 == 403
```

Suggested different approaches to solve this. 
* Freeze flask-login, only
* Freeze all flask associated packages
* Find bug or workaround

mikaell:
The root cause seems to be that session globals have moved from `flask._request_ctx_stack` to `flask.g` (of reasons for me unclear). 

Altering below functions will make the tests pass i.e. return unauthorized

```
flask_login/login_manager.py 

    def _update_request_context_with_user(self, user=None):
        """Store the given user as ctx.user."""
        # 0.5.0 Works
        ctx = _request_ctx_stack.top
        ctx.user = self.anonymous_user() if user is None else user

        # 0.6.2:DOES NOT WORK
        # if user is None:
        #     user = self.anonymous_user()

        # g._login_user = user
```

and
```
flask_login/utils.py

# 0.6.2: DOES NOT WORK
# def _get_user():
#     print('***GETUSER***')
#     if has_request_context():
#         if "_login_user" not in g:
#             current_app.login_manager._load_user()

#         return g._login_user
#     print('- - -')
#     return None

# 0.5.0 WORKS
def _get_user():
    print('***GETUSER 0.5.0***')
    if has_request_context() and not hasattr(_request_ctx_stack.top, "user"):
        current_app.login_manager._load_user()

    return getattr(_request_ctx_stack.top, "user", None)



```

mikaell:
Breaks here. So, do we want to move to Flask 2.2?
```
commit 359fb004f153635952a7d32ded1bbd6d74ec2561
Author: David Lord <davidism@gmail.com>
Date:   Mon Jul 25 07:27:59 2022 -0700

    store user on g instead of app_ctx_stack.top

    Fixes compatibility with Flask 2.2
```

northwestwitch:
[...]

PR #3514, opened 2022-08-08
Matching causatives found in research variants from other cases

### LINKED ISSUE #3512: Matching causatives missing
We have a case where Matching Causatives from other cases, is not visible on the first case page . But we found a variant that has been classified as causative in a earlier case, which is also evident from the variant page where the text "Matching causatives from other cases: xxx" is seen at the top of the page. 

Why is it missing on the first case page? Bug or related to the specific case/variant?  It seems to work on other cases we have (different variants). 
/Sofie

northwestwitch:
Fix #3512 

Current code performs a strict matching of variants _ids (created as md5 keys using chr, start, ref, alt, **clinical or research**), so when a case has clinical variants there won't be any matches for those variants in another case that has research variants marked as causative. 

I've changed the code to look for causatives also among research variants by using the 2 possible display names of a variant:
- chr_start_ref_alt_clinical
- chr_start_ref_alt_research

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test, locally**:
1. Load the the demo case 643595: `scout --demo load case scout/demo/643595.config.yaml`
2. Load research variants for the demo case 643594: `scout --demo load research -c internal_id -i cust000 --force`
3. Go to case 643594, **research variants**, and mark as causative the usual POT1 var: http://localhost:5000/cust000/643594/80e1bdac78e83339117e4929f116a147
4. On main branch, if you go to the corresponding var of the close case 643595, it won't show any matching causatives
5. Switch to this branch and check that now the matching causative (research var on case 643594) is shown

**Expected outcome**:

**Review:**
- [x] code approved by DN 


northwestwitch:
This PR is blocked by #3510. Let's see what happens after fixing this issue

scout/adapter/mongo/variant.py
```diff
@@ -524,17 +545,31 @@ def check_causatives(self, case_obj=None, institute_obj=None, limit_genes=None):
...
+
+            # Collect other causative variant's coords and change (exclude if it's clinical or research)
+            other_var_displ_name = var_event.get("subject").split("_")[
+                0:4
+            ]  # example: [ "17", "7577559", "G" "A"]
 
-            if other_causative_id in other_case.get("partial_causatives", {}).keys():
-                positional_variant_ids.add(var_event["variant_id"])
+            other_display_names.add("_".join(other_var_displ_name + ["clinical"]))
+            other_display_names.add("_".join(other_var_displ_name + ["research"]))
```
dnil, line 565: Wouldn't it be faster and easier to md5sum these into variant_ids and uses as positional_variant_ids?

(I'm still annoyed by the docs btw; we have sooo many places where it says variant_id is just positional id, nothing about clinical/research type. Still, this version, with type in addition to the simple id is correct according to tests.)

dnil (changes requested):
This works, and solves the issue, but I'm afraid it would get slow with time, querying against unindexed fields on the variant collection. I suggest re-hashing the display names you generate in this PR to variant_ids to search indexed instead.

scout/adapter/mongo/variant.py
northwestwitch: hehe that's how I did the PR in the beginning, but for some technical problem I've changed it. I'll check again!

northwestwitch: It wasn't that hard after all. I've changed the code to use md5 keys instead. Check it out when you have time!

scout/models/variant/variant.py
```diff
@@ -3,12 +3,9 @@
...
     # document_id is a md5 string created by institute_genelist_caseid_variantid:
     _id=str,  # required, same as document_id
-    document_id=str,  # required
-    # variant_id is a md5 string created by chrom_pos_ref_alt (simple_id)
-    variant_id=str,  # required
-    # display name is variant_id (no md5)
-    display_name=str,  # required
-    # chrom_pos_ref_alt
+    document_id=str,  # required. Same as _id
+    variant_id=str,  # required. A md5 string created by chrom_pos_ref_alt_[clinical/research]
```
dnil, line 7: 👍

scout/models/variant/variant.py
```diff
@@ -3,12 +3,9 @@
...
+    display_name=str,  # required. no md5. chrom_pos_ref_alt_[clinical/research]
```
dnil, line 8: Maybe add a comment for simple_id as well now?

scout/models/variant/variant.py
dnil: (you could call that `[c/r]` variant_type as well; defined below)

dnil (approved):
Thank you! This looks easier to maintain.

scout/models/variant/variant.py
northwestwitch: Done!

scout/models/variant/variant.py
northwestwitch: Done!

PR #3518, opened 2022-08-10
Cases with Sanger validation ordered search

### LINKED ISSUE #3505: Search by "verification ordered" on first page
Be able to search for cases where sanger verification has been ordered (independent who ordered the verification). Currently only Sanger validations to evaluate that have been ordered by me as a user are seen on the first page. 
Maybe just add this as a search criteria in the case page list?

northwestwitch:
Fix #3505 => Add a checkbox in case search form to retrieve all cases with Sanger validation ordered (**by any user**) but not yet assessed

Form changed from this:

image

To this:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Install on cg-vm1 and run tests by searching cases with Sanger vars for one or more institutes

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
This is tested on stage and ready for review now

scout/adapter/mongo/case.py
```diff
@@ -391,16 +368,89 @@ def cases(
...
+        }
+        group = {
+            "$group": {
+                "_id": {
+                    "case": "$case_id",
+                },
+            }
+        }
+        pipeline = [match_query, group]
+        for res in self.variant_collection.aggregate(pipeline):
```
dnil, line 441: Ho, didn't you also suggest removing the index for sanger_ordered on the variant collection? This is bound to get very slow if run without index. Wouldn't it be better to ask the the event collection first, and confirm the current status with the variants that have ever had Sanger order events?

dnil: Compare the current adapter `VariantEventHandler sanger_ordered()` which does basically this?

dnil (changes requested):
Seems like it should not query the variant collection directly, but instead use the events?

Also, I have trouble when testing it currently - seems it still returns all cases, not really filtering for the ones with sanger ordered?!

scout/adapter/mongo/case.py
northwestwitch: I agree with you that having also VariantEventHandler sanger_ordered() is confusing. But they are different in this:

- this function collects the sanger ordered but not validated - the other one is collecting the sanger ordered only (could be validated or not)
- this function is collecting variant objects (from variant collection - I've reintroduced the index in this PR, btw :) ) and not variant IDs
- this function is not selecting the sanger ordered vars for a user - the other does it.

I'll try to modify the code to make it more clear (and eventually remove redundant code!)

dnil: Hehe, I didn't see that, sorry! I may not have reloaded this morning. Still, it is a fairly big index if it is only for this?

Right, there are certainly things that are specific here; just saying going via the events (small, indexed) would be faster than unindexed variants. I guesstimate it would be faster doing that plus checking status than searching indexed variants. But, the index does exist so if you must.. 🤷

northwestwitch: Fixed by re-using the existing function!

northwestwitch:
I'm ready for another review for this PR

dnil (approved):
Nice; well done! 👍

PR #3539, opened 2022-08-22
Fix broken filter cases by pinned gene

### LINKED ISSUE #3538: Filter cases by pinned gene doesn't work
While filtering by causative gene does. Weird!

northwestwitch:
Fix #3538.

Locally or on stage, compare the behavior when you try to filter cases by pinned genes.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Pin a variant and go to the institute cases. Filter by pinned gene using the gene symbol of the pinned variant
2. On main branch the search shouldn't return any case
3. On this branch the search should return one or more cases

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -112,7 +112,6 @@ class CaseHandler(object):
             query_field(str) "pinned" or "causative"
             query_term(str) example:"POT1"
         """
-        LOG.warning(f"IN GENES OF INTEREST--->{query_field}:{query_term}")
         hgnc_id = self.hgnc_id(hgnc_symbol=query_term)
         if hgnc_id is None:
             LOG.debug(f"No gene with the HGNC symbol {query_term} found.")
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -112,21 +112,21 @@ class CaseHandler(object):
             query_field(str) "pinned" or "causative"
             query_term(str) example:"POT1"
         """
+        LOG.warning(f"IN GENES OF INTEREST--->{query_field}:{query_term}")
         hgnc_id = self.hgnc_id(hgnc_symbol=query_term)
         if hgnc_id is None:
             LOG.debug(f"No gene with the HGNC symbol {query_term} found.")
             query["_id"] = {"$in": []}  # No result should be returned by query
+            return
 
         unwind = "$causatives"
         lookup_local = "causatives"
-        lookups_as = "causative_variant"
-        match = "causative_variant.hgnc_ids"
-
         if query_field == "pinned":
             unwind = "$suspects"
             lookup_local = "suspects"
-            lookups_as = "suspect_variant"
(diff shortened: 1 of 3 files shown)


scout/adapter/mongo/case.py
```diff
@@ -112,21 +112,21 @@ def _set_genes_of_interest_query(self, query, query_field, query_term):
...
         unwind = "$causatives"
         lookup_local = "causatives"
-        lookups_as = "causative_variant"
-        match = "causative_variant.hgnc_ids"
-
         if query_field == "pinned":
             unwind = "$suspects"
             lookup_local = "suspects"
-            lookups_as = "suspect_variant"
-            match = "suspect_variant.hgnc_ids.hgnc_ids"
```
northwestwitch, line 129: This typo is the cause of the bug; `hgnc_ids.hgnc_ids ` instead of just `hgnc_ids`

scout/server/blueprints/institutes/templates/overview/cases.html
```diff
@@ -29,7 +29,7 @@
       </div>
       <div class="col-4">
           <label class="control-label" for="search_term">Search cases</label>
-          <input class="form-control" id="search_term" name="search_term" placeholder="example:18201" type="text" pattern="[^<>?!=\\/]*" title="Characters ^<>?!=\/ are not allowed">
+          <input class="form-control" id="search_term" name="search_term" value="{{form.search_term.data}}" placeholder="example:18201" type="text" pattern="[^<>?!=\\/]*" title="Characters ^<>?!=\/ are not allowed">
```
northwestwitch, line 32: This is to display what was searched after the form is submitted and the results are returned

scout/adapter/mongo/case.py
dnil: Did `suspect_variant` collide somewhere else or why `lookedup_variant`? 😊

northwestwitch: It's just to give it a common name that can be used by both suspects or pinned. We can can rename it something else if you don't like it.

dnil: Hehe, no, by all means! It's just very deep in the loop layers so to speak. If I read it right, it's like a variable for what the mongodb query temp name will be. We won't really do anything more with it, more than snarf all the case_ids returned, so it might just as well be a literal constant in the query? 😊

scout/server/blueprints/institutes/templates/overview/cases.html
dnil: ✅

dnil (approved):

scout/adapter/mongo/case.py
northwestwitch: Yes, you can name it whatever you want, it's just a temp variable used when running the query, I'll name it as a constant!

PR #3549, opened 2022-08-25
Toggle SVs' genes in causatives and validated and fix causatives export button

### LINKED ISSUE #3546: Causative and validated variants pages are not easy to read when they contain huge SVs
Example:

image

### LINKED ISSUE #3547: Button for exporting causative variants is missing
It should be like this (as in validated variants page)

image

But it looks like this:

image

northwestwitch:
Fix #3547 and fix #3546 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -17,7 +17,7 @@ from scout.constants import (
     PHENOTYPE_GROUPS,
 )
 from scout.parse.clinvar import clinvar_submission_header, clinvar_submission_lines
-from scout.server.blueprints.variant.utils import update_representative_gene
+from scout.server.blueprints.variant.utils import predictions, update_representative_gene
 from scout.server.extensions import beacon, store
 from scout.server.utils import institute_and_case, user_institutes
 from scout.utils.md5 import generate_md5_key
@@ -86,11 +86,12 @@ def verified_vars(institute_id):
     """
     verified = []
     for variant_obj in store.verified(institute_id=institute_id):
+        variant_genes = variant_obj.get("genes", [])
         if variant_obj["category"] in ["snv", "cancer"]:
             update_representative_gene(
-                variant_obj, variant_obj.get("genes", [])
+                variant_obj, variant_genes
             )  # required to display cDNA and protein change
-
+        variant_obj.update(predictions(variant_genes))
         verified.append(variant_obj)
 
     return verified
@@ -142,10 +143,12 @@ def causatives(institute_obj, request):
     causatives = []
 
     for variant_obj in store.check_causatives(institute_obj=institute_obj, limit_genes=hgnc_id):
+        variant_genes = variant_obj.get("genes", [])
         if variant_obj["category"] in ["snv", "cancer"]:
             update_representative_gene(
-                variant_obj, variant_obj.get("genes", [])
+                variant_obj, variant_genes
             )  # required to display cDNA and protein change
+        variant_obj.update(predictions(variant_genes))
         case_obj = store.case(variant_obj["case_id"])
         if not case_obj:
             continue
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -393,25 +393,20 @@ def cases(store, request, institute_id):
         )
     data["name_query"] = name_query
     limit = int(request.args.get("search_limit")) if request.args.get("search_limit") else 100
-    skip_assigned = request.args.get("skip_assigned")
     data["form"] = populate_case_filter_form(request.args)
-    data["skip_assigned"] = skip_assigned
-    is_research = request.args.get("is_research")
-    data["is_research"] = is_research
+
     prioritized_cases = store.prioritized_cases(institute_id=institute_id)
     all_cases = store.cases(
         collaborator=institute_id,
         name_query=name_query,
-        skip_assigned=skip_assigned,
-        is_research=is_research,
+        skip_assigned=request.args.get("skip_assigned"),
+        is_research=request.args.get("is_research"),
+        verification_pending=request.args.get("validation_ordered"),
     )
     all_cases = _sort_cases(data, request, all_cases)
 
     data["nr_cases"] = store.nr_cases(institute_id=institute_id)
-
-    sanger_unevaluated = get_sanger_unevaluated(store, institute_id, current_user.email)
-    if len(sanger_unevaluated) > 0:
-        data["sanger_unevaluated"] = sanger_unevaluated
+    data["sanger_unevaluated"] = get_sanger_unevaluated(store, institute_id, current_user.email)
 
     case_groups = {status: [] for status in CASE_STATUSES}
     nr_cases = 0
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -9,13 +9,7 @@ from flask import current_app, flash, url_for
 from flask_login import current_user
 from pymongo import ASCENDING, DESCENDING
 
-from scout.constants import (
-    ACMG_COMPLETE_MAP,
-    ACMG_MAP,
-    CASE_SEARCH_TERMS,
-    CASE_STATUSES,
-    PHENOTYPE_GROUPS,
-)
+from scout.constants import CASE_SEARCH_TERMS, CASE_STATUSES, PHENOTYPE_GROUPS
 from scout.parse.clinvar import clinvar_submission_header, clinvar_submission_lines
 from scout.server.blueprints.variant.utils import predictions, update_representative_gene
 from scout.server.extensions import beacon, store
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -337,6 +337,7 @@ def update_institute_settings(store, institute_obj, form):
         cohorts=cohorts,
         loqusdb_ids=form.getlist("loqusdb_id"),
         alamut_key=form.get("alamut_key"),
+        check_show_all_vars=form.get("check_show_all_vars"),
     )
     return updated_institute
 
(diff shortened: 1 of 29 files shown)


northwestwitch:
The fix is working to toggle gene names, but the tables still look ugly:

image

That long list of repeated functional annotations is not really helpful. What about compressing it to a set of unique annotations instead? That would fix the looks of the tables!

dnil:
> The fix is working to toggle gene names, but the tables still look ugly:
> 
> image
> 
> That long list of repeated functional annotations is not really helpful. What about compressing it to a set of unique annotations instead? That would fix the looks of the tables!

At least fold it like the gene names? If there is only one kind, there is also absolutely no need for multiple rows. The downside with compressing if there are many consequences is you wouldn't know what belongs to what gene - as it is now I assume they are at least on the same row.

northwestwitch:
> At least fold it like the gene names? If there is only one kind, there is also absolutely no need for multiple rows. The downside with compressing if there are many consequences is you wouldn't know what belongs to what gene - as it is now I assume they are at least on the same row.

They were on the same row, but this PR fixes the code of the first column (genes) and sorts them alphabetically, so the functional anno doesn't follow the genes any more. I suggest let's just collect unique functional anno descriptions, to give a general idea of the consequences, and then if the user is interested clicks on the variant and checks it out in detail

northwestwitch:
I see that variants table on SVs page is built in a smarter way:

image

We should try to do the same here

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -21,6 +21,25 @@
   </div>
 {% endmacro %}
 
+{% macro svs_funct_anno_cell(variant) %}
```
northwestwitch, line 24: OK, Much better now!

image

northwestwitch:
Ok, I think this is ready for a first review now!

northwestwitch:
Mmm I still wonder way there is not a functional annotation available for each gene present in the large SVs:

image

dnil:
> Mmm I still wonder why there is not a functional annotation available for each gene present in the large SVs:
> 
> image

Interesting! Clinical list vs all genes at some point would be my first guess?

dnil:
And if so, which it seems, the clinical list ones are very likely the interesting ones anyway, even if the total gene count can be a nice indicator of size.

northwestwitch:
> And if so, which it seems, the clinical list ones are very likely the interesting ones anyway, even if the total gene count can be a nice indicator of size.

I think it could be because the predictions (and among those functional annotations) are based on the limited number of genes present in the SVs (30ish,  field  `variant["genes"]` ) while the list of genes present in the gene names cell is complete (field `variant["hgnc_symbols"]` ) because it's based on. 

I don't know how to fix it without slowing down the loading of the whole page (because then you need to pull every gene and for each of them check the prediction)

northwestwitch:
> I don't know how to fix it without slowing down the loading of the whole page (because then you need to pull every gene and for each of them check the prediction)

Actually even pulling the general gene info is useless, because these annotations are saved when loading the variants, and these variants have only 30 something complete genes saved..

scout/server/blueprints/institutes/controllers.py
```diff
@@ -80,11 +80,12 @@ def verified_vars(institute_id):
...
     verified = []
     for variant_obj in store.verified(institute_id=institute_id):
+        variant_genes = variant_obj.get("genes", [])
         if variant_obj["category"] in ["snv", "cancer"]:
             update_representative_gene(
-                variant_obj, variant_obj.get("genes", [])
+                variant_obj, variant_genes
             )  # required to display cDNA and protein change
-
+        variant_obj.update(predictions(variant_genes))
```
northwestwitch, line 88: This is to have SVs functional anno available

scout/server/blueprints/institutes/templates/overview/causatives.html
```diff
@@ -38,21 +38,5 @@
 {% block scripts %}
   {{ super() }}
   {{ db_table_external_scripts() }}
-  <script type="text/javascript">
```
northwestwitch, line 41: Move to shared macro

scout/server/blueprints/institutes/templates/overview/verified.html
```diff
@@ -86,28 +86,8 @@
 </div>
 {% endmacro %}
 
-
 {% block scripts %}
   {{ super() }}
   {{ db_table_external_scripts() }}
-  <script type="text/javascript">
```
northwestwitch, line 93: Moved to shared macro

northwestwitch:
ready for review again! Took me a while to fix that duplicated code issue

scout/server/blueprints/institutes/static/variants_list_scripts.js
```diff
@@ -0,0 +1,12 @@
+$(document).ready(function() {
+  $('#variants_table').DataTable( {
+      paging: false,
+      dom: 'Bfrtip',
```
dnil, line 4: Cool, now I've learnt something!

scout/server/blueprints/institutes/templates/overview/causatives.html
```diff
@@ -38,21 +38,11 @@
...
+  <script>
     var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'))
     var tooltipList = tooltipTriggerList.map(function (tooltipTriggerEl) {
       return new bootstrap.Tooltip(tooltipTriggerEl)
     })
-
-    $(document).ready(function() {
-      $('#causatives_table').DataTable( {
-          paging: false,
-          dom: 'Bfrtip',
```
dnil, line 50: ..that I apparently maybe should have known already! 😆

scout/server/blueprints/institutes/templates/overview/causatives.html
```diff
@@ -38,21 +38,11 @@
...
-          buttons: [
-              'copyHtml5',
-              'excelHtml5',
-          ]
-      } );
-    } );
   </script>
+  <script src="{{ url_for('overview.static', filename='variants_list_scripts.js') }}"></script>
```
dnil, line 47: Again, just for learning, why a static script js file vs a jinja macro js script? Same difference?

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -190,7 +192,7 @@ <h3 class="header">General Institute Settings</h3>
 {% macro variant_list_content(institute, variants, acmg_map, callers) %}
 <div class="card mt-3">
   <div class="card-body overflow-auto">
-    <table id="verified_table" class="table display table-sm">
+    <table id="variants_table" class="table display table-sm">
```
dnil, line 195: 👍

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -258,13 +258,7 @@ <h3 class="header">General Institute Settings</h3>
...
             </td>
             <td><!-- Functional Annotation -->
-              <span class="text-muted">
-              {% for gene in variant.genes or [] %}
-                {{gene.functional_annotation}}&nbsp;
-              {% else %}
-                -
-              {% endfor %}
-              </span>
+              {{ svs_funct_anno_cell(variant) }}
```
dnil, line 261: Same here with the naming.

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -215,9 +217,7 @@ <h3 class="header">General Institute Settings</h3>
         {% for variant in variants %}
           {% set case = variant["case_obj"] %}
           <tr>
-            <td><!-- Gene -->
-              <span class="text-body">{{ variant.hgnc_symbols|join(', ') if variant.hgnc_symbols else '--'}}</span>
-            </td>
+            <td>{{ sv_genes_cell(variant) }}</td>
```
dnil, line 220: This is nit-picking, but the same is actually used for all types of variants right now, so the name could use a generalisation to go beyond its history! 😄

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -21,6 +21,25 @@
...
+  <div class="d-flex justify-content-between align-items-center">
+    <div>
+      {% if variant.functional_annotations|length >= 5 %}
+        <a class="mr-3" data-bs-toggle="collapse" href="#_functanno_{{variant._id}}" role="button" aria-expanded="false" aria-controls="_{{variant._id}}">{% if variant.missing_data %}[first 30 genes]{% else %}[...]{% endif %}</a>
```
dnil, line 28: Yes, we'll just have to live with that; otw this becomes a rather large project. It is now understood, and says so clearly much like on the variant page.

dnil (approved):

scout/server/blueprints/institutes/templates/overview/causatives.html
northwestwitch: It made sense for 2 reasons:
- It contains static code (while jinja macros ofter are used to populate content according to the params that you pass to them)
- JS code is included in static files in many other blueprints, so this would be consequent

scout/server/blueprints/institutes/templates/overview/utils.html
northwestwitch: Ah no, you are right, it makes more sense if it's general. I'll change it!

scout/server/blueprints/variants/templates/variants/components.html
northwestwitch: What we could improve is parsing the large SVs not to include the first 30something genes, but genes of interest

dnil: Yes, definitely - like use OMIM, clinical list, near-ends of balanced events... But better make a nice common routine for that separately, right?!

northwestwitch: yes!

PR #3555, opened 2022-08-30
Matching SV links

dnil:

Hiding variant details a bit longer may or not be good, but linking to the wrong kind of page is just confusing - and e.g. a random STR match may not display well as it was. Reusing macros is nice.

Screenshot 2022-08-30 at 14 13 46

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/utils.html
+++ b/scout/server/blueprints/cases/templates/cases/utils.html
@@ -572,10 +572,7 @@
     <ul class="list-group collapse" id="matchingCausatives">
       {% for variant in other_causatives %}
         <li class="list-group-item">
-          <a href="{{ url_for('variant.variant', institute_id=institute._id,
-                              case_name=case.display_name, variant_id=variant._id) }}">
-            {{ variant.hgnc_symbols|join(', ') or variant.hgnc_ids|join(', ') or variant.str_repid }}
-          </a>
+          {{ pretty_link_variant(variant, case)}}
         </li>
       {% else %}
         <li class="list-group-item">No matching causative variants.</li>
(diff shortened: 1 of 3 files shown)


dnil:
Should be ready for review - I had just missed a merge! 😊

scout/server/blueprints/cases/templates/cases/utils.html
```diff
@@ -572,10 +572,7 @@ <h6><span class="fa fa-map-pin ms-3 mt-2"></span>&nbsp;Pinned variants ({{suspec
     <ul class="list-group collapse" id="matchingCausatives">
       {% for variant in other_causatives %}
         <li class="list-group-item">
-          <a href="{{ url_for('variant.variant', institute_id=institute._id,
-                              case_name=case.display_name, variant_id=variant._id) }}">
-            {{ variant.hgnc_symbols|join(', ') or variant.hgnc_ids|join(', ') or variant.str_repid }}
-          </a>
+          {{ pretty_link_variant(variant, case)}}
```
northwestwitch, line 575: 👍🏻

northwestwitch (approved):
Looks good! 💯 
Weird that the pinned ones had the correct links and causatives and the matching managed no!

scout/server/blueprints/cases/templates/cases/utils.html
dnil: Yeah, I think we (I) might have had the idea to make them a tad more non-informative (before the hiding div) so that the user had one more moment to think about whether they really wanted to triage a variant in that gene for that patient. Some of the CLGs would insta-recognise common managed variants.

dnil: It should be ok with the hiding div I think. Thanks for the review!

PR #3572, opened 2022-09-07
Re-write ClinVar module to prepare for ClinVar API submissions

### LINKED ISSUE #3569: Move ClinVar code to a dedicated blueprint and refactor to use FlaskForm instead of template-based code
And check also that the CSV outfiles are working fine for submitting

### LINKED ISSUE #3602: Include all condition types in ClinVar submission, even if Scout supports only OMIM and HPO
All condition types accepted by the ClinVar API are:  
- OMIM
- MedGen
- Orphanet
- MeSH
- HP
- MONDO

This can be done in the form by introducing a select with the list above and a free text field that can be filled in by the user on the fly when creating the submission

northwestwitch:
Fix #3569 and fix #3602

image

- Doesn't change the functionality, but makes it **more user-friendly**
- Allow for submissions with phenotype type from the following DBs: [OMIM, MedGen, Orphanet, MeSH, HP, MONDO]
- The only change is that it is possible to add one variant at the time to ClinVar submissions
- Simplifies the fields exported in CaseData .CSV files (most of the current stuff is not used anyway)
- Moves all ClinVar export code to a dedicated "clinvar" blueprint. Before is was half under "variant" and half under "institutes"
- Is Using Flask-form to define the ClinVar input fields to be saved for a Variant submission, instead of ugly HTML code
- Created a model under scout.models that describes exactly what the expected formats are when you create and save a submission in the database
- Increases the test coverage

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Install on cg-vm1 and try to save variants to ClinVar submissions and edit/delete ClinVar submissions for an institute
2. Try to download the Variants and CaseData .csv files from the Institute's submissions page and make sure download works

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Tested on stage and fixed what crashed. I think it looks good enough for a review now!

northwestwitch:
Ok, I've made it a bit more user friendly and I've also included the possibility to specify phenotypes other than HPO and OMIM. if tests pass this is ready for review.

scout/models/clinvar.py
```diff
@@ -0,0 +1,73 @@
+from datetime import date, datetime
+
+from bson.objectid import ObjectId
+
+"""Model of the document that gets saved/updated in the clinvar_submission collection
```
dnil, line 5: 🥇

scout/server/blueprints/clinvar/controllers.py
```diff
@@ -0,0 +1,389 @@
...
+    """Sets up values for a SV variant form
+
+    Args:
+        variant_obj(dict) scout.models.Variant
+        case_obj(dict) scout.models.Case
+
+    Returns:
+        var_form(scout.server.blueprints.clinvar.form.SVariantForm)
+    """
+    var_form = SVariantForm()
```
dnil, line 89: 😺 Hihi, like CDisk!

dnil (approved):
Super nice! The flow bar on the page is really cool, the page is simplified and distilled to its essentials, and the use of e.g. flask forms instead of ad hoc html so much easier to maintain! This works on local tests, and is anyway outdated on main (submissions require extensive manual edit): I approve. 

Some things that met my eye when testing:
- [x] dark mode colors are hard to read (light subheadings on light background, dark example strings on dark backgrounds on non-filled form inputs).
On variant details transcripts and hgvs: 
    - [x] the checkbox and label are far from each other, and checkbox is subtle and hard to see on the light background
    - [x] also, no versions for transcripts? Maybe that is only on my test case, but otherwise this will likely fail on submission. If not quite possible automatically as it is, perhaps add a manual edit option or a submit-to-HGVS-checker like mutalyzer.nl to get it right?
- [ ] clinical significance could be pre-populated from ACMG and/or Variant tag - or they could be shown
- [ ] associated conditions, condition type could be prepopulated based on what terms are available (HPO if only HP, OMIM if there is an OMIM condition...)

ClinVar submissions page:
- [ ] not from this PR, but the Submission heading is hard to read in light mode (black on blue), very nice in dark mode though.😊 I have the feeling it was ok before the bs5 change.
- [ ] the spinner is stuck after downloading of the CSV files - that is also not new, so feel even more free to ignore 😁 

Final CSV:
- [ ] you will have to choose to use either reference, chrom, start, stop, ref, alt OR full HGVS (including the transcript, eg NM_015450.3:c.903G>T), otherwise it will generate an error at least on manual submission via the clinvar portal.

I have not tested what happens with old submission objects - have you? Most should be the same I guess, some of the terms changing perhaps? It is probably not super important, just if someone happens to have a larger submission ongoing when we merge...

northwestwitch:
> * also, no versions for transcripts? Maybe that is only on my test case, but otherwise this will likely fail on submission. If not quite possible automatically as it is, perhaps add a manual edit option or a submit-to-HGVS-checker like mutalyzer.nl to get it right?

Right! This appears to be broken also on main branch, but I remember that I fixed it once. Will fix again!

northwestwitch:
Thanks for the nice and detailed review @dnil. I've fixed according all your suggestions (transformed your comment in a to-do list) except for setting up a check of HGVS on mutalyzer.nl. 
I feel like this PR is already huge as it is and perhaps we can have this functionality in a separate one?

>I have not tested what happens with old submission objects - have you? Most should be the same I guess, some of the terms changing perhaps? It is probably not super important, just if someone happens to have a larger submission ongoing when we merge...

I did! I tested using this branch to adding variants and case data to old submission objects. "Mixed" submissions downloads work fine since I haven't removed any item from the old download headers: [variant header](https://github.com/Clinical-Genomics/scout/blob/80af0c5a6b046b82370292e30c5607ab7a39149b/scout/constants/clinvar.py#L2) and [casedata header](https://github.com/Clinical-Genomics/scout/blob/80af0c5a6b046b82370292e30c5607ab7a39149b/scout/constants/clinvar.py#L41)
I did it this was on purpose so the old stuff can still be exported!

northwestwitch:
I think I'm going to merge this PR since I've fixed almost all the things you suggested during the last review @dnil, and you approved also last time. We're going to test it again before the new release anyway. So I can start working on integrating the ClinVar API

PR #3576, opened 2022-09-08
Fix region|function freq. overlapping in SV variants cells

### LINKED ISSUE #3575: Functional annotations overlapping frequencies
When checking the new version, this came up;
Screenshot 2022-09-07 at 14 56 39

We should wrap/fix column width or give prio to freq worst case: it is the more important info.

northwestwitch:
Fix #3575 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - `Scout-REViewer-Service` integration - show `REViewer` picture if available
 ### Changed
 - Better visualization of regional annotation for long lists of genes in large SVs in Variants tables
+- Order of cells in variants tables
 ### Fixed
 - HPO filter button on SV variantS page
 - Spacing between region|function cells in SVs lists
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,12 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ## [unreleased]
 ### Added
 - Button to go directly to HPO SV filter variantS page from case
+- `Scout-REViewer-Service` integration - show `REViewer` picture if available
+### Changed
+- Better visualization of regional annotation for long lists of genes in large SVs in Variants tables
 ### Fixed
 - HPO filter button on SV variantS page
+- Spacing between region|function cells in SVs lists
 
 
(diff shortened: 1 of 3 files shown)


northwestwitch:
I've deployed this branch on stage @dnil , check it out when you have time!

dnil:
Thanks, much better! Is it a little bit funny that the few-genes cells work differently (comma separated list) than the many gene cells (multiple single lines)? Nah, since the many ones are folded from the start, it works, right?

Now that I've thought it, I would still think the frequency column should go earlier in the table, like at least before the region anno. I would make it Length, Frequency, Gene, Region, Function instead, which is more how you think about the variants. Makes sense?

northwestwitch:
> Thanks, much better! Is it a little bit funny that the few-genes cells work differently (comma separated list) than the many gene cells (multiple single lines)? Nah, since the many ones are folded from the start, it works, right?
> 
> Now that I've thought it, I would still think the frequency column should go earlier in the table, like at least before the region anno. I would make it Length, Frequency, Gene, Region, Function instead, which is more how you think about the variants. Makes sense?

Makes sense. I've changed the order of columns like you suggest for SVs and cancer SVs. SNVs and STRs are different and didn't have that order to start with

northwestwitch:
> Thanks, much better! Is it a little bit funny that the few-genes cells work differently (comma separated list) than the many gene cells (multiple single lines)? Nah, since the many ones are folded from the start, it works, right?

Deployed the latest now. Yes the cells with many genes start folded.

dnil:
> Makes sense. I've changed the order of columns like you suggest for SVs and cancer SVs. SNVs and STRs are different and didn't have that order to start with

Sorry for being late - I'm the too-many-meetings person today. Looks great! Yes, SNVs have an order closer to this actually, and we don't do region/consequences for the STRs - in the cases they are truly known, they are sort of clear from the catalog/litterature before calling and annotation. Bonus points for also remembering the cancer SVs!

dnil (approved):

northwestwitch:
> Sorry for being late - I'm the too-many-meetings person today.

No worries, I figured! Thanks for the review!

PR #3585, opened 2022-09-09
Set response headers instead of appending them

mhkc:
This PR fixes a bug where Scout responded with duplicated response header when returning a BAM file causing a 502 error.

Response headers without the fix
```
Content-Type: text/html; charset=utf-8
Content-Length: 26
Content-type: application/octet-stream
Accept-Ranges: bytes
Content-Range: bytes 3285-3310/45531713695
Content-Length: 26
```

Response headers with fix
```
Content-type: application/octet-stream
Content-Length: 26
Accept-Ranges: bytes
Content-Range: bytes 3285-3310/45531713695
```

**How to test**:
1. Print headers before and after alteration, (`print(resp.headers)`)
2. Ensure that IGV can display the alignment track

**Expected outcome**:
1. The Content-Type and Content-Length header should be overwritten and not appended
3. IGV can display alignment tracks again

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - HPO filter button on SV variantS page
 - Spacing between region|function cells in SVs lists
 - Labels on gene panel Chanjo report
+- Fixed ambigious duplicated response headers when requesting a BAM file from /static
 
 ## [4.58.1]
 ### Fixed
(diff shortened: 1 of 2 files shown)


northwestwitch (approved):
Tested on our stage server with this case: https://scout-stage.scilifelab.se/cust000/SWEDAC-2019-HG-05-Trio-NovaSeq/7076347d9ff7b06e8c24104ef14a2152

Alignments work fine. They worked also before the fix for us. Perhaps some different settings allowed the requests to be successful anyway on our site?

Thanks for the fix! 💯

PR #3596, opened 2022-09-23
Rerun by redirecting to orderportal

### LINKED ISSUE #2401: Send users to the order portal when they click "Request rerun"
Re-runs can be ordered in two ways:
1) When customers are viewing a case on Scout they can click "Request rerun"
2) Via the orderportal

I suggest we replace the "Request rerun"-button on Scout with a link that send users to the order portal where they can fill in their order for a re-run. This update will save production a lot of time since re-runs ordered via the order portal starts automatically, while the ones originating from Scout has to be started manually. Additionally orders placed via order portal are more likely to be correct from start. Re-runs requested from scout are often started with incomplete information, leading to that the analysis fail, which in turn cost production a lot of time.

Before implementing this change there are a few things we need to do:
- Discuss with key customers what they think, this change will make placing an order for re-runs a bit more time consuming for them
- Come up with good argument to tell the customers when we do this change. For example we should inform the customers that this change is done so they will get their analysis faster, since request via scout leads to analyses that often fail, while re-runs ordered via the order portal are more likely to finish without problems
- Consider how to handle the re-run orders for customers that do not have access to the order portal. Are there such customers (Lund?)? Can we give them access to the order portal?

northwestwitch:
- Do not send a ticket by redirect user to orderportal (fix #2401)
- Since it's not automatic any more, allow users to edit rerun status of a case by clicking on new button

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN


docs/admin-guide/server.md
```diff
@@ -29,6 +31,15 @@ MAIL_USE_SSL = False
 # If null no badge is displayed in scout
 ACCREDITATION_BADGE = "swedac-1926-iso17025.png"
 
+# LDAP login Settings
```
northwestwitch, line 34: There were a bunch on missing params here, so I've just copy-pasted the current confg file

northwestwitch:
Ready for review now!

scout/adapter/mongo/case_events.py
```diff
@@ -388,10 +388,13 @@ def request_rerun(self, institute, case, user, link):
         Return:
             updated_case
         """
-        if case.get("rerun_requested"):
-            raise ValueError("rerun already pending")
+        rerun_status = True
```
dnil, line 391: Renaming this var could make it an easier read. The toggle part next is a bit dense on first read, especially with the function name still being request_rerun. Perhaps the function should also simply be renamed `update_rerun_status`, like the endpoint? With a new name and docstring the var name is more obvious.

scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html
```diff
@@ -304,17 +304,28 @@
...
+  <form action="{{ url_for('cases.update_rerun_status', institute_id=institute._id, case_name=case.display_name) }}">
+    <button class="btn btn-info btn-sm form-control" type="submit" name="rerun_update">Reset rerun status</button>
+  </form>
 {% else %}
   <div id='submenu3' class="collapse">
     {% if 'RERUN_MONITOR' in config and config.RERUN_MONITOR == True %}
       {{ rerun_monitor_interact(institute, case) }}
     {% endif %}
-    <button type="button" class="btn btn-warning btn-sm form-control" data-bs-toggle="modal" data-bs-target="#rerun-modal">Request rerun</button>
+    {% if "RERUN_URL" in config %}
```
dnil, line 315: 👍

dnil (approved):
I am not a big fan of dropping this before an integration with rerunning cases is in place, but this has been long requested and so far no vocal protests so I guess we go with it.

The code works of course, so approved! The `request_rerun` adapter function changed its usage a bit, so at least a docstring update is needed, but even better change the name as for the endpoint!

scout/adapter/mongo/case_events.py
northwestwitch: I agree. Let's modify the function name, since the function purpose is different now!

northwestwitch: Renamed also variable to is_rerun

PR #3601, opened 2022-09-28
FIX: Remove Animation and Decrease Image Resolution For Speed

mikaell:
This PR adds a better user experience.

Drawing Chromogaph images could be slow. This PR adds a CSS class that disables animated accordions –these could be slow/CPU consuming when redrawing large screen real estate– and adds smaller included Chronograph images (300 DPI instead of 1000) (*).

(*) Chromgraph version soon to be published on GitHub.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

Expand the Chromograph images on a case page. Compare this PR towards Main -it should be faster.

Screenshot 2022-09-28 at 10 41 45

**Expected outcome**:
The drawing of the page should be faster.

Screenshot 2022-09-28 at 10 40 03

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -339,7 +339,7 @@ <h5 class="modal-title" id="case_pick_dialog">Add other case to case group ({{ l
 <!-- Iterate through Individuals, if chromograph_files is presenet add a panel for images -->
 {% set chromograph = namespace(available=false) %}
 {% macro insert_multi_image_panel() %}
-<div class="panel-group" id="accordion">
+<div class="panel-group no-animation" id="accordion">
```
northwestwitch, line 342: Where does this no-animation class come from? Is it a custom style or a bootstrap one? If it's custom I don't find it anywhere

dnil (changes requested):
Nice! Feels faster already! 🏎️ 💨 

We should probably sort out the case origins for the files before merging - there could be test runs from MIP11.1 to use soonish perhaps?

scout/server/blueprints/cases/templates/cases/case.html
dnil: Reviewer 2 is always right! 😊  I was sure it was in the css above, but probably two different versions that diverged a little. The class should ideally be added to the css instead of disabling transition for all collapses on the page.

northwestwitch: Reviewer 2 is such an evil person, when it comes to manuscript revisions at least! 😆

mikaell: > Reviewer 2 is always right! 😊 I was sure it was in the css above, but probably two different versions that diverged a little. The class should ideally be added to the css instead of disabling transition for all collapses on the page.

Yes. That was the plan och also the implementation. Might differ from committed code.

mikaell: Now committed code and intention are the same.

mikaell:
> Nice! Feels faster already! 🏎️ 💨
> 
> We should probably sort out the case origins for the files before merging - there could be test runs from MIP11.1 to use soonish perhaps?

Yes. I think I have found a good candidate case.

mikaell:
Done!

dnil (approved):
Nice, thank you!

PR #3610, opened 2022-09-30
Avoid duplicate analyses and showing cases as rerun when first upload failed

### LINKED ISSUE #3608: Cases marked as reruns which are not reruns
Check for instance cases F0047911 and F0047867 for cust087 (you see a rerun tag on caseS page)

northwestwitch:
Attempt at fix #3608 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test 1 --> cases that are not rerun should not be marked as rerun**:
1. Locally, recreate the database situation illustrated in https://github.com/Clinical-Genomics/scout/issues/3608#issuecomment-1263513504.
2. In main branch go to list of cases and make sure that demo case will be marked as rerun
3. Switch to this case and make sure that case is not marked as rerun

**How to test 2 --> update functionality not compromised by new code**:
5. run scout setup demo and start from a fresh demo instance
6. Go to the demo case config file and change the analysis date
7. Re-upload the demo case (with -u option) and make sure that the new analysis is saved correctly and the old one is written under "analyses" in the case document (case mongo collection)  

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by DN

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -877,44 +877,49 @@ class CaseHandler(object):
                 if ind.get("tissue_type") is None:
                     ind["tissue_type"] = old_ind.get("tissue_type")
 
+        analysis_date = case_obj["analysis_date"]
+        update_actions = {
+            "$addToSet": {"collaborators": {"$each": case_obj["collaborators"]}},
+            "$set": {
+                "analysis_date": analysis_date,
+                "custom_images": case_obj.get("custom_images"),
+                "cnv_report": case_obj.get("cnv_report"),
+                "coverage_qc_report": case_obj.get("coverage_qc_report"),
+                "delivery_report": case_obj.get("delivery_report"),
+                "gene_fusion_report": case_obj.get("gene_fusion_report"),
+                "gene_fusion_report_research": case_obj.get("gene_fusion_report_research"),
+                "genome_build": case_obj.get("genome_build", "37"),
+                "has_strvariants": case_obj.get("has_strvariants"),
+                "has_svvariants": case_obj.get("has_svvariants"),
+                "individuals": case_obj["individuals"],
+                "is_research": case_obj.get("is_research", False),
+                "madeline_info": case_obj.get("madeline_info"),
+                "mme_submission": case_obj.get("mme_submission"),
+                "multiqc": case_obj.get("multiqc"),
+                "panels": case_obj.get("panels", []),
+                "rank_model_version": case_obj.get("rank_model_version"),
+                "rerun_requested": case_obj.get("rerun_requested", False),
+                "research_requested": case_obj.get("research_requested", False),
+                "smn_tsv": case_obj.get("smn_tsv"),
+                "status": case_obj.get("status"),
+                "sv_rank_model_version": case_obj.get("sv_rank_model_version"),
+                "track": case_obj.get("track", "rare"),
+                "updated_at": updated_at,
+                "vcf_files": case_obj.get("vcf_files"),
+            },
+        }
+        analysis_dates = [
+            analysis.get("date", [{"date": analysis_date}]) for analysis in old_case.get("analyses")
+        ]
+        if analysis_date not in analysis_dates:  # avoid duplicates due to failed uploads
+            update_actions["$addToSet"]["analyses"] = {
+                "date": old_case["analysis_date"],
+                "delivery_report": old_case.get("delivery_report"),
+            }
+
         updated_case = self.case_collection.find_one_and_update(
             {"_id": case_obj["_id"]},
-            {
-                "$addToSet": {
-                    "collaborators": {"$each": case_obj["collaborators"]},
-                    "analyses": {
-                        "date": old_case["analysis_date"],
-                        "delivery_report": old_case.get("delivery_report"),
-                    },
-                },
-                "$set": {
-                    "analysis_date": case_obj["analysis_date"],
-                    "custom_images": case_obj.get("custom_images"),
-                    "cnv_report": case_obj.get("cnv_report"),
-                    "coverage_qc_report": case_obj.get("coverage_qc_report"),
-                    "delivery_report": case_obj.get("delivery_report"),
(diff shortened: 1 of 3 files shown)


dnil:
✅ Tested ok locally.

Before:
Screenshot 2022-10-04 at 09 01 34
After:
Screenshot 2022-10-04 at 09 02 09

scout/server/blueprints/institutes/controllers.py
```diff
@@ -424,13 +424,13 @@ def populate_case_obj(case_obj):
...
         # Check if case was re-runned
-        analyses = case_obj.get("analyses", [])
-        now = datetime.datetime.now()
-        case_obj["is_rerun"] = (
-            len(analyses) > 1
-            or analyses
-            and analyses[0].get("date", now) < case_obj.get("analysis_date", now)
-        )
+        last_analysis = case_obj.get("analysis_date", datetime.datetime.now())  # it's a datetime
+        all_analyses = set()  # a set of datetimes
```
dnil, line 428:
```suggestion
        all_analyses_dates = set()
```

dnil: Better with a descriptive var name and no comment!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -424,13 +424,13 @@ def populate_case_obj(case_obj):
...
+        for analysis in case_obj.get("analyses", [{"date": last_analysis}]):
+            all_analyses.add(analysis.get("date", last_analysis))
+
+        case_obj["is_rerun"] = len(all_analyses) > 1 or last_analysis > max(all_analyses)
```
dnil, line 432: Why max if len == 0 or 1?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -424,13 +424,13 @@ def populate_case_obj(case_obj):
...
         ]
         # Check if case was re-runned
-        analyses = case_obj.get("analyses", [])
-        now = datetime.datetime.now()
-        case_obj["is_rerun"] = (
-            len(analyses) > 1
-            or analyses
-            and analyses[0].get("date", now) < case_obj.get("analysis_date", now)
-        )
+        last_analysis = case_obj.get("analysis_date", datetime.datetime.now())  # it's a datetime
```
dnil, line 427:
```suggestion
        last_analysis_date = case_obj.get("analysis_date", datetime.datetime.now())  
```

scout/adapter/mongo/case.py
```diff
@@ -877,44 +877,50 @@ def update_case(self, case_obj, keep_date=False):
...
+                "track": case_obj.get("track", "rare"),
+                "updated_at": updated_at,
+                "vcf_files": case_obj.get("vcf_files"),
+            },
+        }
+        analysis_dates = [
+            analysis.get("date", [{"date": analysis_date}])
+            for analysis in old_case.get("analyses", [])
+        ]
+        if analysis_date not in analysis_dates:  # avoid duplicates due to failed uploads
```
dnil, line 915: Write a line in the docstring about what the function does while we still know! 😊 In addition to the comment if you feel the latter still helps reading, but could be the comments here become redundant if one can read the overall goal in the docstring!

dnil (approved):
Nice, it works like a charm! When reading I noticed the classics for python readability: extending var names and docstring instead of adding comments.

scout/adapter/mongo/case.py
northwestwitch: Fixed. Fixed also the condition for pushing the old analysis to the "analysis" list. Now like this:

```
if analysis_date != old_analysis_date and old_analysis_date not in analysis_dates:
            update_actions["$addToSet"]["analyses"] = {
                "date": old_analysis_date,
                "delivery_report": old_case.get("delivery_report"),
            }
```

scout/server/blueprints/institutes/controllers.py
northwestwitch: It was written like that in the code before (kind of). But I think it is sufficient to just have this:
`if last_analysis_date > max(all_analyses_dates)`

I'll test with this code on stage!

northwestwitch: The code had a bug before, and I remember that in some cases we could have the last analysis (case["analysis_date]) also inside the analyses list.  I think that's why we have 1 there

dnil: Sorry, I don't follow? Either its longer than 1, or we compare to the max value of the one (or 0)?

northwestwitch: With the bug from some time ago (current analysis in analyses list) last_analysis > the_only_analysis might be False

northwestwitch: In that case you need len(all_analyses) > 1

dnil: But its an `or` not an `xor`! If one of them is true, the expression will be true. So if there is more than one analysis, it will be true regardless of what is in the second term. Hence enough to test the one remaining value.

northwestwitch: What is `xor`?

What if you have the following situation saved in database (old bug):

```
case.analysis_date : 2022-10-04
case.analyses : [{date:2021-10-04}, {date:2022-10-04}] 
```

Then if you check with ONLY 
if `last_case_analysis > all_cases_analyses`

You get False. 
But it's rerun, because it has 2 analyses

Or am I not understanding your question?

dnil: No, I think not? The issue is the use of the `max()` function. First you test if there are more than one unique date. Then you take the max of all the unique dates. Thing is, if there is more than one, the whole expression is true anyway!

northwestwitch: To sum up the discussion: we leave it as it is since it works. When we touch the code again, modify this line to:

case_obj["is_rerun"] = len(all_analyses) > 1 or last_analysis > all_analyses[0]

Issue #3600, opened 2022-09-28
Matching causatives


dnil:
Try to make the matching causatives sharper in differentiating causatives in relevant panels from panels likely to result in incidental findings for the institute.

ehre:
A related thought (I can open a separate issue if you prefer): Should perhaps "managed variants" also be visible only in relevant gene panels? I had a discussion last week with a Stockholm colleague and we seem to have the same problem with incidental findings for this category.

ehre:
Not sure what you mean with "ran with OMIM", but yes, it is definitely a question for us as well. To answer your question, it is important to understand all use cases, but from my perspective perhaps one way would be to only show matching causatives and managed variants in genes that are included in the chosen gene panel(s) per default, with an option to "open up" if you are in exploratory mode?

dnil:
I think you started with more open mendeliome analysis, less panels, but nice that things converge!

Right, that is kind of the way it is intended currently: an extra click is needed to look, and the causatives from within your selected panel would be visible on the variant lists anyway. The (even more open) initial use-case was to avoid missing causative variants from phenotypically partially overlapping panels.

ehre:
Ah ok. Well, that is the typical procedure for complex trios, but all our single WGS analyses run on panels (i.e., the majority of requests). The thing with managed variants, as you know, is that they are displayed separately so you currently see them on all investigations, but if you can make them to just show up on panels where the gene is included would be good, IMHO.

dnil:
We'll give it another shot at a proper balance point. The thing is that several customers (typically with a few, somewhat related panels) find it useful. At the very least we could add an optional limit-to-panels on a per institute basis.

dnil:
Comment from Sofie S: also ensure that the list of matching causatives only includes variants that are currently marked causative, not other cases that just have the variant. 

This could be in line with a more restrictive use of the functionality, and should probably be optional. It has been very practical a few times to go directly to old cold cases when a new causative is established.

northwestwitch:
> A related thought (I can open a separate issue if you prefer): Should perhaps "managed variants" also be visible only in relevant gene panels? I had a discussion last week with a Stockholm colleague and we seem to have the same problem with incidental findings for this category.

[...]

PR #3611, opened 2022-10-03
Limit incidental findings by filtering matching causatives and managed vars by gene panels - Case page fix -

northwestwitch:
Fixes partially #3600 

- Introduces a new institute-level key/value to create a list of "safe panels" for searching for matching causatives and managed vars
- Fixes the search for matching causatives and managed vars to not display matchings of variants outside the institute safe panels

This is just 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/institute.py
+++ b/scout/adapter/mongo/institute.py
@@ -172,6 +172,21 @@ class InstituteHandler(object):
 
         return institute_obj
 
+    def safe_genes_filter(institute_id):
+        """Returns a list of "safe" HGNC IDs to filter variants with. These genes are retrieved from the institute.gene_panels_matching
+        Can be used to limit secondary findings when retrieving other causatives or matching managed variants
+
+        Args:
+            institute_id(str): _id of an institute
+
+        Returns:
+            hgnc_list(list of HGNC ids)
+        """
+        safe_genes = []
+        for panel_name in institute_obj.get("gene_panels_matching", {}).keys():
+            safe_genes += self.panel_to_genes(panel_name=panel_name, gene_format="hgnc_id")
+        return safe_genes
+
     def institutes(self, institute_ids=None):
         """Fetch all institutes.
 
--- a/scout/adapter/mongo/institute.py
+++ b/scout/adapter/mongo/institute.py
@@ -180,7 +180,7 @@ class InstituteHandler(object):
             institute_id(str): _id of an institute
 
         Returns:
-            hgnc_list(list of HGNC ids)
+            safe_genes(list of HGNC ids)
         """
         safe_genes = []
         for panel_name in institute_obj.get("gene_panels_matching", {}).keys():
--- a/scout/adapter/mongo/institute.py
+++ b/scout/adapter/mongo/institute.py
@@ -64,7 +64,7 @@ class InstituteHandler(object):
             remove_sanger(str): Email adress for sanger user to be removed
             phenotype_groups(iterable(str)): New phenotype groups
             gene_panels(dict): a dictionary of panels with key=panel_name and value=display_name
-            gene_panels_matching(dict): panels to limit search of matching variants (managed, causatives, tiered) to. Dict with key=panel_name and value=display_name
+            gene_panels_matching(dict): panels to limit search of matching variants (managed, causatives) to. Dict with key=panel_name and value=display_name
             group_abbreviations(iterable(str))
             add_groups(bool): If groups should be added. If False replace groups
             sharing_institutes(list(str)): Other institutes to share cases with
--- a/scout/adapter/mongo/institute.py
+++ b/scout/adapter/mongo/institute.py
@@ -43,6 +43,7 @@ class InstituteHandler(object):
         remove_sanger=None,
         phenotype_groups=None,
         gene_panels=None,
+        gene_panels_matching=None,
         group_abbreviations=None,
         add_groups=None,
         sharing_institutes=None,
@@ -63,6 +64,7 @@ class InstituteHandler(object):
             remove_sanger(str): Email adress for sanger user to be removed
             phenotype_groups(iterable(str)): New phenotype groups
             gene_panels(dict): a dictionary of panels with key=panel_name and value=display_name
+            gene_panels_matching(dict): panels to limit search of matching variants (managed, causatives, tiered) to. Dict with key=panel_name and value=display_name
             group_abbreviations(iterable(str))
             add_groups(bool): If groups should be added. If False replace groups
             sharing_institutes(list(str)): Other institutes to share cases with
@@ -94,9 +96,6 @@ class InstituteHandler(object):
             )
             updates["$push"] = {"sanger_recipients": sanger_recipient}
 
-        if sanger_recipients is not None:
-            updates["$set"]["sanger_recipients"] = sanger_recipients  # can be empty list
-
         if remove_sanger:
             LOG.info(
                 "Removing sanger recipient {0} from institute: {1}".format(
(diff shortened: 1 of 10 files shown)


scout/adapter/mongo/institute.py
```diff
@@ -192,6 +172,24 @@ def institute(self, institute_id):
...
+
+        Args:
+            institute_id(str): _id of an institute
+
+        Returns:
+            safe_genes(list of HGNC ids)
+        """
+        safe_genes = []
+        institute_obj = self.institute(institute_id)
+        if not institute_obj:
```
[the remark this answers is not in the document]
dnil: ...and remove the following line and reindent? Up to you; this may actually be more Pythonesque, but I find the c-style easier here since it avoids a negation and an extra line, plus a little confusion as to the value of `safe_genes == []` for just one indent level extra. 🤷‍♂️

scout/server/blueprints/cases/controllers.py
```diff
@@ -329,12 +329,19 @@ def case(store, institute_obj, case_obj):
...
+    limit_genes = store.safe_genes_filter(institute_obj["_id"])
+
     data = {
         "institute": institute_obj,
         "case": case_obj,
         "status_class": STATUS_MAP.get(case_obj["status"]),
-        "other_causatives": [var for var in store.check_causatives(case_obj=case_obj)],
-        "managed_variants": [var for var in store.check_managed(case_obj=case_obj)],
+        "other_causatives": [
+            var for var in store.check_causatives(case_obj=case_obj, limit_genes=limit_genes)
```
dnil, line 340: nicely prepared! 😁

scout/server/blueprints/institutes/controllers.py
```diff
@@ -274,6 +274,7 @@ def populate_institute_form(form, institute_obj):
     for panel in available_panels:
         panel_set.add((panel["panel_name"], panel["display_name"]))
     form.gene_panels.choices = sorted(panel_set, key=lambda tup: tup[1])
+    form.gene_panels_matching.choices = sorted(panel_set, key=lambda tup: tup[1])
```
dnil, line 277: 👍

dnil (approved):
I think we can go with this as a test rocket. It will not solve the whole issue for big institutes with many panels and medical responsibility areas, but will aid the discussion forward. If they want to stay this inclusive, while worried about seeing variants from other part areas, a similar feature could move to the case area.

scout/server/blueprints/cases/controllers.py
northwestwitch: I'm just wondering if the search and subsequent loading of case page will be slower, especially when having a long list of limit_genes. Have you tried it on stage? I haven't

dnil: I was thinking the same, but then assumed you had tested and checked for index use! 😅

dnil: I did try it with multiple panels, and it didn't grind to a halt or anything.

scout/adapter/mongo/institute.py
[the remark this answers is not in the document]
northwestwitch: I think I'd leave it this was, it's how we handle these situations in most of the code (whenever we touch it). But I could add a comment perhaps?

northwestwitch:
Before merging this, don't you think we should add something on case page that notifies when the institute has one or more safe panels set? So people would know that search for matching causatives and managed vars is not performed all all vars? 

Would be nice to add also a list of panels where the search is done, as a popover of something? Think about the situation when an institute adds a panel and forgets to update the list of safe panels

scout/server/blueprints/cases/controllers.py
dnil: But do check that it hits indexes!

dnil:
> Before merging this, don't you think we should add something on case page that notifies when the institute has one or more safe panels set? So people would know that search for matching causatives and managed vars is not performed on all vars?
> 
> Would be nice to add also a list of panels where the search is done, as a popover of something? Think about the situation when an institute adds a panel and forgets to update the list of safe panels

Some info is always nice. We will have to inform very thoroughly either way!

Just be careful: one would not want a situation where the match is inadvertently disclosed just for showing the information about what panels are allowed.

northwestwitch:
> Just be careful: one would not want a situation where the match is inadvertently disclosed just for showing the information about what panels are allowed.

Yeah, I know. I meant just an info saying safe gene search is on or something

scout/server/blueprints/cases/controllers.py
northwestwitch: `check_causatives` is first looking into the events to collect causatives and partial causatives, then loops thru them to see if they meet the requirements.

`check_managed` is using the positional_variant_ids to collect all variants that match. We should be fine I think

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -54,9 +54,17 @@ <h4 class="">
...
-            <div class="col-sm-8">
+            <div class="col-sm-6">
               {{ variants_buttons() }}
             </div>
+            <div class="col-sm-2">
+              Safe variant matching:
+              {% if institute.gene_panels_matching %}
+                <span class="badge rounded-pill bg-warning text-dark" data-bs-toggle="tooltip" data-bs-placement="left" title="To avoid incidental findings, eventual matching causatives and managed variants displayed on case page are filteres using a list of 'safe' panels. Modify institute settings to edit list of 'safe' panels. Current safe panels: {{ institute.gene_panels_matching.values()|list }}">on</span>
+              {% else %}
+                <span class="badge rounded-pill bg-info" data-bs-toggle="tooltip" data-bs-placement="left" title="Matching causatives and managed variants displayed on case page are NOT filtered by gene panel. Use caution to avoid incidental findings.">off</span>
```
northwestwitch, line 65: image

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -54,9 +54,17 @@ <h4 class="">
...
               </h4>
             </div>
-            <div class="col-sm-8">
+            <div class="col-sm-6">
               {{ variants_buttons() }}
             </div>
+            <div class="col-sm-2">
+              Safe variant matching:
+              {% if institute.gene_panels_matching %}
+                <span class="badge rounded-pill bg-warning text-dark" data-bs-toggle="tooltip" data-bs-placement="left" title="To avoid incidental findings, eventual matching causatives and managed variants displayed on case page are filteres using a list of 'safe' panels. Modify institute settings to edit list of 'safe' panels. Current safe panels: {{ institute.gene_panels_matching.values()|list }}">on</span>
```
northwestwitch, line 63: image

scout/adapter/mongo/institute.py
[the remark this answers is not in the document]
dnil: Nah, in that case make it a function call! But its all ok, its just a few words - I can still read it! 😁

dnil:
I'll put it on stage to test!

scout/server/blueprints/cases/templates/cases/case.html
dnil: Tested; looks good!

scout/server/blueprints/cases/controllers.py
dnil: Ok, so could slow down with increasing counts of relevant events then, but still, it is matching all variants first, so as long as the gene check is done only after match to the variant, it must be a very small contribution. Things should be all green.

dnil (approved):
Tested ok on stage methinks!

PR #3625, opened 2022-10-07
Prepare version 4.60

northwestwitch:
### Added
- [x] Mitochondrial deletion signatures (mitosign) can be uploaded and shown with mtDNA report
- [x] A `Type of analysis` column on Causatives and Validated variants pages
- [x] List of "safe" gene panels available for matching causatives and managed variants in institute settings, to avoid secondary findings
- [x] svdb_origin as a synonym for FOUND_IN to complement set for variants found by all callers
### Changed
- [x] Hide removed gene panels by default in panels page
- [x] Removed option for filtering cancer SVs by Tumor and Normal alt AF
- [x] Hide links to coverage report from case dynamic HPO panel if cancer analysis
- [x] Remove rerun emails and redirect users to the analysis order portal instead
- [x] Updated clinical SVs igv.js track (dbVar) and added example of external track from https://trackhubregistry.org/
### Fixed
- [x] If trying to load a badly formatted .tsv file an error message is displayed.
- [x] Avoid showing case as rerun when first attempt at case upload failed
- [x] Dynamic autocomplete search not working on phenomodels page
- [x] Callers added to variant when loading case (but `set=Intersection` from MIP/svdb)
- [x] Now possible to update managed variant from file without deleting it first
- [x] Missing preselected chromosome when editing a managed variant
- [x] Preselected variant type and subtype when editing a managed variant
- [x] Typo in dbVar ClinVar track, hg19

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by


northwestwitch:
I've installed this branch on stage, so we can test it

northwestwitch:
Oh no, the Hg19 SVs track is not found any more, I'll fix!

image

image

northwestwitch:
I forgot that we have to modify the config files of Scout also if we want to redirect to the order portal to rerun samples. But we should modify at the same time that we do the deploy of the new release, otherwise the old release will still look for the params to send the email in the config files!

northwestwitch:
> I forgot that we have to modify the config files of Scout also if we want to redirect to the order portal to rerun samples. But we should modify at the same time that we do the deploy of the new release, otherwise the old release will still look for the params to send the email in the config files!

Perhaps we can have both rerun URL to the orderportal and  and rerun email params for a while. When the new release of scout is deployed in scout the rerun email params won't be used any more and we can removed them

northwestwitch:
> Perhaps we can have both rerun URL to the orderportal and and rerun email params for a while. When the new release of scout is deployed on prod the rerun email params won't be used any more and we can removed them

This is done --> https://github.com/Clinical-Genomics/servers/pull/914

dnil:
Loading `justhusky` on `scout-stage` from `hasta`, with mitodel files manually added to load config.
✅ Screenshot 2022-10-11 at 17 12 57
Also tests caller load:
Screenshot 2022-10-11 at 17 11 13
Screenshot 2022-10-11 at 17 10 54
Looks good, but also saw this:
Screenshot 2022-10-11 at 17 12 03
This is correct in a way from Scout, since the vcf has `set=Intersection;` (but no `FOUND_IN` tag). In contrast e.g. `svdb_origin=tiddit|cnvnator|manta;` gives a better picture. It is fine for the one or two caller set's but breaks when all three are just given as Intersection. Lets give this a think. We should forward the issue to MIP again - perhaps it is already fixed with the latest of 11.1? But since it is bugged in Scout also already, some level of fixing now is also on the table. We could temp disable a couple of currently unused callers, move the caller constants to config, or add parsing of `svdb_origin`.

dnil:
✅ 
Screenshot 2022-10-11 at 17 30 49

dnil (approved):

dnil:
✅ Deployed on hasta:
```
Installing collected packages: scout-browser
  Attempting uninstall: scout-browser
    Found existing installation: scout-browser 4.58
    Uninstalling scout-browser-4.58:
      Successfully uninstalled scout-browser-4.58
Successfully installed scout-browser-4.60
θ61° [daniel.nilsson@hasta:~/scout_update_221012] [P_scout] $ 
```

northwestwitch:
Deployed on cg-prod-services:

image

image

PR #3630, opened 2022-10-12
Fix #3629 - svdb_origin tags

### LINKED ISSUE #3629: svdb_origin
SVDB uses a VCF tag `svdb_origin` that appears synonymous in function to `FOUND_IN` - a pipe separated list of caller tags.
Parsing this is useful in particular for variants found by all of multiple callers, with set=Intersection.
Screenshot 2022-10-11 at 17 12 03

dnil:
This PR adds a functionality.

`svdb_origin` is now parsed in addition to `FOUND_IN` and `set`.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. pick a case merged with a recent svdb version, producing svdb_origin tags, and a variant found by all callers
2. note it is shown as called by some callers not used in the analysis
3. reload with this patch, and see only the actual callers found

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/parse/variant/callers.py
+++ b/scout/parse/variant/callers.py
@@ -50,10 +50,9 @@ def parse_callers(variant, category="snv"):
             elif call in set(callers.keys()):
                 callers[call] = "Pass"
 
-    # The following is parsing of a custom made merge
-
     if raw_info or svdb_origin or other_info:
         return callers
+
     if category == "snv":
         # cyvcf2 FILTER is None if VCF file column FILTER is "PASS"
         filter_status = "Pass"
--- a/scout/parse/variant/callers.py
+++ b/scout/parse/variant/callers.py
@@ -34,7 +34,7 @@ def parse_callers(variant, category="snv"):
             elif call in set(callers.keys()):
                 callers[call] = "Pass"
     # The following is parsing of a custom made merge
-    other_info = variant.INFO.get("FOUND_IN")
+    other_info = variant.INFO.get("FOUND_IN") or variant.INFO.get("svdb_origin")
     if other_info:
         for info in other_info.split(","):
             called_by = info.split("|")[0]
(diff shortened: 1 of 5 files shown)


dnil:
Should be ready for review, given that the tests pass naturally. Fixes the issue on 4.60 testing (really it was there before summer as well, when callers were still parsed) with set=Intersection adding in callers known to Scout but not really part of the pipeline.

scout/parse/variant/callers.py
```diff
@@ -8,6 +8,11 @@
 def parse_callers(variant, category="snv"):
     """Parse how the different variant callers have performed
 
+    Caller information can be passed in one of three ways, in order of priority:
```
northwestwitch, line 11: Very good with the explanation, it's a complicated matter

northwestwitch (approved):
👍🏻 Looks good!

Would be nice to add something about svdb also in the docs, perhaps this one: https://github.com/Clinical-Genomics/scout/blob/main/docs/user-guide/variants.md

Doesn't have to be in this PR though

dnil:
Looks good so far. Here is justhusky loaded on stage with an "Intersection" variant from all three callers:
Screenshot 2022-10-12 at 09 30 14

dnil:
> Would be nice to add something about svdb also in the docs, perhaps this one: https://github.com/Clinical-Genomics/scout/blob/main/docs/user-guide/variants.md

I reformatted that part about SVDB a little; it also has a link to the repo. I'm all for figuring out a good way to connect tests and documentation here; should be possible to use some auto-doc function to get more of the docstrings from tests out. Then we could get going documenting VCF info that Scout can actually display, and the content expected. But another PR.

PR #3632, opened 2022-10-17
Add filter RD variants by GT call

### LINKED ISSUE #3537: Filtering/hover info for genotype
Hi all!

One of our collaborators using scout asked after a possible feature request.

We are currently working with some singletons, which means we don't have the same level of inheritance information that you'd have with cohort. For example, we won't have the "Inheritance model" section of scout the filter variants by. However, these singletons still have genotype information that is viewable on a given variant page (in the "GT call" box).

We were wondering if it would be possible to do one of the following:

- have an option to filter variants by genotype call (make the most sense to me)
- have the GT calls viewable on the variant summary page for example as a hover box
- have the ability to filter variants by the inheritance model of the gene the variant is in (this option seems a bit harder since it relies on gene and inheritance model for genes annotations being available)

This would provide the benefit of being able to filter for variants or quickly see if variants are hetrozygous/homozygous on the variant summary page for singletons.

We're currently working with scout version  4.57.4

What are your thoughts?

northwestwitch:
Attempt at fix #3537 

- In RD cases, SNVs and SVs pages add a select that allows to return variants with a specific genotype:

image

-  Can be used in a combo with this checkbox to return only variants for the affected samples that have the chosen genotype:
 
image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN


scout/adapter/mongo/query.py
```diff
@@ -715,3 +722,16 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
...
+
+def _get_query_genotype(query):
+    """Query helper that returns the specific genotype selected by the user for a variantS query
+
+    Args:
+        query(dict): form dictionary with variantS query terms
+    """
+    q_value = query.get("genotypes")
+    if q_value == "other":
+        return {"$nin": ["0/1", "1/1", "0/0", "./."]}
```
dnil, line 735: Most may at some point be interested also in phased GTs, but I guess we and the other centers are not quite there yet?

scout/constants/variant_tags.py
```diff
@@ -96,6 +96,15 @@
     ("XD_dn", "X Linked Dominant De Novo"),
 )
 
+# Used for filter variants in blueprints.variants
+VARIANT_GENOTYPES = (
+    ("", ""),
+    ("0/1", "Heterozygous (0/1)"),
```
dnil, line 102: Heterozygous could also be "1/0". 😊 There would be several options to fix; e.g. having a list to match in, or a single pattern for matching that allows both..

scout/constants/variant_tags.py
```diff
@@ -96,6 +96,15 @@
...
+    ("1/1", "Homozygous (1/1)"),
+    ("0/0", "Homozygous ref (0/0)"),
+    ("./.", "Allele missing (./.)"),
```
dnil, line 105: "Allele missing" is ambiguous. It could also mean one allele missing, in which case the pattern could be any combination with a "." in it.

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -390,27 +403,7 @@
...
-  <div class="col-2">
-    {{ form.functional_annotations.label(class="control-label") }}
-    {{ form.functional_annotations(class="selectpicker", data_style="btn-secondary") }}
-  </div>
-  <div class="col-2">
-    {{ form.genetic_models.label(class="control-label") }}
-    {{ form.genetic_models(class="selectpicker", data_style="btn-secondary") }}
-  </div>
-</div>
+  {{ variants_common_filters(form, "sv") }}
```
dnil, line 406: 😸

tests/adapter/mongo/test_query.py
```diff
@@ -131,6 +131,18 @@ def test_gene_panel_query(adapter, case_obj):
     }
 
 
+def test_genotype_query(adapter, case_obj):
```
dnil, line 134: 👍

dnil (changes requested):

scout/constants/variant_tags.py
northwestwitch: Yes, but in this case I mean both alleles missing. I'll write like this!

scout/adapter/mongo/query.py
northwestwitch: No, let's keep it simpler!

scout/constants/variant_tags.py
northwestwitch: I think I wrote it like this because the vast majority of calls in our database are of type 0/1 (84%)

image

Only 2% is the other heterozygous type (1/0)

image

So this 2% gets collected when you filter for "other".

But I agree with you: it's confusing and if you want heterozygous calls then you want them for both alleles. Will fix!

scout/adapter/mongo/query.py
```diff
@@ -715,3 +722,18 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
...
+def _get_query_genotype(query):
+    """Query helper that returns the specific genotype selected by the user for a variantS query
+
+    Args:
+        query(dict): form dictionary with variantS query terms
+    """
+    q_value = query.get("genotypes")
+    if q_value == "other":
+        return {"$nin": ["0/1", "1/1", "0/0", "1/0", "./."]}
+    elif q_value == "0/1 or 1/0":
```
dnil, line 736: 👍

dnil (approved):
👍

PR #3641, opened 2022-10-24
Loqus priv

dnil:

The issue here is that `getlist` will return the `[]` also if the control does not exist on the form, not `None`. Actually check if user is admin, else set it to `None`.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. With current main (4.60), as an admin user, ensure an institute has a loqusdb instance selected. Switch to an unprivileged user, change some other random setting, and notice that loqus db connectivity is gone for the institute (`loqusdb_ids` is `[]`). 
2. After patch, unprivileged user saving other settings should not reset loqusdb_ids.
3. Also ensure that privileged users can still change loqusdb_ids when needed.

 how to test it, possibly with real cases/data

**Expected outcome**:

**Review:**
- [x] code approved by CR

### DIFF
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -328,6 +328,10 @@ def update_institute_settings(store, institute_obj, form):
     for cohort in form.getlist("cohorts"):
         cohorts.append(cohort.strip())
 
+    loqusdb_ids = form.getlist("loqusdb_id")
+    if current_user.is_admin is False:
+        loqusdb_ids = None
+
     updated_institute = store.update_institute(
         internal_id=institute_obj["_id"],
         sanger_recipients=sanger_recipients,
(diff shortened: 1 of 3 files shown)


dnil:
Seems to work as normal for admins. Picking a variant for testing:
Screenshot 2022-10-24 at 15 59 05
When dropping admin privs:
Screenshot 2022-10-24 at 16 00 11
And making a random change to gene lists for the institute:
Screenshot 2022-10-24 at 16 00 02
The variant can still be seen for the institute (though not the additional matching case for the non-admin user):
Screenshot 2022-10-24 at 16 00 34

scout/server/blueprints/institutes/controllers.py
```diff
@@ -328,6 +328,10 @@ def update_institute_settings(store, institute_obj, form):
     for cohort in form.getlist("cohorts"):
         cohorts.append(cohort.strip())
 
+    loqusdb_ids = form.getlist("loqusdb_id")
+    if current_user.is_admin is False:
```
northwestwitch, line 332: 👍🏻

northwestwitch (approved):
Looks good. Nice fix! 👍🏻

PR #3656, opened 2022-11-03
Validate ClinVar submission refseq:HGVSs using VariantValidator

northwestwitch:
Pre-validate the refseq:hgvs items shown in the form ClinVar add variant:

image

Validation is performed using the [VariantValidator API](https://rest.variantvalidator.org/)

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:
1. Locally or on stage, create a ClinVar submission (or add to an existing one) with a SNV variant. Make sure that transcripts for this variant are validated during the submission process

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
@dnil this one can wait until the next release

dnil:
> @dnil this one can wait until the next release

Yeah, while it looks rather good, it would not hurt to brainstorm handling of external request first I guess. This is not Quart (which looks a bit tempting: https://quart.palletsprojects.com/en/latest/)

dnil (approved):
We can't have this hanging around forever. It works just fine, and the service is super fast, assuming it is actually working. We have worse in other places in the code. It would be very nice to implement it to finish async and update the page via e.g. js instead, but let us leave that until it becomes a problem. 

It does have a layout issue though; the check boxes can apparently end up outside the bounding cell. Perhaps due to the use of the `<P>` element in the row-col-layout?
Screenshot 2022-12-15 at 15 17 43

northwestwitch:
> We can't have this hanging around forever. It works just fine, and the service is super fast, assuming it is actually working. We have worse in other places in the code. It would be very nice to implement it to finish async and update the page via e.g. js instead, but let us leave that until it becomes a problem.
> 
> It does have a layout issue though; the check boxes can apparently end up outside the bounding cell. Perhaps due to the use of the `<P>` element in the row-col-layout? Screenshot 2022-12-15 at 15 17 43

I can't reproduce the ugly layout thing. I'm testing on stage and the page looks like this:

image

What browser and what variant did you test?

dnil:
Let's see, it was one from the demo case, probably this one:

http://localhost:8000/cust000/643594/e24b65bf27feacec6a81c8e9e19bd5f1

It would also nearly always be Safari on the first try for me, and it does appear the same if I just pin and submit another [variant](http://localhost:8000/cust000/643594/b78204ac5f3a119bbaf9a4a20fcf9a94):

Screenshot 2022-12-19 at 12 47 05

They are tentatively good on Chrome and Firefox though!

northwestwitch:
> Let's see, it was one from the demo case, probably this one:
> 
> http://localhost:8000/cust000/643594/e24b65bf27feacec6a81c8e9e19bd5f1
> 
> It would also nearly always be Safari on the first try for me, and it does appear the same if I just pin and submit another [variant](http://localhost:8000/cust000/643594/b78204ac5f3a119bbaf9a4a20fcf9a94):
> 
> Screenshot 2022-12-19 at 12 47 05
> 
> They are tentatively good on Chrome and Firefox though!

Ah OK, that must be because I've tried on chrome! Will fix, thanks!

dnil:
It was pretty with the colors, but this is nice and uniform - and doesn't break!

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
```diff
@@ -100,14 +100,16 @@ <h2 class="fs-title">Variant Details</h2>
                         {% set ns.label = item.label.text | replace(" (validated)", "") %}
 
                         <td style="width: 30%; text-align: left;">
+                          <p
                           {% if ns.label|length > 25 %}
-                          <p data-bs-toggle="tooltip" title="{{ns.label}}">{{ ns.label|truncate(25,true,'..') }}</p>
```
dnil: 👍

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
```diff
@@ -85,24 +85,34 @@ <h2 class="fs-title">Variant Details</h2>
                     <span class="text-danger" data-bs-toggle='tooltip' title="If you do not provide any HGVS expression, chromosome coordinates will be used to describe this variant instead (automatic). HGVS expressions were validated using VariantValidator"><strong>?</strong></span>
                     <span class="badge bg-primary float-end"><a class="text-white" href="https://variantvalidator.org/service/validate/" target="_blank" rel="noopener">VariantValidator</a></span>
                     <br><br>
+                    <table style="width:100%; table-layout: auto;">
```
dnil: Hm, not the biggest fan: but I'm sure it will work - they did in like 1995. 😉

dnil: Oh, we are supposed to add some aria style suggestion to tell it is just a layout table, not a content one, aren't we?

northwestwitch:
checkbox-text alignment is fixed now: 

image

northwestwitch:
SonarCloud doesn't like it:

image

On the other hand I've been trying to replace the table with better code (div-based) but looked ugly. I like the fact that specifying the role is helping assistive technologies, so I'm going to keep it and I shut up SonarCloud!

PR #3664, opened 2022-11-11
Colorize inheritance pattern on variant page

### LINKED ISSUE #3631: One colour per inheritance pattern?
It would be nice to show different inheritance patterns in different colours
In this view 
image
and in this view
image
to make sure one doesn´t misstake for example a hom variant for a het.

northwestwitch:
Fixes partially #3631 : display different colors for different inheritance patterns on Variant page.

Please note that the genetic models used on [VariantS page](https://github.com/Clinical-Genomics/scout/blob/a494edd64090fd4f613c72308ff5623442792af1/scout/constants/variant_tags.py#L86) are different from the inheritance patterns shown on on [Variant page](https://github.com/Clinical-Genomics/scout/blob/a494edd64090fd4f613c72308ff5623442792af1/scout/constants/gene_tags.py#L5)

For this reason I think it would be best to split #3631 into 2 tasks

From this:

image

To this:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


northwestwitch:
Looks OK on stage: 🆗 

image

image

dnil (changes requested):
These colors are nice, but to keep it maintainable, let's try with the bootstrap colors! If it gets too weird, we can always use this.

dnil:
Super! Yeah, I'm still hopeful is possible to just use the ten in that other `colors` palette; I've never tried growing a palette out of their other predefined colors, though it should be possible based on their sass/css docs (https://getbootstrap.com/docs/5.1/customize/color/#theme-colors). The dark mode also overloads some color variables, but not all...

northwestwitch:
> These colors are nice, but to keep it maintainable, let's try with the bootstrap colors! If it gets too weird, we can always use this.

I've been reading the docs on [bootstrap colors page](https://getbootstrap.com/docs/5.1/customize/color/#theme-colors). Looks to me that first 6 colors that are included in the bootstrap theme (black and white are not optimal) are usable for this purpose:

image

But you need to use other colors to compose a complete palette for the [10 types of inheritance]()https://github.com/Clinical-Genomics/scout/blob/13431e694c8299786a9efe2656025cac676f619f/scout/constants/gene_tags.py#L5

Since [other colors](https://getbootstrap.com/docs/5.1/customize/color/#all-colors) are not included in bootstrap, you must generate them. For instance here is how you generate a gradient (here's how: https://www.youtube.com/watch?v=CTmubyW4uYo&ab_channel=5Balloons)

Then you have to [add these styles](https://getbootstrap.com/docs/5.1/customize/color/#example) either on the main layout that is inherited by all these HTML pages or on each page where these colors must be used. Looks a bit cumbersome to me and not easy to maintain. 

What about using another hardcoded palette instead? Hardcoded is easier and maintainable all right I think. Here are a few other examples:

https://coolors.co/palettes/trending

dnil:
In a way I can agree that it is easier to have absolute colors defined where they are used, and that are sure not to change unless we really want them to. The problem is to maintain them nicely together with the other colors, that are defined in css (if manually or via sass or dito). A compromise could be to at least make classes for them instead, and define a desired custom palette in css. That way we at least have all colors in one place, and dont have to remember that they were defined in a constants file?

If you agree and define them in CSS, it is possible to access the other standard bootstrap colors via e.g. `color: var(--bs-purple)` etc, so it would be possible to test how it looks with bootstrap color? 😊

northwestwitch:
> A compromise could be to at least make classes for them instead, and define a desired custom palette in css. That way we at least have all colors in one place, and dont have to remember that they were defined in a constants file?
> 
> If you agree and define them in CSS, it is possible to access the other standard bootstrap colors via e.g. `color: var(--bs-purple)` etc, so it would be possible to test how it looks with bootstrap color? 😊
> 
> See e.g. https://getbootstrap.com/docs/5.0/customize/css-variables/ for those that are always exposed.

Ok, sounds like an easier thing to do. I'll draft something using the colors that are always exposed!

northwestwitch:
Switched to bootstrap colors now:

image

northwestwitch:
What's wrong with the tests now? 

```
 python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
      [end of output]
```

scout/server/static/bs_styles.css
```diff
@@ -49,6 +49,44 @@ a.dropdown-item:visited {
...
+.bg-blue {
+ background-color: var(--bs-blue);
+}
+
+.bg-teal {
+ background-color: var(--bs-teal);
+}
+
+.bg-green {
+ background-color:: var(--bs-green);
```
dnil, line 66:
```suggestion
 background-color: var(--bs-green);
```

dnil: Dunno; seems to work either way, so most parsimonious is single colon. 🤷

scout/constants/gene_tags.py
```diff
@@ -15,6 +15,20 @@
     ("other", "other - Other"),
 )
 
+# Associates an inheritance model with a color, for displaying inheritance on Variant page
+INHERITANCE_PALETTE = {
+    "AD": {"bgcolor": "bg-red", "text_color": "text-white"},
```
dnil, line 20: Much nicer to my eyes! And one could still add an absolute color in the css if the var(bs-) ones are not good enough in the end.

scout/constants/gene_tags.py
```diff
@@ -15,6 +15,20 @@
...
+    "AR": {"bgcolor": "bg-blue", "text_color": "text-white"},
```
dnil, line 21: Ah, and for the models, one would want `AR_hom` and `AD_dn` to really stand out. Here, for the litterature inheritance pattern for the gene, it does not matter so much, as long as it is easy to compare with what the individual actually has.

dnil (approved):

scout/server/static/bs_styles.css
northwestwitch: ah yes, a typo. Thanks!

scout/constants/gene_tags.py
northwestwitch: We can still keep them blue and red. But the bootstrap colors are basically these, so if we want to color by categories ( a gradient) then we need to define these colors in the css. Let's see!

PR #3665, opened 2022-11-11
Colorize genetic models on variantS page

### LINKED ISSUE #3631: One colour per inheritance pattern?

northwestwitch:
Second partial fix #3631. **TO BE REVIEWED PREFERABLY AFTER #3664**, because contains code from it.

Changes RD variantS table from this:

image

To this:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by


scout/constants/variant_tags.py
```diff
@@ -96,6 +98,20 @@
     ("XD_dn", "X Linked Dominant De Novo"),
 )
 
+# Associates an inheritance model with a color, using the same colors categories as scout.constants.gene_tags.INHERITANCE_PALETTE
+# The colors of this palette are the same as the MatPlotLib Tableau palette (https://matplotlib.org/3.4.1/gallery/color/named_colors.html)
+GENETIC_MODELS_PALETTE = {
+    "AR_hom": INHERITANCE_PALETTE["AR"],
```
dnil, line 104: Here I think we will want a graded intensity. AR should be more visible than AR_comp (which anyway has a Compund button connected, and will be nice with a thematic color), AD_dn more intense than AD etc.

scout/constants/variant_tags.py
```diff
@@ -96,6 +98,20 @@
     ("XD_dn", "X Linked Dominant De Novo"),
 )
 
+# Associates an inheritance model with a color, using the same colors categories as scout.constants.gene_tags.INHERITANCE_PALETTE
+# The colors of this palette are the same as the MatPlotLib Tableau palette (https://matplotlib.org/3.4.1/gallery/color/named_colors.html)
+GENETIC_MODELS_PALETTE = {
```
dnil, line 103: 👍 Good with the same base colors!

dnil (changes requested):

northwestwitch:
I'm working on the graded intensity and I've found a solution using opacity. Works nicely on light mode:

image


But it doesn't render OK on dark mode: 😢 

image

northwestwitch:
Ok, I've modified the genetic models (variantS oage) like this:

image

And the inheritance models (variant page) like this:

image

dnil (approved):
Looks good! If you want, I could have a look at your opacity fix - maybe I could coerce the dark mode - but otherwise it is ok like this, isn't it?

northwestwitch:
> Looks good! If you want, I could have a look at your opacity fix - maybe I could coerce the dark mode - but otherwise it is ok like this, isn't it?

The opacity fix looks really nice on the light theme. I have tried to force the badges to stay in light mode also if the system is in dark mode with this css style: `color-scheme: only light;` but didn't work. If you have time yes, give it a try (on another branch perhaps?) You can pull one of the intermediate commits of this branch in the new branch and test. Thanks!

dnil:
> > Looks good! If you want, I could have a look at your opacity fix - maybe I could coerce the dark mode - but otherwise it is ok like this, isn't it?
> 
> The opacity fix looks really nice on the light theme. I have tried to force the badges to stay in light mode also if the system is in dark mode with this css style: `color-scheme: only light;` but didn't work. If you have time yes, give it a try (on another branch perhaps?) You can pull one of the intermediate commits of this branch in the new branch and test. Thanks!

Hm, you may only have done this on your local? Couldn't find the commit? Anyway, its nice enough this way so agreed, new branch and at QOL priority. Meeting ahead...

northwestwitch:
> Hm, you may only have done this on your local?

Perhaps! A disadvantage of **opacity** is that the default style applied to badge color but also badge text. 

I remember using the --bs-bg-opacity style: https://getbootstrap.com/docs/5.2/utilities/background/#opacity

dnil:
> > Hm, you may only have done this on your local?
> 
> Perhaps! A disadvantage of **opacity** is that the default style is applied to badge color but also badge text. Example: https://getbootstrap.com/docs/5.1/utilities/opacity/
> 
> I remember using the `bg-opacity-x` style: https://getbootstrap.com/docs/5.2/utilities/background/#opacity

Looks super sweet! Let's see, aren't we/darkmode still on 5.0? Perhaps we should update..

dnil:
> > Perhaps! A disadvantage of **opacity** is that the default style is applied to badge color but also badge text. Example: https://getbootstrap.com/docs/5.1/utilities/opacity/
> > I remember using the `bg-opacity-x` style: https://getbootstrap.com/docs/5.2/utilities/background/#opacity
> 
> Looks super sweet! Let's see, aren't we/darkmode still on 5.0? Perhaps we should update..

Yeah, well, bootstrap dark stopped main at 5.1.3, since the 5.2-beta color changes were rather big. There is now also a less intrusive dark mode version with only the deltas, not the whole base css:  bootstrap-dark-plugin, but I think its also just at 5.1.3. 

It seems right now they will not fix the core issue with 5.2 in the 5.x, delaying it for v6 https://github.com/twbs/bootstrap/issues/36458. Hmm, maybe absolute colors is not so bad... 😆

dnil:
> It seems right now they will not fix the core issue with 5.2 in the 5.x, delaying it for v6 [twbs/bootstrap#36458](https://github.com/twbs/bootstrap/issues/36458). Hmm, maybe absolute colors is not so bad... 😆

Ah, actually he does have a dev branch for 5.2: https://github.com/vinorodrigues/bootstrap-dark-5/tree/dev/v1.2.0
So we could do that, but...

PR #3673, opened 2022-11-22
Fix #442: allow uploading reference info and exe ver files.

### LINKED ISSUE #442: Report versions of MIP, callers & references
It would be prudent in general to allow access to the version report for a case through scout. If nothing else, versions of MIP, crucial references such as ExAC, OMIM, genome, gene models (i e transcript set) and aligners/callers would be expected. This would probably fit well in the discussion of displaying deltas to reruns in #381.

dnil:
This PR adds a functionality: yaml version reports from the pipeline.

Before:
Screenshot 2022-11-22 at 15 56 46

Load supplied demo reports:
Screenshot 2022-11-22 at 15 57 53

After:
Screenshot 2022-11-22 at 16 36 53

YAML will be the native download. PDF export is added, and PDFs generated from the yaml now look like this:
Screenshot 2022-11-22 at 15 59 20

Screenshot 2022-11-22 at 15 59 54

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
I can review this one today if we want to include it in the new release

dnil:
No rush really, its been a few years in the thinking.. 😆 But we might as well. I'll look through yours as well; was testing the clinvar one when I noticed main at 4.61. The colorise could be included as well, right?

northwestwitch:
> No rush really, its been a few years in the thinking.. 😆 But we might as well. I'll look through yours as well; was testing the clinvar one when I noticed main at 4.61. The colorise could be included as well, right?

Yes, let's add the colorize one. The clinVar one is easy, so we might as well add it

dnil:
Ok, now fixes the CNV report on stage as well. It was all ready to just change the config! 😸 

Screenshot 2022-11-23 at 16 49 20

Still will need a load-from-config on hasta-stage as well though.

scout/constants/case_tags.py
```diff
@@ -3,8 +3,9 @@
 # keys are used in the load report cli, while key_name is saved in case database documents
 CUSTOM_CASE_REPORTS = {
     "delivery": {"key_name": "delivery_report", "format": "HTML", "pdf_export": True},
-    "cnv": {"key_name": "cnv_report", "format": "PDF", "pdf_export": True},
+    "cnv": {"key_name": "cnv_report", "format": "PDF", "pdf_export": False},
     "cov_qc": {"key_name": "coverage_qc_report", "format": "HTML", "pdf_export": True},
+    "exe_ver": {"key_name": "exe_ver", "format": "YAML", "pdf_export": True},
```
northwestwitch, line 8: The displayed report name could be more self explanatory here, look at the other report types:

image


```suggestion
    "exe_ver": {"key_name": "bioinfo_software", "format": "YAML", "pdf_export": True},
```

northwestwitch: Or pipeline software, ..?

docs/admin-guide/loading-case.md
```diff
@@ -75,6 +75,8 @@ Available types or case reports:
 - **delivery**: Delivery Report (HTML)
 - **cnv**: Copy Number Variants Report (PDF), available only for cancer cases
 - **cov_qc**: Coverage QC Report (HTML), available only for cancer cases
+- **exe_ver**: Pipeline detailed software versions (YAML)
+- **reference_info**: Pipeline detailed reference file versions (YAML)
```
northwestwitch, line 79: This key could also be abbreviated to `ref_info`, but it's fine also like this

scout/constants/case_tags.py
northwestwitch, line 8: The downloaded PDF version of this report (when you click on the right side button) doesn't look that good:

image

Perhaps just keep the other button and serve the yaml?

scout/constants/case_tags.py
```diff
@@ -13,6 +14,7 @@
         "format": "PDF",
         "pdf_export": False,
     },
+    "reference_info": {"key_name": "reference_info", "format": "YAML", "pdf_export": True},
```
northwestwitch, line 17: Also the PDF export of this one looks messy:

image

The easiest solution would be to just set pdf_export to False

northwestwitch (changes requested):
Nice and useful new feature! 🥇 

I have a few suggestion for improvement, easy ones. Most importantly the exported PDF reports don't render well.
Would also be nice to update the docs, [here](https://github.com/Clinical-Genomics/scout/blob/main/docs/admin-guide/loading-case.md#adding-custom-reports-to-the-case)

docs/admin-guide/loading-case.md
dnil: It just feels easiest to keep it consistent all the way through MIP-deliverables/Hermes/Housekeeper/CG/Scout, though CG/Hermes could certainly map it for us.

scout/constants/case_tags.py
dnil: Very valid point. I changed it to preformatted instead, so this:
Screenshot 2022-11-24 at 08 22 01

scout/constants/case_tags.py
dnil: Mm, I did the same here. It is still not super easy to read, but at least the ws is native:
Screenshot 2022-11-24 at 08 27 35
One can still dl the yaml if that is preferred.

docs/admin-guide/loading-case.md
northwestwitch: No, but then it's OK

scout/constants/case_tags.py
northwestwitch: Looks great now!

scout/adapter/mongo/case.py
```diff
@@ -1028,12 +1033,15 @@ def update_case(self, case_obj, keep_date=False):
         for key in [
```
northwestwitch, line 1033: Could this one be collected from the constants instead? So we avoid hardcoding every new key

scout/constants/case_tags.py
```diff
@@ -3,8 +3,9 @@
 # keys are used in the load report cli, while key_name is saved in case database documents
 CUSTOM_CASE_REPORTS = {
     "delivery": {"key_name": "delivery_report", "format": "HTML", "pdf_export": True},
-    "cnv": {"key_name": "cnv_report", "format": "PDF", "pdf_export": True},
+    "cnv": {"key_name": "cnv_report", "format": "PDF", "pdf_export": False},
     "cov_qc": {"key_name": "coverage_qc_report", "format": "HTML", "pdf_export": True},
+    "exe_ver": {"key_name": "pipeline_software_versions", "format": "YAML", "pdf_export": True},
```
northwestwitch, line 8: 👍🏻

northwestwitch (approved):
Looks great now!

northwestwitch:
> The keynames got a bit long for display. 😊 Are you still ok with "pipeline_version" and "reference_info"? We could also add display names to the `CUSTOM_CASE_REPORTS` - but maybe not until we need to...

Yes, the names you suggested are OK!

> We could also add display names to the `CUSTOM_CASE_REPORTS` - but maybe not until we need to...

nonono, I like it as it is!

PR #3670, opened 2022-11-22
Dashboard labels

dnil:
This PR fixes a small layout issue.

From 
Screenshot 2022-11-22 at 10 26 21
To 
Screenshot 2022-11-22 at 10 26 44

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the Pax software available at [https://pax.scilifelab.se/](https://pax.scilifelab.se). The resource you are going to call dibs on is `scout-stage` and the server is `cg-vm1`.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, log out from `cg-vm1` and log in again in the `hasta` server, repeat the `hasta` and `paxa` procedure, which will release the allocated resource (scout-stage) to be used for testing by other users.
</details>

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #3674, opened 2022-11-23
Revert "Dashboard labels"

dnil:
Reverts Clinical-Genomics/scout#3670
Accidentally got the 4.61 version PR on there anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,10 +26,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Non-admin users saving institute settings would clear loqusdb instance selection
 - Layout of variant position, cytoband and type in SV variant summary
-- Broken `Build Status - GitHub badge` on GitHub README page
+- Broken `Build Status - GitHub badge` on GitHub README page 
 - Visibility of text on grey badges in gene panels PDF exports
-- Labels for dashboard search controls
-
 
 ## [4.60]
 ### Added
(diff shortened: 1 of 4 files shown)


northwestwitch (commented):
Isn't it easier to just revert the header of the changelog to [unreleased] and the __version__.py to version = 4.60 ??

dnil:
A lot, but that would mean accepting a git history error - and making a new PR. It would have been done already if it wasn't for the ubuntu / setuptools change.

northwestwitch:
> A lot, but that would mean accepting a git history error - and making a new PR. It would have been done already if it wasn't for the ubuntu / setuptools change.

This one doesn't even pass the tests.. I can approve, it's fine, but I would go the easiest way

northwestwitch:
> A lot, but that would mean accepting a git history error

It's always good to accepts from one's errors 😆

dnil:
> It's always good to accept from one's errors 😆

Ah, yes, but we must not mix up our minds, hearts and souls with the git log.

northwestwitch (approved):
Looks OK! 👍🏻

dnil:
> > It's always good to accept from one's errors 😆
> 
> Ah, yes, but we must not mix up our minds, hearts and souls with the git log.

Importantly it will now go down in the log as having been acknowledged, reverted and hopefully learned from.

northwestwitch:
> Ah, yes, but we must not mix up our minds, hearts and souls with the git log.

Git log is a product of our minds and self indulgence is a mindful concept. So I see no problem in this :blossom:

PR #3681, opened 2022-11-23
Make Safe variant matching badge more visible on case page

northwestwitch:
Fix this badge on case page:

image

To be like this:

image

and this:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Do not export `Assertion method` and `Assertion method citation` to ClinVar submission files according to changes to ClinVar's submission spreadsheet templates.
 - Simplified code to create and download ClinVar CSV files
 - Colorize inheritance models badges by category on VariantS page
+- `Safe variants matching` badge more visible on case page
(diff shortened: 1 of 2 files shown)


northwestwitch:
> I see what you mean. In a way its more visible in dark mode as it is now, but on the light mode page this is both more consistent (matches with Prioritized e.g.) and more visible. Adapting might actually be best for both: one way would be to use `class=btn-info`, or keep badge as is and use "class=text-body" color for the foreground. Now that I think about it, wouldn't it also just work if you skip the fg/bg and go for badge-info directly?

I'll try!

northwestwitch:
No, badge-info doesn't work. I think it's bootstrap 4.

image

dnil:
> No, badge-info doesn't work. I think it's bootstrap 4.
> 
> image

Right you are; and it will be text-bg-info in bs 5.2. They do like to keep us on our toes.

dnil (approved):
💯

PR #3523, opened 2022-08-12
On diagnoses page, show number of OMIM terms available and datetime of last update

northwestwitch:
Fix #3241 

To achieve that, I've replaced the OMIM terms `_id` of type string (example OMIM:67655) with a pymongo ObjectID

From this:

image

To this:

image

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the paxa software installed on `hasta`:
 - Log in into hasta: `ssh <USER.NAME>@hasta.scilifelab.se`
 - Activate the staging environment with the command `us` (Use Stage)
 - Run the command `paxa` and follow the instructions. Make sure to specify scout-stage as the resource you request and the server cg-vm1 as server.
1. Log out from the hasta server.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, log out from `cg-vm1` and log in again in the `hasta` server, repeat the `hasta` and `paxa` procedure, which will release the allocated resource (scout-stage) to be used for testing by other users.
</details>

**How to test locally**:
1. Locally, run `scout setup demo`
2. Go to diagnoses page

**How to test on hasta and cg-vm1**:
1. Install the branch on cg-vm1 and surf the pages which might be affected by changes in OMIM terms document structure
- assign and removing OMIM terms from cases
- Go to OMIM diagnoses page (should crash and only show number of OMIM terms, no last update date)
- Create and edit phenomodels containing OMIM terms
2. Install the branch on hasta stage and update the OMIM terms
3. Repeat 1

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #3685, opened 2022-11-24
Try joblib..

dnil:

Test using joblib for loading variants.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Hihi, its not working, and its actually slower for small files. We can perhaps add a check for file size/variant count above some threshold to save a little time on small runs. But it is fun to see the different threads running. I think I might give it some more time tomorrow to better evaluate if it can be fixed to be really safe - and is worth the effort.

northwestwitch:
Out of curiosity, do tests pass on you local instance @dnil?

dnil:
> Out of curiosity, do tests pass on you local instance @dnil?

The failing ones do, but I'll restart a full sweep..

dnil:
> > Out of curiosity, do tests pass on you local instance @dnil?
> 
> The failing ones do, but I'll restart a full sweep..

```
FAILED tests/commands/export/test_export_variant_cmd.py::test_export_variants - AssertionError: assert 1 == 0
```

Nah, another one failed on local, and that might just be the order. There are more issues. A guess would be more "unique" things like that VCF object in the loop, that exist before it fans out. Or that are left for after.

dnil:
> ```
> FAILED tests/commands/export/test_export_variant_cmd.py::test_export_variants - AssertionError: assert 1 == 0
> ```
> 
> Nah, another one failed on local, and that might just be the order. There are more issues. A guess would be more "unique" things like that VCF object in the loop, that exist before it fans out. Or that are left for after.

You know what, that was actually a bugged function that the test didn't catch since it reads the variants in the same order every time. With another reader, the order was shook up an voila, exposed.

dnil:
Ah, found one - the variant_rank differs! 😊 Will fix!

dnil:
And then one where the test relied on pulling a random variant to modify would end up on the first page of the variants view. 🤷

dnil:
So, this is no super revolution so far; possibly due to IO constraints, possibly due to eh other limitations on the hasta head node: 
With one thread:
```
[daniel.nilsson@hasta:/home/proj/stage/housekeeper-bundles/justhusky/2022-06-08] [base] $ time scout load case -u scout_load_rna_multiqc.yaml -t 1 -c 1
real	14m47.344s
user	9m3.261s
sys	0m26.479s
```
Just threading the reader doesn't really seem to help here.
```
 time scout load case -u scout_load_rna_multiqc.yaml -t 1 -c 8
real	15m34.738s
user	9m57.206s
sys	0m32.295s
```

With a few loader threads, trivially parallel by chromosome:
 ```
time scout load case -u scout_load_rna_multiqc.yaml -t 8 -c 2
real	9m11.434s
user	9m16.059s
sys	1m8.683s
```
The process gets like 120% CPU vs 60% with the 1 thread. It's a bit more complex, but it shaves 7 mins off the load; thats in the ballpark of doubling the speed. I'm very sure there would be tweaks to stuff that needs to be done fewer times (e.g. in the compound update interval trees).

File: scout/commands/export/variant.py [hunk 1]
```diff
@@ -161,32 +161,32 @@ def get_vcf_entry(variant_obj, case_id=None):
     Returns:
         variant_string(str): string representing variant in vcf format
     """
-    if variant_obj["category"] == "snv":
+    if variant_obj.get("category") == "snv":
```
northwestwitch, line 164: Do we expect variants without category?

scout/commands/export/variant.py
```diff
@@ -161,32 +161,32 @@ def get_vcf_entry(variant_obj, case_id=None):
...
 
     variant_string = "{0}\t{1}\t{2}\t{3}\t{4}\t{5}\t{6}\t{7}".format(
         variant_obj["chromosome"],
         variant_obj["position"],
-        variant_obj["dbsnp_id"],
-        variant_obj["reference"],
-        variant_obj["alternative"],
-        variant_obj["quality"],
-        ";".join(variant_obj["filters"]),
+        variant_obj.get("dbsnp_id"),
```
northwestwitch, line 179: Here as well, was this export failing for one or more categories?

scout/commands/load/case.py
```diff
@@ -51,6 +48,22 @@
 @click.option("--peddy-ped", type=click.Path(exists=True), help="path to a peddy.ped file")
 @click.option("--peddy-sex", type=click.Path(exists=True), help="path to a sex_check.csv file")
 @click.option("--peddy-check", type=click.Path(exists=True), help="path to a ped_check.csv file")
+@click.option(
+    "-t",
+    "--threads",
+    type=int,
+    default=LOADER_THREADS,
```
northwestwitch, line 55: Nice to be able to customize it, also in the other functions

northwestwitch (commented):
Nice new feature! I have some question and suggestions. Haven't tried to load variants, since I believe you did and we'll need to test it properly on stage. I am mostly concerned with the complexity of the code, that wasn't very readable from before actually..

Check also the Vulture warnings, there are unused imports and passed params

scout/constants/__init__.py
```diff
@@ -90,6 +90,10 @@
     VERIFIED_VARIANTS_HEADER,
 )
 
+# Default number of threads for variant loading
+LOADER_THREADS = 8
```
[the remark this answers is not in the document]
dnil: They are mostly there as defaults, and would presumably just be aliased in, but it would arguably be nice to be able to set them in config as well. In that case we better use the old CLI-one for now. I'll add a suggestion!

scout/adapter/mongo/variant_loader.py
```diff
@@ -387,10 +398,15 @@ def _load_variants(
             nr_inserted(int)
         """
         build = build or "37"
-        genes = [gene_obj for gene_obj in self.all_genes(build=build)]
-        gene_to_panels = self.gene_to_panels(case_obj)
-        hgncid_to_gene = self.hgncid_to_gene(genes=genes, build=build)
-        genomic_intervals = self.get_coding_intervals(genes=genes, build=build)
+
+        if not genes:
```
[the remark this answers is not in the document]
dnil: I'm not the biggest fan of comments, but it should be possible to split out some more sub functions, with some nice names and docstrings as a comment substitute! The thing that mostly bothers me is the long list of parameters - it rather feels like a datastructure / class is missing.

scout/adapter/mongo/variant_loader.py
```diff
@@ -695,24 +713,78 @@ def load_variants(
...
-                institute_id=institute_id,
-                build=build,
-                rank_results_header=rank_results_header,
-                vep_header=vep_header,
-                category=category,
-                sample_info=sample_info,
-                custom_images=custom_images,
-                local_archive_info=local_archive_info,
-            )
+            if region == "":
```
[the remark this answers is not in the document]
dnil: Its more simple but slightly less efficient, as the variant vcf obj has to be reset an extra time. But it can certainly be made more plain! I'll give it a shot!

scout/adapter/mongo/variant_loader.py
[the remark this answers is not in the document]
northwestwitch: > it rather feels like a datastructure / class is missing.

Indeed!

northwestwitch:
Once in a while the tests of this one fail, the error due to another variant returned instead of the expected one.. 🤔

dnil:
> Once in a while the tests of this one fail, the error due to another variant returned instead of the expected one.. 🤔

Mm, could of course be an issue with the parallel load, but I would bet against it. I did correct several tests where it turned out we were just relying on the load to always return a certain variant as first in order. Could be that there is one or a few left that I just randomly didn't find. Will check later!

scout/commands/load/case.py
```diff
@@ -64,11 +80,12 @@ def case(
     ped,
     update,
     config,
-    no_variants,
```
dnil, line 67: I have never used this option, the organisation does not use it and it does nothing. Looks like it was a leftover from a dev process to implement a test for loading cases without variants? See
https://github.com/Clinical-Genomics/scout/commit/c5ee18c9bd431d2a8b488844470cf3a095090091

northwestwitch: I'm not surprised, we have quite a number of those!

dnil:
And did you see, "All checks have passed". Next time things calm down again I'll add suggestions and refactor a little bit more. Never give up.. 😝 
Screenshot 2023-10-25 at 16 06 19

dnil:
We are going to get serious conflicts on this now with changes to file loading for first `nf-core/raredisease`, then `tomte`. Closing this stubborn test with the conclusion that it is feasible to at least somewhat speed up loading with trivial parallelisation on chromosome even on the current setup, and even more so if we were to spin loading out to cluster nodes instead of running it on a head node (we'd have to fix e.g. db connectivity, but..). The biggest hurdle is really the tests with variant load order dependence. Those are silly, and should be changed, but it looks like it will involve a bit of a pain to do it.

northwestwitch:
It stayed open all that time, perhaps longer than #3523

image

PR #3701, opened 2022-12-05
Reset and change placeholder for conditions in ClinVar when condition ID changes

### LINKED ISSUE #3699: Set condition_type in ClinVar upload based on condition terms used
For adding a clinvar submission, if an OMIM diagnosis is selected for the case, it is not possible to add other types of phenotype term, like picking an HPO instead. It would make sense to lock in the condition type based on that instead then, since there is a strict if clause anyway. 

Screenshot 2022-12-05 at 09 04 17

As it is, the user must take care not to do this, which would fail verification.
Screenshot 2022-12-05 at 09 13 07

northwestwitch:
Fix #3699 
When condition ID changes in ClinVar form:
- Reset the list of conditions already provided
- Modify conditions placeholder to reflect the format of the condition ID

I'm not adding any validation that the provided values for the conditions are consistent with the condition ID type, just a simpler explanation of the expected values. On the other hand all the other values in this form are not validated either. **Validation of these values will be performed in the future by the ClinVar API**, which will return specific error messages when validation fails.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/server/blueprints/clinvar/controllers.py
+++ b/scout/server/blueprints/clinvar/controllers.py
@@ -56,7 +56,7 @@ def _set_var_form_common_fields(var_form, variant_obj, case_obj):
         for hpo in case_obj.get("phenotype_terms", [])
     ]
     var_form.omim_terms.choices = [
-        (omim.get("disease_id"), " - ".join([omim.get("disease_id"), omim.get("description")]))
+        (omim.get("disease_nr"), " - ".join([str(omim.get("disease_nr")), omim.get("description")]))
         for omim in case_obj.get("diagnosis_phenotypes", [])
     ]
 
--- a/scout/server/blueprints/clinvar/form.py
+++ b/scout/server/blueprints/clinvar/form.py
@@ -60,7 +60,7 @@ class ClinVarVariantForm(FlaskForm):
     condition_type = SelectField(
         "Condition ID type", choices=[(key, value) for key, value in PHENO_DBS.items()]
     )
-    conditions = SelectMultipleField("Condition ID values, comma separated)")
+    conditions = SelectMultipleField("Condition ID values, comma-separated)")
 
     # Extra fields:
     assertion_method = StringField("Assertion method", default=ASSERTION_METHOD)
--- a/scout/server/blueprints/clinvar/form.py
+++ b/scout/server/blueprints/clinvar/form.py
@@ -58,9 +58,9 @@ class ClinVarVariantForm(FlaskForm):
     omim_terms = MultiCheckboxField("Case-associated OMIM terms", choices=[])
     condition_comment = TextAreaField("Additional comments describing condition")
     condition_type = SelectField(
-        "Condition type", choices=[(key, value) for key, value in PHENO_DBS.items()]
+        "Condition ID type", choices=[(key, value) for key, value in PHENO_DBS.items()]
     )
-    conditions = SelectMultipleField("List of conditions, comma-separated)")
+    conditions = SelectMultipleField("Condition ID values, comma separated)")
 
     # Extra fields:
     assertion_method = StringField("Assertion method", default=ASSERTION_METHOD)
--- a/scout/server/blueprints/clinvar/templates/multistep_add_variant.html
+++ b/scout/server/blueprints/clinvar/templates/multistep_add_variant.html
@@ -206,7 +206,7 @@
                         {% for term, _ in variant_data.var_form.omim_terms.choices %}
                           <option value="{{term}}" selected>{{term}}</option>
                         {% endfor %}
-                      {% elif variant_data.var_form.hpo_terms.choices%}
+                      {% elif variant_data.var_form.hpo_terms.choices %}
                         {% for term, _ in variant_data.var_form.hpo_terms.choices %}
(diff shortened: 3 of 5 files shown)


scout/server/blueprints/clinvar/controllers.py
```diff
@@ -56,7 +56,7 @@ def _set_var_form_common_fields(var_form, variant_obj, case_obj):
         for hpo in case_obj.get("phenotype_terms", [])
     ]
     var_form.omim_terms.choices = [
-        (omim.get("disease_id"), " - ".join([omim.get("disease_id"), omim.get("description")]))
+        (omim.get("disease_nr"), " - ".join([str(omim.get("disease_nr")), omim.get("description")]))
```
northwestwitch, line 59: So OMIM values will become xxx, yyy instead of OMIM:xxx, OMIM:yyy

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
```diff
@@ -184,7 +183,7 @@ <h3 class="fs-subtitle">The interpretation, or clinical significance, of the var
...
+                <h3 class="fs-subtitle">The condition for which the variant is interpreted. Examples available at <a href="https://www.ncbi.nlm.nih.gov/clinvar/docs/spreadsheet/#condition" target="_blank" rel="noopener">ClinVar</a>"</h3>
```
northwestwitch, line 186: This is an easier explanation to the conditions formats

scout/server/blueprints/clinvar/form.py
```diff
@@ -58,9 +58,9 @@ class ClinVarVariantForm(FlaskForm):
     omim_terms = MultiCheckboxField("Case-associated OMIM terms", choices=[])
     condition_comment = TextAreaField("Additional comments describing condition")
     condition_type = SelectField(
-        "Condition type", choices=[(key, value) for key, value in PHENO_DBS.items()]
+        "Condition ID type", choices=[(key, value) for key, value in PHENO_DBS.items()]
     )
-    conditions = SelectMultipleField("List of conditions, comma-separated)")
+    conditions = SelectMultipleField("Condition ID values, comma-separated)")
```
dnil, line 63: 👍

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
```diff
@@ -375,12 +374,25 @@ <h3 class="fs-subtitle">Information provided by filling in these fields will be
 	});
 });
 
+var conditionsPlaceHolders = {"OMIM": "e.g. 100800, ..", "HPO": "e.g. HP:0002839, ..", "MedGen": "e.g. C0001080, ..", "MeSh": "e.g. D000130, ..", "MONDO": "e.g. MONDO:0007947, ..", "Orphanet": "e.g. ORPHA155, .."};
```
dnil, line 377: 👍

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
```diff
@@ -375,12 +374,25 @@ <h3 class="fs-subtitle">Information provided by filling in these fields will be
...
+
+// reset and modify conditions field's placeholder when condition ID type changes
+$(function(){
+  $("#condition_type").change(function(){
+    $('#condition_tags').val(null).trigger('change');
+    selectedCondId = $("#condition_type option:selected").text();
+    $('#condition_tags').select2({
+      placeholder: conditionsPlaceHolders[selectedCondId]
+    });
+  });
```
dnil, line 395: Now all that is missing is an update of the actual choices? 😉 
Its fine without as well: if one has OMIM terms, it is more than reasonable to use them by default rather than HPO terms for like for an unknown disorder. That is presumably kind of why ClinVar only accept one kind anyway?

But as the placeholders look like one can swap, I can imagine someone being disappointed they can't just switch back and forth between OMIM and HPO:

Screenshot 2022-12-05 at 13 36 47

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
dnil: 👍

dnil (changes requested):
Super, almost there! 

1. 🕶️ 
Screenshot 2022-12-05 at 13 34 31
1. The dynamic placeholder and content clear on the conditions select is super cool, but would be even better if the actual options also updated.. 😁 This one is very optional.

scout/server/blueprints/clinvar/templates/multistep_add_variant.html
northwestwitch: I've been try to fix it as you suggest, but still doesn't work. Changing values of the select2 lib elements is not so straightforward as modifying a select. I'll try again tomorrow.

northwestwitch: Ok, now it should work as intended!

PR #3666, opened 2022-11-14
Remove Assertion Method fields from ClinVar submission files

northwestwitch:
Fix #3642 - it's mid November!

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the Pax software available at [https://pax.scilifelab.se/](https://pax.scilifelab.se). The resource you are going to call dibs on is `scout-stage` and the server is `cg-vm1`.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, log out from `cg-vm1` and log in again in the `hasta` server, repeat the `hasta` and `paxa` procedure, which will release the allocated resource (scout-stage) to be used for testing by other users.
</details>

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [X] code approved by DNIL
- [x] tests executed by CR

PR #3716, opened 2022-12-13
Only ACMG terms in ClinVar clinsig when user selects ACMG 2015 as assertion criteria

### LINKED ISSUE #3715: LImit variant classif in ClinVar to ACMG terms if user selects ACMG assertion criteria
I've received an email with feedback from a test submission using the API (I guess there is still manual inspection when they see that a submission, even if it's a test one, returns error).

Reason why the submission failed:
- It is not expected to find an rs# in the Accession field (To describe the variant
please provide the assembly, chromosome, chromosome coordinate, and reference and alternate alleles.)

Additionally:
- the ACMG Guidelines, 2015 do not include 'protective' as a classification term. Therefore, please
exclude the ACMG guidelines as your assertion criteria.

Honestly I just selected a random term in order to test the process, but this comment is useful. Perhaps we should start by limiting the available classification options in that select to what's available in the ACMG? That would be less of a headache for the users!

northwestwitch:
Fix #3715. Fixes also the order of the items in that select, which now is random.

### Default assertion criteria is ACMG, so clinsig will just show the ACMG criteria by default:

image

### When user removes the ACMG assertion method citation field:

image

### Then all clinsig options will be shown:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/clinvar/form.py
+++ b/scout/server/blueprints/clinvar/form.py
@@ -50,7 +50,9 @@ class ClinVarVariantForm(FlaskForm):
         "Inheritance model",
         choices=[("", "-")] + [(item, item) for item in CLINVAR_INHERITANCE_MODELS],
     )
-    clinsig = SelectField("Clinical Significance", choices=[(item, item) for item in CLNSIG_TERMS])
+    clinsig = SelectField(
+        "Clinical Significance", choices=[(item, item) for item in CLNSIG_TERMS[:5]]
+    )
     clinsig_comment = TextAreaField("Comment on clinical significance")
     clinsig_cit = TextAreaField("Clinical significance citations (with identifier)")
     last_evaluated = DateField("Date evaluated")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Source link out for MIP 11.1 reference STR annotation
 - Avoid duplicate causatives and pinned variants
+- ClinVar clinical significance displays only the ACMG terms when user selects ACMG 2015 as assertion criteria
 
 
 ## [4.61.1]
--- a/scout/constants/clinvar.py
+++ b/scout/constants/clinvar.py
@@ -74,7 +74,7 @@ CASEDATA_HEADER = {
 # The following items are defined in the ClinVar API specifications:
 # https://www.ncbi.nlm.nih.gov/clinvar/docs/api_http/
 ####################################################################
-CLNSIG_TERMS = {
+CLNSIG_TERMS = [
     "Pathogenic",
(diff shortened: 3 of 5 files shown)


scout/server/blueprints/clinvar/form.py
```diff
@@ -50,7 +50,9 @@ class ClinVarVariantForm(FlaskForm):
         "Inheritance model",
         choices=[("", "-")] + [(item, item) for item in CLINVAR_INHERITANCE_MODELS],
     )
-    clinsig = SelectField("Clinical Significance", choices=[(item, item) for item in CLNSIG_TERMS])
+    clinsig = SelectField(
+        "Clinical Significance", choices=[(item, item) for item in CLNSIG_TERMS[:5]]
```
dnil, line 54: If a user adds several variants in the same submission batch, and sets different submission criteria, how will these be handled by preClinVar? 😊

dnil (approved):
Nice touch! 🙌

scout/server/blueprints/clinvar/form.py
northwestwitch: You can't have submissions with more than one criterion, it's one of the changes introduced in the API last month. From the email that they sent to the submitters:


Dear ClinVar API Submitter,

Thank you for using the ClinVar Submission API! We are making changes to the API submission schema to streamline the submission process:

* We will move 'assertionCriteria' to the top level. Note that only one assertion criteria (a citation or a file) may be provided for a submission. This assertion criteria will be applied to all records in the submission.

northwestwitch: Forgot to add that this is valid also for the submissions via file. PreClinVar handles it fine, in the sense that collects the assertion criteria from the extra args sent by the request, that in Scout are in turn collected from the first variant of the submission [code](https://github.com/Clinical-Genomics/scout/blob/f4b4064591f2888250d0e9aadc42f371017c84c2/scout/server/blueprints/clinvar/controllers.py#L344)

dnil: For file I'm thinking the user is responsible and can edit their spreadsheet as they wish. 

For api, the first in the sub is quite acceptable. The ideal representation would naturally be to lock the criteria and citation edit plus the select once the first variant is in the sub. That is load the select with data from the current submission if one exists instead of defaulting it to ACMG. But don't worry - I have a feeling for most it will not matter, they will just pick the same. 😸

northwestwitch: We removed the relative columns from the spreadsheet in #3666, by your suggestion, just saying.. 😆

Issue #3658, opened 2022-11-07
New RNA reports in Scout


northwestwitch:
RNA data already supported by Scout:
- A pdf report produced by Arriba, loaded at the case level, ~~with the key `gene_fusion_report ` (or `gene_fusion_report_research ` if it's based on research data)~~ Suggested name: `arriba_report`
- `splice_junctions_bed` and `rna_coverage_bigwig` -> igv tracks saved at the case.individual level

New reports and data to be included:
- [ ] Another fusion report file in HTML format produced by FusionReport (this is the fusion summary that collects metrics from different methods: arriba, pizzly, starfusion, fusioncatcher, squid). This should be the primary summary report. Suggested name: `fusion_report`. 
- [ ] Another fusion report HTML file produced by FusionInspector (this provides a detailed visualization of all fusions found). Suggested name: `fusion_inspector`. 
- [ ] RNA multiqc file, also HTML file, work in progress, see #3650 
- [ ] .tsv file from Fusion Inspector -> convert into variants-like data and load as if they were scout variants

fevac:
I am uncertain about the names for each report. To me `fusion_report` and `fusion_inspector` are the most important/informative reports since they collect the information from several tools. I would suggest renaming the report produced by Arriba, is this something you want to avoid (as this is already implemented)?

dnil:
Yeah, exactly: I’m afraid changing it retroactively for all the already loaded ones is a bit of a pain. It’s doable, but if it’s just for the name we would rather spend the effort where it counts. 

Perhaps the pipeline or programs used to produce the reports have more names / parts to them?

northwestwitch:
To rename the key `gene_fusion_report` and `gene_fusion_report_research` to `arriba_report_something` one could always write a script to add a new key/value to the cases and then remove the old keys/values. This is totally doable for us, but what if other facilities started to use those key/values as well? We would mess up with their data..

fevac:
ok, then, what about:
- leave arriba as: `gene_fusion_report`
- fusion report: `gene_fusion_summary`
- fusion inspector: `gene_fusion_inspector`

@jemten what do you think?

dnil:
It would not hurt with a little "RNA" in the key. The pipeline is named `nf-core/rnafusion`, so perhaps 
they could be `RNAfusion_report` and `RNAfusion_inspector`?

fevac:
Good point. I like that.

northwestwitch:
[...]

Issue #3648, opened 2022-10-28
RNA cases from nf-core/rnafusion


pbiology:
The aim is to implement nf-core/rnafusion in production at CG Stockholm.
The output from this is for now a fusion-report and multiQC.

If we could view RNA cases as separate from DNA (but perhaps with a link to existing DNA case) this would be a good start.

Example of rnafusion case is `soldifawn`
Files can be found at `/home/proj/stage/rnafusion/cases/solidfawn`

dnil:
Three parts I guess, though the first two have the lowest hanging fruit.

1. As discussed, for the ones with existing DNA cases, the easiest and likely best value solution is to just add the reports to the case, much as with reports from MIP RNA. Let us know the keys and we can fix / guide you to fix!

2. For cases from either RNA pipe, allowing scout to load a case without VCFs can be useful when DNA was not available. We would then initially only show reports and IGV tracks from the case page. This would have to be coordinated with CG producing appropriate scout load configs, but let's start that one in Scout.

3. If the pipeline can provide some kind of VCF (perhaps an SV VCF like one with BNDs for fusions) we could load that one as well. This would be great, since users could pin, mark and assess the variants the usual way in Scout, not just view the report.

pbiology:
I don't think there is any VCF file produced unfortunately. And I guess it's not entirely uncomplicated to add once since the output is an aggregate of multiple callers. So one would have to make some kind of decisions on which fusions to include in such a file.

But perhaps that is something to think about to do downstream in a CG specific add-on or something like that

northwestwitch:
> 2. For cases from either RNA pipe, allowing scout to load a case without VCFs can be useful when DNA was not available. We would then initially only show reports and IGV tracks from the case page. This would have to be coordinated with CG producing appropriate scout load configs, but let's start that one in Scout.

I've tested now locally and it's already possible to load a case without any associated DNA VCF files. So this is already covered

dnil:
🥳 
![Screenshot 2023-02-03 at 15 04 56](https://user-images.githubusercontent.com/758570/216623174-d6dc3f7c-0207-4c0b-95e1-3b975cf03c95.png)

PR #3747, opened 2023-01-03
Add rnafusion reports

fevac:
This PR adds support for RNAfusion specific reports (`RNAfusion_report` and `RNAfusion_inspector`). 

Related to: https://github.com/Clinical-Genomics/scout/issues/3658 and https://github.com/Clinical-Genomics/scout/issues/3648

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/constants/case_tags.py
+++ b/scout/constants/case_tags.py
@@ -15,7 +15,11 @@ CUSTOM_CASE_REPORTS = {
         "pdf_export": False,
     },
     "reference_info": {"key_name": "reference_info", "format": "YAML", "pdf_export": True},
-    "RNAfusion_inspector": {"key_name": "RNAfusion_inspector", "format": "HTML", "pdf_export": False},
+    "RNAfusion_inspector": {
+        "key_name": "RNAfusion_inspector",
+        "format": "HTML",
+        "pdf_export": False,
+    },
     "RNAfusion_report": {"key_name": "RNAfusion_report", "format": "HTML", "pdf_export": False},
 }
 
--- a/scout/constants/case_tags.py
+++ b/scout/constants/case_tags.py
@@ -15,6 +15,8 @@ CUSTOM_CASE_REPORTS = {
         "pdf_export": False,
     },
     "reference_info": {"key_name": "reference_info", "format": "YAML", "pdf_export": True},
+    "RNAfusion_inspector": {"key_name": "RNAfusion_inspector", "format": "HTML", "pdf_export": False},
+    "RNAfusion_report": {"key_name": "RNAfusion_report", "format": "HTML", "pdf_export": False},
 }
 
 CASE_REPORT_CASE_FEATURES = [


dnil (commented):
Thank you - good start! Could you also update the docs (https://github.com/Clinical-Genomics/scout/blob/main/docs/admin-guide/load-config.md#configuration-parameters) to highlight the differences from `gene_fusion` and `gene_fusion_research`!

Although this new dict covers quite a lot I'm afraid we also still have a few explicit mentions of the report keys in other functions. If you like, have a search for e.g. "gene_fusion" and see which ones you would update. Or just shout and we try to fix!

fevac:
@dnil I see that for gene fusion reports produced by Arriba there is a difference between `clinical` and `research`. The first one for clinically relevant fusions (subset of all detected fusions) and the second one for all detected fusions. Currently the rnafusion pipeline does not filter for clinically relevant fusions and I haven't heard any discussions about this either. Do you think it's ok to only include the `RNAfusion_inspector` and `RNAfusion_report` to add all detected fusions? Or should we also include a `_research` version of those to keep consistency?

dnil:
> @dnil I see that for gene fusion reports produced by Arriba there is a difference between `clinical` and `research`. The first one for clinically relevant fusions (subset of all detected fusions) and the second one for all detected fusions. Currently the rnafusion pipeline does not filter for clinically relevant fusions and I haven't heard any discussions about this either. Do you think it's ok to only include the `RNAfusion_inspector` and `RNAfusion_report` to add all detected fusions? Or should we also include a `_research` version of those to keep consistency?

It depends a bit on if you do clinical relevance filtering in the pipeline or not I guess, and if the underlying data is whole transcriptome or a subset capture? If full transcriptome, I can answer for constitutional cases (RD) that we would very much not like to see any fusions that do not involve at least one partner that is on the clinical gene panel for the case. Ideally actually on the currently selected panel, but hey, one step at the time; we will need to import structured data, not PDF for that.

I think I speak also for the somatic side saying that they would not like to see fusions between genes that have never been described in a relevant somatic disorder. That would best case just take time, worst case warrant an in depth investigation into incidental findings. 

So, research and clinical seem like a good starting point!

scout/demo/rnafusion.load_config.yaml
```diff
@@ -0,0 +1,22 @@
+---
+
+owner: cust000
```
northwestwitch, line 3: I'm giving this demo track a try and it loads! 🥇 

image

Perhaps we don't worry about the HTML files for now, they are not created by Scout anyway!

northwestwitch (commented):
Would be nice to assign a type of sequencing, so the cases with these reports would stick out when you look all cases of an institute.

Now the look like this on cases page:

image

scout/demo/rnafusion.load_config.yaml
dnil: Agreed - I've flagged them down in Sonarclound for now, but we should alert the rna-fusion devs as well - or is that you now @fevac?

dnil:
> Would be nice to assign a type of sequencing, so the cases with these reports would stick out when you look all cases of an institute.
> 
> Now the look like this on cases page:
> 
> image

Yes, let's! What is it called in StatusDB / CG-web / Orderportalen? Simply "RNA"?

fevac:
This is weird, I did add a sequencing type that I see. I just added colours now but I'm not sure why you don't see it at all:

image

northwestwitch:
> Yes, let's! What is it called in StatusDB / CG-web / Orderportalen? Simply "RNA"?

One of these? https://github.com/Clinical-Genomics/cg/blob/5ce32b0f971d309869e27a36aa68a585a7b8d859/cg/constants/constants.py#L72

scout/demo/rnafusion.load_config.yaml
fevac: The thing is that the nf-core/rnafusion pipeline is using other external tools to create these reports.  So I don't think the rnafusion developers will be the ones addressing this but instead the developers of those external tools. I could have a look at their repos and make an issue in those.

northwestwitch:
> This is weird, I did add a sequencing type that I see. I just added colours now but I'm not sure why you don't see it at all:
> 
> image

AHA! Do you see what I see??? 😆 

image


We need to fix the style in light mode!

dnil:
I mean, who runs light mode anyway? 😜

northwestwitch:
> I mean, who runs light mode anyway? 😜

I'm sure it's just the 2 of you to use it in dark mode. But good, so we see bugs in both styles

henrikstranneheim:
> > Yes, let's! What is it called in StatusDB / CG-web / Orderportalen? Simply "RNA"?
> 
> One of these? Wts perhaps? https://github.com/Clinical-Genomics/cg/blob/5ce32b0f971d309869e27a36aa68a585a7b8d859/cg/constants/constants.py#L72

I think that one is more LIMS oriented. SequencingMethod in https://github.com/Clinical-Genomics/cg/blob/5ce32b0f971d309869e27a36aa68a585a7b8d859/cg/constants/sequencing.py is perhaps a better one. Unfortunaltely, the code in cg mixes these concepts a bit, which is very confusing. However, It is still going to be `wts` I guess.

northwestwitch:
Fixed! 
image

I took the liberty to push to this branch @fevac

scout/constants/case_tags.py
```diff
@@ -1,4 +1,4 @@
-ANALYSIS_TYPES = ("wgs", "wes", "mixed", "unknown", "panel", "panel-umi", "external")
+ANALYSIS_TYPES = ("wgs", "wes", "mixed", "unknown", "panel", "panel-umi", "external", "wts")
```
northwestwitch, line 1: Yeah, it was here as well!

northwestwitch (approved):
I definitely approve! 🥇

scout/adapter/mongo/case.py
```diff
@@ -454,6 +454,10 @@ def rna_cases(self, owner):
                 {"gene_fusion_report_research": EXISTS_NOT_NULL},
                 {"individuals.splice_junctions_bed": EXISTS_NOT_NULL},
                 {"individuals.rna_coverage_bigwig": EXISTS_NOT_NULL},
+                {"RNAfusion_report": EXISTS_NOT_NULL},
```
northwestwitch, line 457: 👏🏻

PR #3754, opened 2023-01-13
Manual rank updates: add risk factor, show on matching variants in other cases

### LINKED ISSUE #2414: Classification category for Risk Factor (RF), also visible on variantS page
image

Can we have a grey button next to the other 5 ACMG Classification Buttons (on the Variant page under "Bascics") and for SV-variants as an option in the Variant tag drop down list? This would make it easy to annotate a variant as a "Risk Factor" for disease (for instance if a variant increases cancer risk by a few percent above background risk, this should be allowed within the ACMG-guidelines, as far as we understand in Lund). If you agree the button could be grey (when pressed) and have the letters RF on it.
What do you think? All the best Tord

dnil:

Fix #2414. 

Adds options for risk factors in line with ACMG recommendations to keep this outside the classification, and consider separating into different categories for the strength of evidence for the risk increase (RF, likely RF and uncertain RF).
Screenshot 2023-01-13 at 17 39 09

These are shown as usual manual ranks, e.g.:
Screenshot 2023-01-13 at 17 38 55

In order to make them more useful, this PR allows to show them on matching variants from other cases, on the variantS list:
Screenshot 2023-01-13 at 17 37 12

And then with details on the variant page:
Screenshot 2023-01-13 at 17 38 26

All pretty much as for the matching tiered variants from the cancer side.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. classify a variant shared between two cases as a risk factor (or other manual rank)
2. see it displayed on the case
3. see it displayed on the matching case

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -295,6 +295,24 @@ CANCER_TIER_OPTIONS = {
 }
 
 MANUAL_RANK_OPTIONS = {
+    12: {
+        "label": "URF",
+        "name": "Uncertain Risk Factor",
+        "description": "Uncertain risk allele - uncertain evidence for a small risk increase",
+        "label_class": "default",
+    },
+    11: {
+        "label": "LRF",
+        "name": "Likely Risk Factor",
+        "description": "Likely risk allele - some evidence for a small risk increase",
+        "label_class": "default",
+    },
+    10: {
+        "label": "RF",
+        "name": "Risk Factor",
+        "description": "Established risk allele - strong evidence for a small risk increase",
+        "label_class": "default",
+    },
     9: {
         "label": "VUS",
         "name": "Unknown Significance",
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -386,6 +386,24 @@ MANUAL_RANK_OPTIONS = OrderedDict(
             },
         ),
         (
+            13,
+            {
+                "label": "PLP",
+                "name": "Pathogenic Low Penetrance",
+                "description": "Pathogenic allele with low or reduced penetrance - robust evidence for an intermediate risk increase",
+                "label_class": "warning",
+            },
+        ),
+        (
+            14,
+            {
+                "label": "LPLP",
+                "name": "Likely Pathogenic Low Penetrance",
+                "description": "Likely pathogenic allele with low or reduced penetrance - some evidence for an intermediate risk increase",
+                "label_class": "warning",
+            },
+        ),
+        (
             10,
             {
                 "label": "RF",
--- a/scout/constants/variant_tags.py
+++ b/scout/constants/variant_tags.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from .gene_tags import INHERITANCE_PALETTE
 
 VARIANT_REPORT_VARIANT_FEATURES = [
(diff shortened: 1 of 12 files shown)


dnil:
Note that one may want to keep track of variants with new tag levels on them while testing, since lookup to the constant dicts are not under try except and will crash if one reverts to a version that does not have the new ones.

scout/constants/variant_tags.py
```diff
@@ -295,6 +295,24 @@
 }
 
 MANUAL_RANK_OPTIONS = {
```
northwestwitch, line 297: I was just comparing this list with the [clinsigs recommended](https://www.ncbi.nlm.nih.gov/clinvar/docs/clinsig/) by ClinVar, just for fun.

Why not including all criteria from ClinGen instead, it's more like a concept, instead of 3 useful but random terms from it.. Some of these terms make more sense to me than others from our list :)

image

scout/constants/variant_tags.py
```diff
@@ -295,6 +295,24 @@
...
+        "label": "LRF",
+        "name": "Likely Risk Factor",
+        "description": "Likely risk allele - some evidence for a small risk increase",
+        "label_class": "default",
+    },
+    10: {
+        "label": "RF",
+        "name": "Risk Factor",
+        "description": "Established risk allele - strong evidence for a small risk increase",
+        "label_class": "default",
```
northwestwitch, line 314: Perhaps here I would also do something about the order in which these terms are shown in the select.

image

 Risk factor tags are not tags used a lot (otherwise they would already be there) so perhaps shown them on the bottom or something? To do so one could convert MANUAL_RANK_OPTIONS to an OrderedDict

scout/server/blueprints/variants/templates/variants/indicators.html
```diff
@@ -51,6 +51,17 @@
   {% endif %} <!-- end of if variant.matching_tiered -->
 {% endmacro %}
 
+{% macro matching_manual_rank(variant) %}
+  {% if variant.matching_ranked %}
+    <span class="popovers badge bg-dark" data-bs-toggle="popover" data-bs-html="true" title=""
```
northwestwitch, line 56: The popover doesn't seem to work for me? Tested locally with the 2 RD demo cases

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -72,7 +72,7 @@
     </div>
     {% endif %}
 
-    <div class="row"><div class="col-12">{{ matching_variants(managed_variant, causatives) }}</div></div>
+    <div class="row"><div class="col-12">{{ matching_variants(managed_variant, causatives, variant.matching_ranked) }}</div></div>
```
northwestwitch, line 75: This works but the alignment is off

image

northwestwitch (changes requested):
Nice and useful fix that will be very appreciated by those who requested it (and hopefully others) 🥇 

I've noticed a couple of this that don't work as intended yet (small things, that's why I request changes). Check also my comments for the suggestion to include all ClinGen terms!

scout/server/blueprints/variant/templates/variant/variant.html
dnil: Ah! Same as the Tiered, but looks a little odd once you see it. I'll see what can be done.

dnil: Like so:
Screenshot 2023-01-31 at 11 15 46

scout/server/blueprints/variants/templates/variants/indicators.html
dnil: Hm, I so convinced it worked before, but if I could have few bucks for each of those... 😄 Now it works, hopefully again. 😸 
Screenshot 2023-01-31 at 11 15 13

scout/constants/variant_tags.py
dnil: Good call!
Screenshot 2023-01-31 at 12 08 17

scout/constants/variant_tags.py
dnil: Good idea re-checking. This "system" we have is for historical and practical reasons modelled after ESHG and DeCipher in particular, by way of the aCGH classifications we use for SVs. The risk factors are not included from there though, rather they are picked a bit ad hoc from ClinVar and the ACMG/AMP variant classification system body text (they are not in their recommendation as such yet). There is definitely merit to naming them "risk alleles" as ClinGen do; that would be my formal go-to. As the issue also asked for "risk factor", which is also what most people say, I stuck with it, and used the ClinGen language in the description to clarify.

The additional "low penetrance" ones could potentially be useful: they are not called for, but the ClinGen make a good recommendation for them in https://clinicalgenome.org/site/assets/files/4531/clingenrisk_terminology_recomendations-final-02_18_20.pdf. Let's include them!

scout/server/blueprints/variants/templates/variants/indicators.html
dnil: This could be a little confusing with new terms I guess. Here's a verbose version:
Screenshot 2023-01-31 at 12 36 14

scout/constants/variant_tags.py
```diff
@@ -294,74 +296,151 @@
...
-        "description": "Known pathogenic, previously known pathogenic in ClinVar, HGMD, literature, etc",
-        "label_class": "danger",
-    },
-    7: {
-        "label": "P",
-        "name": "Pathogenic",
-        "description": (
-            "Pathogenic, novel mutation but overlapping phenotype with known pathogenic, "
-            "no further experimental validation needed"
+MANUAL_RANK_OPTIONS = OrderedDict(
```
northwestwitch, line 299: 🚀

northwestwitch (approved):
Looks good to me now. We're going to check it again anyway before the new release. but let's merge I think! 💯

PR #3772, opened 2023-01-19
Fix causatives page

### LINKED ISSUE #3771: Causatives page mixes in matching causatives
At some point it appears we started pulling matching causative onto the CausativeS page. Can be verified locally with e.g. two cases with the same variants.
Screenshot 2023-01-18 at 20 38 26

northwestwitch:
Fix #3771  - Show only causative and not matching causatives

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Locally, with the 2 RD demo cases.
2. On main branch, go to a variant of a case and mark it as causative
3. Go to causatives page and check that the same variant from the other case is shown (also if it's not causative)
4. Repeat on this branch

**Expected outcome**:
This branch should only show 1 variant on the Causatives page (the original causative)
 
**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -512,7 +512,7 @@ class VariantHandler(VariantLoader):
             limit_genes (list): list of gene hgnc_ids to limit the search to
 
         Returns:
-            causatives(list): a list of variant objects
+            a list of variant objects
         """
         causatives = []
         causative_events = self.event_collection.find(
@@ -553,44 +553,29 @@ class VariantHandler(VariantLoader):
             causatives(list[Variant])
         """
         causatives = []
-
-        var_causative_events = self.event_collection.find(
+        causative_events = self.event_collection.find(
             {
                 "institute": case_obj["owner"],
                 "verb": {"$in": ["mark_causative", "mark_partial_causative"]},
-                "category": "variant",
+                "category": "case",
             }
         )
 
-        for var_event in var_causative_events:
-            if var_event["case"] == case_obj["_id"]:
-                # exclude causatives from the same case
-                continue
-            other_case = self.case(var_event["case"])
-            if other_case is None:
-                # Other variant belongs to a case that doesn't exist any more
-                continue
-            other_link = var_event["link"]
-            # link contains other variant ID
-            other_causative_id = other_link.split("/")[-1]  # md5-key _id of a variant
+        for case_event in causative_events:
 
-            if (
-                other_causative_id not in other_case.get("causatives", [])
-                and other_causative_id not in other_case.get("partial_causatives", {}).keys()
-            ):
+            causative_ids = set()
+            other_case = self.case(case_event.get("case"))
+
+            if other_case is None or other_case["_id"] == case_obj["_id"]:
                 continue
 
-            other_var_simple = var_event.get("subject").split("_")[
-                0:4
-            ]  # example: [ "17", "7577559", "G" "A"]
+            for var_id in other_case.get("causatives", []):
+                causative_ids.add(var_id)
 
-            positional_variant_ids = set()
-            for variant_type in ["clinical", "reseach"]:
-                positional_variant_ids.add(generate_md5_key(other_var_simple + [variant_type]))
+            for var_id, _ in other_case.get("patial_causatives", {}).items():
+                causative_ids.add(var_id)
 
-            causatives += [
-                var for var in self.match_affected_gt(case_obj, positional_variant_ids, limit_genes)
-            ]
+            causatives += self.match_affected_gt(case_obj, causative_ids, limit_genes)
 
         return causatives
 
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -502,7 +502,7 @@ class VariantHandler(VariantLoader):
 
         return self.variant_collection.find(filters)
 
-    def instutute_causatives(institute_obj, limit_genes=None):
+    def instutute_causatives(self, institute_obj, limit_genes=None):
         """Return all causative variants for an institute - eventually within the provided list of genes
 
         Args:
@@ -515,7 +515,7 @@ class VariantHandler(VariantLoader):
         causatives = []
         var_causative_events = self.event_collection.find(
             {
-                "institute": institute_id,
+                "institute": institute_obj["_id"],
                 "verb": {"$in": ["mark_causative", "mark_partial_causative"]},
                 "category": "variant",
             }
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -16,7 +16,9 @@ from scout.utils.md5 import generate_md5_key
 from .variant_loader import VariantLoader
 
 LOG = logging.getLogger(__name__)
+
 MATCHQ = "$match"
+CARRIER = r"[12]"  # same as re.compile()"1|2")
 
 
 class VariantHandler(VariantLoader):
@@ -491,15 +493,11 @@ class VariantHandler(VariantLoader):
         if len(affected_ids) == 0:
             return []
 
-        has_allele = re.compile(
-            "1|2"
-        )  # a non wild-type allele is called at least once in this sample
-
         filters["case_id"] = case_obj["_id"]
         filters["samples"] = {
             "$elemMatch": {
                 "sample_id": {"$in": affected_ids},
-                "genotype_call": {"$regex": has_allele},
+                "genotype_call": {"$regex": CARRIER},
             }
         }
 
@@ -959,11 +957,6 @@ class VariantHandler(VariantLoader):
         Returns:
             result(iterable(Variant))
         """
-        LOG.info("Retrieving variants for subject : {0}".format(sample_name))
-        has_allele = re.compile(
-            "1|2"
-        )  # a non wild-type allele is called at least once in this sample
-
         query = {
             "$and": [
                 {"_id": variant_id},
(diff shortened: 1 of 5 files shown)


scout/adapter/mongo/variant.py
```diff
@@ -552,10 +553,18 @@ def check_causatives(self, case_obj=None, institute_obj=None, limit_genes=None):
                 0:4
             ]  # example: [ "17", "7577559", "G" "A"]
 
+            positional_variant_ids = set()
             for variant_type in ["clinical", "reseach"]:
                 positional_variant_ids.add(generate_md5_key(other_var_simple + [variant_type]))
 
-        return self.match_affected_gt(case_obj, institute_obj, positional_variant_ids, limit_genes)
```
northwestwitch, line 558: The problem was that when invoked from causatives controller case_obj here is None. Then `match_affected_gt` queries for variants in all cases of the institute_obj, also those that are not causative.

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -357,8 +357,7 @@ def test_case_variant_check_causatives_carrier(app, real_variant_database):
         other_causatives = adapter.check_causatives(
             case_obj=other_case, institute_obj=institute_obj
         )
-        LOG.debug("other causatives: {}".format(other_causatives))
-        assert sum(1 for i in other_causatives) == 0
```
northwestwitch, line 361: Was this wrong from the start? 🤔

dnil (approved):
Tests ok locally, and code looks good for this use case so I approve! If I wasn't at the meeting and this wasn't a bug fix I would try to understand the other entries into the adapter as well. In particular Im a little worried that we are overlooking something when the test outcome changes - though it certainly looks more consistent with the comment now? Perhaps we should also add an issue to investigate/refactor it?

northwestwitch:
> though it certainly looks more consistent with the comment now? Perhaps we should also add an issue to investigate/refactor it?

O think that before merging I'll try to understand why the test was formulated like that. The easiest explanation is that the test was reflecting the pre-existent error in the functionality?

northwestwitch:
> Tests ok locally, and code looks good for this use case so I approve! If I wasn't at the meeting and this wasn't a bug fix I would try to understand the other entries into the adapter as well. In particular Im a little worried that we are overlooking something when the test outcome changes - though it certainly looks more consistent with the comment now? Perhaps we should also add an issue to investigate/refactor it?

Now that I look at the code, looks more likely that the test was correct, and the comment was wrong.

I think it should be like this:
- Case A with a causative 
- Case B with causative in an affected individual
-----> check_causatives returns the causative

- Case A with a causative 
- Case B with causative in an UNaffected individual
-----> check_causatives DOES NOT return the causative

scout/adapter/mongo/variant.py
```diff
@@ -560,7 +563,7 @@ def check_causatives(self, case_obj=None, institute_obj=None, limit_genes=None):
             causatives += [
                 var
                 for var in self.match_affected_gt(
-                    other_case, institute_obj, positional_variant_ids, limit_genes
+                    case_obj, institute_obj, positional_variant_ids, limit_genes
```
northwestwitch, line 566: This was an error introduced yesterday by me.

Basically case_obj is the case you are analysing while other_case is the case with the original causative. Of course if you use other case in match_affected_gt it will return the variant, but you are interested in checking of the causative from the other case is found in one affected individual of the case you are analysing... Does it make any sense?

northwestwitch:
Ok, it still doesn't pass the tests. I understood why but perhaps instead of a dirty fix I'll try refactoring to make the code more simple. We have at least 3 functions that return causatives and it's not clear what does what and why. I'll try!

tests/server/blueprints/variant/test_variant_controllers.py
northwestwitch: Nope, the test was right, it was the comment to be wrong!

scout/adapter/mongo/variant.py
```diff
@@ -469,69 +471,94 @@ def _find_affected(self, case_obj):
...
+        }
+
         if limit_genes:
             filters["genes.hgnc_id"] = {"$in": limit_genes}
 
         return self.variant_collection.find(filters)
 
-    def check_causatives(self, case_obj=None, institute_obj=None, limit_genes=None):
-        """Check if there are any variants that are previously marked causative
+    def institute_causatives(self, institute_obj, limit_genes=None):
```
northwestwitch, line 505: This is a completely new function called from institute controllers to get variants for the Cauaatives page

scout/adapter/mongo/variant.py
```diff
@@ -469,69 +471,94 @@ def _find_affected(self, case_obj):
...
+            for var_id, _ in case_obj.get("patial_causatives", {}).items():
+                causative_ids.add(var_id)
+
+        filters = {"_id": {"$in": list(causative_ids)}}
+        if limit_genes:
+            filters["genes.hgnc_id"] = {"$in": limit_genes}
+
+        return self.variant_collection.find(filters)
+
+    def case_matching_causatives(self, case_obj=None, limit_genes=None):
```
northwestwitch, line 541: This is the old `check_causatives`, that now is used only to retrieve matching causatives from other cases

dnil:
Seems like a good idea to split it. I'll have a look if you wish, but the specified test looks good again ofcourse: 

Before:
Screenshot 2023-01-31 at 13 46 36
After:
Screenshot 2023-01-31 at 13 47 13

northwestwitch:
> Seems like a good idea to split it. I'll have a look if you wish, but the specified test looks good again ofcourse:
> 
> Before: Screenshot 2023-01-31 at 13 46 36 After: Screenshot 2023-01-31 at 13 47 13

Code-wise it is more safe now. I've tested it as well and looked fine. Shall I merge?

dnil:
I've just opened it, and can have a look since you asked before. But yeah, I would definitely think splitting this into two functions instead of having a single multipurpose adapter with many conditionals is safer. The only reason to not duplicate in that case is basically to have better maintainability, but if we dont remember all the uses it has, it can't be maintained anyway!

I guess this means the we've had this for about 5 months then? Or have we had something else hide it? I guess it was just in a period when no-one was wanting causatives stats, or only partial such and didn't notice?

scout/adapter/mongo/variant.py
```diff
@@ -469,69 +471,94 @@ def _find_affected(self, case_obj):
...
+            }
+        )
+        causative_ids = set()
+        for case_event in causative_events:
+
+            case_obj = self.case(case_event.get("case"))
+            if case_obj is None:
+                continue
+
+            for var_id in case_obj.get("causatives", []):
```
dnil, line 529: 🥇

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -371,10 +375,11 @@ def test_case_variant_check_causatives_carrier(app, real_variant_database):
     other_case["_id"] = "other_case"
     other_case["internal_id"] = "other_case"
     other_case["display_name"] = "other_case"
-    # GIVEN another case with the same variant for an unaffected individual,
-    other_case["individuals"][0]["phenotype"] = 1
-    other_case["individuals"][1]["phenotype"] = 1
-    other_case["individuals"][2]["phenotype"] = 1
+
+    # GIVEN that the another case has the variant in an unaffected individual
+    for ind in other_case["individuals"]:
```
dnil, line 380: 🥇

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -341,22 +346,21 @@ def test_case_variant_check_causatives(app, real_variant_database):
...
+    # THEN an function will find the matching causative
     with app.test_client() as client:
         resp = client.get(url_for("auto_login"))
-        other_causatives = adapter.check_causatives(
-            case_obj=other_case, institute_obj=institute_obj
+        other_causatives = adapter.case_matching_causatives(
+            case_obj=other_case,
         )
-        LOG.debug("other causatives: {}".format(other_causatives))
-        assert sum(1 for i in other_causatives) > 0
```
dnil, line 351: 👀

tests/server/blueprints/variant/test_variant_controllers.py
```diff
@@ -341,22 +346,21 @@ def test_case_variant_check_causatives(app, real_variant_database):
...
+        assert sum(1 for _ in other_causatives) == 1
```
dnil, line 355: 👍

scout/adapter/mongo/variant.py
dnil: You should probably ditch the `None` defaults here? It would just crash with `case_obj == None` wouldn't it?

dnil (approved):
Looks better than before, and great with the test clarification. Just a small thing regarding the `case_obj` passing that now looks mandatory to me in the "old" adapter function.

dnil:
> I think you are right, it must have been like this for a few months. At least since when I refactored so that Causatives and Verified share the same code. Perhaps even longer 😟

The salient point I guess being that the risk was/is incidental findings in other cases, but since we haven't heard of any, I suppose we were lucky. 😅 Silly with that `>0` test as well; if that had just been `==1` from the get-go we would probably have caught it.

scout/adapter/mongo/variant.py
northwestwitch: Right! Thanks!

northwestwitch:
> had just been `==1` from the get-go we would probably have caught it

Yes! 😞

PR #3802, opened 2023-02-07
PanelApp Green Genes panel

### LINKED ISSUE #3780: Genepanel PanelApp Green genes
We would like to add the genes from Panel App (validated green genes) to OMIM-AUTO (or as a separate gene panel that are always included where OMIM-AUTO is). https://panelapp.genomicsengland.co.uk/

According to Klinisk Genetik in Lund Hans E has possible written / has asked someone else to write a code for this to be performed (@dnil might know more).

According to Klinisk Genetik in Lund (ie Tord) it may be as many as 600 genes or so that are in this gene list that are not present i OMIM-AUTO.

Thank You!  :=)

northwestwitch:
Fix #3780  --> Adds a command line to create/update a gene panel containing all PanelApp green genes

I figured we already have code to perform the basic tasks required by this feature. I've just rearranged to reuse it!

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
Locally, run 
- `scout --demo update genes --build 37`
- `scout --demo update panelapp-green -i cust000`

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


scout/load/panel.py
```diff
@@ -151,6 +173,47 @@ def load_panelapp_panel(adapter, panel_id=None, institute="cust000", confidence=
...
+        "version": old_panel["version"] + 0.1 if old_panel else 1,
+        "date": datetime.now(),
+    }
+
+    genes = set()  # avoid duplicate genes from different panels
+    # Loop over all PanelApp panels
+    for _ in panel_ids:
+        # And collect their green genes
+        parsed_panel = _parse_panelapp_panel(adapter, _, institute, "green")
+        genes.update({(gene["hgnc_id"], gene["hgnc_symbol"]) for gene in parsed_panel.get("genes")})
```
northwestwitch, line 203: This is a workaround because we want unique genes but you dicts are unhashable and can't be added to a set

scout/load/panel.py
```diff
@@ -114,32 +152,17 @@ def load_panelapp_panel(adapter, panel_id=None, institute="cust000", confidence=
...
-    hgnc_map = adapter.ensembl_to_hgnc_mapping()
-
     panel_ids = [panel_id]
 
     if not panel_id:
         LOG.info("Fetching all panel app panels")
-        json_lines = fetch_resource(base_url.format("list_panels"), json=True)
+        panel_ids = _panelapp_panel_ids()
 
-        panel_ids = [panel_info["Panel_Id"] for panel_info in json_lines["result"]]
```
northwestwitch, line 129: I've moved this code to another function so we can reuse it in new feature

northwestwitch:
Tested it locally with a scout instance containing genes in build 37. It saves 3586 genes

image

northwestwitch:
Would be nice to populate the comment field with the panel(s) where the gene is taken from. But perhaps it' an overkill?

image

dnil:
> Would be nice to populate the comment field with the panel(s) where the gene is taken from. But perhaps it' an overkill?

Agreed, probably useful, though I imagine some genes will appear on a lot of panels...

dnil:
Super! From a quick count, some 1054 genes unique to OMIM-AUTO and 198 to PANELAPP-GREEN. Rumors may have been slightly exaggerated, but still a very solidly useful addition!

dnil:
```
scout export panel PANELAPP-GREEN > panelapp_green.txt
grep -v \# panelapp_green.txt |cut -f 1 |sort -n > panelapp_green.ids.txt
scout export panel OMIM-AUTO > omimauto.txt
grep -v \# omimauto.txt |cut -f 1 |sort -n > omimauto.ids.txt
diff -y  --suppress-common-lines omimauto.ids.txt panelapp_green.ids.txt |cut -f 1 | sort |uniq|wc -l
diff -y  --suppress-common-lines panelapp_green.ids.txt omimauto.ids.txt |cut -f1  |sort |uniq |wc -l
```

scout/load/panel.py
```diff
@@ -151,6 +174,46 @@ def load_panelapp_panel(adapter, panel_id=None, institute="cust000", confidence=
...
+    # Loop over all PanelApp panels
+    for _ in panel_ids:
+        # And collect their green genes
+        parsed_panel = _parse_panelapp_panel(adapter, _, institute, "green")
+        genes.update({(gene["hgnc_id"], gene["hgnc_symbol"]) for gene in parsed_panel.get("genes")})
+
+    green_panel["genes"] = [{"hgnc_id": tup[0], "hgnc_symbol": tup[1]} for tup in genes]
+
+    # Do not update panel is new version contains less genes
+    if old_panel and len(old_panel.get("genes", [])) > len(green_panel["genes"]):
```
dnil, line 208: Hm, that could happen from one day to the next, though it is absolutely not likely to be that way with longer update intervals. Add a force flag perhaps?

dnil (approved):
I most certainly approve! Before merging, either adding a force flag or making the test for not decreasing the number of genes on the panel a sharp warning would speak to me, as it COULD happen from one week to another if say some old panels get more stringent reviews while no-one reads new papers. But for almost all most updates, it will be just fine as is.

scout/load/panel.py
northwestwitch: I wasn't convinced of this check myself. Sounds good to keep it like this (think about when something goes wrong and the new panel has no genes) and add a force option, so load it anyway on a second command run!

scout/load/panel.py
```diff
@@ -151,6 +174,48 @@ def load_panelapp_panel(adapter, panel_id=None, institute="cust000", confidence=
...
+    LOG.info("Fetching all panel app panels")
+    panel_ids = _panelapp_panel_ids()
+
+    # check and set panel version
+    old_panel = adapter.gene_panel(panel_id=PANEL_NAME)
+    green_panel = {
+        "panel_name": PANEL_NAME,
+        "display_name": "PanelApp Green Genes",
+        "institute": institute,
+        "version": old_panel["version"] + 0.1 if old_panel else 1,
```
northwestwitch, line 194: This increment of version by 0.1 is something that doesn't really work as one would expect. After the first couple of updates you get this:

image

Which is due to this:

https://stackoverflow.com/questions/11640902/incremented-floats-do-not-equal-each-other

I'll fix

dnil: Nice, I didn't think about that! We should have a structured type there really - I mean, it isn't strictly speaking a float, it is more like a tuple of ints.

northwestwitch: I'm copying the code we use for the OMIM panel: https://github.com/Clinical-Genomics/scout/blob/a998f0941583cd64b9e495319e22f1532d62c850/scout/adapter/mongo/panel.py#L67

dnil: Also nice in its own right that they increment in the same way.

northwestwitch: > Also nice in its own right that they increment in the same way.

yes!

northwestwitch:
Ok, I think it's good to merge now. You have 5 mins to stop me @dnil 😁

dnil:
I had (have?!) a network issue with eduroam and Tunnelblick, but it amounts to the same response. Good job!

PR #3836, opened 2023-03-02
Fix #3792 - thin and allow closing dismiss bar

### LINKED ISSUE #3792: Enable closing of the dismiss bar (variantS page)
Dear developers,

Would it be possible to allow users to close or minimize the "Dismiss selected variants" bar on the variantS page? When looking on a laptop screen it takes up a significant portion of the screen :)

Thank you!

dnil:

Allow temporarily dismissing (until page reload) dismiss bar.

Screenshot 2023-03-02 at 11 57 17

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/views.py
+++ b/scout/server/blueprints/variants/views.py
@@ -305,20 +305,20 @@ def sv_variants(institute_id, case_name):
         "",
     ]
     return dict(
-        institute=institute_obj,
         case=case_obj,
-        dismiss_variant_options=DISMISS_VARIANT_OPTIONS,
-        variant_type=variant_type,
-        form=form,
-        filters=available_filters,
         cytobands=cytobands,
-        severe_so_terms=SEVERE_SO_TERMS,
+        dismiss_variant_options=DISMISS_VARIANT_OPTIONS,
+        expand_search=expand_search,
+        filters=available_filters,
+        form=form,
+        institute=institute_obj,
         manual_rank_options=MANUAL_RANK_OPTIONS,
         page=page,
-        expand_search=expand_search,
         result_size=result_size,
+        severe_so_terms=SEVERE_SO_TERMS,
         show_dismiss_block=show_dismiss_block,
         total_variants=variants_stats.get(variant_type, {}).get(category, "NA"),
+        variant_type=variant_type,
         **data,
     )
 
@@ -454,7 +454,7 @@ def cancer_sv_variants(institute_id, case_name):
     if request.form.get("hpo_clinical_filter"):
         case_obj["hpo_clinical_filter"] = True
 
-    if request.form.getlist("dismiss"):  # dismiss a list of variants
+    if "dismiss_submit" in request.form:  # dismiss a list of variants
         controllers.dismiss_variant_list(
             store,
             institute_obj,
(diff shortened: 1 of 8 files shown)


northwestwitch:
It's thinner and nicer, but you have to close it in every page when you go to the next page or the previous one.. Would be nice if it remembers, at least within the same session. Is it complicated? Uu perhaps by setting a session variable when you close it?

northwestwitch:
But also, would be nice to leave something on the bottom to re-open it again, if the user changes their mind

dnil:
Yes, if it lasts for more than just the page an opening control is needed. It could go on session, but we kind of need the chanjo issue resolved first. As is we need to save space on session, do t we? Maybe on the user object: we don't have too much on it and I don't think they share the session allocation? I guess we could put a control around the pagination block - or possibly in the filter footer, like near the "hide dismissed" toggle?

northwestwitch:
> Yes, if it lasts for more than just the page an opening control is needed. It could go on session, but we kind of need the chanjo issue resolved first. As is we need to save space on session, do t we? Maybe on the user object: we don't have too much on it and I don't think they share the session allocation? I guess we could put a control around the pagination block - or possibly in the filter footer, like near the "hide dismissed" toggle?

To be saved at the user level is totally doable but feels like an overkill to me. I would go on session and use a boolean variable. Shouldn't have that much impact I think.. But either way is fine. And yes, we need to fix that chanjo thing. it's good that we remember of it. I might as well take a look at in the next days.

dnil:
> > Yes, if it lasts for more than just the page an opening control is needed. It could go on session, but we kind of need the chanjo issue resolved first. As is we need to save space on session, do t we? Maybe on the user object: we don't have too much on it and I don't think they share the session allocation? I guess we could put a control around the pagination block - or possibly in the filter footer, like near the "hide dismissed" toggle?
> 
> To be saved at the user level is totally doable but feels like an overkill to me. I would go on session and use a boolean variable. Shouldn't have that much impact I think.. But either way is fine. And yes, we need to fix that chanjo thing. it's good that we remember of it. I might take a look at in the next days.

Sounds good - I'll make a session var then; with a whole panel of gene ids out of the session there should be plenty of space for a few bytes of temporary settings.

dnil:
Ok, the chanjo-report session thing is hopefully fixed, so we must have some space free on the session cookie.
I placed an additional toggle in the filter form footer, in addition to the close button on the dismiss bar itself:
Screenshot 2023-03-16 at 15 34 56

I solved the storage issue by an on click call to a new endpoint, rather than mess with storing this over to different forms.

I note we do not yet have a dismiss block bar on the STR variantS page, but that is a separate issue. 😁

scout/server/blueprints/variants/views.py
```diff
@@ -126,6 +125,8 @@ def variants(institute_id, case_name):
     if request.form.get("export"):
         return controllers.download_variants(store, case_obj, variants_query)
 
+    show_dismiss_block = controllers.get_show_dismiss_block()
```
dnil, line 128: These are the new ones..

scout/server/blueprints/variants/views.py
dnil, line 1: I sorted a little after adding the same lines in many views. This file has lots of unnecessary duplication. But lets finish that off in #3854 instead of here.

dnil:
Thank you - I think this is ready for re-review!

scout/server/blueprints/variants/views.py
```diff
@@ -279,9 +282,10 @@ def sv_variants(institute_id, case_name):
     # Populate chromosome select choices
     controllers.populate_chrom_choices(form, case_obj)
 
-    cytobands = store.cytoband_by_chrom(case_obj.get("genome_build"))
+    genome_build = "38" if "38" in str(case_obj.get("genome_build")) else "37"
```
northwestwitch, line 285: 🥇

scout/server/blueprints/variants/views.py
northwestwitch: These files are heavily duplicated, but perhaps avoid these lines by calling directly controllers.get_show_dismiss_block() later, when you return the data?

scout/server/blueprints/variants/views.py
```diff
@@ -147,6 +148,7 @@ def variants(institute_id, case_name):
         cytobands=cytobands,
         page=page,
         expand_search=expand_search,
+        show_dismiss_block=show_dismiss_block,
```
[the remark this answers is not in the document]
northwestwitch: From here I mean

scout/server/blueprints/variants/views.py
```diff
@@ -299,19 +305,20 @@ def sv_variants(institute_id, case_name):
         "",
     ]
     return dict(
-        institute=institute_obj,
```
northwestwitch, line 302: Nice with ordered items in here

northwestwitch (approved):
Cool improvement! I really like the new dismiss bar layout and the functionality works. 

One thing that I've noticed that happens also in current main is that when you navigate the variantS pages, the variants' filters section is expanded, even if you were not actively using filters. This behavior is a bit annoying because if you were trying to shrink the dismiss bar to see more variants, the variants filter occupy A LOT of space on top of the variant's list than the said bar. Would be nice to fix. But we can do it in another PR!

scout/server/blueprints/variants/views.py
dnil: Sure, it's not liket they get modified more!

scout/server/blueprints/variants/views.py
dnil: Sure, it's not liket they get modified more!

PR #3837, opened 2023-03-02
Hide panels improvements

### LINKED ISSUE #3793: Removed gene panels still visible
Hi,

I'm not sure whether this may be a per-institute setting thing, but I'll start with reporting here. I noticed that deleted gene panels are still visible both on the variantS page (see screenshot) and also on the corresponding variant page (under "gene panels"). In this screen shot, all the panels "Utvecklingsstorning", "panelapp-intellectual-disability" and "panelapp-intellectual-disability-v3" were removed since long before this case was loaded into Scout, and none of those panels are visible in the Gene panels dropdown list on the variantS page.

image

### LINKED ISSUE #3794: Removed panel collapse on panels page is not so pretty on Safari..
Show:
Screenshot 2023-02-02 at 14 01 33
Hide:
Screenshot 2023-02-02 at 14 01 15

😝 Looks fine in Chrome of course. Not a prio, but will look silly at some point if we have many of them. I'm guessing on the abundant use of new tbodies.

dnil:

- Fix #3794: browser support for visibility:collapse is rather sketchy, with a lot of them only accepting certain elements, and essentially just doing visibility:hidden instead, giving the empty-line phenomenon in the panelS list. Use bootstrap collapse instead.

Screenshot 2023-03-02 at 14 31 55
Screenshot 2023-03-02 at 14 31 16
Instead of blank lines for the hidden ones.

- Fix #3793: this was a bit more complicated than anticipated, since several aspects of the panel view on e.g. the variantS page relies on the case version of the panel, which will not have been updated with recent panel hiding/removal since the time of case creation. Add updates.

Looks like before with the case panel active: 
Screenshot 2023-03-02 at 16 49 26

Then if we remove the panel, count drops and the removed panel is not shown.
Screenshot 2023-03-02 at 16 50 01
Screenshot 2023-03-02 at 16 50 54

Also, from the same issue, on the variant page, with panel removed:
Screenshot 2023-03-03 at 08 43 45

and panel shown again:
Screenshot 2023-03-03 at 08 44 15

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. on stage, pick a case with several panels
1. hide some of them and note the hidden ones shown and counted on variantS page, and on the variant page
1. also see that the collapse of the panel rows is rendered as hidden on e.g. Safari
1. apply patch and note the count change and panels hide appropriately
1. see that the collapse of the panel rows is now rendered as collapsed on e.g. Safari
1. make sure the panel sorting is alphabetical on affected views

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/panels/templates/panels/panels.html
+++ b/scout/server/blueprints/panels/templates/panels/panels.html
@@ -157,10 +157,10 @@
             <th></th>
           </tr>
         </thead>
+        <tbody>
       {% for panel in panels|sort(attribute="display_name",case_sensitive=False) %}
         <!--create table for each panel-->
-        <tbody>
-          <tr {% if panel.hidden %} class="hidableRow" style="visibility:collapse;" {% endif %}>
+          <tr {% if panel.hidden %} class="hidableRow collapse" {% endif %}>
             <td>
               <a href="{{ url_for('panels.panel', panel_id=panel._id) }}">{{ panel.display_name }}</a>
               {% if panel.hidden %}
@@ -182,9 +182,10 @@
             </td>
           </tr>
           {{ modify_panel(panel._id, panel) }}
-        </tbody>
+
         {{ remove_gene_panel_modal(institute, panel) }}
-      {% endfor %}
+        {% endfor %}
+        </tbody>
       </table>
     </div>
   </div>
@@ -268,18 +269,6 @@
 <script src="https://cdn.datatables.net/1.12.0/js/jquery.dataTables.min.js" integrity="sha512-fu0WiDG5xqtX2iWk7cp17Q9so54SC+5lk/z/glzwlKFdEOwGG6piUseP2Sik9hlvlmyOJ0lKXRSuv1ltdVk9Jg==" referrerpolicy="no-referrer" crossorigin="anonymous"></script>
 <script type="text/javascript">
 
-function toggleRow(hideCheck) {
-  var els = document.getElementsByClassName("hidableRow");
-  for(var i=0;i<els.length;i++){
-    if(hideCheck.checked){
-      els[i].style.visibility = 'collapse';
-    }
-    else{
-      els[i].style.visibility = 'visible';
-    }
-  }
-}
-
 $('.history_btn').on('click', function(){
     var bid = $(this)[0].id;
     var sel = '#paneldiv_' + bid;
@@ -302,6 +291,20 @@ $('.modify_btn').on('click', function(){
     }
 });
 
+function toggleRow(hideCheck) {
+  var els = document.getElementsByClassName("hidableRow");
+  for(var i=0;i<els.length;i++){
+    collapse = bootstrap.Collapse.getOrCreateInstance(els[i])
+    if(hideCheck.checked){
+      collapse.hide();
+    }
+    else{
+      collapse.show();
+    }
+  }
+}
+
+
 $(function(){
       function getNameTerms(query, process) {
         $.get("{{ url_for('genes.api_genes') }}", {query: query}, function(data) {
@@ -314,6 +317,8 @@ $(function(){
         source: getNameTerms,
         minLength: 3,
       });
+
+
   });
 
 </script>
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -20,7 +20,11 @@ from scout.constants import (
     MOSAICISM_OPTIONS,
     VERBS_MAP,
 )
-from scout.server.blueprints.variant.utils import update_representative_gene
+from scout.server.blueprints.variant.utils import (
+    update_representative_gene,
+    update_variant_case_panels,
+)
+from scout.server.blueprints.variants.utils import update_case_panels
 from scout.server.extensions import cloud_tracks, gens
 from scout.server.links import get_variant_links
 from scout.server.utils import (
@@ -245,6 +249,10 @@ def variant(
     # add default panels extra gene information
     panels = default_panels(store, case_obj)
     add_gene_info(store, variant_obj, gene_panels=panels, genome_build=genome_build)
+
+    # Update some case panels info from db and populate it on variant to avoid showing removed panels
+    update_case_panels(store, case_obj)
+    update_variant_case_panels(store, case_obj, variant_obj)
 
     # Add information about bam files and create a region vcf
     if add_case:
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -252,6 +252,7 @@ def variant(
 
     # Update some case panels info from db and populate it on variant to avoid showing removed panels
     update_case_panels(store, case_obj)
+    # The hierarchical call order is relevant: cases are used to populate variants
     update_variant_case_panels(store, case_obj, variant_obj)
 
     # Add information about bam files and create a region vcf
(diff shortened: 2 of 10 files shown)


northwestwitch:
Panel removed from main branch - Safari - after removing a panel named panel3:

PanelS page:

image

VariantS page:

image

Variant page:

image


From this branch: 🆗 

image

image

image

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -344,8 +344,8 @@ <h6 class="font-weight-bold">{{ image.title }}</h6>
       <ul class="list-group">
         <div class="list-group-item">
           <strong>Gene panels:</strong><br>
-          {% for panel_id in variant.panels|sort(case_sensitive=False) %}
-            <a href="{{ url_for('panels.panel', panel_id=panel_id) }}">{{ panel_id }}</a>&nbsp;&nbsp;
+          {% for panel in variant.case_panels|sort(attribute='panel_name', case_sensitive=False)|rejectattr('removed') %}
```
northwestwitch, line 347: 👍🏻

scout/server/blueprints/panels/templates/panels/panels.html
```diff
@@ -302,6 +291,20 @@ <h5 class="modal-title">Remove gene panel "{{panel.display_name}}"?</h5>
     }
 });
 
+function toggleRow(hideCheck) {
+  var els = document.getElementsByClassName("hidableRow");
+  for(var i=0;i<els.length;i++){
+    collapse = bootstrap.Collapse.getOrCreateInstance(els[i])
+    if(hideCheck.checked){
+      collapse.hide();
```
dnil, line 299: This is the change here: using bootstrap collapse instead of the html style that is not fully implemented on all browsers.

scout/server/blueprints/variant/controllers.py
```diff
@@ -246,6 +250,10 @@ def variant(
     panels = default_panels(store, case_obj)
     add_gene_info(store, variant_obj, gene_panels=panels, genome_build=genome_build)
 
+    # Update some case panels info from db and populate it on variant to avoid showing removed panels
+    update_case_panels(store, case_obj)
```
northwestwitch, line 254: If I understand what the code does, this might be a bit dangerous if you invert the order of these 2 functions. Because in that case the case panels will still contain the hidden panels. It could happen when developing, by mistake because we won't remember that the first function called does a filtering that is required in the second..

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -96,15 +96,24 @@
           </span>
         {% endif %}
       </a>
-      {% if variant.panels|count > 0 %}
```
dnil, line 99: The code here repeats a lot between SNV and SV, but I left it outside the scope of the PR to merge it.

scout/server/blueprints/variant/utils.py
```diff
@@ -148,6 +148,26 @@ def update_transcripts_information(variant_gene, hgnc_gene, variant_obj, genome_
...
+    Args:
+        store(adapter.MongoAdapter)
+        case_obj(dict)
+        variant_obj(dict)
+    """
+
+    variant_panel_names = variant_obj.get("panels") or []
+    case_panel_objs = [
+        panel
+        for panel in (case_obj.get("panels") or [])
```
northwestwitch, line 164: I would still make it check that the panels are not hidden, since relies of a filtering happening before that could not be in place once we forget about it

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -96,15 +96,24 @@
...
+
+      {% set ns = namespace(panel_count = 0) %}
+      {% for panel in variant.case_panels %}
+        {% if not panel.removed %}
+          {% set ns.panel_count = ns.panel_count + 1 %}
```
northwestwitch, line 103: Just a question, can't |rejectattr('removed') used to count as well?

northwestwitch (commented):
Tested and works fine. Got some comments!

scout/server/blueprints/variant/controllers.py
dnil: To be clear, there is no filtering happening here until we reach the template, only updating. But I agree there is an order dependence that could be made explicit. The cost of e.g. calling `update_case_panels` from `update_variant_case_panels` would be some extra db calls for the panels for every variant, instead of having it done once for the case, then calling for each variant. For the variant page, that is nothing, but for the variantS it gets a little costly. I can add a line to the `update_variant_case_panels` docstring to note this perhaps?

northwestwitch: Update them both together?

scout/server/blueprints/variants/templates/variants/components.html
dnil: It doesn't work with the jinja `length` operation after `rejectattr`, unfortunately, but I think it should work instead of the `if not panel.removed` certainly!

scout/server/blueprints/variant/controllers.py
dnil: Yes, we could, but to avoid the performance penalty when used in the variantS view, we would have to take a list of variant_obj to it then. And to do that, consume the db iterator a little earlier and store all variant objs during the variantS loop below. It is fine I guess, just costs a little bit of memory. And for the variant feels a little silly to knowingly pass a list of one all the time for the variant. But making the code easy to read and maintain can be worth it. I would go with the docstring here, for a measurable gain in either cpu or memory over code complexity. But it is not like we are super starved for either CPU or RAM, so just insist and I will change it.

northwestwitch: It is fine with a docstring, but perhaps for good measure add also a comment in the function below (I never read the docstrings..)

scout/server/blueprints/variants/templates/variants/components.html
dnil: Lets see the reason it doesn't work with `length` appears to be that `rejectattr` returns a generator object. So perhaps `|rejectattr|list|length` could work!?

scout/server/blueprints/variant/controllers.py
dnil: Hihi, I took that to heart from one of the code analysis tools we used (Sourcery or Code factor). When one is compelled to comment, its most often because better function names, or an additional function with a docstring, is needed. The corollary becomes to read docstrings if in doubt about what the function does. 😸

northwestwitch: Then the fact I never read docstrings means that we use good names for the functions!

scout/server/blueprints/variant/controllers.py
```diff
@@ -246,6 +250,11 @@ def variant(
...
+    # The hierarchical call order is relevant: cases are used to populate variants
```
northwestwitch, line 255: Thank you! 🙏🏻 😄

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -132,14 +136,19 @@
           href="{{ url_for('genes.gene', hgnc_id=gene.hgnc_id) }}">{{ gene.hgnc_symbol or gene.hgnc_id }}
         </a>
       {% endfor %}
-      {% if variant.panels|count > 0 %}
+
+
+      {% set panel_count = variant.case_panels|rejectattr('removed')|list|count %}
```
northwestwitch, line 141: 🥇

northwestwitch (approved):
Nice fixes! 💯

PR #3838, opened 2023-03-03
Gens individual

raysloks:
### Adds
- `individual_id`-parameter to both Gens links.
### Fixes
- Changes the name of the `sample_id`-field in the `gt_call`-model to the correct `individual_id`. https://github.com/Clinical-Genomics/scout/blob/6ae6114d3fdf2e88b55d61e799f4cfedc6d5e8b8/scout/parse/variant/genotype.py#L76

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Check that the links produce URLs with the correct parameters.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/models/variant/variant.py
+++ b/scout/models/variant/variant.py
@@ -98,7 +98,7 @@ compound = dict(
 clinsig = dict(value=int, accession=str, revstat=str)
 
 gt_call = dict(
-    sample_id=str,
+    individual_id=str,
     display_name=str,
     genotype_call=str,
     allele_depths=list,  # int
--- a/scout/models/variant/variant.py
+++ b/scout/models/variant/variant.py
@@ -98,7 +98,7 @@ compound = dict(
 clinsig = dict(value=int, accession=str, revstat=str)
 
 gt_call = dict(
-    individual_id=str,
+    sample_id=str,
     display_name=str,
     genotype_call=str,
     allele_depths=list,  # int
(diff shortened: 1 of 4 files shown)


scout/models/variant/variant.py
```diff
@@ -98,7 +98,7 @@
 clinsig = dict(value=int, accession=str, revstat=str)
 
 gt_call = dict(
-    sample_id=str,
+    individual_id=str,
```
dnil, line 101: Hm, indeed: you are not wrong. But, if you want to change it throughout you would need to track it down through the `build` path as well. And if you actually change the name in `build`, keep going, because some other stuff like `is_affected` would break. Up to you, but I would not mind e.g. just adding `individual_id` to the model and making an issue for it rather than adding something that may have consequences that are hard to test?

dnil (approved):
Agreed on the gens link side. Models are more of a guideline anyway, but I would suggest just adding `individual_id` and making an issue about the  `gt_call` `sample_id` and `individual_id` usage?

scout/models/variant/variant.py
raysloks: Took a closer look, and realized I was wrong about this. Got confused because `gt_call` is being used as a name for both pre and post build step objects.

dnil: I would not say wrong, but those are very annoying. If we ever have a lot of developer time, refactoring that mechanism and making nice clear storage objects and a couple of kinds of lightly and heavily populated display objects is definitely on the wish list.

PR #3850, opened 2023-03-14
Prepare new minor version 4.65

northwestwitch:
### Added
- [x] Option to mark a ClinVar submission as submitted
- [x] Docs on how to create/update the PanelApp green genes as a system admin
- [x] individual_id`-parameter to both Gens links
- [x] Download a gene panel in TXT format from gene panel page
- [x] Panel gene comments on variant page: genes in panels can have comments that describe the gene in a panel context
### Changed
- [x] Always show each case category on caseS page, even if 0 cases in total or after current query
- [x] Improved sorting of ClinVar submissions
- [x] Pre-populate SV type select in ClinVar submission form, when possible
- [x] Show comment badges in related comments tables on general report
- [x] Updated version of several GitHub actions
- [x] Migrate from deprecated `pkg_resources` lib to `importlib_resources`
- [x] Dismiss bar on variantS pages is thinner.
- [x] Dismiss bar on variantS pages can be toggled open or closed for the duration of a login session.
### Fixed
- [x] Fixed Sanger order / Cancel order modal close buttons
- [x] Visibility of SV type in ClinVar submission form
- [x] Fixed a couple of creations where now was called twice, so updated_at and created_at could differ
- [x] Deprecated Ubuntu version 18.04 in one GitHub action
- [x] Panels that have been removed (hidden) should not be visible in views where overlapping gene panels for genes are shown
- [x] Gene panel test pointing to the right function

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
I was testing the gene panel txt download feature and I've noticed that the name of the downloaded panel could be a bit weird if the use hasn't provided a meaningful name (ID field), for instance take these 2 panels:


image



image

When downloaded, they result in these files:

image

Perhaps it is fine, I don't know..

dnil:
Ah, oops! That was the intention. I must have had the setting stored already when testing. Will fix!

northwestwitch:
Testing the `Panel gene comments on variant page: genes in panels can have comments that describe the gene in a panel context` feature. Question/observation: I've tested adding a comment to the latest OMIM panel for LIG1 one this gene:

image

Then I picked one case with LIG1: https://scout-stage.scilifelab.se/cust000/15026-mip7/2fc4fe9b557989504831957e6188ae7e

And the comment DOESN'T show on that variant's page.

Now I don't remember the code but perhaps it's fishing comments for the same version of the panel the case was loaded with? In that case we should specify it

dnil:
> Testing the `Panel gene comments on variant page: genes in panels can have comments that describe the gene in a panel context` feature. Question/observation: I've tested adding a comment to the latest OMIM panel for LIG1 gene:
> 
> image
> 
> Then I picked one case with LIG1: https://scout-stage.scilifelab.se/cust000/15026-mip7/2fc4fe9b557989504831957e6188ae7e
> 
> And the comment DOESN'T show on that variant's page.
> 
> Now I don't remember the code but perhaps it's fishing comments for the same version of the panel the case was loaded with? In that case we should specify it

Yes, without looking, I think it does, kind of because it will retrieve the case default panels. It would be exactly the same way the other things like disease transcripts, manual inheritance patterns are fetched as well, link outs from the page to the panel etc. It did seem like a good idea at the time, given it was what the case was run with, but there is also good merit in fetching updated panels and gathering latest changes as much as we can. Let's check that it actually works when the comment is on the actual panel, and make an issue to rework the whole mechanism perhaps?

northwestwitch:
> Let's check that it actually works when the comment is on the actual panel, and make an issue to rework the whole mechanism perhaps?

Agreed!

northwestwitch:
> Let's check that it actually works when the comment is on the actual panel,

It's a bit tricky to test, since as soon as you add a comment to the old panel it becomes a new panel. So first you have to make the comment and then load the case.

northwestwitch:
> Let's check that it actually works when the comment is on the actual panel, and make an issue to rework the whole mechanism perhaps?

We could also avoid mentioning the feature for the time being, otherwise people will be confused when they expect comments to show up and they don't

dnil:
Hm, I wasn't able to recreate the issue with the dismiss bar? If I clear session, and log in again, it appears on? Is there some weird other value thing when one is already logged in but have not visited the page?

northwestwitch:
>

northwestwitch:
> Hm, I wasn't able to recreate the issue with the dismiss bar? If I clear session, and log in again, it appears on? Is there some weird other value thing when one is already logged in but have not visited the page?

Ok,  but perhaps then it was my session. I just logged out and in again

dnil:
> > Hm, I wasn't able to recreate the issue with the dismiss bar? If I clear session, and log in again, it appears on? Is there some weird other value thing when one is already logged in but have not visited the page?
> 
> Ok, but perhaps then it was my session. I just logged out and in again

At least I can't recreate it: I tried on stage as well, and it behaves as expected: shown on first login after session clear, toggles are saved between cases, page reloads etc.

You made me worried that there is an additional special case for returning users that have a valid session, but no variable set (like some None to undefined issue), but it also seems to be working when switching between branches on local.

I'm happy to try to fix it if we can recreate, but perhaps it was just from having used it last week?

northwestwitch:
> I'm happy to try to fix it if we can recreate, but perhaps it was just from having used it last week?

I think it might be that, never mind then!

dnil:
> I was testing the gene panel txt download feature and I've noticed that the name of the downloaded panel could be a bit weird if the user hasn't provided a meaningful name (ID field), for instance take these 2 panels:
> 
> image
> 
> image
> 
> When downloaded, they result in these files:
> 
> image
> 
> Perhaps it is fine, I don't know..

It is better than leaving it blank at least! But since its just in one place we could make it like the panel name if the id is not set? This in particular looks like it might be a PanelApp panel: perhaps there is rather an issue with how we set the id for them, like it would be better re-using the panelApp panel numerical "id".

northwestwitch:
> It is better than leaving it blank at least! But since its just in one place we could make it like the panel name if the id is not set? This in particular looks like it might be a PanelApp panel: perhaps there is rather an issue with how we set the id for them, like it would be better re-using the panelApp panel numerical "id".

Id is always set, and yes, it might be a PanelApp thing, like if panel ID is not found when building it, then it is set to internal panel ID. Let's investigate and keep it like this for now!

dnil:
> > Let's check that it actually works when the comment is on the actual panel, and make an issue to rework the whole mechanism perhaps?
> 
> We could also avoid mentioning the feature for the time being, otherwise people will be confused when they expect comments to show up and they don't

I had a quick check on local again, and it seems all good with the latest version. I feel it is defendable being the same way as e.g. disease causing transcripts and panel inheritance patterns? I had a brief chat with the somatic people, though not exactly the ones with the issue. They found it good like this; they have a high turnover, and having changes affect old cases is not interesting.

northwestwitch:
> They found it good like this; they have a high turnover, and having changes affect old cases is not interesting.

Perfect! 👍🏻

dnil:
I'm ticking the gens one as well: if we look at a case that is loaded on gens-stage https://scout-stage.scilifelab.se/cust003/21339-trio this patch goes from the likes of
Screenshot 2023-03-20 at 11 27 03
to
Screenshot 2023-03-20 at 11 27 07
Index files are still missing, so actually loading would take until next week I guess.

dnil (approved):
Looks good! I took the liberty to revert something I happened to include at some point while testing in a demo case before releasing. 

What of #3857?

northwestwitch:
> What of #3857?

I completely forgot about it. We could include a fix for it as well?

dnil:
> > What of #3857?
> 
> I completely forgot about it. We could include a fix for it as well?

Ok, I'll have a look again!

dnil:
> > > What of #3857?
> > 
> > 
> > I completely forgot about it. We could include a fix for it as well?
> 
> Ok, I'll have a look again!

For reference, it is likely due to this, and its previous incarnations:  https://github.com/Clinical-Genomics/scout/blob/e9ce44b0f61b504b79c75f212f3c61d9139a6917/scout/server/blueprints/variants/controllers.py#L781

This is by no means a new issue, but one we will adress in the future, perhaps even refactoring some of the build stuff while we are at it.

PR #3864, opened 2023-03-21
Fix PDF delivery report

### LINKED ISSUE #3862: Delivery report pdf conversion off?
It seems the pdf conversion of delivery reports died. 

The pdf download button downloads html, much as if it had had `pdf_export: False`.
Screenshot 2023-03-21 at 10 18 37

But it doesn't, apparently: 
https://github.com/Clinical-Genomics/scout/blob/6a597662854de7f75d08420446f7fbe7ddaa94f8/scout/constants/case_tags.py#L5

northwestwitch:
Fix #3862 
I am under the impression that the PDF version hasn't been working for a while

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Run scout setup demo and make sure that the case contains 2 new reports that can be seen/downloaded from the case sidebar -> pipeline_version and reference_info
2. Try to download the delivery report in the PDF format

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


scout/constants/case_tags.py
```diff
@@ -2,7 +2,7 @@
 
 # keys are used in the load report cli, while key_name is saved in case database documents
 CUSTOM_CASE_REPORTS = {
-    "delivery": {"key_name": "delivery_report", "format": "HTML", "pdf_export": True},
+    "delivery_report": {"key_name": "delivery_report", "format": "HTML", "pdf_export": True},
```
dnil, line 5: 🙄

northwestwitch: Ok, I'd rather keep that name and change the template, wait for the review

northwestwitch: Naa it's more comoplicated. I'd like to keep it like this instead

northwestwitch:
Yes, it has been broken since [Nov 22nd](https://github.com/Clinical-Genomics/scout/commit/aa235cdaede62285ffc8a490bd33ac5b0a21fde6), when [this code](https://github.com/Clinical-Genomics/scout/blob/79e171597bfdadea788832c763b510817b4855f7/scout/server/blueprints/cases/views.py#L728) was introduced.

northwestwitch:
Ok, it should be fixed now. I wanted to fix the Vulture issues but I've realized that I can't: those --update are parameters that were present before in the cli and they aren't used any more, but if we remove them then it becomes a breaking change (major). We marked that cli function as deprecated for this reason.

dnil:
> Yes, it has been broken since [Nov 22nd](https://github.com/Clinical-Genomics/scout/commit/aa235cdaede62285ffc8a490bd33ac5b0a21fde6), when [this code](https://github.com/Clinical-Genomics/scout/blob/79e171597bfdadea788832c763b510817b4855f7/scout/server/blueprints/cases/views.py#L728) was introduced.

Hm? But it should get this: https://github.com/Clinical-Genomics/scout/blob/63b99eba44ad8873673f34b8fe6b7a6c77bf3b83/scout/server/blueprints/cases/templates/cases/collapsible_actionbar.html#L132, which should be the key_name to use?

dnil (approved):
This is fine; it solves the issue, and it would not be the first time we use explicit file endings (bam and cram files and their indexes come to mind). But see alternative solution to keep using the const in the other PR. I'll leave it to you!

northwestwitch:
To better test the reports  (pipeline version and  reference_info ) I've included them in the demo config file, so they are loaded by default in the demo case. But I've noticed that the pipeline version one doesn't show on the case sidebar. It was due to a different naming of the key in the stored case object, now it is fixed!

northwestwitch:
I see what do you mean for the test @dnil , I'll refactor to cover all the situations!

dnil:
> To better test the reports (pipeline version and reference_info ) I've included them in the demo config file, so they are loaded by default in the demo case. But I've noticed that the pipeline version one doesn't show on the case sidebar. It was due to a different naming of the key in the stored case object, now it is fixed!

Yes, if I recall correctly we kept it at optional load since demand for them was limited, but agreed, why not load them when we have them anyway. 

I do not think we should change the actual config file key name or db field name of any of them - that's synced with other cg services and has been in rotation for a bit. But maybe I misunderstood, I'll check the code.

northwestwitch:
> I do not think we should change the actual config file key name or db field name of any of them - that's synced with other cg services and has been in rotation for a bit. But maybe I misunderstood, I'll check the code.

I didn't change it (I think), the expected key was exe_ver and still is. But yes, take a look because it's a tricky one

scout/models/case/case.py
```diff
@@ -54,7 +54,6 @@
     diagnosis_phenotypes=list,  # List of dictionaries with OMIM disease data
     display_name=str,  # required. This is the case name that will be shown in scout.
     dynamic_gene_list=list,  # List of genes
-    exe_ver=str,  # Path to the pipeline executable versions report file
```
northwestwitch, line 57: I think the key saved on the database for the case is pipeline_version, at least it's the one that is expected on the [formats constant](https://github.com/Clinical-Genomics/scout/blob/79e171597bfdadea788832c763b510817b4855f7/scout/constants/case_tags.py#L8)

tests/server/blueprints/cases/test_cases_views.py
```diff
@@ -14,46 +11,60 @@
 TEST_TOKEN = "test_token"
 
 
-@pytest.mark.parametrize("report_types", list(CUSTOM_CASE_REPORTS.keys()))
-def test_custom_report(app, institute_obj, case_obj, report_types):
+def test_custom_report(app, institute_obj, case_obj, tmp_path):
```
northwestwitch, line 14: To make sure that this test is good I've tried reverting the code in cases_views and indeed the error is caught:

image

The tests fails with this message:

image

That happens when downloading the PDF report for the delivery reports -> instead of the PDF report you get an error page

scout/build/case.py
```diff
@@ -34,6 +34,17 @@ def build_phenotype(phenotype_id, adapter):
     return phenotype
 
 
+def _populate_pipeline_info(case_obj, case_data):
```
dnil, line 37: Perhaps also add the `reference_info` then if we decided to load them systematically. They complement each other well for MIP?

scout/models/case/case.py
dnil: But hold on, this is still weird: if this is the db key, then we have a bunch of other places to change it as well, don't we? Like https://github.com/Clinical-Genomics/scout/blob/b0cfc8795392ed7e1b223eeaa5ad5590fc6e8892/scout/adapter/mongo/case.py#L1000?

scout/build/case.py
dnil, line 37:
```suggestion
def _assign_pipeline_info(case_obj, case_data):
```
Or just `build_pipeline_info` - its just language, but we've generally called the addition of data to the in memory objects for web rendering 'population' in contrast to the 'build' for converting the objects from memory for db storage, though there are some unfortunate reuses.. 😓

dnil (commented):
Sorry, this is getting complicated, but thank you for adding the test fix; I think that was very relevant. The test would have caught the issue if it was right from the start, and it would also have made it more obvious what the different dictionary and database key names were.

scout/build/case.py
northwestwitch: Yes the other key is already collected fine, because the case key has the same name that is taken from the config file

scout/models/case/case.py
northwestwitch: Ok it looks weird but if you try the command then it updates the path correctly, perhaps we can change the comments in there to be less confusing

northwestwitch: The cli command to update that field is not invoking that command it seems, but I fixed the adapter so that whenever you make another update in the case you keep also the pipeline_version

dnil (commented):
Ok, looks good to me now, but I think let us test a bit with different loading and updating?

northwestwitch:
Merging this one now

PR #3865, opened 2023-03-22
Alt fix pdf delivery

dnil:

I mostly wanted to make sure I understood the issue. Your solution is easier, so we can go with it as well! It just feels a little annoying to have hard coded file ending usage, when we anyway have a format tag in the describing dict. Making a reverse lookup dict for this key is a little annoying but cheap. 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
I feel the risk of the ego arguing here, but I'd say the test fix at least should be added: it makes sense to use the test with the parameter it would be passed when actually using the endpoint. Ego aside, it kind of still works out either way, as both test versions actually produce a delivery report pdf.

scout/constants/case_tags.py
```diff
@@ -33,6 +33,13 @@
     },
 }
 
+CUSTOM_CASE_REPORT_KEY_NAMES = dict(
```
northwestwitch, line 36: For me it is a bit hard to understand and follow what the field of this variable are. What is it, it's very nested in the end

scout/server/blueprints/cases/views.py
```diff
@@ -725,7 +724,11 @@ def custom_report(institute_id, case_name, report_type):
         try:
             with open(os.path.abspath(report_path), "r") as html_file:
                 source_code = html_file.read()
-                if CUSTOM_CASE_REPORTS[report_type]["format"] == "YAML":
+
+                if (
+                    CUSTOM_CASE_REPORTS[CUSTOM_CASE_REPORT_KEY_NAMES[report_type]]["format"]
```
northwestwitch, line 729: This works, but the syntax is quite complex with a very nested dictionary. I don't like much my solution but I find it simpler to follow and to maintain?

northwestwitch (commented):
Check my comments!

scout/server/blueprints/cases/views.py
dnil: I prefer this one, but I'm ok with your solution just checking file ending. But I would prefer it if we at least also fix the endpoint test then to use the keys the endpoint actually receives...

northwestwitch: I can add some parametrized tests tomorrow!

dnil: It has a parametrised test already; just needs the proper key - but you can e.g. reuse the reverse dict from here, or make a list directly from the original dict member dicts key_name members!

northwestwitch: But Ok, I'll check the code tomorrow!

dnil:
The test is now fixed in the parent PR as well, so closing this. While I still think a reverse lookup was not a bad two-with-one-stone solution, checking filename endings explicitly and skipping the test parametrisation for an explicit loop also works.

northwestwitch:
> The test is now fixed in the parent PR as well, so closing this. While I still think a reverse lookup was not a bad two-with-one-stone solution, checking filename endings explicitly and skipping the test parametrisation for an explicit loop also works

After the third level of dictionary nesting it should be considered machine code 😄

PR #3892, opened 2023-04-04
Collapse filters except when filtering

### LINKED ISSUE #3855: Keep variants filters section closed when navigating thru pages
Right now when you navigate the variants page the section keeps expanding also if you close it

northwestwitch:
Fix #3855 by keeping filters always collapsed, except when filtering

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Deploy on web server stage and navigate the variants page. Try also submitting some search using filters

**Expected outcome**:
- [x] Filters window should stay closed, except when manually expanded 

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
> Did you get some feedback about this?

No, I just wanted to simplify the feature. Like you click filter and it filters, then if you want to inspect which filters you used you can expand the filtering window

dnil (commented):
Simple is good, but let's run the functionality with a couple of users then! I have a feeling it will be quite ok for the RD-users, but a few extra unhelpful clicks for the somatic side.

northwestwitch:
> Then opening the filter becomes redundant between each click. Did you get some feedback about this? Otherwise one option could be to keep the filter open for the first round after hitting Filter or Clinical filter, but not after a Next/Previous, clear filter, or navigate from case page?

Perhaps one could just write something like "3 filters applied", near the number of variants shown? Would that be helpful instead of expanding the filters view?

dnil:
> > Then opening the filter becomes redundant between each click. Did you get some feedback about this? Otherwise one option could be to keep the filter open for the first round after hitting Filter or Clinical filter, but not after a Next/Previous, clear filter, or navigate from case page?
> 
> Perhaps one could just write something like "3 filters applied", near the number of variants shown? Would that be helpful instead of expanding the filters view?

I thing the counts are fine there - it's the quick access to changing filter (i.e. the Load and perhaps Audit filters buttons and preset filter selection dropdown). I suppose we could expose those outside the collapse on the cancer pages..

dnil (changes requested):
Ok, as per chat, especially the cancer side kind of go through several filters per case, so they like it as is. A compromise could be to allow extended filters to collapse on next/previous but not after Filter etc.

northwestwitch:
> Ok, as per chat, especially the cancer side kind of go through several filters per case, so they like it as is. A compromise could be to allow extended filters to collapse on next/previous but not after Filter etc.

I've fixed it so it should follow this behavior now!

northwestwitch:
I suspect that the SonarCloud analysis is mostly failing because of a duplicated block of code that is there since before. I don't want to address this well known issue in this PR

scout/server/blueprints/variants/controllers.py
```diff
@@ -43,6 +44,21 @@
 LOG = logging.getLogger(__name__)
 
 
+def get_variants_page(form=ImmutableMultiDict) -> int:
```
dnil, line 47: 👍

scout/server/blueprints/variants/templates/variants/cancer-variants.html
```diff
@@ -42,7 +42,7 @@
         <strong><a data-bs-toggle="collapse" data-parent="#accordion" href="#collapseFilters">Filters</a></strong>
       </div>
       <!--Expand filters form if filters were used in a previous POST request or the previous POST request form was not validated-->
-      <div class="card-body panel-collapse collapse {{ 'show' if request.args.get('expand_search') == "True" or expand_search is sameas true }}" id="collapseFilters" >
+      <div class="card-body panel-collapse collapse {{ 'show' if expand_search is sameas true }}" id="collapseFilters" >
```
dnil, line 45: 👍

scout/server/blueprints/variants/views.py
```diff
@@ -508,6 +487,7 @@ def cancer_sv_variants(institute_id, case_name):
         show_dismiss_block=controllers.get_show_dismiss_block(),
         total_variants=variants_stats.get(variant_type, {}).get(category, "NA"),
         variant_type=variant_type,
+        expand_search=controllers.get_expand_search(request.form),
```
dnil, line 490: Wasn't alphabetical order kind of nice? 😊

scout/server/blueprints/variants/controllers.py
dnil: Does change current ops a little with not checking args.page from an initial GET, but afaik that should never want anything other than page 1 anyway - so good riddance!

dnil (approved):
🌟

scout/server/blueprints/variants/views.py
northwestwitch: I didn't notice, I'll fix it!

scout/server/blueprints/variants/controllers.py
dnil: Exactly, though before it was technically possible to ask for another page via get. Though I do not think we use that in code anymore, since a looong time really.

scout/server/blueprints/variants/views.py
dnil: Hihi, not all of them are, as evident from this. But thanks! 😁

PR #3898, opened 2023-04-11
Speed up institutes loading

### LINKED ISSUE #3897: Inefficient db queries makes /institutes slow to load
The `/institutes` view in Scout takes several seconds to load which is likely caused by inefficient query design.

To build up the view Scout

1. Query the user collection to get which institutes the logged-in user has access to
2. For each institute query the case collection for all cases belonging to the institute
3. [Iterate over each resulting case to count the number of cases](https://github.com/Clinical-Genomics/scout/blob/c91a7b7a7317e2778f5cd6fe8b9604c4679c1294/scout/server/blueprints/cases/views.py#L56)

Each query conducted with `store.cases(collaborator=institute_obj["_id"])))` takes roughly 300 ms so the query time glows rapidly as the user has access to more institutes.

**Suggested solutions**

I suggest that we either modify `store.cases` to accept kwargs for data projections so we can use a covered query instead. By using a covered query mongo wouldn't need to read documents from disk which are likely to speed up the query and reduce the amount of data being returned and casted into python types.

Another solution would be to add a collection of aggregated data that are rarely updated. Keeping track of the number of cases per institute could be done when data is added and removed from the database. With this solution, there wouldn't be any need to continuously compute a number that rarely changes every time a user logs in or goes to `/institutes`

northwestwitch:
Tentative fix #3897 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -30,6 +30,7 @@ from scout.server.utils import (
     institute_and_case,
     jsonconverter,
     templated,
+    user_cases,
     user_institutes,
     zip_dir_to_obj,
 )
@@ -52,12 +53,14 @@ cases_bp = Blueprint(
 @templated("cases/index.html")
 def index():
     """Display a list of all user institutes."""
-    institute_objs = user_institutes(store, current_user)
-    institutes_count = (
-        (institute_obj, sum(1 for i in store.cases(collaborator=institute_obj["_id"])))
-        for institute_obj in institute_objs
-        if institute_obj
-    )
+    institute_cases = user_cases(store, current_user)
+    institutes_count = []
+    for item in institute_cases:
+        institute_obj = store.institute(item["_id"]["institute"])
+        if not institute_obj:
+            continue
+        institutes_count.append((institute_obj, item["count"]))
+
     return dict(institutes=institutes_count)
 
 
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -53,9 +53,8 @@ cases_bp = Blueprint(
 @templated("cases/index.html")
 def index():
     """Display a list of all user institutes."""
-    institute_cases = user_cases(store, current_user)
     institutes_count = []
-    for item in institute_cases:
+    for item in user_cases(store, current_user):
         institute_obj = store.institute(item["_id"]["institute"])
         if not institute_obj:
             continue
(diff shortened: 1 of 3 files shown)


scout/server/utils.py
```diff
@@ -133,6 +133,35 @@ def institute_and_case(store, institute_id, case_name=None):
...
+        store(scout.adapter.mongo.base.MongoAdapter)
+        login_user(werkzeug.local.LocalProxy)
+
+    Returns:
+        pymongo.Cursor
+    """
+
+    match_query = {
+        "$match": {
+            "collaborators": {"$in": [login_user.institutes]},
```
dnil, line 149: Looks good, but needs an admin exception, like in https://github.com/Clinical-Genomics/scout/blob/4f210fbd15dcc8ad53d9fbd1bb29b59347b96b5f/scout/server/utils.py#L136

northwestwitch: Does it? As an admin using this code you should have access to all cases of all institutes anyway. And you'd get a feeling of how much times it takes for a non-admin user with many cases in many institutes

northwestwitch: Ah no, you are right, if you are not a collaborator then you don't get that institute. Will fix!

scout/server/blueprints/cases/views.py
```diff
@@ -52,12 +53,13 @@
...
-        for institute_obj in institute_objs
-        if institute_obj
-    )
+    institutes_count = []
+    for item in user_cases(store, current_user):
+        institute_obj = store.institute(item["_id"]["institute"])
+        if not institute_obj:
+            continue
+        institutes_count.append((institute_obj, item["count"]))
+
```
dnil, line 62: We loose the institute sort now; I guess it was implicit before from looping over the institutes, but now they appear in user case order? Anyway, institute list needs a sort. `internal_id` is convenient enough for admins, and I guess regular users usually don't have that many, otherwise `display_name` sorting would be in order. Either is fine!

dnil (approved):
Let's fix the sort order as well, otherwise I think it looks good! One could test the new utils function, but coverage stays ok so very much up to the author.

northwestwitch:
Testing fine on stage so i'm merging. Give it a try as soon as you have the chance @mhkc!

PR #3907, opened 2023-04-13
Optimise cases

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,10 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Prepare to filter local SV frequency
 ### Changed
-- Speed up instituteS page loadiding by refactoring cases/institutes query
+- Speed up instituteS page loading by refactoring cases/institutes query
 - Clinical Filter for SVs includes `splice_polypyrimidine_tract_variant` as a severe consequence
 - Clinical Filter for SVs includes local variant frequency freeze ("old") for filtering, starting at 30 counts
+- Speed up caseS page loading by adding status to index and refactoring totals count
 ### Fixed
 - Page crashing when a user tries to edit a comment that was removed
 - Warning instead of crashed page when attempting to retrieve a non-existent Phenopacket
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -576,7 +576,7 @@ class CaseHandler(object):
         """
 
         pipeline = []
-        if institute_id:å
+        if institute_id:
             pipeline.append({"$match": {"collaborators": institute_id}})
         pipeline.append({"$count": "nr_cases"})
         result = {"nr_cases": res["nr_cases"] for res in self.case_collection.aggregate(pipeline)}
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -574,12 +574,17 @@ class CaseHandler(object):
         Returns:
             nr_cases(int)
         """
-        query = {}
+
+        pipeline = []
 
         if institute_id:
-            query["collaborators"] = institute_id
-
-        nr_cases = sum(1 for i in self.case_collection.find(query))
+            pipeline.append({"$match": {"collaborators": institute_id}})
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,9 +8,10 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Prepare to filter local SV frequency
 ### Changed
-- Speed up instituteS page loadiding by refactoring cases/institutes query
```
northwestwitch, line 11: 🤣

scout/adapter/mongo/case.py
```diff
@@ -574,12 +574,16 @@ def nr_cases(self, institute_id=None):
         Returns:
```
northwestwitch, line 574: I wonder if this query can be entirely skipped. One could sum up the numbers returned by [this other function](https://github.com/Clinical-Genomics/scout/blob/aecafa3219e3be066dbc137896d3d1b603dcd144/scout/server/blueprints/institutes/controllers.py#L420) (that is necessary) instead?

scout/constants/indexes.py
```diff
@@ -123,8 +123,8 @@
         IndexModel([("synopsis", TEXT)], default_language="english", name="synopsis_text"),
         IndexModel([("causatives", ASCENDING)], name="causatives"),
         IndexModel(
-            [("collaborators", ASCENDING), ("updated_at", ASCENDING)],
-            name="collaborators_updated_at",
+            [("collaborators", ASCENDING), ("status", ASCENDING), ("updated_at", ASCENDING)],
```
northwestwitch, line 126: This speeds up the other query (nr_cases_by_status), no? Nice!

northwestwitch (approved):
It works. Check my comment about removing one query entirely!

scout/adapter/mongo/case.py
dnil: Yes, I thought the same. But it should be really quick, and then we have to ensure the others are not overlapping, and the function is not used elsewhere.

scout/constants/indexes.py
dnil: Exactly!

scout/adapter/mongo/case.py
dnil: Ok, does not seem to be used elsewhere, and the `$group` should make it as non-overlapping as it gets.

scout/constants/indexes.py
dnil: What is left to speed up if needed later is the third query, from `cases()` that actually cursors the documents.

scout/adapter/mongo/case.py
northwestwitch: Yes, I checked, it's used only in there

dnil: Heh, it was surprisingly slightly slower when removing it; possibly my slower python juggling of dict-array conversions than the rather to the point db summing, or just a different day. But both still clocked faster than the original.

Screenshot 2023-04-18 at 10 44 10

I would still keep the latest without the double db request, just to keep things simple, and we see how it performs on a large db?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -416,8 +416,8 @@ def cases(store, request, institute_id):
     )
     all_cases = _sort_cases(data, request, all_cases)
 
-    data["nr_cases"] = store.nr_cases(institute_id=institute_id)
     data["status_ncases"] = store.nr_cases_by_status(institute_id=institute_id)
+    data["nr_cases"] = sum(data["status_ncases"].values())
```
northwestwitch, line 420: 👍🏻

scout/adapter/mongo/case.py
northwestwitch: Ok!

northwestwitch (approved):
👍🏻

PR #3979, opened 2023-06-08
Fix assigned name when id is different from email

maehler:
As @dnil got to see last week, our Scout instance did not show the assigned user correctly in the case view. This is what we had:

image

After digging around, it seemed to be due to the user ID not being the same as the user email. We are using LDAP for login, and would like to keep the ID separate from the email. This PR addresses this issue by making sure that the user ID is used when assigning a user to a case, and that the assigned users are fetched properly when the case is processed for viewing.

image

I noticed more places where emails are used like user IDs, but decided to only modify the bare minimum to solve this particular problem. Hopefully this doesn't introduce any issues elsewhere. Looking forward to your feedback!

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -558,7 +558,7 @@ def assign(institute_id, case_name, user_id=None, inactivate=False):
     if user_id:
         user_obj = store.user(user_id=user_id)
     else:
-        user_obj = store.user(current_user.email)
+        user_obj = store.user(email=current_user.email)
     if request.form.get("action") == "DELETE":
         store.unassign(institute_obj, case_obj, user_obj, link, inactivate)
     else:
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -556,7 +556,7 @@ def assign(institute_id, case_name, user_id=None, inactivate=False):
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
     link = url_for(".case", institute_id=institute_id, case_name=case_name)
     if user_id:
-        user_obj = store.user(user_id)
+        user_obj = store.user(user_id=user_id)
     else:
         user_obj = store.user(current_user.email)
     if request.form.get("action") == "DELETE":
--- a/tests/server/blueprints/cases/test_cases_templates.py
+++ b/tests/server/blueprints/cases/test_cases_templates.py
@@ -137,7 +137,7 @@ def test_sidebar_cnv_report(app, institute_obj, cancer_case_obj, user_obj):
 
 
 def test_sidebar_assign(app, institute_obj, case_obj, user_obj):
-    """Test the case sidebar macro items for a cancer case"""
+    """Test the case sidebar macro items for a case"""
     # GIVEN an initialized app
     with app.test_client() as client:
         # GIVEN that the user could be logged in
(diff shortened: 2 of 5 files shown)


northwestwitch:
Hi @maehler, sorry for this issue. We aren't using the LDAP login system so we don't see these missing emails all over the place. Thanks for the fix! 🙏🏻 
I can review tomorrow if @dnil hasn't started yet!

dnil:
> Hi @maehler, sorry for this issue. We aren't using the LDAP login system so we don't see these missing emails all over the place. Thanks for the fix! 🙏🏻
> I can review tomorrow if @dnil hasn't started yet!

Super! I got too close to the Hackathon vortex and got dragged in, so have not had much time for Scout the last couple of days. 🙄

northwestwitch:
> Super! I got too close to the Hackathon vortex and got dragged in, so have not had much time for Scout the last couple of days.

I figured as much, no worries! :)

maehler:
> Hi @maehler, sorry for this issue. We aren't using the LDAP login system so we don't see these missing emails all over the place. Thanks for the fix! 🙏🏻 I can review tomorrow if @dnil hasn't started yet!

Don't be sorry; thanks for all your hard work! It's not really an LDAP issue per se, just that there seem to be places in the code where IDs and emails are assumed to be the same. I could get around it by having the user IDs in the database be the user email addresses and setting the LDAP login attribute to email, but I'm reluctant to change our setup at this point in order to cure the symptoms.

scout/server/blueprints/cases/views.py
```diff
@@ -556,7 +556,7 @@ def assign(institute_id, case_name, user_id=None, inactivate=False):
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
     link = url_for(".case", institute_id=institute_id, case_name=case_name)
     if user_id:
-        user_obj = store.user(user_id)
+        user_obj = store.user(user_id=user_id)
     else:
         user_obj = store.user(current_user.email)
```
northwestwitch, line 561:
```suggestion
        user_obj = store.user(email=current_user.email)
```

northwestwitch:
> Don't be sorry; thanks for all your hard work! It's not really an LDAP issue per se, just that there seem to be places in the code where IDs and emails are assumed to be the same. I could get around it by having the user IDs in the database be the user email addresses and setting the LDAP login attribute to email, but I'm reluctant to change our setup at this point in order to cure the symptoms.

No, you're right, we should fix the code to use differently _id and email. I'm testing this fix locally (without LDAP login) and works fine! 💯

tests/server/blueprints/cases/test_cases_templates.py
```diff
@@ -132,3 +134,23 @@ def test_sidebar_cnv_report(app, institute_obj, cancer_case_obj, user_obj):
         # It should show the expected items:
         assert "Cnv Report" in html
         assert "Coverage Qc Report" in html
+
+
+def test_sidebar_assign(app, institute_obj, case_obj, user_obj):
+    """Test the case sidebar macro items for a cancer case"""
```
northwestwitch, line 140:
```suggestion
    """Test the case sidebar macro items for a case"""
```

northwestwitch (approved):
Looks good, great fix! 🥇

PR #4054, opened 2023-08-28
Prepare minor version 4.71

northwestwitch:
### Added
- [x] Added Balsamic keys for SweGen and loqusdb local archive frequecies, SNV and SV
- [x] New filter option for Cancer variantS: local archive RD loqusdb
- [x] Show annotated observations on SV variantS view, also for cancer somatic SVs
- [x] Revel filter for variantS
- [x] Show case default panel on caseS page
- [ ] CADD filter for Cancer Somatic SNV variantS - show score
- [x] SpliceAI-lookup link (BROAD, shows SpliceAI and Pangolin) from variant page
- [x] BioNano Access server API - check projects, samples and fetch FSHD reports
### Fixed
- [x] Name of reference genome build for RNA for compatibility with IGV locus search change
- [x] Howto to run the Docker image on Mac computers in `admin-guide/containers/container-deploy.md`
- [x] Link to Weasyprint installation howto in README file
- [x] Avoid filling up disk by creating a reduced VCF file for every variant that is visualized
- [x] Remove legacy incorrectly formatted CODEOWNERS file
- [x] Restrain variant_type requests to variantS views to "clinical" or "research"
- [x] Visualization of cancer variants where cancer case has no affected individual
- [x] ProteinPaint gene link (small StJude API change)
- [x] Causative MEI variant link on causatives page
- [x] Bionano access api settings commented out by default in Scout demo config file.
- [x] Do not show FSHD button on freshly loaded cases without bionano_access individuals
### Changed
- [x] Remove function call that tracks users' browser version
- [x] Include three more splice variant SO terms in clinical filter severe SO terms
- [x] Drop old HPO term collection only after parsing and validation of new terms completes
- [x] Move score to own column on Cancer Somatic SNV variantS page
- [x] Refactored a few complex case operations, breaking out sub functionalities

**Review:**
- [ ] code approved by
- [ ] tests executed by DN, CR


northwestwitch:
I've deployed the branch on cg-vm1 for testing now!

northwestwitch:
Mmm I'm testing the change "Drop old HPO term collection only after parsing of new one completes" and seems not to be working as intended:

image

I say yes but then it fails because of the HPO bug:

image

dnil:
> But Ok, perhaps this is the intended behaviour after all. If it's just checking that the files for updating the HPO are available, and not that they can be parsed without errors?
> 
> > Mmm I'm testing the change "Drop old HPO term collection only after parsing of new one completes" and seems not to be working as intended:
> > image
> > I say yes but then it fails because of the HPO bug:

Yes, that is as intended, given that it didn't actually drop your collection? 😊 It asks permissions to drop up front, then proceeds to check all files are ok.

northwestwitch:
> Yes, that is as intended, given that it didn't actually drop your collection? 😊 It asks permissions to drop up front, then proceeds to check all files are ok, and commits to dropping only after.

The collection is dropped a bit later, but not after it checks for parsing problems. Like the error occurs on this line: https://github.com/Clinical-Genomics/scout/blob/deb5cb6de3b4e1cc65e6fe6b98659618e2dc07c4/scout/load/hpo.py#L134

While the collection is dropped a few lines above(line 126).

So basically the collection is dropped even if the loading of new terms is failing. I can try to write a fix real quick!

northwestwitch:
Hi @khurrammaqbool, do you have any cancer case with SweGen and loqusdb local archive frequencies keys that we could load into scout to test the new features?

dnil:
> Hi @khurrammaqbool, do you have any cancer case with SweGen and loqusdb local archive frequencies keys that we could load into scout to test the new features?

There are some in /home/proj/stage/cancer/validation/2023-07-27, like `civilsole`, but last time I checked they were not in cg-stage. I can check in the meantime..

northwestwitch:
Ok, I've just re-uploaded a cancer case without affected individual -> https://scout-stage.scilifelab.se/cust000/829728fam14 . Variants open fine!

dnil:
For tracking, I uploaded a version of [civilsole](https://scout-stage.scilifelab.se/cust059/G1A1471p10_Balsamic). It is v12.0.2, so has swegen and some loqus annotations but no cadd if I got it right. The SNV loqus annotations are made with so few test cases loaded to loqus that we dont really get and variants in the end, but they are visible on the SV side.

northwestwitch:
Testing the fix to the /tmp folder that fills up when visiting variants page. 

- Deployed last version's branch: 4.70
- Visited a bunch of variants:

image

- Deployed this branch: 4.71
- Visited a bunch of variants:

The only file present in the temp folder is a certificate:

image

dnil:
I deployed the servers patch for the BioNano config to stage, so we are live. Tested with 
Screenshot 2023-09-05 at 15 11 15.

Works well - see https://github.com/Clinical-Genomics/servers/pull/1107.

dnil:
Also updated to the latest version of this PR on hasta stage, to get the update individuals fix tested in the same go.

northwestwitch:
> Also updated to the latest version of this PR on hasta stage, to get the update individuals fix tested in the same go.

only one test to go now, the cadd one!

dnil:
> > Also updated to the latest version of this PR on hasta stage, to get the update individuals fix tested in the same go.
> 
> only one test to go now, the cadd one!

Yes. It does work well on local, and is really same-same as the RD one, so I would feel reasonably comfortable letting this one slip, but better to have a live test naturally. If that happens to cover also some Balsamic loqusdb archive SNVs I would also be more comfortable. Still, for both of those, it's really primarily features for a dev target for the pipeline. We are in good shape for prod.

dnil:
Ok, with that reasoning - nothing more breaking with the update, everything tested on stage except the Balsamic 13 features, which have been tested as far as possible locally, we approve the release.

dnil (approved):
🆗

northwestwitch:
Deployed on hasta prod:

image

PR #4091, opened 2023-09-21
Validate internal links action

northwestwitch:
A github action that makes sure that docs pages don't contain broken internal links. Works well on atlas!

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Break a link and see if action finds out

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/.github/workflows/validate_internal_docs_links.yml
+++ b/.github/workflows/validate_internal_docs_links.yml
@@ -1,0 +1,24 @@
+name: Validate internal documentation links
+
+"on": [pull_request]
+
+jobs:
+    tests:
+        name: mkdocs serve strict
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4.0.0
+
+            - name: Set up Python
+              uses: actions/setup-python@v4.7.0
+              with:
+                  cache: "pipenv"
+
+            - name: Set up pipenv
+              run: pip install pipenv
+
+            - name: Install package
+              run: pipenv install --dev
+
+            - name: launch mkdocs in strict mode
+              run: timeout 20 pipenv run mkdocs serve --dev-addr 127.0.0.1:8000 --strict || code=$?; if [[ $code -ne 124 && $code -ne 0 ]]; then exit $code; fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
+###
+- A GitHub action that checks for broken internal links in docs pages
 ### Fixed
 - Documentation nav links for a few documents
 - Slightly extended the BioNano Genomics Access integration docs


.github/workflows/validate_internal_docs_links.yml
```diff
@@ -0,0 +1,23 @@
+name: Validate internal documentation links
+
+"on": [pull_request]
+
+jobs:
+    tests:
+        name: mkdocs serve strict
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4.0.0
```
dnil, line 10: How about we keep it at major versions so fixes appear automagically? 😊 Which also reminds that it's time to update a few actions, like checkout - thats like two weeks out.. 😁 
```suggestion
            - uses: actions/checkout@v4
```

.github/workflows/validate_internal_docs_links.yml
```diff
@@ -0,0 +1,23 @@
...
+
+            - name: Set up Python
+              uses: actions/setup-python@v4.7.0
```
dnil, line 13:
```suggestion
              uses: actions/setup-python@v4
```
Same here; really doubt we have anything that would break with minor version changes.

.github/workflows/validate_internal_docs_links.yml
```diff
@@ -0,0 +1,23 @@
...
+              with:
+                  python-version: 3.8
+
+            -   name: Install deps
+                run: |
+                    pip install --upgrade --upgrade-strategy eager -r requirements-dev.txt -e .
+                    pip check
+
+            - name: launch mkdocs in strict mode
+              run: timeout 20 mkdocs serve --dev-addr 127.0.0.1:8000 --strict || code=$?; if [[ $code -ne 124 && $code -ne 0 ]]; then exit $code; fi
```
dnil, line 23: Nice! Isn't the `serve` and wait redundant? Like, if you just do `mkdocs build -s`, wouldn't it croak with any link issue?

dnil (changes requested):
Nice addition! In keeping with the other action uses here, let's keep to major versions.
I have a feeling the local serve is redundant - a `mkdocs build -s` or similar should be all that is needed, and doesn't stop to fire up a server and wait if all is fine?

.github/workflows/validate_internal_docs_links.yml
northwestwitch: Agreed!

.github/workflows/validate_internal_docs_links.yml
northwestwitch: I vaguely remember that using mkdocs build is where I started from in my Atlas PR, but it didn't work, I don't remember why. If you use the serve, then the timeout is necessary to set a stop at the check, because the action otherwise never ends (because it's a server running). But I'll try again with the build instead.

northwestwitch: Yes, it was this one, basically there is always this warning and it makes the action end with error:

image

northwestwitch: Mmm but wait, after I break a link the action doesn't fail. I need to fix this

northwestwitch: Ah no, there was a release (1.5) this summer where they changed the behavior of the links validation! --> https://github.com/mkdocs/mkdocs/releases

northwestwitch: Ok this is the behavior with a broken link and new validation settings in mkdocs.yml:

mkdocs build never really gets to the link validation step, because it fails before:

image

mkdocs serve works: 🆗 

image

dnil: We talked about it, but to document, I think this can be fixed with using a localhost dev_addr, just like you do in the validation with serve. It should be transparent to local users, and not affect the final site.

dnil:
I updated the config a little for `mkdocs` 1.5 - it is much more flexible on the settings for what should fail and not on build with strict enabled!

dnil:
Ok, see what you think, but I think this works and is actually better now than with 1.4 and mkdocs serve.

Missing nav:
Screenshot 2023-09-22 at 10 18 45

Reverted:
Screenshot 2023-09-22 at 10 22 02

dnil (approved):

northwestwitch:
> Ok, see what you think, but I think this works and is actually better now than with 1.4 and mkdocs serve.
> 
> Missing nav: Screenshot 2023-09-22 at 10 18 45
> 
> Reverted: Screenshot 2023-09-22 at 10 22 02

I didn't want to mess up with the mkdocs file, but apparently we are kind of forced now. I like your solution, so let's merge! 👍🏻

PR #4099, opened 2023-09-29
Reduce case view loading time by speeding up case retrieval in case_matching_causatives (maybe).

alkc:
Hej and greetings from CMD in Lund.

I'm not sure if this PR will improve case loading times for everyone, or is just very specific to our local build, so I'm submitting this PR as a "draft".

### TL;DR:

Two changes to case retrieval in `case_matching_causatives` in `scout.adapter.mongo.variant` reduces (in our build) the loading times of case pages from 10-12s to 3s in our largest institute.

### Background

We have an issue with our scout setup where browser `GET` requests for individual cases in our largest institute take around 10-12 seconds to load. It seems to be due to some sort of yet to be described hardware problem/curse that affects the data transfer rate between our mongo database and scout on our end. Still, I decided to dig around in the code to see if I could speed things up a little until the larger issue is solved.

To this end, I did some benchmarking of simple `<institute>/cases` `GET` requests to scout using `cProfile` and found that the call to `case_matching_causatives` took around 8s out of the aforementioned 10-12 seconds. Through some tinkering with how the case data is handled in that function I've managed to bring down the time to load a page in the browser to around ~3s.

### Fix

I've sketched how I went about it in the following gist: https://gist.github.com/alkc/99ca1068d42092af398f1e035224c566

This is the output of the linked script when run against our largest institute:

```
Testing queries on institute w/ 1093 events linked to 951 unique cases.
Time to process events/cases:
find_one case by case: 13.914s
find_one case by case with minimal projection: 1.357s
Read all cases into memory + minimal projection: 0.076s
find_all approach takes less time than find_one case by case at 19 cases and 0.027s
```

In short, `case_matching_causatives` iterates over all events and fetches each case together with all its data one by one via repeated `self.case()` calls. 

In the submitted update: 

1) All distinct cases associated with the event list are pulled out in one go and read into memory. 
2) A projection is applied to reduce the `other_case` objects to only contain the `_id`, `partial_causatives` and `causatives`, that is only the fields required for `case_matching_causatives` to do its work.

Again, on our end this alone reduces case loading times from 10-12s to 2-3s, but I've no idea what or if any difference it might make for other scout users. 

The script suggests that on our setup the change starts to make a difference at around 20 events/cases, but that's of course without taking into account all the other steps when building case views.

If this update is interesting, then I'm probably going to need help with the testing procedures below.

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
[...]

PR #4100, opened 2023-09-29
Reduce case view loading time by speeding up case retrieval in case_matching_causatives (maybe).

alkc:
Copying over the description from #4099 :

Hej and greetings from CMD in Lund.

I'm not sure if this PR will improve case loading times for everyone, or is just very specific to our local build, so I'm submitting this PR as a "draft".

### TL;DR:

Two changes to case retrieval in `case_matching_causatives` in `scout.adapter.mongo.variant` reduces (in our build) the loading times of case pages from 10-12s to 3s in our largest institute.

### Background

We have an issue with our scout setup where browser `GET` requests for individual cases in our largest institute take around 10-12 seconds to load. It seems to be due to some sort of yet to be described hardware problem/curse that affects the data transfer rate between our mongo database and scout on our end. Still, I decided to dig around in the code to see if I could speed things up a little until the larger issue is solved.

To this end, I did some benchmarking of simple `<institute>/cases` `GET` requests to scout using `cProfile` and found that the call to `case_matching_causatives` took around 8s out of the aforementioned 10-12 seconds. Through some tinkering with how the case data is handled in that function I've managed to bring down the time to load a page in the browser to around ~3s.

### Fix

I've sketched how I went about it in the following gist: https://gist.github.com/alkc/99ca1068d42092af398f1e035224c566

This is the output of the linked script when run against our largest institute:

```
Testing queries on institute w/ 1093 events linked to 951 unique cases.
Time to process events/cases:
find_one case by case: 13.914s
find_one case by case with minimal projection: 1.357s
Read all cases into memory + minimal projection: 0.076s
find_all approach takes less time than find_one case by case at 19 cases and 0.027s
```

In short, `case_matching_causatives` iterates over all events and fetches each case together with all its data one by one via repeated `self.case()` calls. 

In the submitted update: 

1) All distinct cases associated with the event list are pulled out in one go and read into memory. 
2) A projection is applied to reduce the `other_case` objects to only contain the `_id`, `partial_causatives` and `causatives`, that is only the fields required for `case_matching_causatives` to do its work.

Again, on our end this alone reduces case loading times from 10-12s to 2-3s, but I've no idea what or if any difference it might make for other scout users. 

The script suggests that on our setup the change starts to make a difference at around 20 events/cases, but that's of course without taking into account all the other steps when building case views.

If this update is interesting, then I'm probably going to need help with the testing procedures below.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
Hello, I'm taking a look at this PR. Your solution works (haven't really quantified the time it takes to load the cases on our stage server) but definitely doesn't break anything. 

Then I noticed that the function `case_matching_causatives` is called twice in a row [here](https://github.com/Clinical-Genomics/scout/blob/035544c4dc5bef1b8b60f9474fe80abf04b3c84a/scout/server/blueprints/cases/controllers.py#L402), line 402 and 406. Perhaps might be possible to speed up stuff even more by having just one call to retrieve both other_causatives and default_other_causatives (what a bad naming this variable has!)?

dnil:
Cool, well done! I would have guessed an `$in: []` with and N of thousands would be slower than retriving the same N one by one by index. The projection of a what was reasonably large doc shall apparently not be underestimated. I have a feeling this is exacerbated by the working set not quite fitting in memory. Also a bit curious what would happen if you reverse the action in the test script so that the cases in question have not been touched lately, in particular for the diff between 1-by-1 and minimal projection? 😸

alkc:
Thank you for looking at the PR @northwestwitch and @dnil, and sorry for the time it took me to reply (I've been AFK/on vacation the entirety of the previous week).

I'll do my best to post an update this week, starting with your request @dnil to reverse the test script as that's the fastest to implement.

@northwestwitch: I'll have a look at reducing the two calls to one and see if I come up with a solution!

alkc:
@dnil:

Switching the order of execution yields the following results:

```
Time to process events/cases:
Read all cases into memory + minimal projection: 0.101s
find_one case by case with minimal projection: 1.405s
find_one case by case: 13.694s
Queries tested on institute w/ 1098 events from 955 unique cases.
```

The `Read all cases into memory + minimal projection` time estimation is the first time in the script that the event and case collections are accessed.

Is this what you wanted to know?

dnil:
Good enough methinks, thanks! There is still one that hasn't been first (one by one, projection?) but given the magnitude of the differences, the order will very likely remain.

alkc:
@northwestwitch What do you think about the following approach?

```python
   def case(...):
    ...
    # Limit secondary findings according to institute settings
    limit_genes = store.safe_genes_filter(institute_obj["_id"])

    limit_genes_default_panels = _limit_genes_on_default_panels(
        case_obj["default_genes"], limit_genes
    )

    # Fetch secondary findings in one db call:
    case_causative_gene_filter = list(set(limit_genes + limit_genes_default_panels))
    matching_causatives = store.case_matching_causatives(
        case_obj=case_obj, limit_genes=case_causative_gene_filter
    )

    # Filter secondary findings:
    other_causatives = []
    other_causatives_in_default_panels = []
    for causative in matching_causatives:
        hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
        # Check for overlap:
        if set(hgnc_ids) & set(limit_genes):
            other_causatives.append(causative)
        if set(hgnc_ids) & set(limit_genes_default_panels):
            other_causatives_in_default_panels.append(causative)

    data = {
        "institute": institute_obj,
        "case": case_obj,
        "other_causatives": other_causatives,
        "default_other_causatives": other_causatives_in_default_panels,
         ...
    } 
```

`limit_genes` and `limit_genes_default_panels` are combined into a set and used to retrieve the causatives that match both lists in one db query.

The causatives are sorted into `other_causatives` and `default_other_causatives` based on overlap with `limit_genes` and `limit_genes_default_panels` respectively.

northwestwitch:
> @northwestwitch What do you think about the following approach?
> 
> ```python
>    def case(...):
>     ...
>     # Limit secondary findings according to institute settings
>     limit_genes = store.safe_genes_filter(institute_obj["_id"])
> 
>     limit_genes_default_panels = _limit_genes_on_default_panels(
>         case_obj["default_genes"], limit_genes
>     )
> 
>     # Fetch secondary findings in one db call:
>     case_causative_gene_filter = list(set(limit_genes + limit_genes_default_panels))
>     matching_causatives = store.case_matching_causatives(
>         case_obj=case_obj, limit_genes=case_causative_gene_filter
>     )
> 
>     # Filter secondary findings:
>     other_causatives = []
>     other_causatives_in_default_panels = []
>     for causative in matching_causatives:
>         hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
>         # Check for overlap:
>         if set(hgnc_ids) & set(limit_genes):
>             other_causatives.append(causative)
>         if set(hgnc_ids) & set(limit_genes_default_panels):
>             other_causatives_in_default_panels.append(causative)
> 
>     data = {
>         "institute": institute_obj,
>         "case": case_obj,
>         "other_causatives": other_causatives,
>         "default_other_causatives": other_causatives_in_default_panels,
>          ...
>     } 
> ```
> 
> `limit_genes` and `limit_genes_default_panels` are combined into a set and used to retrieve the causatives that match both lists in one db query.
> 
> The causatives are sorted into `other_causatives` and `default_other_causatives` based on overlap with `limit_genes` and `limit_genes_default_panels` respectively.

Sounds like a neat approach! And should be faster, given that you skip one query. Push it to this branch so we can test it!

northwestwitch:
@alkc sorry for the failed tests due to formatting stuff. Must be a bit frustrating when you start contributing to the repo.. 🙏🏻

scout/server/blueprints/cases/controllers.py
```diff
@@ -390,19 +390,28 @@ def case(store, institute_obj, case_obj):
...
-            var
-            for var in store.case_matching_causatives(case_obj=case_obj, limit_genes=limit_genes)
-        ],
-        "default_other_causatives": [
-            var
-            for var in store.case_matching_causatives(
-                case_obj=case_obj, limit_genes=limit_genes_default_panels
-            )
-        ],
+        "other_causatives": other_causatives,
```
northwestwitch, line 413:
```suggestion
        "other_causatives": [causative for causative in matching_causatives if set(causative.get("genes", [])) & set(limit_genes)]
```

northwestwitch: It's the same thing, but perhaps you could avoid a few lines of code by using this syntax instead?

scout/server/blueprints/cases/controllers.py
```diff
@@ -390,19 +390,28 @@ def case(store, institute_obj, case_obj):
...
+        "default_other_causatives": other_causatives_in_default_panels,
```
northwestwitch, line 414: Same here

northwestwitch (commented):
Looks good to me. I have a small suggestion to cut some lines of code.

I'll try the branch on our stage server now!

northwestwitch:
Tested on stage server. For instance this case having 4 matching causatives (all in default panels) - This is main branch:

image


When you use this branch becomes like this:

image


Which is kind of expected, given the changed code.

Let me find another example with causatives outside default panels!

scout/server/blueprints/cases/controllers.py
alkc: My idea with building `other_causatives` and `default_other_causatives` in the same for-loop was to only have to iterate once over the `matching_causative` cursor.

If it's not a bad idea then I can do a `matching_causatives = list(matching_causatives)` and iterate over it twice!

northwestwitch: I don't think it would take so much time to du the iteration twice. It won't be the thing that slows down the loading of the page at least. You can also consider to move the lines of code above 

```
 # Limit secondary findings according to institute settings
    limit_genes = store.safe_genes_filter(institute_obj["_id"])

    limit_genes_default_panels = _limit_genes_on_default_panels(
        case_obj["default_genes"], limit_genes
    )

    # Fetch secondary findings in one db call:
    case_causative_gene_filter = list(set(limit_genes + limit_genes_default_panels))
    matching_causatives = store.case_matching_causatives(
        case_obj=case_obj, limit_genes=case_causative_gene_filter
    )

    # Filter secondary findings:
    other_causatives = []
    other_causatives_in_default_panels = []
    for causative in matching_causatives:
        hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
        # Check for overlap:
        if set(hgnc_ids) & set(limit_genes):
            other_causatives.append(causative)
        if set(hgnc_ids) & set(limit_genes_default_panels):
            other_causatives_in_default_panels.append(causative)
```

Into a separate function that might return a tuple (other_causatives, other_causatives_in_default_panels)

That way the code would be more easy to test with a unit test..

northwestwitch:
Mmm I'm testing the branch locally and the behaviour of the matching causatives is different than in main branch:

Main branch:

To reproduce on a local instance:
1. Create. a new gene panel named panel2 based on demo panel 1.

image

2. Remove gene IFT43 from panel 2.
3. Edit the config file of case 643595 (cout/demo/643595.config.yaml) to use panel2 instead of panel1
4. Load case 643595 case using `scout --demo load scout/demo/643595.config.yaml`
6. Go to default demo case 643594 and mark causative the first SNV variant (POT1 gene)
7.  Go to default demo case 643594 and mark causative the first SNV variant (IFT43 gene)

At this point if you go to case 643595 using main branch you see this:

image

<hr>

If you instead use this branch you have this (you are basically missing the variant outside the default panel):

image

alkc:
Humm. I don't understand why the matching causatives disappear. Will look into it though, and move the new code to a testable function in the same module.

Thanks for the detailed steps to reproduce the bug!

scout/server/blueprints/cases/controllers.py
```diff
@@ -1372,3 +1368,55 @@ def matchmaker_match(request, target, institute_id, case_name):
         )
 
     return ok_responses
+
+
+def _matching_causatives(
```
northwestwitch, line 1373: 🥇

scout/server/blueprints/cases/controllers.py
```diff
@@ -1372,3 +1368,55 @@ def matchmaker_match(request, target, institute_id, case_name):
...
+    store, case_obj, other_causatives_filter, other_causatives_in_default_panels_filter
```
northwestwitch, line 1374:
```suggestion
    store, case_obj, other_causatives_filter=[], other_causatives_in_default_panels_filter=[]
```

scout/server/blueprints/cases/controllers.py
```diff
@@ -1372,3 +1368,55 @@ def matchmaker_match(request, target, institute_id, case_name):
...
+                The subset of all secondary findings found in default gene panels
+        )
+    """
+    matching_causatives_filter = []
+
+    # Combine into one query if both filters defined, otherwise don't limit genes in query:
+    if other_causatives_filter and other_causatives_in_default_panels_filter:
+        matching_causatives_filter = list(
+            set(other_causatives_filter + other_causatives_in_default_panels_filter)
+        )
```
northwestwitch, line 1403:
```suggestion
    matching_causatives_filter =  list(set(other_causatives_filter + other_causatives_in_default_panels_filter))
```

northwestwitch (approved):
Looks great and works as it should now! 💯 

image

Thanks for this fix!

alkc:
The issue with missing non default causatives was was based in combining the two gene filters `limit_genes` and `limit_genes_default_panels` into a set of unique ids when `limit_genes` was just an empty list. Passing an empty list to `case_matching_causatives` should fetch an unfiltered list of matching causatives, which is not what happens when you pass a new filter such as `new_limit_genes = list(set([] + ['foo', 'bar', 'baz'])) x_x

northwestwitch:
I've re-triggered the tests and all of them are green now. It's good to merge I think!

alkc:
> I've re-triggered the tests and all of them are green now. It's good to merge I think!

Nice! I'll merge it in a bit then.

Thank you for all the help and input!

dnil:
> > I've re-triggered the tests and all of them are green now. It's good to merge I think!
> 
> Nice! I'll merge it in a bit then.
> 
> Thank you for all the help and input!

Thank you for contributing!

PR #4109, opened 2023-10-04
Simplify and possibly speed up case report

northwestwitch:
This PR adds a functionality or fixes a bug:
- Simplify creation of case report by not filtering out many keys from case and variants

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -48,7 +48,7 @@
       </li>
     </ul>
     <span class="navbar-text">
-      <a href="{{ url_for('cases.pdf_case_report', institute_id=institute._id, case_name=case.display_name) }}"><button class="btn btn-sm btn-secondary" type="button">Download PDF</button></a>
+      <button class="navbar-brand" onclick="window.print();">Create PDF</button>
```
dnil, line 51: Hm, I like it being very lightweight and potentially saving us a lot of hassle with the pdf generation stuff. But wasn't there reason we do it like this that pdf generation was missing from vanilla windows? In MacOS, the pdf generation from "print" is rather accessible, but on windows was previously dependent on additional programs. Seems latest versions do have a "Microsoft Print to PDF", one or two dialogs into the print menu depending on which browser you use (the system print dialog has it, not eg MS Edge print dialog). The quality is not great. Most colors from the report disappear. I suspect some of this could be fixed by adjusting html to fit their capabilities, much like other pdf generators we know and pain over. But at least some of that would be needed - the users will protest as is.

dnil (changes requested):
Yes! Feels so good to reduce the complexity. But the pdf print for windows needs some attention. The quality of pdf-print from the current reports is not good. Perhaps there are some easy compatibly fixes we could do? It would save some CPU on CGVS to leave this browser-side, but e.g. warning colors on the report have to work. The first page of the report is also all some medical users see of the system, which also makes me want it nice looking.

scout/server/blueprints/cases/templates/cases/case_report.html
northwestwitch: I agree that the report shouldn't lose quality when downloaded with explorer. How did you try?

There is this simulation that you can make using safari: https://setapp.com/how-to/use-internet-explorer-for-mac

And if you export it checking the "print backgrounds" thing

image

It looks good: 

image




But of course we should test using a real explorer browser!

dnil: The thing is not the browser, it’s the pdf generator on the OS. I’m pretty sure that IE-for-Mac uses the macOS pdf print. I tested with a (hospital) windows machine, looks the same with both Chrome and Edge with “MS Print to pdf”. I did not see any flags for “background” as in your Safari simulation. If we could find a setting like that on win (ideally without the user having to find it) maybe it would look better!

northwestwitch: I bet that's because people in general don't want to export background colors and figures.. 

Just of of curiosity, if you go into the explorer settings, like advanced, is there an option to on printing of background colors and images? There should be right? And how does it look like if you check that option?

dnil: Oooh, you know what: there is an option to include "Bakgrundsgrafik" both in Chrome and Edge, which makes the printouts pretty ok. It's off by default, and hidden a few menus in, but perhaps possible to explain!

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -1404,3 +1404,30 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-warning">Part
...
+  <script>
+    function printPDF(){
+      {% if config.SQLALCHEMY_DATABASE_URI %} // include coverage report
+        fetch('{{ url_for('cases.case_coverage_data', institute_id=institute._id, case_name=case.display_name) }}', {
+          method: 'GET',
+          headers: {'Content-Type': 'application/json'}
+          })
+          .then(resp => resp.text())
+          .then(html => {
+            document.body.innerHTML = document.body.innerHTML + "<hr>" + html;
```
northwestwitch, line 1419: Added a js function that adds the coverage data to the report when you print the PDF, I don't know if it works tho, I'll test it!

northwestwitch:
Coverage content is sooooo working now: 🍾 🔥 

image

dnil (changes requested):
Oooh yes, coverage loads just great! You almost have me here, although this is going to take some user convincing. One more showstopper: the filenames. Currently the default is something like the whole URI. Should be back to the current main with just family name methinks?

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -1408,7 +1408,16 @@ <h6 class="card-subtitle mb-2 text-muted"><span class="badge badge-warning">Part
...
   <script>
+    function setDocumentName()
+    {
+      let todayDate =  new Date().toISOString().slice(0, 10);
+      document.title= '{{case.display_name}}_'+todayDate+'_scout_report'
+    }
+
     function printPDF(){
+
+      setDocumentName();
```
dnil, line 1419: 💯

northwestwitch:
> Oooh yes, coverage loads just great! You almost have me here, although this is going to take some user convincing. One more showstopper: the filenames. Currently the default is something like the whole URI. Should be back to the current main with just family name methinks?

Right! I've fixed it so the downloaded document name is the same as the old downloaded document, i-e. caseName_date_scout_report.pdf.

There is just one thing that I think I need to fix: when you download the PDF report, the body of the HTML changes (the coverage reports gets included in the new body). I'm afraid that if you download the the report again the coverage part gets duplicated. I'll fix!

northwestwitch:
Reverted the code to generate the PDF to how it was previously. It's slower than the javascript solution, but it makes sense because it's recreating the report data and formatting the madeline before exporting.

scout/server/blueprints/cases/views.py
```diff
@@ -216,26 +216,31 @@ def case_synopsis(institute_id, case_name):
...
+    institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
+    data = controllers.case_report_content(
+        store=store, institute_obj=institute_obj, case_obj=case_obj
+    )
     return dict(format="html", **data)
 
 
 @cases_bp.route("/<institute_id>/<case_name>/pdf_report", methods=["GET"])
 def pdf_case_report(institute_id, case_name):
     """Download a pdf report for a case"""
```
northwestwitch, line 239: @dnil I've reverted this function to how it originally was for the sake of keeping this link:

image

We can see how to improve/change how this report is exported to PDF in the future

dnil (approved):
Ok, let's go with that! I still think you could have left the new js, leaving a "faster-but-more-clicks" route for those that like it. It's more to maintain after all. We can still pick it up later after migrating chanjo-report: then there will be stronger motivation for removing it. Feels good with the reduced complexity! 💯

northwestwitch:
> I still think you could have left the new js, leaving a "faster-but-more-clicks" route for those that like it.

It's redundant code. I think it's better to download the PDF in one of the two ways, otherwise might be confusing, for maintenance and the users

scout/server/blueprints/cases/controllers.py
```diff
@@ -566,131 +562,63 @@ def check_outdated_gene_panel(panel_obj, latest_panel):
...
-            evaluated_variants[vt].append(var_obj)
-
-    data["variants"] = {}
-
-    for var_type in evaluated_variants:
-        decorated_variants = []
-        for var_obj in evaluated_variants[var_type]:
-            # We decorate the variant with some extra information
-            filtered_var_obj = {}
-            for feat in VARIANT_REPORT_VARIANT_FEATURES:
```
northwestwitch, line 613: @dnil this line made sure that the variant had the keys specified in VARIANT_REPORT_VARIANT_FEATURES (acmg_classification among them). The key was there but the value was None

dnil: And then the template test would work. It checked if there was a defined `acmg_classification.short_whatever`, and that wouldn't be defined for a `None` key, but not illegal enough to crash. Which not having `acmg_classification` when reaching down into its keys would be. Ok, then we have it!?

dnil: I dont think we need to keep the empty keys, do we, as long as we are careful to check them?

northwestwitch: Yes! Perhaps let's make sure that using any of the other case or variant keys won't have the same effect. We can do a quick check tomorrow, but al least now we have a list of things to check.

northwestwitch: > I dont think we need to keep the empty keys, do we, as long as we are careful to check them?

Nono I would check that they exist instead

northwestwitch: I think we should be fine with all he other keys. I gave them a quick check anyway, but I'm sure you remember that the report in the beginning used to work with case and variants objects fished from the database and passed to the template. Then we made some changes in order to just pass the necessary keys when we discussed creating the mono repo together with the web developers. That macro `somatic_variant_table(variants` was created after and all keys were there, so it was never tested with unprocessed case and variants directly from the database.

PR #4114, opened 2023-10-10
Validate disease terms and don't drop old terms before validation ends

### LINKED ISSUE #4128: Move load_phenotypes to setup
`load_phenotypes()` (previously `load.hpo.load_hpo()`) is called only during setup. It is implemented with calls to `load_hpo_terms` and `load_disease_terms`. In other scenarios we call these separately. Either remove this function entirely, making two calls directly during setup, or move it there as a helper.

northwestwitch:
- When updating disease terms, check that all terms are consistent with a DiseaseTerm model before dropping the old collection
- Renamed `scout.load.hpo` module to `scout.load.phenotype` since it's loading both HPO terms and diseases.
- Renamed `scout.load.hpo.load_hpo` funtion to `scout.load.phenotype.load_phenotypes since` it's loading both HPO terms and diseases.
- Fix #4128 

This could be of help when @TereseBo starts working with the disease terms

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Load OMIM terms as you would be in a production DB update

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


scout/load/phenotype.py
```diff
@@ -19,7 +20,7 @@
 LOG = logging.getLogger(__name__)
 
 
-def load_hpo(
+def load_phenotypes(
```
northwestwitch, line 23: Renamed, because otherwise it's confusing. this function loads BOTH HPO and OMIM terms

scout/commands/update/disease.py
```diff
@@ -16,7 +16,7 @@
 from flask.cli import current_app, with_appcontext
 
 from scout.constants import UPDATE_DISEASES_RESOURCES
-from scout.load.hpo import load_disease_terms
```
northwestwitch, line 19: scout.load.hpo was also loading OMIM terms, so I've renamed it to avoid confusion

northwestwitch:
Tested locally by loading diseases using non-demo files: 🆗 

image

scout/load/phenotype.py
dnil: Right, this is where it really convoluted, since we rely on HPO to provide us connection to diseases (OMIM/ORPHANET - maybe POSSUM?) and use both for the same command.  This one is really the setup command. We should just move that step into `scout setup` (maybe even have the setup do a download everything) and have it call the separated `scout.load.phenotype.load_hpo_terms` and `scout.load.disease.load_disease_terms` from there.

dnil, line 23:
```suggestion
def setup_phenotypes_and_diseases(
```

scout/load/phenotype.py
```diff
@@ -116,7 +117,7 @@ def load_hpo_terms(
 
         hpo_term = hpo_terms[hpo_id]
 
-        if not "genes" in hpo_term:
+        if "genes" not in hpo_term:
```
dnil, line 120: 👍

scout/load/phenotype.py
```diff
@@ -208,10 +210,9 @@ def load_disease_terms(
                 disease_id="OMIM:{0}".format(disease_number),
                 hpo_term_to_symbol=hpo_term_to_symbol,
             )
+            disease_objs.append(build_disease_term(disease_info, genes))
 
-            disease_obj = build_disease_term(disease_info, genes)
-
-            adapter.load_disease_term(disease_obj)
+    adapter.load_disease_terms(disease_objs)
```
dnil, line 215: We have the RAM nowadays. This should be faster as well, in addition to allow testing. Wouldn't about here be a good point to drop the old as well?

scout/load/phenotype.py
```diff
@@ -6,6 +6,7 @@
 
 from scout.build.disease import build_disease_term
 from scout.build.hpo import build_hpo_term
+from scout.models.disease_term import DiseaseTerm
```
dnil, line 9:
```suggestion
```

scout/commands/update/disease.py
```diff
@@ -99,11 +99,11 @@ def diseases(downloads_folder, api_key):
...
             raise click.Abort()
 
-    _check_resources(resources)
-
     LOG.info("Dropping DiseaseTerms")
     adapter.disease_term_collection.delete_many({})
-    LOG.debug("DiseaseTerms dropped")
+    LOG.debug("Disease terms dropped")
+
+    _check_resources(resources)
```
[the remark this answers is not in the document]
dnil: Oh, didn't you want to wait with dripping until after reading in the disease term files to validate them first?

scout/models/disease_term.py
```diff
@@ -0,0 +1,22 @@
+from typing import List, Optional, Union
+
+from pydantic import BaseModel, field_validator
+
+
+class DiseaseTerm(BaseModel):
```
dnil, line 6: 💯
Good stuff! But shouldn't this be in like a general `models/loading_models.py`, in analogy with the  
`models.case.case_loading_models` and `models.case.case`?

Not needed, but it would make it clearer to me to move the old class DiseaseTerm, imported from `models.phenotype_term` into a `models.disease_term`.

dnil (changes requested):
Really a Good Idea! I have some further suggestions in the same spirit of separating out pheno and disease as clearly as possible. If you feel a move the joint load to setup gets too complex, we can do that as another PR.  Setting it as "Request changes" mostly as I didn't quite get how this tests all before dropping, as the CHANGELOG states, but that should be a simple change, if any! 😊

scout/load/phenotype.py
northwestwitch: It's super fast, not so many items to load. Good point, we can drop the old terms here, so we also avoid to have the check function before, I think?

scout/commands/update/disease.py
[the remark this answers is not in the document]
northwestwitch: Yeah, I forgot to move the lines of code for dropping the collection later in the code! Well spotted! :)

scout/models/disease_term.py
northwestwitch: There is a difference between a case model (we don't have it in place yet) and a case loading model. After loading a case, you can set several key/values that are missing in the loading model..

But the disease term is pretty much the same. I think it's fine to leave it like this

dnil: Ah, but we do: https://github.com/Clinical-Genomics/scout/blob/8d212fd0863c8c22130ea3db2ec583d66ca32ee9/scout/models/phenotype_term.py#L34
I think I will loose track if the disease load one is in a file of its own, and the db data model one together with the phenotypes.

scout/load/phenotype.py
dnil: If you mean the file check function I guess it doesn't matter as much when there is validation prior to dropping, but it sure doesn't hurt just to check if the files are there before doing anything. 😊

scout/models/disease_term.py
dnil: But, I can drop an issue on it instead anyway and we can do that separately - it doesn't have to be now.

scout/load/phenotype.py
northwestwitch: I'll reintroduce that code. It would have crashed anyway 😆

scout/load/phenotype.py
northwestwitch: Let's do this in another PR, I feel one is already messy as it is..

scout/commands/update/disease.py
northwestwitch: If you check the code it should already be split, lines 54 and 60, no?

scout/commands/update/disease.py
```diff
@@ -16,7 +16,7 @@
 from flask.cli import current_app, with_appcontext
 
 from scout.constants import UPDATE_DISEASES_RESOURCES
-from scout.load.hpo import load_disease_terms
+from scout.load.phenotype import load_disease_terms
```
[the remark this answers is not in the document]
northwestwitch: I think the old scout.load.hpo should be named something general, because at the moment contains a function to load both HPO and diagnoses, one to load the HPO and one to load the OMIM (diagnoses), That's why I named it phenotype

northwestwitch:
Ok, this is ready for another review now. I've left load.phenotype.load_phenotypes that now has very little code and invokes load.disease.load_hpo_terms and load.hpo.load_hpo_terms

dnil (commented):
As on chat, all good to save the move to setup stuff, it's so well contained now. I still have some gripes with the old Class living around though. 😅

scout/adapter/mongo/omim.py
```diff
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 import logging
+from typing import List
 
 from pymongo import ASCENDING, ReturnDocument
 from pymongo.errors import DuplicateKeyError
 
 from scout.exceptions import IntegrityError
+from scout.models.disease_term import DiseaseTerm
```
dnil, line 9: This is unused, isn't it?

tests/build/test_build_disease.py
```diff
@@ -18,7 +18,7 @@ def test_build_disease_term(adapter, test_disease):
     assert disease_obj["genes"] == [17978]
     assert disease_obj["source"] == "OMIM"
 
-    assert isinstance(disease_obj, DiseaseTerm)
+    assert isinstance(disease_obj, dict)
```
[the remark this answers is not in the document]
northwestwitch: I remember that function returning a dictionary because the Pydantic class can't have a _id (or anything with underscore) among its fields, but I'll check it again, thanks!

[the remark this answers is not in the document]
northwestwitch: I mean, it can have fields with underscore, but they are ignored..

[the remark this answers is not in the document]
dnil: Ah did you intend to switch the imports?
We are at 
```from scout.models.phenotype_term import DiseaseTerm```
both here and on main, so this is presently still a test for the use of the old class `scout.models.phenotype_term.DiseaseTerm`, as it was before.

File: scout/build/disease.py [hunk 11]
```diff
@@ -72,12 +36,14 @@ def build_disease_term(disease_info, alias_genes={}):
...
-            "The following gene symbols could not be found in database: %s", hgnc_symbols_not_found
+            "The following gene symbols could not be found in database: %s",
+            hgnc_symbols_not_found,
         )
 
     disease_obj["genes"] = list(hgnc_ids)
 
-    if "hpo_terms" in disease_info:
-        disease_obj["hpo_terms"] = list(disease_info["hpo_terms"])
+    DiseaseTerm(**disease_obj)
```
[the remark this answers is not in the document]
northwestwitch: It's invoking DiseaseTerm (pydantic model class) just for the sake of validating the disease dictionary. It's not returning the pydantic class itself because it would be missing the _id key/value, required when you save the dictionary in the database

tests/build/test_build_disease.py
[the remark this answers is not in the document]
northwestwitch: Wait, the DiseaseTerm Class has changed:

Main branch:

image


This branch:

image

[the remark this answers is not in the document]
northwestwitch: AHHHH Now I see, there are 2 in 2 different files. Yes, confusing. I'll fix!

scout/build/disease.py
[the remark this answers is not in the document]
dnil: Well, yes, now it does. 😉

scout/models/phenotype_term.py
```diff
@@ -27,49 +27,3 @@ class HpoTerm(BaseModel):
...
-        self["description"] = description
-        self["source"] = source
-        self["genes"] = genes or []
-        self["inheritance"] = inheritance or []
-        self["hpo_terms"] = hpo_terms or []
-
-
-# phenotype_term is a special object to hold information on case level
-# This one might be deprecated when we skip mongoengine
-phenotype_term = dict(
```
dnil, line 71: You could drop it in `models/case/case.py` for now. 🤷‍♂️ Its not very informative, but there would be no descriptions of the db dict anywhere if it goes.

scout/models/phenotype_term.py
```diff
@@ -27,49 +27,3 @@ class HpoTerm(BaseModel):
     def set_hpo_number(self) -> "HpoTerm":
         self.hpo_number = int(self.hpo_id.split(":")[1])
         return self
-
-
-# Disease terms represent diseases collected from omim, orphanet and decipher.
-# Collected from OMIM
-class DiseaseTerm(dict):
```
dnil, line 34: See if you can't find a way to preserve the description of the db mapping object if you are not doing two different ones. Doc string on the Pydantic DiseaseTermLoader-object might be a little confusing? Just a non-functional `dict`in `scout/models/disease_term.py` perhaps?

scout/load/phenotype.py
dnil: https://github.com/Clinical-Genomics/scout/issues/4128

dnil (approved):
Ok, I think this is now in a good state. I would prefer to keep a description of the db objs as well, in addition to the new DiseaseTerm Loader Pydantic one, though rather similar, so one (probably me) does not think that is what is in the db when only seeing it. 😊

scout/models/phenotype_term.py
northwestwitch: But it's not a case model, we wouldn't even been looking for it in there. I'll add some description to the pydantic one

dnil: Just trying to be helpful here: we would want db compatible obj model descriptions for
- [ ] `phenotype_term` 
- [ ] `DiseaseTerm`

scout/models/disease_term.py
```diff
@@ -0,0 +1,23 @@
...
+    """Values that populate this class are parsed from files downloaded from OMIM:
+    https://www.omim.org/
+    """
+
+    # _id: str # same as disease_id,  will be present in the disease term saved in the database
```
northwestwitch, line 11: With this line the pydantic term is exactly describing what is saved in the database. I think it's fine. Also, what is the point of using pydantic to describe an object if you jave to have another dictionary class (that is not used anywhere) to do just the same thing

image

dnil: Well, as you like to point out, there can be a difference between the Loader versions of classes, that basically model and validate parsing of what is present on file, and models of what is supposed to be in the db or in memory. There is no set truth, strong typing or strict db model here: the things morph as we go, but good to describe where we are.

dnil:
And I saw that "Remove load_phenotype"... 😂

northwestwitch:
> And I saw that "Remove load_phenotype"... 😂

🤣 All tests strangely passed: amazing! I think I'm going to merge this monster before something gets red

scout/models/disease_term.py
northwestwitch: Yes, the load_case_model has a sense, because case load model and case db model are different. That's what I meant!

PR #4127, opened 2023-10-16
Wtforms unfreeze

dnil:

Right, so we have https://wtforms.readthedocs.io/en/3.1.x/forms/#the-form-class, saying that the formdata passed to Form _init_ needs to be pre-coerced to types. This seems to be working more or less, but not for our "betterDecimalField". Lets see what we can do..

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. use the filter form with better decimal fields set to both decimal point and commas (e.g. before switching to this branch)
1. switch to this branch and update installation, in particular of wtforms
1. retrieve both the so created filters and see them working
1. attempt to save a filter with betterdecimalfields set to both decimal point and commas

**Expected outcome**:

**Review:**
- [x] code approved by
- [x] tests executed by DN


northwestwitch:
@dnil would you like to include this code in the new release?

dnil:
It's not a priority, and it is a little finicky. 

I've tried a couple of versions of explicit coercing before Form init now and still nothing that really does the trick out of the box. All work in the sense of getting the data and correct filtering, but not rendering in the html-control. So does main I believe. I'm currently thinking about overriding the field rendering.... Enabling locales/babel would probably work in the end, but make things worse before they get better.

dnil:
Hmmm, this actually works, finally. I will try to make a minimal viable product out of it, and then extend to other `BetterDecimalFields` that get stored on `Filters`.

dnil:
Looks like this works now. Saving with comma
Screenshot 2023-10-30 at 10 59 37
loads with decimal point
Screenshot 2023-10-30 at 10 59 45

Switching to Decimal for the form is an option, and should now work as we are testing for `str` format when coercing.

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -689,11 +689,11 @@
       </div>
       <div class="col-2">
         {{ form.tumor_frequency.label(class="control-label") }}
-        {{ form.tumor_frequency(class="form-control") }}
+        {{ form.tumor_frequency(class="form-control", step="0.01", min="0", max=1) }}
```
dnil, line 692: These are not really in play - they belong to `Decimal()`, which we may or may not want to switch back to.

scout/server/blueprints/clinvar/form.py
```diff
@@ -3,6 +3,7 @@
 from flask_wtf import FlaskForm
 from wtforms import (
     BooleanField,
+    DateField,
```
northwestwitch, line 6: 👍🏻

scout/server/blueprints/variants/controllers.py
```diff
@@ -1488,7 +1489,38 @@ def persistent_filter_actions(
...
+                        filter_decimal_value_form_field in filter
+                        and type(filter[filter_decimal_value_form_field][0]) is str
+                    ):
+                        try:
+                            raw_value = decimal.Decimal(
+                                filter[filter_decimal_value_form_field][0].replace(",", ".")
+                            )
+                        except (decimal.InvalidOperation, ValueError):
+                            raw_value = filter[filter_decimal_value_form_field][0]
+                            raise ValueError("Not a valid decimal value")
```
northwestwitch, line 1516: Won't this make the page crash?

scout/server/blueprints/variants/controllers.py
```diff
@@ -1488,7 +1489,38 @@ def persistent_filter_actions(
...
         if filter_obj is not None:
-            form = FiltersFormClass(MultiDict(filter_obj))
+
+            def _coerce_formdata(filter: dict) -> MultiDict:
+                """According to wtforms docs https://wtforms.readthedocs.io/en/3.1.x/forms/#the-form-class,
+                formdata passed to Form init needs to be pre-coerced to types. Hence especially the replace
+                of locale specific chars in numbers on class process_formdata would not be enough for the
+                better_decimal form fields.
+                """
+                for filter_decimal_value_form_field in [
```
northwestwitch, line 1499:
```suggestion
                for decimal_value_form_field in [
```

scout/server/blueprints/variants/controllers.py
```diff
@@ -1488,7 +1489,38 @@ def persistent_filter_actions(
...
+                    "cadd_score",
+                    "revel",
+                    "gnomad_frequency",
+                    "control_frequency",
+                    "tumor_frequency",
+                ]:
+                    if (
```
northwestwitch, line 1506: You can skip the try and catch if you check that the string is a number.

Something like this perhaps?

```
if filter_decimal_value_form_field not in filter:
  continue

filter_value = filter_decimal_value_form_field][0]

if isinstance(filter_value, str) and filter_value.replace(",","").isnumeric(): # note that I'm replacing comma with empty string
```

scout/server/blueprints/variants/forms.py
```diff
@@ -76,13 +76,17 @@ class BetterDecimalField(DecimalField):
     """
 
     def process_formdata(self, valuelist):
-        if valuelist:
+        if not valuelist:
```
northwestwitch, line 79: 👍🏻

northwestwitch (commented):
Looks good, but I have a bad feeling about that raise ValueError. I would remove it

scout/server/blueprints/variants/controllers.py
dnil: I didnt think so - we used to catch them, but now I moved it up slightly to before the form cerationg, so could be that the catching happens in the WTForm. I'll modify - doesn't hurt!

scout/adapter/mongo/filter.py
```diff
@@ -30,16 +32,18 @@ def retrieve_filter(self, filter_id):
...
+    ) -> str:
         """Store away filter settings for later use.
 
         Arguments:
-            filter_obj(MultiDict)
-            institute_obj(dict)
-            user_obj(dict)
-            case_obj(dict)
-            category(str): in ['cancer', 'snv', 'str', 'sv']
+            category(str): in ['cancer', 'snv', 'str', 'sv', ...]
```
northwestwitch, line 46: 👍🏻

scout/server/blueprints/variants/controllers.py
```diff
@@ -1551,7 +1552,41 @@ def persistent_filter_actions(
...
-            form = FiltersFormClass(MultiDict(filter_obj))
+
+            def _coerce_formdata(filter: dict) -> MultiDict:
+                """According to wtforms docs https://wtforms.readthedocs.io/en/3.1.x/forms/#the-form-class,
+                formdata passed to Form init needs to be pre-coerced to types. Hence especially the replacement
+                of locale specific chars in numbers on class process_formdata would not be enough for the
+                better_decimal form fields.
+                """
+                for decimal_value_form_field in [
+                    "cadd_score",
```
northwestwitch, line 1563: To have them hardcoded like this is a bit dangerous I think. For instance now that we added other filters like the fusion ones, or any other field that we's add in the future won't be covered if we don't remember to update this list. What about looping through the key/values of the filter and fix them if needed instead?

northwestwitch (commented):
Check out my comment!

scout/server/blueprints/variants/controllers.py
dnil: Hm, dunno, this is working and not very strange. The alternative would be to hard code the filter types this should go with since the "pre" construction doesn't allow overriding some method on the specific Class. But sure, we can give it a spin, I'm fine with either!

scout/server/blueprints/variants/controllers.py
```diff
@@ -1551,7 +1554,47 @@ def persistent_filter_actions(
...
+                of locale specific chars in numbers on class process_formdata would not be enough for the
+                better_decimal form fields.
+
+                Check all filters in the current FiltersFormClass by making an instance, finding any BetterDecimalField
+                or DecimalFields. Replace "," with "." in any string value, and check that the remaining string will
+                convert to a number before finally instantiating the form with values.
+                """
+
+                test_form = FiltersFormClass()
+                for decimal_value_form_field in dir(FiltersFormClass):
```
northwestwitch, line 1570: 🥇

northwestwitch (approved):
Looks god. Thanks for the fix! I assume you tested this latter change.

PR #4134, opened 2023-10-18
Stricter genome build validation when loading case + missing tests

northwestwitch:
- Do a stricter validation of what's provided as genome build on case load
- Genome build param is mandatory and not defaulted to either "37" or "38"

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DNIL
- [x] tests executed by CR


scout/demo/643594.config.yaml
```diff
@@ -140,4 +140,4 @@ rank_model_version: '1.20'
 sv_rank_model_version: '1.5'
 rank_score_threshold: -100
 analysis_date: 2016-10-12 14:00:46
-human_genome_build: 37
+human_genome_build: '37'
```
northwestwitch, line 143: Let's set a good example

docs/admin-guide/loading-case.md
```diff
@@ -60,7 +60,7 @@ rank_model_version: '1.12'
 sv_rank_model_version: '1.5'
 rank_score_threshold: -100
 analysis_date: 2016-10-12 14:00:46
-human_genome_build: 37
+human_genome_build: '37'
```
dnil, line 63: 👍

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Upgrade pymongo to version 4
 - When updating disease terms, check that all terms are consistent with a DiseaseTerm model before dropping the old collection
 - Better separation between modules loading HPO terms and diseases
+- Stricter validation of mandatory genome build key when loading a case. Allowed values are ['37','38',37,38]
```
dnil, line 37: Hm, why make it mandatory suddenly if we have had a default forever? I would not have minded it from the start - it is rather crucial info - but isn't the change sudden?

scout/models/case/case_loading_models.py
```diff
@@ -381,7 +381,7 @@ class CaseLoader(BaseModel):
     gene_fusion_report: Optional[str] = None
     gene_fusion_report_research: Optional[str] = None
     gene_panels: Optional[List[str]] = []
-    genome_build: Union[str, int] = Field(38, alias="human_genome_build")
+    genome_build: str
```
dnil, line 384: I don't follow: wouldn't this drop the "human_genome_build" completely? Not that we use it locally, but it did exist? Perhaps deprecate first?

scout/models/case/case_loading_models.py
```diff
@@ -466,18 +466,19 @@ def set_synopsis(cls, synopsis: Optional[Union[str, List]]) -> Optional[str]:
             synopsis = ". ".join(synopsis)
         return synopsis
 
-    @field_validator("genome_build", mode="before")
+    @model_validator(mode="before")
     @classmethod
-    def format_build(cls, genome_build: Union[str, int]) -> str:
+    def format_build(cls, values) -> "CaseLoader":
```
dnil, line 471: Ah, ok, so this runs first, and coerces ints to str?! It takes a little longer to read fully, but it makes it very clear at a glance which type is preferred while coping with other values more silently. That is good Pydantic to me.

tests/parse/test_parse_case.py
```diff
@@ -452,3 +452,57 @@ def test_missing_mandatory_config_key(scout_config, key):
     ## THEN calling parse_case_config() will raise ConfigError
     with pytest.raises(ValueError):
         parse_case_config(scout_config)
+
+
+@pytest.mark.parametrize("key", ["genome_build", "human_genome_build"])
+def test_load_case_wrong_genome_build(scout_config, key):
+    """Test loading a case with a genome build of wrong format."""
```
dnil, line 459: 👍

scout/models/case/case_loading_models.py
```diff
@@ -466,18 +466,19 @@ def set_synopsis(cls, synopsis: Optional[Union[str, List]]) -> Optional[str]:
...
         """Format the genome build so it will be saves as either '37' or '38'."""
-        str_build = str(genome_build)
+        str_build = str(values.get("genome_build") or values.get("human_genome_build", ""))
```
dnil, line 473: And same here I guess, though the alias is hidden rather deep. Considering it is the key we use both in demos, docs and in CG prod I find it may need a little explanation - maybe in that docstring above?

tests/conftest.py
```diff
@@ -1376,7 +1376,6 @@ def minimal_config(request, scout_config):
     config.pop("rank_model_version")
     config.pop("rank_score_threshold")
     config.pop("sv_rank_model_version")
-    config.pop("human_genome_build")
```
dnil, line 1379: I find this uneasy at this point, though in principle it would have been better from the start. But we have strict value checking now, and the default hg38 is right for most.

tests/parse/test_parse_case.py
```diff
@@ -452,3 +452,57 @@ def test_missing_mandatory_config_key(scout_config, key):
...
+def test_load_case_missing_genome_build(scout_config):
+    """Test loading a case with without providing a genome build."""
+
+    ## GIVEN a scout_config (dict) containing user case information
+    ## WHEN deleting the genome_build key
+    scout_config.pop("human_genome_build")
+
+    ## THEN calling parse_case_config() will raise ConfigError
+    with pytest.raises(ValueError):
+        parse_case_config(scout_config)
```
dnil, line 480: Again, I find this change a bit abrupt.

dnil (changes requested):
Good fix, but please convince me why we should drop the default build now?
Otherwise ok, just extend the docstring to describe what was moved from the more obvious model to the validator!

northwestwitch:
> Good fix, but please convince me why we should drop the default build now?

You would undoubtedly concur that the genome build is one of the most critical parameters within a case. It plays a huge role in virtually all variants pages. In the past we had many issues that arose when the key was either missing or wrongly configured.
What prompted  to make this key mandatory for me  was the fact that when the key/value was missing, one of the many 37 cases we have on the stage server was loaded without errors. Perhaps you don't realise that the key is missing and you would treat this case as a case with build 37, but in the database the case is classified as 38. This is likely to introduce errors in the variants pages, or when fetching the variants genes 

Having full control over such a critical parameter is a comforting thought I think. And ot's not that you don't know the build of a case when you write down its config file.

scout/models/case/case_loading_models.py
northwestwitch: Yes, it fixes it before the validation happens

scout/models/case/case_loading_models.py
northwestwitch: Nono, it's still used, check the validation down, it collects the build from wither genome_build or human_genome_build

scout/models/case/case_loading_models.py
northwestwitch: Fixed the docstriing

dnil:
> > Good fix, but please convince me why we should drop the default build now?
> 
> You would undoubtedly concur that the genome build is one of the most critical parameters within a case. It plays a huge role in virtually all variants pages. In the past we had many issues that arose when the key was either missing or wrongly configured. What prompted to make this key mandatory for me was the fact that when the key/value was missing, one of the many 37 cases we have on the stage server was loaded without errors. Perhaps you don't realise that the key is missing and you would treat this case as a case with build 37, but in the database the case is classified as 38. This is likely to introduce errors in the variants pages, or when fetching the variants genes
> 
> Having full control over such a critical parameter is a comforting thought I think. And it's not that you don't know the build of a case when you write down its config file.

Right, I asked for an argument and got it. I can envision quite a few issues for ppl after installing this, but they will at least not be silent, and they can hopefully revert quickly if need be.

northwestwitch:
> > > Good fix, but please convince me why we should drop the default build now?
> > 
> > 
> > You would undoubtedly concur that the genome build is one of the most critical parameters within a case. It plays a huge role in virtually all variants pages. In the past we had many issues that arose when the key was either missing or wrongly configured. What prompted to make this key mandatory for me was the fact that when the key/value was missing, one of the many 37 cases we have on the stage server was loaded without errors. Perhaps you don't realise that the key is missing and you would treat this case as a case with build 37, but in the database the case is classified as 38. This is likely to introduce errors in the variants pages, or when fetching the variants genes
> > Having full control over such a critical parameter is a comforting thought I think. And it's not that you don't know the build of a case when you write down its config file.
> 
> Right, I asked for an argument and got it. I can envision quite a few issues for ppl after installing this, but they will at least not be silent, and they can hopefully revert quickly if need be.

This change would merely prevent the loading of cases if a facility generates config files without specifying the genome build. How often have we had to modify our software due to changes in its dependencies? I see this change as a security enhancement.

PR #4142, opened 2023-10-19
Patch version 4.72.1

northwestwitch:
### Fixed
- [x] Jinja filter that renders long integers
- [x] Case cache when looking for causatives in other cases causing the server to hang

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
Tested on stage, this branch:

image


Main branch:

image

dnil:
> Tested on stage, this branch:
> 
> image
> Main branch:
> 
> image

Those spaced-number columns are a bit overfull and not really readable - we should really fix it or revert!

vwirta:
> > Tested on stage, this branch:
> > image
> > Main branch:
> > image
> 
> Those spaced-number columns are a bit overfull and not really readable - we should really fix it or revert!

what about adding a comma sign?

dnil:
> > > Tested on stage, this branch:
> > > image
> > > Main branch:
> > > image
> > 
> > 
> > Those spaced-number columns are a bit overfull and not really readable - we should really fix it or revert!
> 
> what about adding a comma sign?

Was that passive agressive Valtteri, or in good faith? Ill bite a little.. Commas, apostrophes or multiplication signs in numbers are really weird things and frowned upon by IUPAC, IBWM etc. See e.g. https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping for a starting discussion. 😁

dnil:
> Commas or dots on the other hand would solve the issue of the numbers being displayed on different lines..

And having no space, let alone no comma or other number format disrupting symbol in there would also solve that one, while being shorter and easier to read.

northwestwitch:
> and easier to read.

Not easier to read for me,. but shorter, yes!

dnil (approved):
We still need to fix those spaces-in-numbers, but lets go with this once we have the patch in.

PR #4155, opened 2023-10-20
Fix #4145 - lighten by projecting case

### LINKED ISSUE #4145: Partially reintroduce case load speedup with projection
While for us, with 15k cases in institutes, memory ran out for loading the full case set, project based thin cases as an option on the adapter should be a very decent compromise, especially for Lund considering #4136.

dnil:

Lighten case use by projecting queries to only return actually used vars.

I certainly do not mind more testing; automatic tests pass, clicking around on local and scout-stage seems fine (only some views really get a big speedup, but memory looks a bit better on `top` - few peaks.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN


scout/adapter/mongo/variant.py
```diff
@@ -557,9 +570,9 @@ def case_matching_causatives(self, case_obj, limit_genes=None):
                 # exclude causatives from the same case
                 continue
 
-            other_case = self.case(var_event["case"])
+            other_case = self.case(var_event["case"], CASE_CAUSATIVES_PROJECTION)
```
dnil, line 573: This corresponds to the almost as good solution with projection from the gist given in #4100.

dnil:
There are soo many more places one could project down the objects. This is a start. If you see some that you'd like fixed, just tell me or make an issue. I'm going to call it now as well, and leave this up for review.

scout/adapter/mongo/case.py
```diff
@@ -686,18 +697,17 @@ def update_dynamic_gene_list(
         )
         return updated_case
 
-    def case(self, case_id=None, institute_id=None, display_name=None):
+    def case(
+        self,
+        case_id: str = None,
+        institute_id: str = None,
+        display_name: str = None,
+        projection: Dict = None,
```
alkc, line 705: Very good idea!

dnil: Good call; we haven't had to worry much about memory on the web server, only db server before, but it is definitely becoming an issue. Very timely!

scout/adapter/mongo/case.py
```diff
@@ -45,7 +46,10 @@ def cases_by_phenotype(self, phenotype_terms, owner, case_id=None):
             set_1 = set_1.union(set(hpo_term.get("all_ancestors", [])))
         # Need to control what cases to look for here
         # Fetch all cases with phenotypes
-        for case in self.cases(phenotype_terms=True, owner=owner):
+        CASES_BY_PHENOTYPE_PROJECTION = {"phenotype_terms": 1}
```
northwestwitch, line 49: 👍🏻

scout/adapter/mongo/case.py
```diff
@@ -288,28 +297,29 @@ def _update_case_id_query(self, query, id_list):
...
-        group=None,
-        pinned=False,
-        cohort=False,
-        name_query=None,
-        yield_query=False,
-        within_days=None,
-        assignee=None,
-        verification_pending=None,
-        has_clinvar_submission=None,
+        owner: str = None,
```
northwestwitch, line 300: 💯 Perhaps add Optional where the key is optional )most of them I guess?)

scout/adapter/mongo/case.py
```diff
@@ -288,28 +297,29 @@ def _update_case_id_query(self, query, id_list):
...
+        group: ObjectId = None,
+        pinned: bool = False,
+        cohort: bool = False,
+        name_query: str = None,
+        yield_query: bool = False,
+        within_days: int = None,
+        assignee: str = None,
+        verification_pending: bool = None,
+        has_clinvar_submission: bool = None,
+        projection: Dict[str, Any] = None,
```
northwestwitch, line 322:
```suggestion
        projection: Optional[Dict[str, int]] = None,
```
Or can the values of projections be different than integers?

scout/adapter/mongo/case.py
```diff
@@ -596,7 +607,7 @@ def case_ids_from_group_id(self, group_id):
         return [
             case["_id"]
             for case in self.case_collection.find(
-                {"group": {"$elemMatch": {"$eq": group_id}}}, {"_id": 1}
+                {"group": {"$elemMatch": {"$eq": group_id}}}, ID_PROJECTION
```
northwestwitch, line 610: 💯

scout/adapter/mongo/case.py
```diff
@@ -686,18 +697,17 @@ def update_dynamic_gene_list(
         )
         return updated_case
 
-    def case(self, case_id=None, institute_id=None, display_name=None):
+    def case(
+        self,
+        case_id: str = None,
```
northwestwitch, line 702: Add Optional to the Optional params perhaps?

scout/adapter/mongo/case.py
```diff
@@ -288,28 +297,29 @@ def _update_case_id_query(self, query, id_list):
...
+        pinned: bool = False,
+        cohort: bool = False,
+        name_query: str = None,
+        yield_query: bool = False,
+        within_days: int = None,
+        assignee: str = None,
+        verification_pending: bool = None,
+        has_clinvar_submission: bool = None,
+        projection: Dict[str, Any] = None,
     ):
```
northwestwitch, line 323:
```suggestion
    ) -> List[dict]:
```

scout/adapter/mongo/variant.py
```diff
@@ -516,14 +527,14 @@ def institute_causatives(self, institute_obj, limit_genes=None):
...
         for case_event in causative_events:
-            case_obj = self.case(case_event.get("case"))
+            case_obj = self.case(case_event.get("case"), projection=CASE_CAUSATIVES_PROJECTION)
             if case_obj is None:
                 continue
 
             for var_id in case_obj.get("causatives", []):
                 causative_ids.add(var_id)
 
-            for var_id, _ in case_obj.get("patial_causatives", {}).items():
```
northwestwitch, line 526: Whoops, was this a bug?

scout/adapter/mongo/variant_events.py
```diff
@@ -187,14 +188,11 @@ def cancel_verification(self, institute, case, user, link, variant):
         )
         return updated_variant
 
-    def sanger_ordered(self, institute_id=None, user_id=None, case_id=None):
+    def sanger_ordered(
+        self, institute_id: str = None, user_id: str = None, case_id: str = None
+    ) -> List[Dict[str, Any]]:
```
northwestwitch, line 193:
```suggestion
    ) -> List[Dict[str, List]]:
```

scout/adapter/mongo/variant_events.py
```diff
@@ -216,14 +214,11 @@ def sanger_ordered(self, institute_id=None, user_id=None, case_id=None):
         sanger_ordered = [item for item in results]
         return sanger_ordered
 
-    def validated(self, institute_id=None, user_id=None, case_id=None):
+    def validated(
+        self, institute_id: str = None, user_id: str = None, case_id: str = None
+    ) -> List[Dict[str, Any]]:
```
northwestwitch, line 219:
```suggestion
    ) -> List[Dict[str, List]]:
```

scout/constants/indexes.py
```diff
@@ -131,3 +131,5 @@
         ),
     ],
 }
+
+ID_PROJECTION = {"_id": 1}
```
northwestwitch, line 135: Good to have it here! 👍🏻

scout/server/blueprints/institutes/controllers.py
```diff
@@ -431,6 +439,20 @@ def cases(store, request, institute_id):
     limit = int(request.args.get("search_limit")) if request.args.get("search_limit") else 100
     data["form"] = populate_case_filter_form(request.args)
 
+    ALL_CASES_PROJECTION = {
```
northwestwitch, line 442: Very nice! But what about the fields that should be in the "share" column of cases page?

northwestwitch (changes requested):
This is going to speed up cases pages quite a bit! I haven't tested all the implications inherent to the code changes so we should certainly do some extensive testing before the new release. One thing that I noticed is that you are not collecting data to include in the "shared" column of cases page. That would result in that column to always be empty

scout/adapter/mongo/case.py
dnil: I've seen `bool` as well; in fact I think that is what they prefer in modern pymongo, but since we have so many places with ints, it still seems to work, and documentation often has `int` since mongosh and others use it I figured we could keep both.

scout/adapter/mongo/case.py
dnil: Can do!

scout/adapter/mongo/variant.py
dnil: Mmm, at least I haven't seen `patial_causatives` elsewhere! 😆 Its in the list, so presumably not much harm done.

scout/server/blueprints/institutes/controllers.py
dnil: Thanks, I'll doublecheck - pretty sure I saw the icon, but maybe that was a different version. 🙄

scout/adapter/mongo/case.py
dnil: It's Optional with a capital O, mind you!

scout/adapter/mongo/case.py
dnil: (I guess maybe it could be Union[int, bool], but when in doubt.. 😬)

scout/adapter/mongo/case.py
dnil: Ah, it used to yield a dict. Lets see...

dnil: Ah, sry, cases. Right, that can do either of two, depending on that yield query parameter. If I had a really quiet week I would refactor that, splitting it into two functions, with one calling the other if needed.

dnil: I suppose it could be an optional[union[List[dict], Dict] but seriously..

scout/server/blueprints/institutes/controllers.py
dnil: I remembered correctly at least:
Screenshot 2023-10-23 at 16 58 08
But, Ill check if Im missing any of them!

dnil:
> This is going to speed up cases pages quite a bit! I haven't tested all the implications inherent to the code changes so we should certainly do some extensive testing before the new release. One thing that I noticed is that you are not collecting data to include in the "shared" column of cases page. That would result in that column to always be empty

Actually the clinvar submissions are added on the fly and are not (supposed to be) stored on the case_collection, hence the column was populated on my test. But you are very right about beacon and mmr_submission - I have added them now.

scout/adapter/mongo/case.py
```diff
@@ -288,29 +297,30 @@ def _update_case_id_query(self, query, id_list):
...
-        pinned=False,
-        cohort=False,
-        name_query=None,
-        yield_query=False,
-        within_days=None,
-        assignee=None,
-        verification_pending=None,
-        has_clinvar_submission=None,
-    ):
+        owner: Optional[str] = None,
```
northwestwitch, line 300: 🥇with the Optional you see right away what is required and what is not!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -431,6 +439,22 @@ def cases(store, request, institute_id):
...
+        "analysis_date": 1,
+        "assignees": 1,
+        "beacon": 1,
+        "case_id": 1,
+        "display_name": 1,
+        "individuals": 1,
+        "is_rerun": 1,
+        "is_research": 1,
+        "mme_submission": 1,
```
northwestwitch, line 451: 👍🏻

northwestwitch (approved):
Looks good. Good job! 💯

scout/adapter/mongo/case.py
dnil: Ah, but that is the thing - you don’t - see the pydantic manual quote. The “Optional[x]” typing term simply means “Union[x,None]” which may or may not coincide with being an optional argument in the Python function sense. Consider eg an optional argument for an integer with an integer default value. It would still not be appropriate to use Optional on it. Or conversely a required parameter that has no default but can legitimately take the None value. It is not optional but it is appropriate to typing mark it Optional. To me, it would have made more sense to name the Union[x, None] something else, say “Discontinuous” or “Punctuated”.

dnil: “Noneable”? 😸

northwestwitch: > “Noneable”? 😸

Open an issue to propose it in their repo! 😆

northwestwitch: > Ah, but that is the thing - you don’t - see the pydantic manual quote. The “Optional[x]” typing term simply means “Union[x,None]” which may or may not coincide with being an optional argument in the Python function sense.

In fact in this code I would have expected all arguments with a default being set to Optional, but whatever!

dnil: Right? Thats one direction it is silly. A bool with a default is not Optional; it can be False but not None. It is however optional - it is not a required argument in a function  call. The difference a capital letter makes…

PR #4189, opened 2023-11-01
Fix #4182 - export managed variants from CLI

### LINKED ISSUE #4182: Export managed variants from scout
MIP uses a hard filter (VAF > 70%) to remove common/reference variants from the vcf file. Still some very common variants are of interest, e.g PEX6 c.*442_445delTAAA, and these gets filtered currently. Scout has a whitelist of variants used to ensure that these are always loaded. 

It would be great if there was a scout command to export this list to vcf file, preferably with a version tag or date of export in the vcf header.

Relating to https://github.com/Clinical-Genomics/MIP/issues/2062

dnil:

Export managed variants from CLI - either as VCF or json. Options to pick institute and build are not really in play yet, but are anticipated for future use.

Screenshot 2023-11-01 at 15 22 56
Screenshot 2023-11-01 at 15 26 03

Perhaps still could use a "category" selector on the CLI command as well from the start, but this closes the original Issue.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN


scout/adapter/mongo/managed_variant.py
```diff
@@ -92,16 +93,16 @@ def managed_variant(self, document_id):
...
 
         Arguments:
-            display_id(str): chrom_pos_ref_alt_category_build
+            managed_variant_id(str): chrom_pos_ref_alt_category_build
                 category: "snv", "cancer_snv" - "sv", "cancer_sv" possible but not expected
                 build: "37" or "38"
 
         Returns:
-            ManagedVariant
+            dict (compare scout.models.ManagedVariant)
```
northwestwitch, line 105: 👍🏻

scout/adapter/mongo/managed_variant.py
```diff
@@ -113,36 +114,36 @@ def find_managed_variant_id(self, variant_id):
         """Fetch eg search for a managed variant with the encoded positional variant_id.
 
         Arguments:
-            display_id(str): chrom_pos_ref_alt_category_build
+            chromosome, position, reference, alternative, "clinical"
+            variant_id(str): md5(chrom_pos_ref_alt_category_type)
                 category: "snv", "cancer_snv" - "sv", "cancer_sv" possible but not expected
-                build: "37" or "38"
+                type: "clinical"
```
northwestwitch, line 120: Perhaps add types also on line 113?

scout/adapter/mongo/managed_variant.py
```diff
@@ -113,36 +114,36 @@ def find_managed_variant_id(self, variant_id):
...
-
         Returns:
             managed_variants(pymongo.Cursor)
 
         """
-
         query = {"category": {"$in": category}}
+        if build:
+            query["build"] = build
+        if institute:
```
northwestwitch, line 145: 🥇

scout/commands/export/variant.py
```diff
@@ -105,6 +105,41 @@ def verified(collaborator, test, outpath=None):
...
+)
+@click.option(
+    "-b",
+    "--build",
+    help="Specify what build to export variants from. Defaults to all variants.",
+)
+@json_option
+@with_appcontext
+def managed(collaborator: str, build: str, json: bool):
+    """Export causatives for a collaborator in .vcf format"""
```
northwestwitch, line 122:
```suggestion
    """Export causatives for a collaborator in .vcf or json format."""
```

scout/commands/export/variant.py
```diff
@@ -105,6 +105,41 @@ def verified(collaborator, test, outpath=None):
     return written_files
 
 
+@click.command("managed", short_help="Export managed variants")
+@click.option(
+    "-c",
+    "--collaborator",
+    help="Specify what collaborator to export variants from. Defaults to all variants.",
+)
+@click.option(
```
northwestwitch, line 114: This could be treated as en eval type of input, so you force the user to provide the right build values from a list, -> https://click.palletsprojects.com/en/8.1.x/options/#choice-options . I know build is not used, but since you added it..

scout/commands/export/variant.py
```diff
@@ -176,11 +210,11 @@ def get_vcf_entry(variant_obj, case_id=None):
     variant_string = "{0}\t{1}\t{2}\t{3}\t{4}\t{5}\t{6}\t{7}".format(
         variant_obj["chromosome"],
         variant_obj["position"],
-        variant_obj["dbsnp_id"],
+        variant_obj.get("dbsnp_id", "."),
```
northwestwitch, line 213: 👍🏻

scout/export/variant.py
```diff
@@ -2,17 +2,19 @@
...
-def export_variants(adapter, collaborator, document_id=None, case_id=None):
+def export_variants(
+    adapter: MongoAdapter, collaborator: str, document_id: str = None, case_id: str = None
+) -> dict:
     """Export causative variants for a collaborator
 
     Args:
-        adapter(MongoAdapter)
-        collaborator(str)
         document_id(str): Search for a specific variant
```
northwestwitch, line 18: You could also remove the Args thing and provide a better docstring for this function.

scout/export/variant.py
```diff
@@ -40,14 +42,46 @@ def export_variants(adapter, collaborator, document_id=None, case_id=None):
         # Add chromosome and position to prepare for sorting
         variants.append((chrom_int, variant_obj["position"], variant_obj))
 
-    # Sort varants based on position
+    # Sort variants based on position
```
northwestwitch, line 45: 😆

scout/export/variant.py
```diff
@@ -40,14 +42,46 @@ def export_variants(adapter, collaborator, document_id=None, case_id=None):
...
+    category: list = ["snv", "sv"],
+) -> ManagedVariant:
+    """Export managed variants, optionally for a given institute or variant category (["snv", "cancer_sv", ...])"""
+
+    # Store the variants in a list of tuples for sorting
+    variants = []
+
+    managed_variants = adapter.managed_variants(category=category, build=build, institute=institute)
+
+    for managed_variant_obj in managed_variants:
```
northwestwitch, line 66: What about sorting the results of the database query by chrom and position? You get 1, 11, 12 and so on? Could [this](https://stackoverflow.com/questions/57406820/mongodb-sort-gives-incorrect-result-with-alphanumeric-values) help in such case?

northwestwitch (approved):
Looks good. Useful feature to move managed vars around. I have a couple of suggestions, check if it's helpful!

scout/adapter/mongo/managed_variant.py
dnil: Not sure I follow: that is a for-quick-reference expansion of the variant_id md5 string stuff, not a formal argument. It might be neat to have a strict type for it I guess. I can add str att least! 😁

scout/export/variant.py
dnil: I think that could work just as well, but not completely convinced it is better. I was about to say it would miss the index, as the index would be string based by default, not locale collated or whatever, but it seems we have no index on it so all good. That does appear a possible oversight to me for other uses! I guess the Big Thing for matching them is the churn of the incoming cases variants, plus the matching managed, but still. This collection is small, but the index would be really small and save some collection scans.

scout/commands/export/variant.py
dnil: Indeed! Only it is actually managed variants.. 😆

scout/export/variant.py
dnil: The more I think about it, the more I like it. I'll just add the collated index. It will take little space, and drop a lot of old code that will run efficiently on the mongo server instead of on the web server.

dnil:
Looks good with collation sort: X is after the numerical ones. 
Screenshot 2023-11-07 at 16 11 02

dnil:
Oh, noooo. You know what will not work? MT. It will be before the sex chromosomes, instead of after as it should. This could actually be bad for some tools, but I guess this file will not be indexed?

dnil:
Meh, I think I will ignore it: the VCF_HEADER is incomplete at least in the sense of missing `#contig`s anyway, so maybe this is overkill. At least we are aware and can adjust if needed. 😊

dnil:
We'll have to invent the `genetics_GENOME` locale.. 😄

scout/commands/export/variant.py
northwestwitch: We almost need the second reviewer 😆

dnil: Hehe, the obnoxious and scary one! Just letting stuff sit for one day until we completely forget what it was about and rereading does wonders. 😆

northwestwitch:
> Meh, I think I will ignore it: the VCF_HEADER is incomplete at least in the sense of missing `#contig`s anyway, so maybe this is overkill. At least we are aware and can adjust if needed. 😊

Yes, let's fix if somebody reports it as an error! 👍🏻

Issue #4269, opened 2023-12-08
Add additional filters for fusion vcf


fevac:
- Add filtering by: gene, chr, junction reads, split reads, score, tools
- Gene filter should work on gene names too (for those fusions not having gene objects associated)

PR #4278, opened 2023-12-12
Update fusion filters

fevac:
This PR closes https://github.com/Clinical-Genomics/scout/issues/4269

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -761,9 +761,7 @@ class QueryHandler(object):
                     {"fusion_score": {"$gte": float(query.get("fusion_score"))}}
                 )
             if criterion == "ffpm":
-                mongo_secondary_query.append(
-                    {"samples.0.ffpm": {"$gte": float(query.get("ffpm"))}}
-                )
+                mongo_secondary_query.append({"samples.0.ffpm": {"$gte": float(query.get("ffpm"))}})
             if criterion == "junction_reads":
                 mongo_secondary_query.append(
                     {"samples.0.read_depth": {"$gte": float(query.get("junction_reads"))}}
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -756,6 +756,23 @@ class QueryHandler(object):
                 mongo_secondary_query.append({"cosmic_ids": {"$exists": True}})
                 mongo_secondary_query.append({"cosmic_ids": {"$ne": None}})
 
+            if criterion == "fusion_score":
+                mongo_secondary_query.append(
+                    {"fusion_score": {"$gte": float(query.get("fusion_score"))}}
+                )
+            if criterion == "ffpm":
+                mongo_secondary_query.append(
+                    {"samples.0.ffpm": {"$gte": float(query.get("ffpm"))}}
+                )
+            if criterion == "junction_reads":
+                mongo_secondary_query.append(
+                    {"samples.0.read_depth": {"$gte": float(query.get("junction_reads"))}}
+                )
+            if criterion == "split_reads":
+                mongo_secondary_query.append(
+                    {"samples.0.split_read": {"$gte": float(query.get("split_reads"))}}
+                )
+
         return mongo_secondary_query
 
 
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -2,9 +2,14 @@ import logging
 import re
 from datetime import datetime, timedelta
 
-from scout.constants import (CLINSIG_MAP, FUNDAMENTAL_CRITERIA,
-                             PRIMARY_CRITERIA, SECONDARY_CRITERIA,
-                             SPIDEX_HUMAN, TRUSTED_REVSTAT_LEVEL)
+from scout.constants import (
+    CLINSIG_MAP,
+    FUNDAMENTAL_CRITERIA,
+    PRIMARY_CRITERIA,
+    SECONDARY_CRITERIA,
+    SPIDEX_HUMAN,
+    TRUSTED_REVSTAT_LEVEL,
+)
 
 LOG = logging.getLogger(__name__)
 
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -770,6 +770,12 @@ class QueryHandler(object):
                 mongo_secondary_query.append(
                     {"samples.0.split_read": {"$gte": float(query.get("split_reads"))}}
                 )
+            if criterion == "fusion_caller":
+                fusion_caller_query = []
+                for caller in query.get("fusion_caller", []):
+                    fusion_caller_query.append({caller: {"$exists": True}})
+                    fusion_caller_query.append({caller: "Pass"})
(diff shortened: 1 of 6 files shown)


scout/adapter/mongo/query.py
```diff
@@ -756,6 +756,26 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
...
 
+            if criterion == "fusion_score":
+                mongo_secondary_query.append(
+                    {"fusion_score": {"$gte": float(query.get("fusion_score"))}}
+                )
+            if criterion == "ffpm":
+                mongo_secondary_query.append({"samples.0.ffpm": {"$gte": float(query.get("ffpm"))}})
+            if criterion == "junction_reads":
+                mongo_secondary_query.append(
+                    {"samples.0.read_depth": {"$gte": float(query.get("junction_reads"))}}
```
northwestwitch, line 767: I don't find the `junction_reads` anywhere in the scout code? Are you sure it is currently parsed from the VCF? What's the VCF field name?

scout/adapter/mongo/query.py
```diff
@@ -756,6 +756,26 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
...
+                )
+            if criterion == "split_reads":
```
northwestwitch, line 769: This is also an integer according to the VCF

scout/constants/query_terms.py
```diff
@@ -54,4 +54,9 @@
     "clinvar_tag",
     "cosmic_tag",
     "tumor_frequency",
+    "fusion_score",
+    "ffpm",
+    "junction_reads",
```
northwestwitch, line 59: Again, is this field parsed already?

scout/server/blueprints/variants/forms.py
```diff
@@ -213,6 +215,15 @@ class FusionFiltersForm(VariantFiltersForm):
     size_shorter = BooleanField("Length shorter than")
     decipher = BooleanField("Decipher")
     clinical_filter = SubmitField(label="Clinical filter")
+    fusion_score = BetterDecimalField(
+        "Fusion score >=", places=2, validators=[validators.Optional()]
+    )
+    ffpm = BetterDecimalField("FFPM >=", places=2, validators=[validators.Optional()])
+    junction_reads = BetterDecimalField(
```
northwestwitch, line 222: ?

northwestwitch (changes requested):
Nice start! I have some doubts about key/value formats and about a filter by a key that I don't find in the code, see my comments!

scout/server/blueprints/variants/forms.py
```diff
@@ -213,6 +215,15 @@ class FusionFiltersForm(VariantFiltersForm):
...
+        "Junction reads >=", places=2, validators=[validators.Optional()]
+    )
+    split_reads = BetterDecimalField("Split reads >=", places=2, validators=[validators.Optional()])
```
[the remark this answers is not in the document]
fevac: true. I'll update it

scout/adapter/mongo/query.py
```diff
@@ -756,6 +756,26 @@ def secondary_query(self, query, mongo_query, secondary_filter=None):
                 mongo_secondary_query.append({"cosmic_ids": {"$exists": True}})
                 mongo_secondary_query.append({"cosmic_ids": {"$ne": None}})
 
+            if criterion == "fusion_score":
+                mongo_secondary_query.append(
+                    {"fusion_score": {"$gte": float(query.get("fusion_score"))}}
+                )
+            if criterion == "ffpm":
```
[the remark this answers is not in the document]
fevac: ffpm is a float according the last vcf version of rnafusion. I'll update where relevant

scout/adapter/mongo/query.py
fevac: juction reads is the read_depth is the "DV" value in the vcf

fevac: it goes into the field `read_depth`

scout/constants/query_terms.py
fevac: yes, as `read_depth`

scout/adapter/mongo/query.py
northwestwitch: Ah OK, but it's not called like that in the variant, so the filter won't work

fevac: what RD is not given in the fusion vcf

northwestwitch: > but the filter does work, which field to take from the mongo db is given on `samples.0.read_depth`

Ah yes, you are right, I got confused, sorry!

scout/server/blueprints/variants/forms.py
```diff
@@ -213,6 +215,15 @@ class FusionFiltersForm(VariantFiltersForm):
     size_shorter = BooleanField("Length shorter than")
     decipher = BooleanField("Decipher")
     clinical_filter = SubmitField(label="Clinical filter")
+    fusion_score = BetterDecimalField(
+        "Fusion score >=", places=2, validators=[validators.Optional()]
```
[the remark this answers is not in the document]
fevac: this happen because of the rounding. In the database these values are 0.061999 so they don't pass the filter

scout/adapter/mongo/query.py
fevac: we want to show DV (junction reads) and RV (split reads) separate. They give different information and I would not add those values

scout/server/blueprints/variants/forms.py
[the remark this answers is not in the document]
northwestwitch: Ahhh OK!

fevac:
I won't do the gene name filtering in this PR since it's more complex. I will open a different issue for the second point:
- Gene filter should work on gene names too (for those fusions not having gene objects associated)

scout/adapter/mongo/query.py
northwestwitch: > we want to show DV (junction reads) and RV (split reads) separate. They give different information and I would not add those values

I understand, but be aware that whenever ref_depth is -1 the it is computed from those 2 values together, see [here](https://github.com/Clinical-Genomics/scout/blob/a95fb214586f87f31460273beb58324b1f500297/scout/parse/variant/genotype.py#L326)

fevac: I see, but it should not be -1 once the bug is fixed, right?

northwestwitch: Other question: for junction reads you mean the paired ends right? Why not naming them like that?

northwestwitch (approved):
Looks good to me! 💯

scout/adapter/mongo/query.py
fevac: these are the reads spanning the junction of the fusion gene. I think paired reads is too vague

PR #4388, opened 2024-02-01
Fix remaining relative paths on load

dnil:

Several relative paths can potentially get stored on db on load, relative to the working directory at load time. This is very seldom what is intended: most endpoints that serve files in fact attempt to convert to absoulte paths when used, but if the server app is started in a different directory, this will fail. For CG Solna, this is also rarely a problem since our configs use abs file paths in general. It is however noticeable on demo setups and potentially confusing to new users.

The downside might be it closes a potential workaround for container volume mounting, where the absolute path might not be the same on the CLI as on the web app.

On the total, making the behaviour predictable (absolute paths) is preferable.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Update demo case, before:
Screenshot 2024-02-01 at 10 08 05

After:
Screenshot 2024-02-01 at 10 08 40

dnil:
Meh, still:
Screenshot 2024-02-01 at 10 10 30
Recurses!

northwestwitch:
> Recurses!

In the sense of cursing again? 😆

dnil:
Better!
Screenshot 2024-02-01 at 10 54 37

dnil:
> > Recurses!
> 
> In the sense of cursing again? 😆

Doubly so! Cursing again because the recursion is cursed again.

northwestwitch:
Wouldn't be easier to fix relative paths directly when the case load config is validated with Pydantic? The validator has [a function](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L245) that checks if the path (relative or complete) exists on disk. Ifwe change that function to not only do the validation, but also setting the paths to the full path then you don't need any extra code I think.

The validator itself has this [tiny function](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L89) that does the trick for demo files

dnil:
> Wouldn't be easier to fix relative paths directly when the case load config is validated with Pydantic? The validator has [a function](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L245) that checks if the path (relative or complete) exists on disk. Ifwe change that function to not only do the validation, but also set the paths to full path then you don't need any extra code I think.
> 
> The validator itself has this [tiny function](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L89) that does the trick for demo files (at least those for which is required a validation at the moment)

I wouldn't say easier, but I'm looking through now, and Im starting to agree that it would make sense to try to do it there when we anyway touch those files/variables. 

As you say, there is already one that is called for some of the demo files, that must just be slightly bugged since we do have tons of relative paths on the demo.

I guess my only fundamental issue with that is that it will confound the role of that class, as it changes the returned values compared to what is in the config file as such, before we get around to storing it on db. If we start doing this there, we could start doing all of the `build_case()` tasks there as well. Which on the whole I really don't mind! In contrast to some of the weirder build functions, `build_case()` is only called on load, so should be somewhat straightforward.

dnil:
So, let's merge this perhaps: unless you find something wrong with it, that would really be the easiest. Then we can refactor the functionality of `build_case()` into the Pydantic `CaseLoader`, which will then graduate from just validation (and silently fixing some castable variable types) into a CaseBuilder.

northwestwitch:
> As you say, there is already one that is called for some of the demo files, that must just be slightly bugged since we do have tons of relative paths on the demo.

I think it just checks that the relative paths are valid, but it doesn't fix it yet.

Regarding the pydantic class role: It does the fixing of other fields, for instance [converting strings to integers](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L222), fixes the path for the [str variants images](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L297), etc. So why not this?

dnil:
> > As you say, there is already one that is called for some of the demo files, that must just be slightly bugged since we do have tons of relative paths on the demo.
> 
> I think it just checks that the relative paths are valid, but it doesn't fix it yet.

Right, right; it checks that the absolute path version of the relative one would be valid.
 
> Regarding the pydantic class role: It does the fixing of other fields, for instance [converting strings to integers](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L222), fixes the path for the [str variants images](https://github.com/Clinical-Genomics/scout/blob/b394398f5b1db069bd3b0a8d00dab234ba6594e8/scout/models/case/case_loading_models.py#L297), etc. So why not this?

No, exactly - I agree! That would probably be the most standard use of the Pydantic Class, and is consistent with it being called to parse the case config. But there are more things that `build_case()` does, that it should do. Like check if the config had a case owner, sorting individuals affected first. Most of the renaming of file keys into db keys would make sense there as well, wouldn't it? I guess we still want a small build, bringing in data from other documents, like panels, phenotype terms etc but much could just be done in the CaseLoader!

PR #4426, opened 2024-02-13
Add option of downloading clinvar submission as .json

TereseBo:
This PR adds the option of downloading a ClinVar submission as a json file, in addition to the possibility of downloading  it as 2 .csv files.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Navigate to the ClinVar submission page, locate the new button for json download and download the file.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/clinvar/static/clinvar_submissions.js
+++ b/scout/server/blueprints/clinvar/static/clinvar_submissions.js
@@ -1,0 +1,43 @@
+async function getJson(url) {
+    try {
+        const baseUrl = window.location.origin
+        console.log(`url: ${baseUrl}${url}`)
+        //Fetch data
+        const response = await fetch(`${baseUrl}${url}`)
+        if (!response.ok) {
+            throw new Error('Failed to fetch')
+        }
+        const jsonRes = await response.json()
+
+				let fileName=url.split("/").pop()
+
+        download(fileName, jsonRes)
+
+    } catch (error) {
+
+        if (error.toString().includes('Failed to fetch')) {
+            console.log(error)
+        }
+    }
+}
+
+function download(filename, object){
+    const json= JSON.stringify(object)
+    let downloadLink=document.createElement('a')
+    downloadLink.setAttribute('href','data:application/json;charset=utf-8,' + encodeURIComponent(json))
+    downloadLink.setAttribute('download', filename)
+    downloadLink.style.display='none'
+    document.body.appendChild(downloadLink)
+    downloadLink.click()
+    document.body.removeChild(downloadLink)
+}
+
+document.querySelector("#download_clinvar_json").addEventListener("click",(e)=>{
+
+getJson(e.target.value)
+
+})
+
+
+
+
--- a/scout/server/blueprints/clinvar/static/clinvar_submissions.js
+++ b/scout/server/blueprints/clinvar/static/clinvar_submissions.js
@@ -1,26 +1,24 @@
 async function getJson(url) {
     try {
+    //Get base url and fetch data
         const baseUrl = window.location.origin
-        console.log(`url: ${baseUrl}${url}`)
-        //Fetch data
         const response = await fetch(`${baseUrl}${url}`)
         if (!response.ok) {
             throw new Error('Failed to fetch')
         }
         const jsonRes = await response.json()
-
 				let fileName=url.split("/").pop()
 
+				//Create and download file
         download(fileName, jsonRes)
-
     } catch (error) {
-
         if (error.toString().includes('Failed to fetch')) {
             console.log(error)
         }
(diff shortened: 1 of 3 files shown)


TereseBo:
On main: 
image


Image of menu for actions related to clinker submission:
image

Downloaded file:
[SUB2222222.json](https://github.com/Clinical-Genomics/scout/files/14258174/SUB2222222.json)

northwestwitch:
Tested it locally with the POT1 variant:

I get an error trying to download the json:

image

Log from preclinvar prod:

image

I'll try to check if the request from scout sends the required params!

scout/server/blueprints/clinvar/static/clinvar_submissions.js
```diff
@@ -0,0 +1,35 @@
+async function getJson(url) {
```
northwestwitch, line 1: You probably wrote this code mostly in  javascript because you are more familiar with this language, but I think the whole download thing should be handled by python instead, being scout flask project.. Of course we use javascript in some parts, but it's minor things in general, like form checks etc. It's going to be a good exercise I think! ;)

scout/server/blueprints/clinvar/views.py
```diff
@@ -116,3 +116,12 @@ def clinvar_download_csv(submission, csv_type, clinvar_id):
         mimetype="text/csv",
         as_attachment=True,
     )
+
+
+@clinvar_bp.route("/<submission>/download/json/<clinvar_id>", methods=["GET"])
+def clinvar_download_json(submission, clinvar_id):
+    """Download a json for a clinVar submission"""
+
+    code, conversion_res = controllers.json_api_submission(submission_id=submission)
```
northwestwitch, line 125: This function does the conversion and returns the json to the javascript code that has been calling it I guess. Try to make it return directly the file as a response. 

Other thing, the downloaded file doesn't have indentation if you open it now:

image

Would be more helpful if you could pre-format the json prior to return it in the response. Check here how to do it: https://stackoverflow.com/questions/9170288/pretty-print-json-data-to-a-file-using-python

northwestwitch (changes requested):
The code works but as I wrote in the comments the response is handled by javascript code. I think this part should be moved to the python code instead. 

To see how to return a file check the other ClinVar views, for instance those that make you download the csv files.

For readability sake would be great to fix the indentation of the json file before it's returned

northwestwitch:
Now testable against preclinvar prod since I've deployed the latest changes on prod

TereseBo:
New testing on stage after moval of download related code:
When download does not work (for any reason):
Screenshot 2024-02-15 at 15 01 42

JSON file:

[SUB4646465-5.json](https://github.com/Clinical-Genomics/scout/files/14297863/SUB4646465-5.json)

image:
Screenshot 2024-02-15 at 15 07 02

scout/server/blueprints/clinvar/views.py
```diff
@@ -116,3 +117,27 @@ def clinvar_download_csv(submission, csv_type, clinvar_id):
...
+
+    if code in [200, 201]:
+        # Write temp CSV file and serve it in response
+        tmp_json = NamedTemporaryFile(mode="a+", prefix=clinvar_id, suffix=".json")
```
northwestwitch, line 130: The suffix doesn't seem to stick when ou download the file..?

image

northwestwitch (commented):
Looks like it's working, except for the fact that the downloaded file has no extension. Tested locally and on cg-vm1

scout/server/blueprints/clinvar/views.py
TereseBo: This was solved by including .json in the name of the file. Have been tested both using Safari (which had the extension also before the change) and Chrome. Now seems to be working for both browsers.

northwestwitch: Yes, it works for me as well now:

image

northwestwitch (approved):
Looks good! 💯

scout/server/blueprints/clinvar/views.py
```diff
@@ -116,3 +117,27 @@ def clinvar_download_csv(submission, csv_type, clinvar_id):
...
+
+        tmp_json.flush()
+        tmp_json.seek(0)
+        return send_file(
+            tmp_json.name,
+            download_name=f"{clinvar_id}.json",
+            mimetype="application/json",
+            as_attachment=True,
+        )
+    else:
```
northwestwitch, line 141: You probably don't even need the else here

PR #4455, opened 2024-02-22
Allow document id variant API connect, and improve user cred check when accessing variants

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Changed the "x" in cases.utils.remove_form button text to red for better visibility in dark mode
 - Update GitHub actions
 - Default loglevel up to INFO, making logs with default start easier to read
-- Refactor view route to allow navigation directly to unique variant document id
+- Refactor view route to allow navigation directly to unique variant document id, improve permissions check
 ### Fixed
 - Removed log info showing hgnc IDs used in variantS search
 - Maintain Matchmaker Exchange and Beacon submission status when a case is re-uploaded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Changed the "x" in cases.utils.remove_form button text to red for better visibility in dark mode
 - Update GitHub actions
 - Default loglevel up to INFO, making logs with default start easier to read
+- Refactor view route to allow navigation directly to unique variant document id
 ### Fixed
 - Removed log info showing hgnc IDs used in variantS search
 - Maintain Matchmaker Exchange and Beacon submission status when a case is re-uploaded
--- a/scout/server/blueprints/api/views.py
+++ b/scout/server/blueprints/api/views.py
@@ -25,10 +25,11 @@ def variant(institute_id, case_name, variant_id):
 
 
 @api_bp.route("/<institute_id>/<case_name>/<variant_id>/pin")
-def pin_variant(institute_id, case_name, variant_id):
+@api_bp.route("/<variant_id>/pin")
+def pin_variant(variant_id, institute_id=None, case_name=None):
     """Pin an existing variant"""
-
-    institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
+    if institute_id and case_name:
+        institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
     variant_obj = store.variant(variant_id)
 
     user_obj = store.user(current_user.email)
--- a/scout/server/blueprints/api/views.py
+++ b/scout/server/blueprints/api/views.py
@@ -27,10 +27,22 @@ def variant(institute_id, case_name, variant_id):
 @api_bp.route("/<institute_id>/<case_name>/<variant_id>/pin")
 @api_bp.route("/<variant_id>/pin")
 def pin_variant(variant_id, institute_id=None, case_name=None):
-    """Pin an existing variant"""
+    """Pin an existing variant
+
+    We always want to check institute_and_case with the case name and institute given on the variant object,
+    not only the one give on the URL, to ensure user permissions are ok.
+    """
+    variant_obj = store.variant(variant_id)
+
+    variant_case_obj = store.case(case_id=variant_obj["case_id"])
+    variant_case_name = variant_case_obj["display_name"]
+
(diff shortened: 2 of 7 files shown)


dnil:
While thinking about this, I noticed we have a few flaws in checking if users are authorised to view things. For starters, the variant api is not really relying on its checks (which are anyway nonexistent).

Screenshot 2024-03-01 at 13 53 07

 Given a malicious authenticated user, it would be possible to give any fake institute and case. It is easy to see the institute and case are not really used in the variant api call: the variant_id (document_id) is complete enough without them.
Screenshot 2024-03-01 at 13 54 54
Screenshot 2024-03-01 at 13 55 51
Screenshot 2024-03-01 at 13 55 57

Even if an insitute_and_case check is introduced, the same goes, so if we had any malicious users, they could just give an institute and case they do have access to. The latter potentially generalises also to other endpoints that do not make direct use of the institute and case supplied for the check.

scout/server/blueprints/variant/views.py
```diff
@@ -41,14 +41,18 @@ def update_tracks_settings():
     return redirect(request.referrer)
 
 
+@variant_bp.route("/document_id/<variant_id>")
 @variant_bp.route("/<institute_id>/<case_name>/<variant_id>")
 @templated("variant/variant.html")
-def variant(institute_id, case_name, variant_id):
+def variant(variant_id, institute_id=None, case_name=None):
```
dnil, line 47: Order change needed to make institute and case optional.

dnil:
I'll add a test as well, but the code should be there for review.

CHANGELOG.md
```diff
@@ -8,8 +8,10 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Added tags for Sniffles and CNVpytor, two LRS SV callers
 ### Changed
+- Refactor view route to allow navigation directly to unique variant document id, improve permissions check
 - In the diagnoses page genes associated with a disease are displayed using hgnc symbol instead of hgnc id
 ### Fixed
+- Be more careful about checking access to variant on api access
```
northwestwitch, line 14:
```suggestion
- Be more careful about checking access to variant on API access
```

scout/server/blueprints/api/views.py
```diff
@@ -17,19 +19,54 @@ def case(institute_id, case_name):
     return Response(json_util.dumps(case_obj), mimetype="application/json")
 
 
+def _lookup_variant(
+    institute_id: str, case_name: str, variant_id: str
```
northwestwitch, line 23:
```suggestion
    institute_id: str, case_name: Optional[str], variant_id: Optional[str]
```

scout/server/blueprints/api/views.py
```diff
@@ -17,19 +19,54 @@ def case(institute_id, case_name):
...
+) -> Optional[Tuple[dict, dict, dict]]:
+    """Lookup variant using variant document id. Return institute, case, and variant obj dicts.
+
+    The institute and case lookup adds a bit of security, checking if user is admin or
```
northwestwitch, line 27: > The institute and case lookup adds a bit of security

But they are optional, no? Because of code on lines 32 and 35 I mean

scout/server/blueprints/api/views.py
```diff
@@ -17,19 +19,54 @@ def case(institute_id, case_name):
...
+
+    if variant_obj.get("case_id") != case_obj.get("_id"):
+        return abort(403)
+
+    return (institute_obj, case_obj, variant_obj)
+
+
 @api_bp.route("/<institute_id>/<case_name>/<variant_id>")
-def variant(institute_id, case_name, variant_id):
+def variant(institute_id: str, case_name: str, variant_id: str) -> Optional[Response]:
```
northwestwitch, line 56: Aren't institute_id and case_name Optional here as well?

northwestwitch, line 56:
```suggestion
def variant(institute_id: str, case_name: str, variant_id: str) -> Response:
```
Why is the response optional? Doesn't it return an error response if lookup fails?

scout/server/blueprints/variant/controllers.py
```diff
@@ -149,17 +149,17 @@ def get_igv_tracks(build="37"):
...
-    case_name,
-    variant_id=None,
-    variant_obj=None,
-    add_other=True,
-    get_overlapping=True,
-    variant_type=None,
-    case_obj=None,
-    institute_obj=None,
-):
+    store: MongoAdapter,
```
northwestwitch, line 152: 💯

scout/server/blueprints/variant/controllers.py
```diff
@@ -194,13 +194,24 @@ def variant(
...
     """
-    if not (institute_obj and case_obj):
-        institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
+
     # If the variant is already collected we skip this part
     if not variant_obj:
         # NOTE this will query with variant_id == document_id, not the variant_id.
         variant_obj = store.variant(variant_id)
 
+    if not (institute_obj and case_obj):
```
northwestwitch, line 203: This code is also in the _lookup_variant function above, would be nice to unify and have a single function handling this code.

northwestwitch (approved):
Looks good to me! 👍🏻  I have some comments for improving the code

scout/server/blueprints/api/views.py
dnil: They could indeed be, but the routing does not really permit them to be empty. Hm, I might just add a document_id route for it as well.

scout/server/blueprints/api/views.py
dnil: Hm, almost; it's the variant that is mandatory. I reorder here as well.

scout/server/blueprints/api/views.py
dnil: I would say no; there are latent `abort()` calls in there, that would get us straight back with an error report. I dont actually know if there formally still is a response being made, but it will not be the one described on the return line.

scout/server/blueprints/variant/controllers.py
dnil: Ok, I agree, I'll refactor this. Should make the other calls involving variant safer as well.

PR #4461, opened 2024-02-26
Modify variants to display MANE info from Ensembl and not VEP annotations

northwestwitch:
- Update variants pages with MANE info from parsed Ensembl genes

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Download the latest files to update the genes (using main branch)
2. Update genes using these files (also using main branch)
3. Go to a variant in build 38 that you know has MANE transcripts and make sure that these transcripts are correct.
4. Even better to try with a variant which had wrong MANE transcripts, so see if this PR fixes them

**Expected outcome**:

**Review:**
- [x] code approved by ALKC
- [x] tests executed by CR, ALKC


northwestwitch:
HI @alkc, Ive created this PR as a fix to the problem you mentioned (geneticists complaining about wrong MANE transcripts). Since we have no cases in build 38 I have no idea if it's working or not, so I was wondering of you could give it a try when you have time?

Note that in order for it to work you need to:
1. Download the latest files to update the genes (using main branch)
2. Update genes using these files

alkc:
Thank you!

Sure thing, I can review this.

alkc:
I set up a dev build of scout with a minimal mirror of our production setup and loaded a VEP 103-annotated case, ran the update process in the main branch (commit 3389ecc1a75c70753d4cea2467eaba0ce1ae6188) and switched over to this branch.

Outcome: I get a 500 error when accessing variants of the hg38 case. The error does not occur for the demo cases.

```
scout-web_1  | [2024-03-01 14:45:05 +0000] [525] [DEBUG] GET /klingen_38/REDACTED/22b761a1772169220ad1eec8c89223bc                                  
scout-web_1  | 2024-03-01 14:45:05 2aebf616477c scout.server.app[525] ERROR Exception on /klingen_38/REDACTED/22b761a1772169220ad1eec8c89223bc [GET]
scout-web_1  | Traceback (most recent call last):                                                                                                    
scout-web_1  |   File "/venv/lib/python3.8/site-packages/flask/app.py", line 2529, in wsgi_app                                                       
scout-web_1  |     response = self.full_dispatch_request()                                                                                           
scout-web_1  |   File "/venv/lib/python3.8/site-packages/flask/app.py", line 1825, in full_dispatch_request                                          
scout-web_1  |     rv = self.handle_user_exception(e)                                                                                                
scout-web_1  |   File "/venv/lib/python3.8/site-packages/flask_cors/extension.py", line 176, in wrapped_function                                     
scout-web_1  |     return cors_after_request(app.make_response(f(*args, **kwargs)))                                                                  
scout-web_1  |   File "/venv/lib/python3.8/site-packages/flask/app.py", line 1823, in full_dispatch_request                                          
scout-web_1  |     rv = self.dispatch_request()                                                                                                      
scout-web_1  |   File "/venv/lib/python3.8/site-packages/flask/app.py", line 1799, in dispatch_request                                               
scout-web_1  |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)                                                          
scout-web_1  |   File "/home/worker/app/scout/server/utils.py", line 86, in decorated_function                                                       
scout-web_1  |     context = func(*args, **kwargs)                                                                                                   
scout-web_1  |   File "/home/worker/app/scout/server/blueprints/variant/views.py", line 50, in variant                                               
scout-web_1  |     data = variant_controller(                                                                                                        
scout-web_1  |   File "/home/worker/app/scout/server/blueprints/variant/controllers.py", line 230, in variant                                        
scout-web_1  |     add_gene_info(store, variant_obj, gene_panels=panels, genome_build=genome_build)                                                  
scout-web_1  |   File "/home/worker/app/scout/server/blueprints/variant/utils.py", line 260, in add_gene_info                                        
scout-web_1  |     update_transcripts_information(variant_gene, hgnc_gene, variant_obj, genome_build)                                                
scout-web_1  |   File "/home/worker/app/scout/server/blueprints/variant/utils.py", line 141, in update_transcripts_information                       
scout-web_1  |     update_transcript_mane(hgnc_transcript=hgnc_transcript, transcript=transcript)                                                    
scout-web_1  |   File "/home/worker/app/scout/server/blueprints/variant/utils.py", line 98, in update_transcript_mane                                
scout-web_1  |     transcript.pop(f"{key}_transcript")                                                                                               
scout-web_1  | KeyError: 'mane_select_transcript'                                                                                                    
```

The minimal setup entailed copying over `gene_panel`, `institute` and `hgnc_gene` collections from our live db into the demo db.

northwestwitch:
> The minimal setup entailed copying over `gene_panel`, `institute` and `hgnc_gene` collections from our live db into the demo db.

Ah yes, sorry, I've just pushed a fix

northwestwitch:
> Ah yes, sorry, I've just pushed a fix

Or not, I don't see it, weird!

northwestwitch:
> > Ah yes, sorry, I've just pushed a fix
> 
> Or not, I don't see it, weird!

Ah there are problems with Github at the moment:

image

scout/server/blueprints/variant/utils.py
```diff
@@ -89,8 +89,20 @@ def update_representative_gene(variant_obj: dict, variant_genes: List[dict]):
         variant_obj["first_rep_gene"] = None
 
 
+def update_transcript_mane(hgnc_transcript: dict, transcript: dict):
+    """Updates MANE key/values for a transcript in genome build 38."""
+    for key in ["mane_select", "mane_plus_clinical"]:
+        if hgnc_transcript.get(key):
+            transcript[f"{key}_transcript"] = hgnc_transcript[key]
+        else:
+            transcript.pop(f"{key}_transcript", None)
```
northwestwitch, line 98: The fix was finally pushed I see. Please test it again when you have time @alkc

alkc: `dict.pop()` - huh, learned something new today!

alkc (approved):

northwestwitch:
> It works!
> 
> I checked the same VEP 103-annotated hg38 case. Updating genes works from the CLI and in the main branch I can access the entry for a gene with a known MANE issue and confirm that the MANE select transcript is correct.
> 
> The annotation is still wrong when I access an associated variant in the case, but this is fixed when I open up the same variant with this branch loaded.

Cool! @dnil had some concerns about stopping the parsing of MANE info from the VEP annotations. Correct me if I'm wrong but I think his main concern is what if the MANE info collected from Ensembl is not good enough? Then we lose the VEP-derived info. What do you think about this @alkc ?

dnil:
> > It works!
> > I checked the same VEP 103-annotated hg38 case. Updating genes works from the CLI and in the main branch I can access the entry for a gene with a known MANE issue and confirm that the MANE select transcript is correct.
> > The annotation is still wrong when I access an associated variant in the case, but this is fixed when I open up the same variant with this branch loaded.
> 
> Cool! @dnil had some concerns about stopping the parsing of MANE info from VEP annotations. Correct me if I'm wrong but I think his main concern is what if the MANE info collected from Ensembl is not good enough? Then we lose the VEP-derived info without a valid alternative from Ensembl. What do you think about this @alkc ?

My main concern was with the interaction of other tools in the pipeline that use the VEP annotation, and then having e.g. one set of most severe consequences, scoring etc that depends on MANE, then another displayed in Scout. 
Having it incorrect/outdated is also not nice. It would have been a real mess with MIP hg38, but since we are on hg37, and nf-raredisease is handling a bit better, I think we are pretty ok. Feature wise, the ideal is keeping both, displaying the updated one where relevant, and the conflicts with the actually parsed one where it matters (like on most severe consequences) - much like for outdated gene panels. But I haven't really looked into how complex it would get.

northwestwitch:
> My main concern was with the interaction of other tools in the pipeline that use the VEP annotation, and then having e.g. one set of most severe consequences, scoring etc that depends on MANE, then another displayed in Scout.
> Having it incorrect/outdated is also not nice. It would have been a real mess with MIP hg38, but since we are on hg37, and nf-raredisease is handling a bit better, I think we are pretty ok. Feature wise, the ideal is keeping both, displaying the updated one where relevant, and the conflicts with the actually parsed one where it matters (like on most severe consequences) - much like for outdated gene panels. But I haven't really looked into how complex it would get.

Would it be OK if I modify this PR so the variants documents will not lose the MANE info from VEP, but instead we show the MANE from Ensembl (code in this PR)? Then if somebody complains that MANE is wrong again we can show the discrepancies from what was saved in the variant (VEP) and what was collected from Ensembl?

dnil:
> Would it be OK if I modify this PR so the variants documents will not lose the MANE info from VEP, but we only show the info from Ensembl (the code in this PR)? I would like to keep it simple for now, hoping that Ensembl is up-to-date, then if somebody complains for the MANE from Ensembl, then we can display the discrepancies from the VEP definition saved in the variant (which would be outdated for old variants I guess)?

For me that is perfectly acceptable: with regards to the transcripts alone, it's as good as we can do. As for other effects, it should not really affect our prod until we switch to hg38 anyway. When pipeline references are kept up to date, and cases are analysed efficiently, it will also be fine anyway.

Up to @alkc!

alkc:
Sorry for my late reply, I didn't have the time yesterday to think this through.

If I understand it correctly, the final decision is to use the data pulled directly from Ensembl to highlight which transcripts are MANE transcripts, instead of using the VEP-annotated data. The corresponding VEP MANE data is still being parsed into the variant objects for access in case there are discrepancies between the two annotation sources.

I see no issue with this. I think we're only interested in keeping the MANE info shown on the variant page up to date. For the purpose of keeping this info up to date, I suspect it's easier to achieve pulling the updates from ensembl rather than keeping VEP and its reference data up to date (weary greetings from an ongoing pipeline VEP update).

If I understand your concern correctly @dnil, the issue is that your pipeline relies in some way on the MANE annotations provided by VEP prior to the case being loaded into scout, and that any discrepancy between MANE before/after scout would be a source of error?

dnil:
> If I understand your concern correctly @dnil, the issue is that your pipeline relies in some way on the MANE annotations provided by VEP prior to the case being loaded into scout, and that any discrepancy between MANE before/after scout would be a source of error?

Yep! Several steps in a pipeline condense information from the transcript level up to the gene level. It is reasonable for a pipeline to prioritise information on a MANE select transcript over a mess of dubious transcripts. Variant rank scores, region and functional consequence come to mind. If you don't, I think we are fine for now, and the baton shifts to the pipeline developers if they change that.

PR #4475, opened 2024-03-05
Reroute case no institute

dnil:

Adds a route to the case view for unique `case_id`. Intended for use with Gens specifically.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/cases/views.py
+++ b/scout/server/blueprints/cases/views.py
@@ -74,7 +74,7 @@ def index():
 
 
 @cases_bp.route("/<institute_id>/<case_name>")
-@cases_bp.route("/case/<case_id>")
+@cases_bp.route("/case/case_id/<case_id>")
 @templated("cases/case.html")
 def case(
     case_name: Optional[str] = None,
@@ -92,10 +92,10 @@ def case(
             flash("Case could not be found: please provide a case ID.")
             return redirect(request.referrer)
 
-        case_obj = store.case(case_id=case_id, projection={"display_name": 1})
+        case_obj = store.case(case_id=case_id, projection={"display_name": 1, "owner": 1})
         if not case_obj:
             flash("Case {} does not exist in database!".format(case_id))
-            return redirect(request.referrer)
+            return abort(404)
 
         case_name = case_obj.get("display_name")
         institute_id = case_obj.get("owner")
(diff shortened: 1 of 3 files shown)


dnil:
I'll add a test as well!

scout/server/blueprints/cases/views.py
```diff
@@ -74,9 +74,32 @@ def index():
...
+    institute_id: Optional[str] = None,
+    case_id: Optional[str] = None,
+):
+    """Display one case.
+    Institute and display_name pairs uniquely specify a case.
+    So do case_id, but we still call institute_and_case again to fetch institute
+    and reuse its user access verification.
+    """
+
+    if not institute_id:
```
northwestwitch, line 90: Why not just:

```
if case_id:
   case_obj = store.case(case_id=case_id, projection={"display_name": 1, "owner": 1})
   case_name = case_obj.get("display_name")
   institute_id = case_obj.get("owner")
```
Then you just go to line 103?

northwestwitch (commented):
The code works, but could be made simple perhaps?

scout/server/blueprints/cases/views.py
dnil: Which is the same except for removing the error handling? 😉

northwestwitch: It's shorter, and if the case doesn't exist or you have no access to it already handled on line 106

scout/server/blueprints/cases/views.py
```diff
@@ -74,9 +74,29 @@ def index():
...
+    institute_id: Optional[str] = None,
+    case_id: Optional[str] = None,
+):
+    """Display one case.
+    Institute and display_name pairs uniquely specify a case.
+    So do case_id, but we still call institute_and_case again to fetch institute
+    and reuse its user access verification.
+    """
+
+    if case_id:
```
northwestwitch, line 90: 👍🏻 This also works. Fine by me if you like to provide the specific error message

northwestwitch (approved):
👍🏻

scout/server/blueprints/cases/views.py
dnil: Removing the double indentation ifs is good for complexity, agreed, also for the price of loosing one error check I think. But having to have multiple if - else on the right side (those post hocs are slow for me to read at least), and loosing another does not seem to justify removing the other `if not`.

If we want it really simple, I could also just add another endpoint for it. That would cause a few lines of duplication, but keep each very easy to read.

PR #4503, opened 2024-03-14
Parse genes partial fusion pairs

### LINKED ISSUE #4286: Gene filter for fusion vcf should work for fusion_gene names
- Gene filter should work for fusion_gene names when fusions not having gene objects associated

### LINKED ISSUE #4507: Fusion specific fields missing from variantS export
It would be nifty to export things like exon number, orientation, maybe split read counts etc, and `fusion_genes` for fallback when they do not correspond to plain HGNC genes that we have loaded.

dnil:

- Partially fix #4286:
  - Fix bug where genes were discarded during load parsing if one of the fusion partners was lacking hgnc_id. These genes constitute the bulk of the issue, and become searchable etc upon case (re)upload.

With the demo case:

Screenshot 2024-03-14 at 14 44 40

Screenshot 2024-03-14 at 14 47 19

- Fix #4507 - added fusion specific fields missing from variantS export. This helps a bit, as the symbols missing a gene are now part of the export.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


dnil:
Ok, testing the front end part for now on [stage](https://scout-stage.scilifelab.se/cust000/ACC8848A2-75M-FUSION/fusion/variants?variant_type=clinical). So far, so good I guess.
Before: 
Screenshot 2024-03-14 at 15 00 18

After:
Screenshot 2024-03-14 at 15 15 47

dnil:
Mm, the loading bug for the fusion partner of a non-gene is fixed, which does quite a bit on the stage case at least, but as evident when searching, the front end symbol/alias lookup part will not allow searches on its own. Try e.g. [POLH-AS1](https://scout-stage.scilifelab.se/genes/40459) on the stage case: https://scout-stage.scilifelab.se/cust000/ACC8848A2-75M-FUSION/fusion/variants

dnil:
While not exactly in the original issue, it would be nice to 
- use the alias lookup for fusion_genes also for export
- perhaps store back added gene info from alias lookup to the variant in db, to allow searching etc.

dnil:
Sorry for the bulky PR, but the issues were connected, and fixing one thing helped resolve issues with the other.. 

Fusion variantS export now includes some fusion specific fields:
Screenshot 2024-03-15 at 14 28 56

scout/constants/variants_export.py
```diff
@@ -14,6 +14,16 @@
 ]
```
northwestwitch, line 15: Not your fault, but have you seen this?

image

🤣 🤣 🤣

dnil:
I would prefer not to refactor those apparently complex functions, but it is kind of trivial to break out different sub functions for the different categories if need be.

scout/constants/variants_export.py
dnil: Yes... Im part guilty now as well, but it's mostly because of a very looong sequence of "if this value exists, do that". While not so DRY perhaps it's also not really complex. We could break it up in one function per variant category to "fool" the counter, but I dunno..

northwestwitch: Let's do it in a dedicated PR!

scout/constants/variants_export.py
```diff
@@ -14,6 +14,16 @@
...
 
 CANCER_EXPORT_HEADER = EXPORT_HEADER + ["VAF TUMOR", "VAF NORMAL", "COSMIC ID"]
+FUSION_EXPORT_HEADER = EXPORT_HEADER + [
```
northwestwitch, line 17: I would have preferred to handle the export of fusion variants in a separate PR, to be reviewed after merging this one (because I understand this is needed to create content for the export)

scout/server/blueprints/variants/controllers.py
```diff
@@ -1108,7 +1182,35 @@ def variant_export_lines(store, case_obj, variants_query):
         else:
             variant_line.append("N/A")
 
-        if case_obj.get("track") == "cancer":
+        if category == "fusion":
```
northwestwitch, line 1185: This code adds quite a lot of complexity to an already complex function and it would be much better to move outside.

Complexity of main branch with radon:

image


Complexity of this branch:

image

scout/server/blueprints/variants/controllers.py
```diff
@@ -331,20 +333,94 @@ def fusion_variants(
...
+    Also, if we do not have any kind of symbols for the fusion genes, there is not much we can do.
+    If the fusion gene apparent symbol is not already among the variant hgnc symbols, attempt to find the gene using
+    first its exact symbol, then the special case of HGNC locus group symbols, ending in `@` and not kept as aliases,
+    then lastly aliases.
+
+    Update variant_obj with the new info.
+    """
+
+    variant_genes = variant_obj.get("genes") or []
+    if len(variant_genes) >= 2:
```
northwestwitch, line 368: What if the gene "a" is actually 2 genes and the gene "b" is missing? Can the situation happen?

scout/server/blueprints/variants/controllers.py
```diff
@@ -331,20 +333,94 @@ def fusion_variants(
...
+        return
+
+    variant_hgnc_ids = variant_obj.get("hgnc_ids") or []
+    variant_hgnc_symbols = variant_obj.get("hgnc_symbols") or []
+
+    fusion_genes = variant_obj.get("fusion_genes") or []
+    if not fusion_genes:
```
northwestwitch, line 375: Perhaps put this on top of the function?

scout/server/blueprints/variants/controllers.py
```diff
@@ -331,20 +333,94 @@ def fusion_variants(
...
+        return
+
+    result_genes = []
+    remaining_fusion_gene_symbols = []
+    for fusion_gene_symbol in fusion_genes:
+        if fusion_gene_symbol in variant_hgnc_symbols:
```
northwestwitch, line 381: It is quite hard to follow what is going on in this function, also thinking about future maintenance. Can some of this stuff be moved to the parsing part instead? So it doesn't have to be on the fly every time one loads the variantS page

scout/server/blueprints/variants/controllers.py
```diff
@@ -331,20 +333,94 @@ def fusion_variants(
...
+        return
+
+    variant_hgnc_ids = variant_obj.get("hgnc_ids") or []
+    variant_hgnc_symbols = variant_obj.get("hgnc_symbols") or []
+
+    fusion_genes = variant_obj.get("fusion_genes") or []
+    if not fusion_genes:
+        return
+
+    result_genes = []
```
northwestwitch, line 378: Type hints could help understanding better what these variables contain. Add to the entire function

scout/server/blueprints/variants/controllers.py
```diff
@@ -1024,21 +1100,17 @@ def generate(header, lines):
...
-
     Returns:
         an HTTP response containing a csv file
     """
-    document_header = variants_export_header(case_obj)
-    export_lines = []
+    document_header = variants_export_header(case_obj, category)
     export_lines = variant_export_lines(
-        store, case_obj, variant_objs.limit(EXPORTED_VARIANTS_LIMIT)
+        store, case_obj, variant_objs.limit(EXPORTED_VARIANTS_LIMIT), category
```
northwestwitch, line 1113: Yes, category might really be the key to simplify the export. But I would keep that code outside this PR

scout/server/blueprints/variants/controllers.py
```diff
@@ -1084,18 +1158,18 @@ def variant_export_lines(store, case_obj, variants_query):
         variant_line.append("_".join([str(position), change]))
 
         # gather gene info:
-        gene_list = variant.get("genes")  # this is a list of gene objects
+        gene_list = variant.get("genes", [])  # this is a list of gene objects
```
northwestwitch, line 1161: 👍🏻

northwestwitch (commented):
Just an initial impression: I appreciate the fix of 3 things but I think the code, the review process and the future maintenance would be much better if you separate the variant export part from the missing fusion genes which is the purpose of this PR. 

That way you could dedicate more attention to include the missing fields perhaps it a more stylish way that won't add so much complexity to an already complex function?

scout/constants/variants_export.py
northwestwitch: By reading this PR I've realised that having a dedicated PR for the export would give that code more justice.. Sorry!

scout/constants/variants_export.py
dnil: Ok, let's do it like that then. It is more work, and doing it this way was good for uncovering missing values.

scout/server/blueprints/variants/controllers.py
dnil: Right, I would still argue it is not complexity, rather something like affine crystallinity or redundancy, but I do agree it should be fixed.

scout/server/blueprints/variants/controllers.py
dnil: Sure, but then we loose the only large benefit with this function, which is fixing old cases without reload. Perhaps it is cleaner just to remove it and force a reload. Searches would still not be fully functional without upserting the fixed objects to db.

scout/server/blueprints/variants/controllers.py
dnil: yes, we should check for the same len as the fusion genes, if the latter exists!

dnil:
Streamlined into a PR to bug fix #4503. Made a separate PR for fusion export #4530. Removed the controller side fusion gene symbol lookups. The latter can be revisited when fixing e.g. #4505 or done in the pipeline by extending/updating the annotation set so that more genes have current gene symbols or hgnc_id.

scout/parse/variant/variant.py
```diff
@@ -271,7 +271,6 @@ def get_genmod_key(case):
     """
     case_id = case["_id"]
     if "-" in case_id:
-        LOG.debug("iCase ID contains '-'. Using display_name as case _id.")
```
northwestwitch, line 274: 👍🏻

scout/server/blueprints/variants/templates/variants/fusion-variants.html
```diff
@@ -127,7 +127,7 @@
       {% if variant.genes %}
         {% for gene in variant.genes %}
           {% for transcript in gene.transcripts %}
-            <div>{{ transcript.exon|int }} </div>
+            <div>{% if "exon" in transcript %}{{ transcript.exon|int }}{% endif %} </div>
```
northwestwitch, line 130: 👍🏻

northwestwitch (approved):
Really nice fix to the missing genes problem! 💯 
And thanks for simplifying the code, I believe it will be easier to maintain in the long run

dnil:
Re-tested on stage. Looks good, and even though I miss the link out to genes found by alias or post-load lookup for old bugged cases, they were a bit of a false friend without permanent storing.

PR #4508, opened 2024-03-15
Fix search for variants in genes present only in build 38 returning no results

### LINKED ISSUE #4506: hg38 only genes not searchable, at least on fusion page?
Cf https://scout-stage.scilifelab.se/genes/51775 on https://scout-stage.scilifelab.se/cust000/Seraseq-W4PBRB-FUSION/fusion/variants. A tentative guess is that it is global for `38` cases, and due to defaulting to `"37"` when we call https://github.com/Clinical-Genomics/scout/blob/92e390ee660fc8013e31f72a9859e9c590b1f6ae/scout/server/blueprints/variants/views.py#L136 rather than checking the case build?

northwestwitch:
- Fix #4506, it's a bug present for all variant types, no just the fusion ones

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. On a local demo instance of Scout, load all genes (both 37 and 38)
1. Go to fusion variants page of the demo RNA case
1. Pick a variant in a gene present in scout, for instance the first one:
image

4. EML4 gene is present for both builds, so remove the document for build 37.
5. On main branch, replicate the issue in #4506 
6. Switch to this branch and make sure the variant in gene EML4 is found using search filters 

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/variants/views.py
+++ b/scout/server/blueprints/variants/views.py
@@ -131,7 +131,7 @@ def variants(institute_id, case_name):
 
     controllers.update_form_hgnc_symbols(store, case_obj, form)
 
-    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "37")) else "37"
     cytobands = store.cytoband_by_chrom(genome_build)
 
     variants_query = store.variants(
@@ -220,7 +220,7 @@ def str_variants(institute_id, case_name):
 
     controllers.activate_case(store, institute_obj, case_obj, current_user)
 
-    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "37")) else "37"
     cytobands = store.cytoband_by_chrom(genome_build)
 
     query = form.data
@@ -303,7 +303,7 @@ def sv_variants(institute_id, case_name):
     # Populate chromosome select choices
     controllers.populate_chrom_choices(form, case_obj)
 
-    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "37")) else "37"
     cytobands = store.cytoband_by_chrom(genome_build)
 
     controllers.update_form_hgnc_symbols(store, case_obj, form)
@@ -397,7 +397,7 @@ def mei_variants(institute_id, case_name):
     # populate available panel choices
     form.gene_panels.choices = controllers.gene_panel_choices(store, institute_obj, case_obj)
 
-    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "37")) else "37"
     cytobands = store.cytoband_by_chrom(genome_build)
 
     controllers.update_form_hgnc_symbols(store, case_obj, form)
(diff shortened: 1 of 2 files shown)


northwestwitch:
The code on variantS views is 41% duplicated. It's a known issue and it needs heavy refactoring. I'll leave it as it is though because it's not an easy task and it's out of the scope of this PR.

scout/server/blueprints/variants/views.py
```diff
@@ -122,7 +123,6 @@ def variants(institute_id, case_name):
             return redirect(request.referrer)
 
         hgnc_symbols_set = set(form.hgnc_symbols.data)
-        LOG.debug("Symbols prior to upload: {0}".format(hgnc_symbols_set))
```
dnil, line 125: 👍

scout/server/blueprints/variants/views.py
```diff
@@ -297,12 +303,14 @@ def sv_variants(institute_id, case_name):
     # Populate chromosome select choices
     controllers.populate_chrom_choices(form, case_obj)
 
-    genome_build = "38" if "38" in str(case_obj.get("genome_build")) else "37"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
```
dnil, line 306: Huh? Use 38 if the key is missing, use 37 if the key is any false value. I feel this warrants a bit of explanation, as it is done differently in other places..

scout/server/blueprints/variants/views.py
```diff
@@ -576,12 +587,14 @@ def cancer_sv_variants(institute_id, case_name):
...
```
dnil, line 590: Hihi, maybe not so many places left now where it is done differently either. 😁

scout/server/blueprints/variants/views.py
```diff
@@ -659,12 +672,14 @@ def fusion_variants(institute_id, case_name):
...
 
-    genome_build = "38"
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "38")) else "37"
     cytobands = store.cytoband_by_chrom(genome_build)
 
     controllers.update_form_hgnc_symbols(store, case_obj, form)
 
-    variants_query = store.variants(case_obj["_id"], category=category, query=form.data)
+    variants_query = store.variants(
+        case_obj["_id"], category=category, query=form.data, build=genome_build
```
dnil, line 681: 👍

dnil (commented):
Looks good with regards to actually passing genome build for the variants query, but what's up with the default build ambiguity?

scout/server/blueprints/variants/views.py
northwestwitch: Line 300 on main branch is going to make the app crash for cases without genome build, but we didn't experience any crash at this level so I'm assuming that it is safe to say that all cases that are currently used in Scout have a build. 
For the cases with build "37" or those with "GRCh37" (super old, or in some other facilities) genome build will be set to "37" so it is fine. It's a way of projecting us towards the future, i.e. build 38 ?

northwestwitch: Ok perhaps line 300 is not causing a crash since you'd have "None", I can put a 37 there, just to be on the safe side, but I don't think we have any of those cases without the build any more

northwestwitch: Or I can keep line 300 as it is and use it in the other places

dnil: Yes, either of those should be fine! I kind of like the first option of making it explicit. It is not super obvious in Python that you will get a string back from `str(None)` - but not a 0 from `int(None)`. But make the default either "37" or "38". This monkey brain is confused and needs to find meaning in the choice if the possible default outcomes differ. 😁

scout/server/blueprints/variants/views.py
```diff
@@ -131,9 +131,12 @@ def variants(institute_id, case_name):
 
     controllers.update_form_hgnc_symbols(store, case_obj, form)
 
-    cytobands = store.cytoband_by_chrom(case_obj.get("genome_build"))
+    genome_build = "38" if "38" in str(case_obj.get("genome_build", "37")) else "37"
```
dnil, line 134: 🙇

dnil (approved):
Nice, thanks!

Issue #1394, opened 2019-10-17
Be explicit about relative / absolute paths config.yaml


dnil:
Relative paths for vcf, bam are confusing. Absolute bam paths do not appear to work?

northwestwitch:
Being clear about how to fill in config files would certainly help. However I think that the issue of today with bam files was caused by a wrong key name -> bam_file instead of bam_path

dnil:
Quite right, #1397 so that doesn't happen again. 
But, my recollection was that strangeness did occur with relative paths that are not also correctly relative to your $PWD when loading. I'll try to replicate that.

dnil:
This is also a general issue with the way delayed file reads are handled. Consider e.g. batch loading research vcfs after some time.. It may have worked when loading each case yams originally, but now you will have to cd back to the each directory where the load was working relatively..

northwestwitch:
Closing this now that we have the Pydantic path validation in place.

Issue #4399, opened 2024-02-02
Refactor `build_case()` and `CaseLoader`


dnil:
As per discussion in #4388, `CaseLoader` handling several of the current duties of `build_case()` would make a lot of sense. Perhaps not the parts involving pulling parts of other documents from the database, but checks and straightforward format changes and name changes would sit snugly!

dnil:
Remember to prune conditional fields on individuals. No use in keeping empty keys from several of the seldom used applications.

PR #4517, opened 2024-03-20
Store demo case file paths as absolute paths

northwestwitch:
- Alternative to #4388 but using the pydantic case checker

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Locally, run `scout setup demo` using main branch and this branch
2. Check on MongoDB compass that file paths for demo cases are all absolute paths (on main branch they are relative paths)

**Expected outcome**:

**Review:**
- [x] code approved by DN


northwestwitch:
Main branch:

image


This branch:

image

dnil:
I see the will here, and it is good, but this does not fully solve either #4399 (it starts it) or #1394, while it conflicts with #4388, plus its a duplication of effort which we have previous little time for. Let me eat something first, but Im currently not a fan. 😆

northwestwitch:
> I see the will here, and it is good, but this does not fully solve either #4399 (it starts it) or #1394, while it conflicts with #4388, plus its a duplication of effort which we have previous little time for. Let me eat something first, but Im currently not a fan.

I don't really understand why we need to refactor case build and case loader (#4399) just to fix the fact that demo case doesn't have absolute but relative paths to the file? I think this is a simplest solution to the problem here. I also understood from the discussion in your [PR](https://github.com/Clinical-Genomics/scout/pull/4388#issuecomment-1923260674) that you agreed in a solution like this?

dnil:
This is not just about the demo case. The problem is most obvious there, since all new devs/users encounter it. It affects all use of relative paths on loading. Some of these are being checked for at various places in the code base, with varying success.

I do not mind seeing a pydantic class fixing the issue, it is great for it: we just need to make the abstraction clear, and move over other similar tasks from build to it as well so one knows where to look for them. Adding one more task to what is currently primarily the pydantic load validator makes its role more confusing. Is it for validation? For initial load conversion to db? Or perhaps for reading db documents for display? As we discussed in [#4388](https://github.com/Clinical-Genomics/scout/pull/4388].

We can refactor it, but I don't see why we should do all that refactoring before merging the first PR?

northwestwitch:
> This is not just about the demo case. The problem is most obvious there, since all new devs/users encounter it. It affects all use of relative paths on loading. Some of these are being checked for at various places in the code base, with varying success.
> 
> I do not mind seeing a pydantic class fixing the issue, it is great for it: we just need to make the abstraction clear, and move over other similar tasks from build to it as well so one knows where to look for them. Adding one more task to what is currently primarily the pydantic load validator makes its role more confusing. Is it for validation? For initial load conversion to db? Or perhaps for reading db documents for display? As we discussed in #4388.
> 
> We can refactor it, but I don't see why we should do all that refactoring before merging the first PR? And I dont see why we should start by solving a part of the original issue with this PR?

Mmm sorry but I don't really like the approach of PR #4388 to fix the issue. I agree with you that having the pydantic validation plus the case build is confusing and stuff might be refactored to be more clear. This probably comes from the fact that we had many checks in build before we switch to pydantic, but since now we have pydantic I think we should male the. most of it. In current main there is a method that already checks if the paths are valid. So instead of having one hundred lines of extra code to fix the paths outside the validation, you can have just 2 in that pydanic function that do just the same? It's way more easy to maintain as well I think?

northwestwitch:
I think in these case we might solve the impasse like this: we leave both PRs open and we see who gets tired to see them laying around and approves one first 🤣

dnil:
Right, neither is the solution we would like in the long run, and we could argue for a long time on wether its better to have something that solves the problem, and how complete an abstraction should be before introducing it. I think this is a multiplication of effort for no reason other than a realisation about a possible future path, but since you made it, I will review it and ask that it at least satisfies solving #1394. Once it works, we could then make sure to solve #4399 before the next release.

dnil (commented):
Looks convincing on VCF, but still does not fix e.g. custom images path. If you want, you could step through the changes in #4388 for the places I know of.

Screenshot 2024-03-20 at 15 44 22

I haven't checked all places downstream that attempt to make an abs path for redundancy. I guess we may still want several of them to cover for old relative paths in the db.

The CaseLoader class now does part of what build_case() did before. I think the division you sketch is really good, but there are several equivalent sub functionalities left in build_case() that should move to CaseLoader to keep us from looking back and forth between the two.

Abstraction wise, we would want things that only needs to be done when parsing the config file in CaseLoader, and they should as per parsing in general also preferably not require the use of other db resources. Build would then handle the additional resources, and not have to worry about file integrity or syntax. We traditionally reserve dual use for the build function, so that they could sometimes be called when storing back items from prepped display object to db, but this is afaik not used for case objects - we only do field-by-field updates on them live.

northwestwitch:
@dnil I've fixed paths to images (cases and vars), chromograph and reviewer. Comparing with the demo case config it looks like that's all

dnil:
> @dnil I've fixed paths to images (cases and vars), chromograph and reviewer. Comparing with the demo case config it looks like that's all

At the seminar now but will check later!

scout/models/case/case_loading_models.py
```diff
@@ -49,11 +49,19 @@
...
+    "reference_info",
     "RNAfusion_inspector",
+    "RNAfusion_inspector_research",
     "RNAfusion_report",
     "RNAfusion_report_research",
 ]
 
+CHROMOGRAPH_PATH_SUFFIX_CHECKS = ["autozygous", "coverage", "upd_regions", "upd_sites"]
+
+REVIEWER_PATH_CHECKS = ["alignment", "alignment_index", "catalog", "vcf"]
```
dnil, line 63:
```suggestion
REVIEWER_PATH_CHECKS = list(REViewer.model_json_schema().get('properties'))
```
Are they are just forgotten or is there a reason why you wouldn't check if the reference and ref index? Otherwise, again, something like

scout/models/case/case_loading_models.py
```diff
@@ -145,6 +148,22 @@ class ChromographImages(BaseModel):
     upd_regions: Optional[str] = None
     upd_sites: Optional[str] = None
 
+    @model_validator(mode="before")
+    def validate_file_path(cls, values: Dict) -> "SampleLoader":
```
dnil, line 152:
```suggestion
    def validate_file_path(cls, config_values: Dict) -> "SampleLoader":
```

dnil (approved):
Thank you, this should now close #1394 afaik, so I approve. I guess we could say it starts #4399 as well, but there are other simpler value changes in `build`. They can also be determined by just looking at the config file (e.g. variable type casts, booleans for valid config values (like `has_svvariants`) and should go with the CaseLoader when allowing it to change values. But we can do them after.

scout/models/case/case_loading_models.py
```diff
@@ -86,18 +94,14 @@ class Sex(str, Enum):
...
-
-def _is_string_path(string_path: str) -> bool:
-    try:
-        path = Path(string_path) or Path(_get_demo_file_absolute_path(partial_path=string_path))
-        return path.is_file()
-    except AttributeError:
-        return False
+def _resource_abs_path(string_path: str) -> str:
+    """Return the absolute path to a resource file."""
+    if exists(string_path) and isabs(string_path):
```
dnil, line 99: you could flatten this conditional as 
```suggestion
    if not exists(string_path):
        raise FileExistsError(f"Path {string_path} could not be found on disk.")
    if isabs(string_path):
        return string_path
    return abspath(string_path)    
```

scout/models/case/case_loading_models.py
```diff
@@ -49,11 +49,19 @@
...
+    "exe_ver",
+    "smn_tsv",
+    "reference_info",
     "RNAfusion_inspector",
+    "RNAfusion_inspector_research",
     "RNAfusion_report",
     "RNAfusion_report_research",
 ]
 
+CHROMOGRAPH_PATH_SUFFIX_CHECKS = ["autozygous", "coverage", "upd_regions", "upd_sites"]
```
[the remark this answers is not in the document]
northwestwitch: Good suggestion! We already have these variables present in the class, so yes!

scout/models/case/case_loading_models.py
```diff
@@ -145,6 +148,22 @@ class ChromographImages(BaseModel):
...
     upd_sites: Optional[str] = None
 
+    @model_validator(mode="before")
+    def validate_file_path(cls, values: Dict) -> "SampleLoader":
+        """Make sure that chromograph paths associated to samples exist on disk and are absolute paths."""
+        for item in CHROMOGRAPH_PATH_SUFFIX_CHECKS:
+            item_path: str = values.get(item)
+            if (
+                item_path
+            ):  # it is an incomplete path -> replace the path to the directory with the abs path to the directory
```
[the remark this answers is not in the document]
northwestwitch: Done

scout/models/case/case_loading_models.py
northwestwitch: Forgotten, thanks! :)

northwestwitch: I think you can get the keys directly of you `for key in cls.model_fields:`

scout/models/case/case_loading_models.py
```diff
@@ -145,6 +148,22 @@ class ChromographImages(BaseModel):
...
+                item_path_dirname: str = dirname(
+                    item_path
+                )  # path to the directory containing multiple files
+                values[item] = item_path.replace(
```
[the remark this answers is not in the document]
northwestwitch: Now I have replace the loop in the model check with `for key in cls.model_fields:`, so I think it's clear that the keys used for collecting the values are the model fields!

scout/models/case/case_loading_models.py
northwestwitch: Yes, I like it!

northwestwitch: Fixed!

northwestwitch:
I tested it again locally and seems to parse stuff fine. Merging when tests pass. Thanks for the useful review @dnil!

PR #4537, opened 2024-03-28
Patch release 4.80.1

northwestwitch:
### Added
-  [ ] Tag for somatic SV IGH-DUX4 detection samtools script
### Fixed
- [x] Buttons layout in HPO genes panel on case page
- [x] Added back old variant rankscore index with different key order to help loading on demo instance - *tested by collaborator*
- [ ] Cancer case_report panel-table no longer contains inheritance information
- [ ] Case report pinned variants card now displays info text if all pinned variants are present in causatives

**Review:**
- [ ] code approved by


dnil (approved):
👍

dnil:
Maybe I should not have merged the latest main here yet to really keep it a patch: it formally has an addition, but that one is really just a label. We could also skip this patch and go directly to 4.81 - fixing the layout and some-types-of-demo-installation-indices isn't a huge rush. :shrug:

northwestwitch:
Closing this, let's create a minor instead!

Issue #4397, opened 2024-02-02
Unable to parse HGNC ID due to presence of 'HGNC:' prefix


lfearnley:
I'm getting the following error loading a case  from nf-raredisease into a fresh install of Scout:

```
Traceback (most recent call last):
  File "/home/fearnleyl/scout/scout/adapter/mongo/variant_loader.py", line 719, in load_variants
    nr_inserted = self._load_variants(
                  ^^^^^^^^^^^^^^^^^^^^
  File "/home/fearnleyl/scout/scout/adapter/mongo/variant_loader.py", line 436, in _load_variants
    parsed_variant = parse_variant(
                     ^^^^^^^^^^^^^^
  File "/home/fearnleyl/scout/scout/parse/variant/variant.py", line 175, in parse_variant
    parsed_transcripts = add_gene_and_transcript_info(parsed_variant, variant, vep_header)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fearnleyl/scout/scout/parse/variant/variant.py", line 565, in add_gene_and_transcript_info
    for parsed_transcript in parse_transcripts(raw_transcripts):
  File "/home/fearnleyl/scout/scout/parse/variant/transcript.py", line 27, in parse_transcripts
    transcript["hgnc_id"] = get_hgnc_id(entry)
                            ^^^^^^^^^^^^^^^^^^
  File "/home/fearnleyl/scout/scout/parse/variant/transcript.py", line 230, in get_hgnc_id
    return int(hgnc_id)
           ^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'HGNC%3A7470'
```

Is this likely to be a raredisease issue, or a Scout one?

dnil:
Hi! Thank you kindly for reporting!As weird as it sounds, we dont run the nf rare-disease pipe in production yet. I imagine this is mostly a VEP version diff, in which case we will definitely update Scout to handle both formats. We'll take a look!

dnil:
I've asked for some VCF-files from the local nf-raredisease crew, and should get them next week. I'm kind of guessing we are timezone challenged here, but whenever it's convenient, perhaps you could post just the culprit line from your VCF? The one with the HGNC 7470. 

[...]

PR #4577, opened 2024-04-23
Fix #4445 - load separate MT VCFs

### LINKED ISSUE #4445: nf-core raredisease separate MT files
- [ ] MT variants VCF
~MT variants rank model key~
~MT variants rank model file (link on e.g. reference-files)~

We keep the same variant type (SNV on the chromosome MT). 
We postpone making a new view until there are more explicit requests for this. The rank models will not match in sorting, but variants are already incorrectly scored with the same model although annotation completeness differs, so chances are this will be better even before introducing normalized scores.

dnil:
This PR adds a functionality:
- Fix #4445

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +668,82 @@ def load_variants(
...
+        if not variant_files:
             raise SyntaxError(
-                "VCF file {} {} does not seem to exist".format(category, variant_type)
+                "VCF files for {} {} does not seem to exist".format(category, variant_type)
             )
 
-        # Check if there are any variants in file
-        try:
+        for variant_file in variant_files:
+            if not self._has_variants_in_file(variant_file):
```
dnil, line 680: moved to a separate method

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +668,82 @@ def load_variants(
...
+                continue
+
             vcf_obj = VCF(variant_file)
-            var = next(vcf_obj)
```
dnil, line 670: This is the same, only it's in a loop for each of the variant files found for a category, type pair.

scout/adapter/mongo/case.py
```diff
@@ -884,25 +884,12 @@ def load_case(self, config_data: dict, update: bool = False, keep_actions: bool
...
-                "category": "cancer_sv",
-            },
-            {"file_name": "vcf_str", "variant_type": "clinical", "category": "str"},
-            {"file_name": "vcf_mei", "variant_type": "clinical", "category": "mei"},
-            {
-                "file_name": "vcf_fusion",
-                "variant_type": "clinical",
-                "category": "fusion",
-            },
+                "file_name": file_type,
```
dnil, line 888: Refactored to reuse the constant.

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +668,82 @@ def load_variants(
...
-        except StopIteration as err:
-            LOG.warning("Variant file %s does not include any variants", variant_file)
-            return nr_inserted
-        # We need to reload the file
-        vcf_obj = VCF(variant_file)
```
dnil, line 675: This is the same, only it's in a loop for each of the variant files found for a category, type pair.

dnil:
Ok, cool, looks like the loading framework is working - but we now see some issues, maybe as reported elsewhere https://github.com/Clinical-Genomics/scout/issues/4397.

Screenshot 2024-04-23 at 12 16 43

I guess I'll at least start debugging them here - makes little sense to have the extra load config option if we cant load real raredisease files.

dnil:
It's probably some new, slightly differently packaged goodies in the domain field. 

Before that, I'll just make sure we dont delete all snvs again when we load the mt ones.. 😆

dnil:
Ok this is blocking for now on the current `nf-core/raredisease` mt annotation. In particular we have html-escaped all(?) `:`. See https://github.com/Clinical-Genomics/raredisease/issues/20.

dnil:
Tentatively unblocked by https://github.com/nf-core/raredisease/pull/538!

dnil:
Does load, but appears to process variants twice. Will update. Also the new variant annotation does not crash now, but isn't exactly mature yet:
Screenshot 2024-05-22 at 13 45 11

northwestwitch:
> Does load, but appears to process variants twice.

Could be worse, like uploading variants 0 times ;)

dnil:
The "long" distance effects should be eliminated by https://github.com/Clinical-Genomics/servers/pull/1367. Will test again once that is through, to see if we see something else, but I think the PR is about ready then; the variant load thing should have been eliminated by using a set instead of an array... 😸

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +666,90 @@ def load_variants(
...
+                    raise SyntaxError("Specify chrom start and end")
+                region = "{0}:{1}-{2}".format(chrom, start, end)
+            else:
+                rank_threshold = rank_threshold or 0
+
+            variants = vcf_obj(region)
 
-        self.update_variant_rank(case_obj, variant_type, category=category)
+            try:
+                nr_inserted = self._load_variants(
```
dnil, line 728: This arg list really calls for a loader class instead... But, it will need changes either way if/when parallelising loading so I guess this is ok for now.

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +666,90 @@ def load_variants(
...
+                    rank_threshold=rank_threshold,
+                    institute_id=institute_id,
+                    build=build,
+                    rank_results_header=rank_results_header,
+                    vep_header=vep_header,
+                    category=category,
+                    sample_info=sample_info,
+                    custom_images=custom_images,
+                    local_archive_info=local_archive_info,
+                    gene_to_panels=gene_to_panels,
```
dnil, line 742: The last three args are here since we can now reuse these for both files. They could actually be made even earlier in `load_case()` but that will be even more passing stuff around (and changes to where and when `build` is defaulted) for relatively little processing.

scout/load/all.py
```diff
@@ -54,11 +54,11 @@ def load_region(adapter, case_id, hgnc_id=None, chrom=None, start=None, end=None
         start = gene_caption["start"]
         end = gene_caption["end"]
 
-    case_file_types = []
+    case_file_types = set()
```
dnil, line 57: Ahem, yes. Not more than one call per (type, category) tuple now.

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +666,90 @@ def load_variants(
...
-            self.delete_variants(case_obj["_id"], variant_type)
-            raise error
+            # Parse the necessary headers from vcf file
+            rank_results_header = parse_rank_results_header(vcf_obj)
+
+            local_archive_info = parse_local_archive_header(vcf_obj)
+
+            vep_header = parse_vep_header(vcf_obj)
+            if vep_header:
+                LOG.debug("Found VEP header %s", "|".join(vep_header))
```
dnil, line 695: Changed to a debug.

scout/adapter/mongo/variant_loader.py
```diff
@@ -658,85 +666,90 @@ def load_variants(
...
-        if not variant_file:
+        if not variant_files:
             raise SyntaxError(
-                "VCF file {} {} does not seem to exist".format(category, variant_type)
+                "VCF files for {} {} does not seem to exist".format(category, variant_type)
             )
 
-        # Check if there are any variants in file
-        try:
+        gene_to_panels = self.gene_to_panels(case_obj)
```
dnil, line 677: Moved these up to save a little compute when we reuse over a couple of files.

scout/adapter/mongo/hgnc.py
```diff
@@ -186,11 +186,11 @@ def all_genes(self, build=None, add_transcripts=False, limit=100000):
         if build == "GRCh38":
             build = "38"
 
-        LOG.info("Fetching all genes")
+        LOG.debug("Fetching all genes")
```
northwestwitch, line 189: 👍🏻

scout/adapter/mongo/case.py
northwestwitch: 🥇

scout/adapter/mongo/variant_loader.py
northwestwitch: 👍🏻

northwestwitch (approved):
Looks good to me! 👍🏻

PR #4600, opened 2024-05-02
Use latest gene panel for panel overlap in variantS, variant view and case report

### LINKED ISSUE #2733: Use latest gene panel for panel overlap in variantS and variant view
It appears that, when filtering the variant list with gene panels it is always using the latest version of the panel instead of the version saved in the case object in the database. I can only confirm that this is happening för SNVs but possibly also for other variant types.

Not sure if this is the indented behaviour or not. If it is the intended behaviour, feel free to close the issue, bit I think it should be reconsidered :)

northwestwitch:
- Fix #2733 -> Use latest gene panel for panel overlap in variantS, variant view and case report

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Use the same example as in the original issue -> https://github.com/Clinical-Genomics/scout/issues/2733#issuecomment-869855261

**Expected outcome**:
- Make sure variantS page, variant page and case report don't show panel1 as overlapping to the variant

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case_report.html
+++ b/scout/server/blueprints/cases/templates/cases/case_report.html
@@ -443,6 +443,18 @@
 </div>
 {% endmacro %}
 
+{% macro overlapping_panels(variant) %}
+{% if variant.case_panels %}
+  {% if variant.case_panels|length <= 3 %}
+    {% for panel in variant.case_panels %}
+      <a style="text-decoration:none;" href="{{ url_for('panels.panel', panel_id=panel_id) }}" target="_blank" class="badge bg-secondary">{{ panel.panel_name }}</a><br>
+    {% endfor %}
+  {% else %}
+    {{ variant.case_panels|length }} panels
+  {% endif %}
+{% endif %}
+{% endmacro %}
+
 {% macro variant_content(variant, index) %}
   {% if variant.category == 'snv' %}
     {{ sn_variant_content(variant, index) }}
@@ -544,15 +556,7 @@
               {% endif %}
             </td>
             <td>
-              {% if variant.panels%}
-                {% if variant.panels|length <= 3 %}
-                  {% for panel_id in variant.panels %}
-                   <a style="text-decoration:none;" href="{{ url_for('panels.panel', panel_id=panel_id) }}" rel="noopener" target="_blank" class="badge bg-secondary">{{ panel_id }}</a><br>
-                  {% endfor %}
-                {% else %}
-                  {{ variant.panels|length }} panels
-                {% endif %}
-              {% endif %}
+              {{ overlapping_panels(variant) }}
             </td>
           </tr>
         </tbody>
@@ -965,15 +969,7 @@
             <td>-</td>
           {% endif %}
           <td>
-            {% if variant.panels %}
-              {% if variant.panels|length <= 3 %}
-                {% for panel_id in variant.panels %}
-                  <a style="text-decoration:none;" href="{{ url_for('panels.panel', panel_id=panel_id) }}" rel="noopener" target="_blank" class="badge bg-secondary">{{ panel_id }}</a><br>
-                {% endfor %}
-              {% else %}
-                {{ variant.panels|length }} panels
-              {% endif %}
-            {% endif %}
+            {{ overlapping_panels(variant) }}
           </td>
           <td>
             {% if variant.callers %}
@@ -1183,15 +1179,7 @@
             {% endif %}
           </td>
           <td>
-            {% if variant.panels %}
-              {% if variant.panels|length <= 3 %}
-                {% for panel_id in variant.panels %}
-                  <a style="text-decoration:none;" href="{{ url_for('panels.panel', panel_id=panel_id) }}" target="_blank" class="badge bg-secondary">{{ panel_id }}</a><br>
-                {% endfor %}
-              {% else %}
-                {{ variant.panels|length }} panels
-              {% endif %}
-            {% endif %}
+            {{ overlapping_panels(variant) }}
           </td>
         </tr>
       </tbody>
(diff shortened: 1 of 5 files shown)


northwestwitch:
## Local test

- Remove gene **IFT43** from demo panel
- Pin [this variant](http://localhost:5000/cust000/643594/d57f0b1933560335fca26e6dd687e60c)

### VariantS page, this branch

image

### Variant page

image

### Case report

image

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -544,15 +556,7 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
...
-              {% if variant.panels%}
-                {% if variant.panels|length <= 3 %}
-                  {% for panel_id in variant.panels %}
-                   <a style="text-decoration:none;" href="{{ url_for('panels.panel', panel_id=panel_id) }}" rel="noopener" target="_blank" class="badge bg-secondary">{{ panel_id }}</a><br>
-                  {% endfor %}
-                {% else %}
-                  {{ variant.panels|length }} panels
-                {% endif %}
-              {% endif %}
+              {{ overlapping_panels(variant) }}
```
dnil, line 559: 🥇

scout/server/blueprints/variant/utils.py
```diff
@@ -3,6 +3,7 @@
 
 from scout.adapter import MongoAdapter
 from scout.constants import ACMG_COMPLETE_MAP, CALLERS, CLINSIG_MAP, SO_TERMS
+from scout.server.extensions import store
```
dnil, line 6: I would suggest passing store as an argument. It is a bit arbitrary. Parse is the most important to keep clean. This particular file will not be involved in an import cycle any day soon. But since the design is different from essentially all other controllers and their util functions: is there any particular reason to import a global module variable rather than pass the instance of it?

scout/server/blueprints/variant/utils.py
```diff
@@ -175,15 +176,16 @@ def update_variant_case_panels(case_obj: dict, variant_obj: dict):
...
-    ]
-
-    variant_obj["case_panels"] = case_panel_objs
+    for panel in case_obj.get("panels") or []:
+        if panel["panel_name"] not in variant_panel_names:
+            continue
+        latest_panel_hgnc_genes: List[int] = store.panel_to_genes(
+            panel_name=panel["panel_name"], gene_format="hgnc_id"
+        )
+        if set(latest_panel_hgnc_genes).intersection(variant_obj["hgnc_ids"]):
```
dnil, line 187: Nice! Is this faster than asking `not isdisjoint`?

dnil (changes requested):
Nicely fixed! Stylewise, unless there is good reason for changing it, we should keep passing the store object rather than global importing it from the module.

scout/server/blueprints/variant/utils.py
northwestwitch: I don't know but I can check. I'm a bit afraid that collecting the latest panel genes for each variant in the variantS list might slow down too much, so I'd like to do some more testing of this PR

dnil: Right, sounds wise. I think it has its given place on the variant page though.

scout/server/blueprints/variants/utils.py
```diff
@@ -1,23 +1,20 @@
 # -*- coding: utf-8 -*-
+from scout.adapter import MongoAdapter
 
 
-def update_case_panels(store, case_obj):
+def update_case_panels(store: MongoAdapter, case_obj: dict):
```
northwestwitch, line 5: This gets called immediately before update_variant_case_panels. And loops thru all gene panels of the case anyway, so this makes so that the latest panels of a case with relative genes are all collected under case["latest panels"] and used to do comparisons with variants genes downstream.

northwestwitch:
This one is ready for another round, next week is also fine @dnil , thanks! :)

scout/server/blueprints/variant/utils.py
```diff
@@ -660,17 +658,15 @@ def callers(variant_obj):
     return list(calls)
 
 
-def associate_variant_genes_with_case_panels(store: MongoAdapter, variant_obj: Dict) -> None:
+def associate_variant_genes_with_case_panels(case_obj: Dict, variant_obj: Dict) -> None:
```
dnil, line 661: 💯

scout/server/blueprints/variants/utils.py
dnil: Yes, trade a bit of storage on the case level for speed when looping over many variantS.

scout/server/blueprints/variants/utils.py
```diff
@@ -1,23 +1,20 @@
...
     If you do not use the returned variants, but rely on the update, please remember
     to call this function before updating the variants.
-
-    store(adapter.MongoAdapter)
-    case_obj(dict)
-
-    Returns:
-        list(panel_info)
     """
+    case_obj["latest_panels"] = []
```
dnil, line 12: 👍

scout/server/blueprints/variants/utils.py
```diff
@@ -1,23 +1,20 @@
 # -*- coding: utf-8 -*-
+from scout.adapter import MongoAdapter
 
 
-def update_case_panels(store, case_obj):
+def update_case_panels(store: MongoAdapter, case_obj: dict):
     """Refresh case gene panels with info on if a panel was removed.
+    Also sets the key "latest_panels" for a case where the value is the latest version of the case panels
 
-    Also return these more populated panels for optional storage on the variant_obj.
```
dnil, line 7: Mm, I have a feeling the intention was to let the caller decide wether to store or not, but this is very acceptable I think.

dnil (approved):
Looks good to me now! There will be a cost in storage and cpu but having a non-conflicting up-to-date gene panel result is satisfying. This solution has merit, but is a little fresh so perhaps let's test it a bit more before merging. In particular, I would like to be sure we don't accidentally store the variant_obj including the new panels on db in some upsert or other.

PR #4602, opened 2024-05-02
Fix diagnoses API search crashing with empty search string

### LINKED ISSUE #4601: Empty API diagnoses query crashes
https://scout.scilifelab.se/api/v1/diagnoses?query=

northwestwitch:
- Fix #4601 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Launch branch locally and go to this link --> http://localhost:5000/api/v1/diagnoses?query=

**Expected outcome**:
- Main branch should crash
- This branch should return many diagnoses

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/diagnoses/views.py
+++ b/scout/server/blueprints/diagnoses/views.py
@@ -39,6 +39,8 @@ def count_diagnoses():
 def api_diagnoses():
     """Return JSON data about diseases in the database."""
     query: Union[str, None] = request.args.get("query") or None
+    if query is None or len(query) < 3:
+        return jsonify({"error": "Diagnose query must be at least 3 characters long."})
     source: Union[str, None] = request.args.get("source") or None
     data: dict = controllers.disease_terms(store, query=query, source=source)
     return jsonify(data)
--- a/scout/server/blueprints/diagnoses/views.py
+++ b/scout/server/blueprints/diagnoses/views.py
@@ -39,8 +39,6 @@ def count_diagnoses():
 def api_diagnoses():
     """Return JSON data about diseases in the database."""
     query: Union[str, None] = request.args.get("query") or None
-    if query is None or len(query) < 3:
-        return jsonify({"error": "Diagnose query must be at least 3 characters long."})
     source: Union[str, None] = request.args.get("source") or None
     data: dict = controllers.disease_terms(store, query=query, source=source)
     return jsonify(data)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Block submission of somatic variants to ClinVar, until we don't introduce the changes needed to harmonise with their changed API
 - Additional control on the format of conditions provided in ClinVar form
 - Errors while loading managed variants from file are now displayed on the Managed Variants page
-- Diagnoses API accepts queries with at least 3 characters
 ### Fixed
 - Submit requests to Chanjo2 using HTML forms instead of JSON data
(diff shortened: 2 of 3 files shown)


dnil:
Hm, having some trouble replicating locally; getting worker timeouts with both main and this branch. I guess you tried on stage? 

It looks like the PR would indeed address the crash but not necessarily the endpoint issue, with fairly long, unintended response times for blank queries. Not sure if the blank one is used to load all for browsing when using the regular endpoint rather than API, but we could at the very least add a check in the API endpoint?

northwestwitch:
> Hm, having some trouble replicating locally; getting worker timeouts with both main and this branch. I guess you tried on stage?
> 
> It looks like the PR would indeed address the crash but not necessarily the endpoint issue, with fairly long, unintended response times for blank queries. Not sure if the blank one is used to load all for browsing when using the regular endpoint rather than API, but we could at the very least add a check in the API endpoint?

No I've tried only locally, where there are very few diagnoses. Must be expiring because of that. Good that you tried (we would have realised when testing for the release anyway, but better fixing it now!). I don't think the blank query is used any more, since the diseases are not pre-loading now. We could just check that the query contains at least 3 chars, like in the typehead thing, perhaps. Will fix tomorrow!

northwestwitch:
Tested now on stage and didn't time out:

image

What do you say? I can still add a check so the query is at least 3 chars

dnil:
Mm, our mongodb is faster than my demo setup. Good. Three chars is nice to keep the lookups small! I don't see that we need the empty queries for the API really, since the diagnoses page has its own endpoint. We could ask that there at least is some kind of content for the query as well. The next level after that might be some default limiting; it is unclear to me when the API ever needs to return that many disorders? Surely not for typeahead lists.

northwestwitch:
After the latest change:

image

scout/server/blueprints/diagnoses/views.py
```diff
@@ -39,6 +39,8 @@ def count_diagnoses():
 def api_diagnoses():
     """Return JSON data about diseases in the database."""
     query: Union[str, None] = request.args.get("query") or None
+    if query is None or len(query) < 3:
```
dnil, line 42: 👍

dnil: Looks good, but I'm afraid you would have to make a couple of accompanying changes to the UI, where this is used? Like either catching that error code, or ensuring the user can't actually submit < 3 chars?
These hang indefinitely without explanation with fewer than three chars. 😊 
Screenshot 2024-05-03 at 09 15 20

northwestwitch: Right. I was thinking at [another recent change](https://github.com/Clinical-Genomics/scout/pull/4592) to the functionality. So now if you search for OMI you have 3 chars but it's a useless query anyway. Why don't we simplify and just keep it so that the user can insert anything, even blank string. I don't think it will hang on stage or prod because of the performance of mongo there. And if it hangs they'll look for something else anyway. It was like this even before..

dnil: Right, and they can easily ask for all diagnoses using the regular endpoint anyway, so I guess it's not that much of a difference if they can do it via the API. We can modify after the first DDOS... 😁

dnil (approved):
Ok, let's not overcomplicate: this would be a lot of fixing for a small thing. Let's revert to not checking for a minimum query length. It then does what it says, preventing crash on empty query.

PR #4636, opened 2024-05-20
Add better error handling for chanjo report fails.

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. set SQLALCHEMY_DATABASE_URI but use a broken chanjo install
2. notice the import error exception in the scout warnings, and no report crash on the case page (but also no reports obviously)
Screenshot 2024-05-20 at 16 20 29
3. set SQLALCHEMY_DATABASE_URI but use a broken chanjo install
4. same-same, with exception error displayed
Screenshot 2024-05-20 at 16 43 10

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/config.py
```diff
@@ -39,7 +39,7 @@
 # )
 
 # Chanjo database connection string - used by chanjo report to create coverage reports
-# SQLALCHEMY_DATABASE_URI = "mysql+pymysql://test_user:test_passwordw@127.0.0.1:3306/chanjo"
+SQLALCHEMY_DATABASE_URI = "mysql+pymysql://test_user:test_passwordw@127.0.0.1:3306/chanjo"
```
northwestwitch, line 42: Uncommenting this line will create links to the report but when you click on them the app will crash. But perhaps we don't care?

scout/server/app.py
```diff
@@ -40,11 +40,11 @@
     from chanjo_report.server.app import configure_template_filters
     from chanjo_report.server.blueprints import report_bp
     from chanjo_report.server.extensions import api as chanjo_api
-except ImportError:
+except ImportError as error:
     chanjo_api = None
     report_bp = None
     configure_template_filters = None
-    LOG.info("chanjo report not installed!")
+    LOG.warning("chanjo-report not properly installed! %s", error)
```
northwestwitch, line 47: raise an error with the right message instead? So we realise immediately that chanjo-report wasn't installed instead of going to case page that crashes and perhaps we have to open a new PR to fix it?

northwestwitch (commented):
This works even if the demo crashes when you click on the report links. 

What we saw so far is that chanjo-report can't be installed for some reason, scout seems to work but crashes on the case page. For this reason I would go a step ahead and directly prevent the app from starting instead of starting with a warning that we won't read until case page crashes, because at that point we would have a bug already

scout/server/app.py
dnil: Ah, but no, we still want it to run without chanjo installed... Thats kind of the reason there is an except at all?

dnil: Ah, but good, lets rephrase the warning (or not) depending on the configuration!

dnil: Which I can't because the app isn't up yet there. Oh well.

scout/server/config.py
dnil: I think we will want to keep it commented out in the config example - this is just for testing!

scout/server/app.py
northwestwitch: I think if you have chanjo settings in the config file (the mysql uri) you want to use coverage. In the situation when chanjo-report has problems (is not installed) the the app shouldn't start at all. This way you'd know immediately that you have to fix it. This is what I mean

scout/server/config.py
dnil: The intention is that we should just as before get no links if the chanjo module could not load, or if there is no `SQLALCHEMY_DATABASE_URI` configured. In case of config time errors, like db connection or password issues, we would still get a report time crash. We could move such errors up earlier by doing like a health check on the connection directly after init, but maybe its a bit overkill?

northwestwitch: overkill, yes!

scout/server/app.py
dnil: I'm not sure I follow you again: are you saying you are able to get the the report with a broken install of chanjo? That should not happen!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Fix several tests that relied on number of events after setup to be 0
 - Removed unused load case function
 - Artwork logo sync sketch with png and export svg
+- Clearer exception handling on chanjo-report setup
```
northwestwitch, line 17:
```suggestion
- App running without coverage reports links instead of crashing when chanjo-report installation fails. 
```

northwestwitch (approved):
Looks good to me!

dnil:
Now that we agree it was in fact working well, lets also incorporate the good idea of failing early:
Screenshot 2024-05-21 at 09 51 08
And actually on tests near you:
Screenshot 2024-05-21 at 09 52 22

PR #4646, opened 2024-05-28
Fix #4645 - benign modifiers

### LINKED ISSUE #4645: ACMG benign modifiers apply to is_likely_pathogenic and is_pathogenic
We noticed when filling out the ACMG classifier that modifiers to benign classes would apply to the pathogenic terms as well.

We had this situation:
PS3_moderate
PP1_moderate
PP3

This correctly defaults to uncertain significance. But, when adding a benign
BS1_supporting

The classification instead turned into likely_pathogenic

It seems like this code snippet does not separate whether or not a term is of pathogenic origin or benign
https://github.com/Clinical-Genomics/scout/blob/d448461287114e5a79d684ed2e67c913aea1906e/scout/utils/acmg.py#L193

You can test this further by just adding benign classifications and giving them strong modifiers. This results in pathogenic and likely_pathogenic outcomes. As I understand it, it's quite rare to have modifiers on benign classes but in this case, following the strict guidelines the geneticist came to this outcome.

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN, also new automated unit test


dnil:
Before:
Screenshot 2024-05-28 at 16 55 29

After:
Screenshot 2024-05-28 at 16 59 29

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Clearer exception handling on chanjo-report setup - fail early and visibly
 - mtDNA report crashing when one or more samples from a case is not in the chanjo database
 - Case page crashing on missing phenotype terms
+- ACMG benign modifiers
```
northwestwitch, line 23:
```suggestion
- ACMG benign modifiers that modified pathogenic terms as well
```

scout/utils/acmg.py
```diff
@@ -192,11 +192,18 @@ def get_acmg(acmg_terms):
     bp_terms = []
     for term in acmg_terms:
         if term.endswith("_Strong"):
-            ps_terms.append(term)
```
northwestwitch, line 195: Nice! Looking at the entire function there is a long list of ifs and I wonder if it could be simplified?

ChatGPT suggests the following, but I haven't tried it:

```
# Create dictionaries to map suffixes and prefixes to corresponding lists
suffix_map = {
    "_Strong": {"P": ps_terms, "B": bs_terms},
    "_Moderate": {"P": pm_terms},
    "_Supporting": {"P": pp_terms, "B": bp_terms},
}

prefix_map = {
    "PVS": (lambda: setattr(self, 'pvs', True)),
    "PS": ps_terms.append,
    "PM": pm_terms.append,
    "PP": pp_terms.append,
    "BA": (lambda: setattr(self, 'ba', True)),
    "BS": bs_terms.append,
    "BP": bp_terms.append,
}

# Iterate through the terms
for term in acmg_terms:
    # Check suffixes first
    for suffix, prefix_dict in suffix_map.items():
        if term.endswith(suffix):
            for prefix, term_list in prefix_dict.items():
                if term.startswith(prefix):
                    term_list.append(term)
                    break
            break
    else:  # Only check prefixes if no suffix match
        for prefix, action in prefix_map.items():
            if term.startswith(prefix):
                action(term)
                break
```

northwestwitch (approved):
Nice fix, but I wonder if the code could be simplified?

scout/utils/acmg.py
dnil: Mm, agree there are lots of if/elif. But dont you think this code is a little eh strained? It looks a bit like it is trying to be LISP. It would do slightly more comparisons, but without testing I would also say it is probably correct (would have to double check e.g. the PVS assignment). It would help a lot to change the pvs and ba to arrays, but then one would have to update a million tests..

northwestwitch: I like the prefix and suffix map - kind of approach, but feel free to use the suggestion or not. Itäs working after all!

PR #4659, opened 2024-06-04
Small event projection

dnil:

I added a very simple hiding functionality for matching causatives. They will now only show (and compute) on the case page if requested. 

Screenshot 2024-06-04 at 13 16 59

And clicking it:
Screenshot 2024-06-04 at 13 17 07

I'm sure this could be rather smoothly done in js as well with a request to fill in that part of the document and does the compute in a new api endpoint. But simple can be good. Thoughts?

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -281,7 +281,7 @@ def sma_case(store, institute_obj, case_obj):
     return data
 
 
-def case(store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching: str) -> dict:
+def case(store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching = "yes": str) -> dict:
     """Preprocess a single case.
 
     Prepare the case to be displayed in the case view.
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -391,19 +391,19 @@ def case(store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching
             "case_images", case_obj["custom_images"].get("case", {})
         )
 
-    # Limit secondary findings according to institute settings
-    limit_genes = store.safe_genes_filter(institute_obj["_id"])
-
-    limit_genes_default_panels = _limit_genes_on_default_panels(
-        case_obj["default_genes"], limit_genes
-    )
-
     other_causatives = []
     other_causatives_in_default_panels = []
     default_managed_variants = []
     managed_variants = []
 
     if hide_matching == "no":
+        # Limit secondary findings according to institute settings
+        limit_genes = store.safe_genes_filter(institute_obj["_id"])
+
+        limit_genes_default_panels = _limit_genes_on_default_panels(
+            case_obj["default_genes"], limit_genes
+        )
+
         other_causatives, other_causatives_in_default_panels = _matching_causatives(
             store, case_obj, limit_genes, limit_genes_default_panels
         )
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -1424,13 +1424,13 @@ def _matching_causatives(
     other_causatives_in_default_panels = []
 
     for causative in matching_causatives:
-        hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
+        hgnc_ids = {gene.get("hgnc_id") for gene in causative.get("genes", [])}
         # Fetch all matching causatives if no causatives_filter defined
         # or only causatives matching the filter:
-        if not other_causatives_filter or (set(hgnc_ids) & set(other_causatives_filter)):
+        if not other_causatives_filter or (hgnc_ids & set(other_causatives_filter)):
             other_causatives.append(causative)
         # Only matching causatives in default gene panels:
-        if set(hgnc_ids) & set(other_causatives_in_default_panels_filter):
+        if hgnc_ids & set(other_causatives_in_default_panels_filter):
             other_causatives_in_default_panels.append(causative)
 
     return other_causatives, other_causatives_in_default_panels
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -281,22 +281,14 @@ def sma_case(store, institute_obj, case_obj):
     return data
 
 
-def case(store, institute_obj, case_obj):
+def case(store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching: str) -> dict:
     """Preprocess a single case.
 
     Prepare the case to be displayed in the case view.
 
-    Args:
-        store(adapter.MongoAdapter)
-        institute_obj(models.Institute)
-        case_obj(models.Case)
-
-    Returns:
-        data(dict): includes the cases, how many there are and the limit.
-
+    The return data dict includes the cases, how many there are and the limit.
     """
     # Convert individual information to more readable format
-
     _populate_case_individuals(case_obj)
 
     case_obj["assignees"] = [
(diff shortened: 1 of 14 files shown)


scout/server/blueprints/cases/controllers.py
```diff
@@ -1424,13 +1424,13 @@ def _matching_causatives(
     other_causatives_in_default_panels = []
 
     for causative in matching_causatives:
-        hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
+        hgnc_ids = {gene.get("hgnc_id") for gene in causative.get("genes", [])}
```
dnil, line 1427: This is only ever used as a set anyway; no need to call the set constructor several times.

dnil:
Checked again, and I can't help think the most salient thing is one huge db query with a a couple of different `$in: my_large_variant_or_gene_list`. Ill see if we can't do it a bit less often..

scout/server/blueprints/cases/controllers.py
```diff
@@ -281,22 +281,16 @@ def sma_case(store, institute_obj, case_obj):
     return data
 
 
-def case(store, institute_obj, case_obj):
+def case(
+    store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching: str = "yes"
```
northwestwitch, line 285:
```suggestion
    store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching: str = "yes"
```
```suggestion
    store: MongoAdapter, institute_obj: dict, case_obj: dict, hide_matching: bool = True
```

northwestwitch: Yes/no kind of data call for a boolean, no?

scout/server/blueprints/cases/controllers.py
```diff
@@ -399,31 +393,40 @@ def case(store, institute_obj, case_obj):
...
-    limit_genes = store.safe_genes_filter(institute_obj["_id"])
+    other_causatives = []
+    other_causatives_in_default_panels = []
+    default_managed_variants = []
+    managed_variants = []
 
-    limit_genes_default_panels = _limit_genes_on_default_panels(
-        case_obj["default_genes"], limit_genes
-    )
+    if hide_matching == "no":
```
northwestwitch, line 401:
```suggestion
    if hide_matching is False:
```

scout/server/blueprints/cases/controllers.py
```diff
@@ -1424,13 +1428,13 @@ def _matching_causatives(
...
     for causative in matching_causatives:
-        hgnc_ids = [gene.get("hgnc_id") for gene in causative.get("genes", [])]
+        hgnc_ids = {gene.get("hgnc_id") for gene in causative.get("genes", [])}
         # Fetch all matching causatives if no causatives_filter defined
         # or only causatives matching the filter:
-        if not other_causatives_filter or (set(hgnc_ids) & set(other_causatives_filter)):
+        if not other_causatives_filter or (hgnc_ids & set(other_causatives_filter)):
             other_causatives.append(causative)
         # Only matching causatives in default gene panels:
-        if set(hgnc_ids) & set(other_causatives_in_default_panels_filter):
```
northwestwitch, line 1433: 💯

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -120,30 +120,39 @@ <h4 class="h4 d-inline-block">
...
-        <div class="row mt-0">
-          <div class="col-xs-12 col-md-12">{{ matching_causatives(other_causatives, institute, case) }}</div>
-        </div>
-      {% endif %}
-      {% if default_other_causatives|length > 0%}
-        <div class="row mt-0">
-          <div class="col-xs-12 col-md-12">{{ matching_causatives(default_other_causatives, institute, case, default=True) }}</div>
-        </div>
-      {% endif %}
+      {% if hide_matching == "no" %}
```
northwestwitch, line 123:
```suggestion
      {% if hide_matching == false %}
```

northwestwitch (approved):
While it doesn't solve the slowness of loading the page when looking for matching causatives, I feel this solution might be a good compromise to load the page when causatives don't have to be shown. I guess not all users would like it, but perhaps we could try it and see the feedback we get?

As I mentioned in my comments I would make that `hide_matching` variable a boolean.

Other thing that could be improved is adding some spacing on the left to text in this panel (it's like this on main like now):

image

scout/server/blueprints/cases/controllers.py
dnil: Well, the trouble is navigating arg conversion; between jinja-browser-wsgi-flask-python booleans are handled slightly differently. We've had that before with "false"/"False"/False/false-representations that suddenly fail in some situations. It is so much easier to pass a string and test for it.

dnil:
> While it doesn't solve the slowness of loading the page when looking for matching causatives, I feel this solution might be a good compromise to load the page when causatives don't have to be shown. I guess not all users would like it, but perhaps we could try it and see the feedback we get?

Thank you for the review! Ill ask around a little today and try to form a final opinion.

> As I mentioned in my comments I would make that `hide_matching` variable a boolean.

Yes, that would be more natural, in the best of worlds. Its pragmatic. 😁 May I say be my guest and refactor that separately if you feel strongly about? 😊  It is nerve wrecking to get that to work in all situations, mostly since flask and jinja treat bools differently (false vs False and the browser needs to play along with "false" /"False") - see comment. 
 
> Other thing that could be improved is adding some spacing on the left to text in this panel (it's like this on main like now):

Yes! And I was thinking maybe to add a little chevron or sth to indicate there is stuff to "fold out" under it?

scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -120,30 +120,39 @@ <h4 class="h4 d-inline-block">
...
+          <div class="row mt-0">
+            <div class="col-sm-12 col-md-12">{{ matching_managed_variants(default_managed_variants, institute, case, default=True) }}</div>
+          </div>
+        {% endif %}
+      {% else %}
         <div class="row mt-0">
-          <div class="col-sm-12 col-md-12">{{ matching_managed_variants(default_managed_variants, institute, case, default=True) }}</div>
+          <div class="col-sm-12 col-md-12">
+           <div data-bs-toggle='tooltip' class="panel-heading" title="If there are any variants in this case
+            that have been marked as causative in another case for this insitute, or are on the managed variants list.">
```
northwestwitch, line 150:
```suggestion
            that have been marked as causative in another case for this institute, or are on the managed variants list.">
```

scout/server/blueprints/cases/views.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import os.path
 import shutil
+from ast import literal_eval
```
dnil, line 7: 😆

dnil:
> > Thank you for the review! Ill ask around a little today and try to form a final opinion.
> 
> 👍🏻
> 
> > > As I mentioned in my comments I would make that `hide_matching` variable a boolean.
> > 
> > 
> > Yes, that would be more natural, in the best of worlds. Its pragmatic. 😁 May I say be my guest and refactor that separately if you feel strongly about it? 😊 It is nerve wrecking to get that to work in all situations, mostly since flask and jinja treat bools differently (false vs False and the browser needs to play along with "false" /"False") - see comment.
> 
> I've pushed some code that seems to work, see how you like it. Feel free to revert if you don't.

Ok, I'll definitely test it out with a couple of browsers since you took the time: I think it has potential! But not entirely sure having to go via `ast` to keep it kind of bool is better than passing a string. 😄

dnil:
Looks good on stage as well: 
Screenshot 2024-06-05 at 12 12 39

PR #4684, opened 2024-06-19
Workaround fix #4682 - sort case file load order alphabetically

### LINKED ISSUE #4682: SV variants and large SNVs collide in variant_id
**Describe the bug**
While normally describing SVs with position and type, e.g. `chr1_123456_T_DUP` or such, an SV caller (manta) may elect to report small variants, and if the variant is small enough, even represent the full allele sequence, e.g. `chr1_123456_T_TT`. These variants typically also occur in the SNV/INDEL file, giving the same variant id. Depending on what file is parsed first, the variant will only appear in that type variant list. This is normally not a problem, but sometimes an institute may elect not to analyse variants of a certain type, e.g. only SNVs, not SVs, in which case the variant could be missed.

**To Reproduce**
1. Load a case with some collisions, e.g. `engaginggrouse`. Note how there are ID warnings:
```
2024-06-19 09:32:06 hasta.scilifelab.se scout.adapter.mongo.variant_loader[12314] WARNING Variant 900f6c92533766eabeb10a8573297f93 already exists in database - modifying
2024-06-19 09:32:06 hasta.scilifelab.se scout.adapter.mongo.variant_loader[12314] WARNING Variant 8fb3f02b57613d36aefa0b690eeb8e8f already exists in database - modifying
```
2. Note how the variant is now only shown on the variant type page that was parsed first, here the `cancer_sv`.

**Expected behavior**
We would (arguably, this is non-obvious) still like to see the duplicated variants on both type views. A workaround might be to ensure SVs are loaded last and modify colliding small SV calls to be of the less informative, big SV type. We should carefully check that the matching causatives behaviour is still acceptable. The variant could now be tallied in two different views, which might get a bit complex.

**Additional context**
Manta, which is the caller we currently primarily see with this behaviour, does not seem to have an option to force long-sv style entries.

dnil:

The load order was perturbed last week with 4.83 introducing sets of file type categories, to accommodate nf-core raredisease MT VCFs. Introducing an alpabetical sort will work around the problem encountered in #4682 in the sense that SNVs will be loaded before SVs, and in particular `cancer` before `cancer_sv`, and the collision resolution should be consistent.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -14,6 +14,7 @@ from scout.constants import ACMG_MAP, FILE_TYPE_MAP, ID_PROJECTION
 from scout.exceptions import ConfigError, IntegrityError
 from scout.parse.variant.ids import parse_document_id
 from scout.utils.algorithms import ui_score
+from scout.utils.sort import get_lowest_load_priority
 
 LOG = logging.getLogger(__name__)
 
@@ -904,7 +905,10 @@ class CaseHandler(object):
                     continue
                 load_variants.add((vcf_file["variant_type"], vcf_file["category"]))
 
-            for variant_type, category in sorted(load_variants, key=lambda tup: tup[1]):
+            for variant_type, category in sorted(
+                load_variants,
+                key=lambda tup: get_lowest_load_priority(variant_type=tup[0], category=tup[1]),
+            ):
                 if update:
                     self.delete_variants(
                         case_id=case_obj["_id"],
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -904,7 +904,7 @@ class CaseHandler(object):
                     continue
                 load_variants.add((vcf_file["variant_type"], vcf_file["category"]))
 
-            for variant_type, category in load_variants:
+            for variant_type, category in sorted(load_variants, key=lambda tup: tup[1]):
                 if update:
                     self.delete_variants(
                         case_id=case_obj["_id"],
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -14,7 +14,7 @@ from scout.constants import ACMG_MAP, FILE_TYPE_MAP, ID_PROJECTION
 from scout.exceptions import ConfigError, IntegrityError
 from scout.parse.variant.ids import parse_document_id
 from scout.utils.algorithms import ui_score
-from scout.utils.sort import get_lowest_load_priority
+from scout.utils.sort import get_load_priority
 
 LOG = logging.getLogger(__name__)
 
@@ -907,7 +907,7 @@ class CaseHandler(object):
 
             for variant_type, category in sorted(
                 load_variants,
-                key=lambda tup: get_lowest_load_priority(variant_type=tup[0], category=tup[1]),
+                key=lambda tup: get_load_priority(variant_type=tup[0], category=tup[1]),
             ):
                 if update:
                     self.delete_variants(
--- a/scout/adapter/mongo/variant_loader.py
+++ b/scout/adapter/mongo/variant_loader.py
@@ -27,7 +27,7 @@ from scout.parse.variant.headers import (
 from scout.parse.variant.ids import parse_simple_id
 from scout.parse.variant.managed_variant import parse_managed_variant_id
 from scout.parse.variant.rank_score import parse_rank_score
-from scout.utils.sort import get_lowest_load_priority
+from scout.utils.sort import get_load_priority
 
 LOG = logging.getLogger(__name__)
 
@@ -216,7 +216,7 @@ class VariantLoader(object):
             intervals = coding_intervals.get(chrom, IntervalTree())
             for var_type, category in sorted(
                 list(load_variants),
-                key=lambda tup: get_lowest_load_priority(variant_type=tup[0], category=tup[1]),
+                key=lambda tup: get_load_priority(variant_type=tup[0], category=tup[1]),
             ):
                 LOG.info(
                     "Updating compounds on chromosome:{0}, type:{1}, category:{2} for case:{3}".format(
--- a/scout/adapter/mongo/variant_loader.py
+++ b/scout/adapter/mongo/variant_loader.py
@@ -203,14 +203,12 @@ class VariantLoader(object):
         case_id = case_obj["_id"]
         # Possible categories 'snv', 'sv', 'str', 'cancer', 'cancer_sv'. Sort according to load order to ensure Cancer SNVs before
         # Cancer SVs, in particular, and keep a consistent variant_id collision resolution order.
-        # Possible variant types 'clinical', 'research':
-        load_variants = set()
-
-        for file_type in FILE_TYPE_MAP:
-            if case_obj.get("vcf_files", {}).get(file_type):
-                load_variants.add(
-                    (FILE_TYPE_MAP[file_type]["variant_type"], FILE_TYPE_MAP[file_type]["category"])
-                )
+        # Possible variant types are 'clinical', 'research'.
+        load_variants = {
+            (FILE_TYPE_MAP[file_type]["variant_type"], FILE_TYPE_MAP[file_type]["category"])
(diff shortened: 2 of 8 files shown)


scout/constants/file_types.py
```diff
@@ -2,11 +2,13 @@
 
 FILE_TYPE_MAP = {
```
northwestwitch, line 3: Make this one an OrderedDict?

dnil: That was my first thought, but it won't really help since we then add fields from the member dicts to separate sets. Its those that need to be ordered.

northwestwitch: While it works it feels that the solution relies entirely on the fact that the naming of the variables, that are OK by chances. What is we need to introduce new categories in the future or we forget why we introduced that sorting? 🤔

scout/constants/file_types.py
```diff
@@ -2,11 +2,13 @@
...
     "vcf_cancer": {"category": "cancer", "variant_type": "clinical"},
```
northwestwitch, line 4: Something like this?

```suggestion
    "vcf_cancer": {"category": "cancer", "variant_type": "clinical", "loading_order":0},
    "vcf_cancer_research": {"category": "cancer", "variant_type": "research", "loading_order":1},
    ...
```

scout/constants/file_types.py
dnil: Well, it's more that it is the same as before than that it is ok: there is really no way to say a 52 bp variant should rather be an sv or an snv.

northwestwitch: Agreed, but I guess it's also OK to define an arbitrary order, because that's what all is about

scout/constants/file_types.py
dnil: Mm, yes, or perhaps a priority. It will seem a bit clunky, but at least very clear that there is a set order. I'll take a look.

scout/constants/file_types.py
dnil: Ok, here goes. With this, we are now free to deviate from alphabetical order, and until we have one that does this could also potentially compete in a contest for obfuscated python. 😸

northwestwitch: Thanks, I'll review first thing tomorrow!

scout/adapter/mongo/variant_loader.py
```diff
@@ -200,87 +201,90 @@ def update_case_compounds(self, case_obj, build="37"):
...
         # Loop over all intervals
         for chrom in CHROMOSOMES:
             intervals = coding_intervals.get(chrom, IntervalTree())
-            for var_type in variant_types:
-                for category in categories:
-                    LOG.info(
-                        "Updating compounds on chromosome:{0}, type:{1}, category:{2} for case:{3}".format(
-                            chrom, var_type, category, case_id
-                        )
+            for var_type, category in sorted(
```
dnil, line 219: This is the change: the rest is just indentation.

scout/constants/file_types.py
```diff
@@ -1,21 +1,90 @@
...
-    "vcf_sv_mt": {"category": "sv", "variant_type": "clinical"},
-    "vcf_sv_research": {"category": "sv", "variant_type": "research"},
-    "vcf_sv_research_mt": {"category": "sv", "variant_type": "research"},
-    "vcf_str": {"category": "str", "variant_type": "clinical"},
-    "vcf_mei": {"category": "mei", "variant_type": "clinical"},
-    "vcf_mei_research": {"category": "mei", "variant_type": "research"},
+    "vcf_cancer": {
+        "category": "cancer",
+        "variant_type": "clinical",
+        "load_priority": 10,
```
northwestwitch, line 8: I'm not following the reasoning behind the numbering here, why 10, 120.. and not 1,2,3 etc?

scout/adapter/mongo/case.py
```diff
@@ -904,7 +905,10 @@ def load_case(self, config_data: dict, update: bool = False, keep_actions: bool
                     continue
                 load_variants.add((vcf_file["variant_type"], vcf_file["category"]))
 
-            for variant_type, category in load_variants:
+            for variant_type, category in sorted(
```
northwestwitch, line 908: Grasping of the lambda function for sorting has never been my forte. Having admitted this, I find it quite complicated to follow when also calling the new `get_lowest_load_priority` function. 
Also the naming **lowest load priority** is confusing to me when I would expect that the **high** prio ones would be first. 

I was wondering if the function is avoidable at all.. Like if at line 887 of this file you have:

```
# load from files
        files = [
            {
                "file_name": file_type,
                "variant_type": FILE_TYPE_MAP[file_type]["variant_type"],
                "category": FILE_TYPE_MAP[file_type]["category"],
                "priority": FILE_TYPE_MAP[file_type]["load_priority"] <--------------------------
            }
            for file_type in FILE_TYPE_MAP.keys()
            if FILE_TYPE_MAP[file_type]["variant_type"] != "research"
        ]
```
And then again at line 905:
`load_variants.add((vcf_file["variant_type"], vcf_file["category"], vcf_file["priority"]))`


And then more simply (NOT simply! 😆 )

`sorted(load_variants, key=lambda tup: tup[2])` ??

But perhaps I am missing something. I haven't spent as much time on it as you did..

northwestwitch (commented):
Thanks for the code update! I have some questions, see comments. This is partly due to my problems with the lambda function, so I apologize in advance 😄

scout/constants/file_types.py
dnil: To be able to quickly add intermediate priorities without having to renumber all.

scout/adapter/mongo/case.py
dnil: It would be really simple if we only ever wanted to sort on category prio. The tricky thing is that it’s two different calls, an outer one asking to load the highest prio category, type and then an inner one asking for the highest prio file. If the call was in one loop like directly enumerating the files it would also be very simple.

dnil: And lowest prio value first is standard in *nix isn’t it? As in lowest nice value goes first. It is also closest to an order numbering without having to make it contiguous.

scout/constants/file_types.py
northwestwitch: Mmm I'd rather change all numbers when we add a category so we have a clear sequence of numbers then having gaps in between. I find it more clear.

northwestwitch (approved):
Approving because it's working and we need a quick fix, but I'm not very fond of the solution. I think it's good for now but we should consider refactoring soon for simplifying so the code would be more readable and maintainable.

But as I said, it works so it's good for me! 👍🏻

scout/adapter/mongo/case.py
northwestwitch: > lowest prio value first is standard in *nix isn’t it?

We are human beings before being programmers, aren't we? ;)

dnil: And humans never say prio 1?

northwestwitch: Whatever, glad midsommar! :)

dnil:
> Approving because it's working and we need a quick fix, but I'm not very fond of the solution. I think it's good for now but we should consider refactoring soon for simplifying so the code would be more readable and maintainable.
> 
> But as I said, it works so it's good for me! 👍🏻

I liked alphabetical better too. :) We could make two ordered dicts, one with categories and one with files. Or we could refactor to make it one call per file instead of the outer and inner thing. It probably doesn’t save much to do a few things only once per category. But not this week! 😸

scout/adapter/mongo/case.py
dnil: This appears a classic design conflict, with no obvious resolution. Lower numer, higher prio is more common (defcon, unix system prio, and as spoken), but it is by no means completely obvious. I added some more explanation on this to the docstrings.

PR #4692, opened 2024-06-28
SV gene variantS search

### LINKED ISSUE #1179: Retrieve also SVs by searching by gene and ranking?

dnil:

- Fix #1179 - With mongodb 5.0 and higher it is possible to hold multiple partial indices on the same collection and keys, given that the partial criteria do not overlap. Use that to be able to set up a gene variantS search for the SV category as well. An initial test seemed decently quick, so maybe we will not need to filter on size, which is a bit complicated with inf and small sizes in the bag. 
- The hits may be a bit messy with large variants, so we will at least add a filter on category to the search page. A "length" filter could be considered as well; could be part of this PR or a later issue if it is needed.

- Backgrounding is not needed with the new mongodb 7.0 indexing logic. They do a blocking lock during the start and finish of the indexing, but otherwise runs a very fast, but still read-write indexing. Neato!

We prepared for this during the db migration, and have a category-less version of this index in place on prod.

The stage collection built quickly, opting for the same style as on prod, ie one index for all variant categories. rankscore >= 5 effectively excludes most of cancer, Mei variants, etc that are not yet ranked.
Screenshot 2024-06-28 at 16 05 07

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. search for a few genes on stage, note both SVs and SNVs in results
2. toggle search to only SNV or only SV and see that the filter behaves

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN


dnil:
I'll just add the search-page filter part here as well I guess; it won't make sense without the index anyway. Draft again for now then.

scout/adapter/mongo/query.py
```diff
@@ -76,7 +77,11 @@ def delete_variants_query(
         return variants_query
 
     def build_variant_query(
-        self, query=None, institute_ids=[], category="snv", variant_type=["clinical"]
+        self,
+        query: Optional[dict] = None,
+        institute_ids: Optional[list] = [],
+        category: Optional[Union[str, list]] = "snv",
+        variant_type: Optional[list] = ["clinical"],
```
northwestwitch, line 84: 🥇

scout/constants/indexes.py
```diff
@@ -12,17 +12,14 @@
         IndexModel(
             [("build", ASCENDING), ("chromosome", ASCENDING)],
             name="build_chromosome",
-            background=True,
```
northwestwitch, line 15: 👍🏻 Hope people using scout are also using a DB version >7 🤔

northwestwitch (commented):
Brilliant! Code-wise looks good and works nicely on stage. Users will be very happy with this feature! 🚀 

I notice a bunch of notifications from CodeQL (It was me adding the code scanning BTW). I guess no need to address them in this PR, but we could perhaps plan a fix (mitigation) in another PR. I'll open an issue.

northwestwitch (approved):
I forgot to select approve :)

northwestwitch:
This one is ready to merge I think!

scout/constants/indexes.py
dnil: Right, but probably best to stick with the times anyway? The "already existing" indices not backgrounding is not going to hurt, and a fresh install would hopefully be made on a current engine. The worst for a potential existing large old mongod install is the new index, but then it will be only that single one, and it would not be production ready immediately anyway.

northwestwitch: Agreed, let's merge!

PR #4697, opened 2024-07-03
Sort select default

### LINKED ISSUE #4695: Removed panels are displayed on case page
On the panel page, we can see that the v1.0 "albinism" has been removed:

scout_removed_panel

But it is still displayed among the panels on the case page:

scout_removed_panel_shown

This is true generally for removed panels.

I suspect that this is a bug - that the removed panels shouldn't be displayed.

dnil:
- Fix #4695 
- Fix sorting of the select default panels dropdown - as discussed in #4695 

Marking this as a bug fix as the display of the "removed" badge was there once upon a time, but now only for the default panels. Refactored a little to do the panel hiding check separate from the default panel check and latest panel genes update.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. remove a panel present on a case
2. look at the case page and its dropdown
✅ Screenshot 2024-07-03 at 11 45 52
✅ Screenshot 2024-07-03 at 11 47 08

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Unfreeze werkzeug again
 ### Fixed
 - Sort "select default panels" dropdown menu options on case page
+- Show gene panel removed status on case page
 
 ## [4.84]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ## [unreleased]
 ### Changed
 - Unfreeze werkzeug again
-
+### Fixed
+- Sort "select default panels" dropdown menu options on case page
 
 ## [4.84]
 ### Changed
--- a/scout/server/blueprints/cases/controllers.py
+++ b/scout/server/blueprints/cases/controllers.py
@@ -63,7 +63,8 @@ JSON_HEADERS = {
 
 COVERAGE_REPORT_TIMEOUT = 20
 
-PANEL_PROJECTION = {"version": 1, "display_name": 1, "genes": 1, "hidden": 1}
+PANEL_PROJECTION = {"version": 1, "display_name": 1, "genes": 1}
+PANEL_HIDDEN_PROJECTION = {"version": 1, "display_name": 1, "hidden": 1}
 
 
 def phenomizer_diseases(hpo_ids, case_obj, p_value_treshold=1):
@@ -338,7 +339,8 @@ def case(
     case_obj["clinvar_variants_not_in_suspects"] = clinvar_variants_not_in_suspects
 
     case_obj["default_genes"] = _get_default_panel_genes(store, case_obj)
-    flash("DEBUG panels {}".format(case_obj["panels"]))
+
+    _set_panel_removed(store, case_obj)
 
     for hpo_term in itertools.chain(
         case_obj.get("phenotype_groups") or [], case_obj.get("phenotype_terms") or []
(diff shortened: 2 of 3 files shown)


Jakob37:
Nice, I'll give it a go!

Jakob37:
OK, testing it out ..

I can confirm that the badges appear on the case page:

badge_apperas

I can also confirm that the sorting now works for the drop-down on the case page:

sorted_panels

Nice!

Jakob37 (approved):
Looks good! Eyed through the code as well, more for my learning. No comments there.

dnil:
> We still would like some way to optionally hide panels from the "Change default gene panels" dropdown, but I guess that was probably not indended to be a part of this PR anyway. Work for the future.

No, exactly. We can certainly add more switches, but as you see every extra condition and use case for the same component is somewhat error prone. I would also suggest a bit of work on the integration side so that perhaps a slightly more reasonable number of panels could be added to each case. Or maybe we take a discussion with some relevant persons at your end and ours to see what could be generalized to help. Would it eg help if we worked in a more fine grained panel app panel handling than the current single “PanelApp Green” directly in scout?

Jakob37:
> No, exactly. We can certainly add more switches, but as you see every extra condition and use case for the same component is somewhat error prone. 

Yes, I agree that more switches will increase complexity both of logic and UI, and that this needs to be weighed against the use benefit.

> I would also suggest a bit of work on the integration side so that perhaps a slightly more reasonable number of panels could be added to each case.

It seems to me our geneticists like to work this way - to load variants broadly, to have the opportunity to expand their search to other panels if needed, without having to re-sequence or reprompt bioinformaticians for recalculating data.

Guessing they won't be super-happy if I ask them to change how they work. But I can give it a try 😄 For instance if they would be fine with loading a narrower set of panels. Would it be possible for them (or us bioinformaticians) to add on extra if needed?

> Or maybe we take a discussion with some relevant persons at your end and ours to see what could be generalized to help. Would it eg help if we worked in a more fine grained panel app panel handling than the current single “PanelApp Green” directly in scout?

I didn't follow the last part. Can you explain a bit more what you mean with "Current single PanelApp Green" ?

It sounds like a wider discussion on panels will be needed at some point. 

Another approach which wouldn't need extra switches would be to show the "removed" tag in text also in the dropdown as such:

```
albinism (removed)
Albinism
all_genes
OldAnemi (removed)
Anemi
... and so on
```

Then it would be clear what is removed, which is the main point. They would still have a cluttered list, but that is a much smaller problem than risking selecting something removed. What do you think?

dnil:
> I didn't follow the last part. Can you explain a bit more what you mean with "Current single PanelApp Green" ?

we currently have in app support for downloading either all panel app panels as separate scout panels (which we avoid in Solna as it gives too much clutter) and also to make one merged pan-disease panel with all PanelApp green genes automagically- same as with all morbid Omim genes. But this could certainly be reworked: we had an early plan to have a ui component to select sets of relevant panel app panels from a foldable tree of them.

> It sounds like a wider discussion on panels will be needed at some point.
> 
> Another approach which wouldn't need extra switches would be to show the "removed" tag in text also in the dropdown as such:
> 
> ```
> albinism (removed)
> Albinism
> all_genes
> OldAnemi (removed)
> Anemi
> ... and so on
> ```
> 
> Then it would be clear what is removed, which is the main point. They would still have a cluttered list, but that is a much smaller problem than risking selecting something removed. What do you think?

Simple and effective! Me like - sounds like a plan!

PR #4711, opened 2024-07-19
Cases advanced search

### LINKED ISSUE #4704: Advanced case search
It would be useful for institutes with many cases to be able to combine more than one criteria when filtering. E.g. both Tags, Status and Panel. 

Either a separate search box or an "advanced" choice on the dropdown, and the ability to search with qualifiers directly, eg 
`tags: medical status: active panel: cardio`

northwestwitch:
- Closes #4704

image

### Toggle advanced search:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Deploy on stage and try some advanced search

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/adapter/mongo/case.py
```diff
@@ -102,24 +103,20 @@ def _set_similar_phenotype_query(
...
 
         if len(similar_cases) == 0:  # No cases similar to given phenotype
             query["_id"] = {"$in": []}  # No result should be returned by query
             return
 
         similar_case_ids = []
-        order = []
         for i in similar_cases:
             similar_case_ids.append(i[0])
-            order.append(i[1])
```
northwestwitch, line 116: This is weird and I don't know how we were thinking when we wrote the function like this. 

Testing locally: `similar_cases` looks like this:

image

So `order` will be:  [1.0, 1.0, 1.0]

This value is returned by the function.

scout/adapter/mongo/case.py
```diff
@@ -494,21 +487,20 @@ def _conditional_set_query_value(query, condition, set_key, set_value):
...
 
         if has_rna_data:
             cases_with_rna = self.rna_cases(collaborator or owner)
-            self._update_case_id_query(query, cases_with_rna)
+            self.update_case_id_query(query, cases_with_rna)
 
         if yield_query:
             return query
 
-        if order:
```
northwestwitch, line 510: Then order is used here, just to say if the array has values use the following projection:

'analysis_date': 1, 'assignees': 1, 'beacon': 1, 'case_id': 1, 'display_name': 1, 'individuals': 1, 'is_rerun': 1, 'is_research': 1, 'mme_submission': 1, 'owner': 1, 'panels': 1, 'status': 1, 'track': 1, 'vcf_files': 1}

I don't think the order thing adds anything by itself in here?

scout/adapter/mongo/case.py
```diff
@@ -102,24 +103,20 @@ def _set_similar_phenotype_query(
                 hpo_terms.append(term.get("phenotype_id"))
             similar_cases = self.cases_by_phenotype(hpo_terms, institute_id, case_obj["_id"])
         else:  # similar HPO terms
-            hpo_terms = list(query_term.replace(" ", "").split(","))
+            hpo_terms = list(query_value.replace(" ", "").split(","))
```
northwestwitch, line 106: Just renamed he variable to be more clear

northwestwitch:
@dnil unless you have started reviewing let's not include this PR in the next release I think

dnil:
> @dnil unless you have started reviewing, let's not include this PR in the next release I think

Sounds like a plan! Will be really good, but may take a little more iterating. Query changes and all that.. 😅

scout/adapter/mongo/case.py
dnil: 💯

dnil: Query term was more logical when the term was in there, and not in a separate field I guess.. 😁

scout/adapter/mongo/case.py
```diff
@@ -201,94 +193,107 @@ def _set_diagnosis_query(self, query: Dict[str, Any], query_term: str):
...
-            self._set_case_name_query(query, query_term)
-
-        if query_field == "exact_pheno":
-            self._set_case_phenotype_query(query, query_term)
-
-        if query_field == "exact_dia":
-            self._set_diagnosis_query(query, query_term)
+    def update_case_query_or_options(self, query: dict, or_options: List[dict]):
+        """Populates the available options in the query $and field"""
+        if or_options and query.get("$or"):
```
dnil, line 216: i
```suggestion
        if not or_options:
            return
       if query.get("$or"):
```

scout/adapter/mongo/case.py
```diff
@@ -201,94 +193,107 @@ def _set_diagnosis_query(self, query: Dict[str, Any], query_term: str):
...
+            query["$or"] += or_options
+        elif or_options:
```
dnil, line 218:
```suggestion
```

scout/adapter/mongo/case.py
```diff
@@ -201,94 +193,107 @@ def _set_diagnosis_query(self, query: Dict[str, Any], query_term: str):
...
+                "user",
+            ]:
+                query_value = name_query.get(query_field)
+                if query_value not in ["", None]:
+                    set_case_item_query(
+                        query=query, query_field=query_field, query_value=query_value.strip()
+                    )
 
-    def _update_case_id_query(self, query, id_list):
+    def update_case_id_query(self, query, id_list):
```
dnil, line 296: Huh, is this really supposed to be imported outside the module now? Or am I being archaic with PEP-8?
Screenshot 2024-08-28 at 08 12 57

scout/adapter/mongo/case.py
```diff
@@ -201,94 +193,107 @@ def _set_diagnosis_query(self, query: Dict[str, Any], query_term: str):
...
-                    {"phenotype_groups": {"$exists": False}},
-                ]
-        if query_field == "cohort":
-            query["cohorts"] = query_term
-
-        if query_term != "" and query_field in ["similar_case", "similar_pheno"]:
-            order = self._set_similar_phenotype_query(
-                query, query_field, query_term, owner or collaborator
-            )
+    def populate_case_query(
```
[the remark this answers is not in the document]
dnil: Also it's not quite working? I tried filling in a couple of search fields: anything genes related gets a 
Screenshot 2024-08-28 at 08 27 39

scout/adapter/mongo/case.py
```diff
@@ -521,7 +522,7 @@ def rna_cases(self, owner):
         Returns:
             list of case _ids
         """
-        EXISTS_NOT_NULL = {"$exists": True, "$nin": [None, ""]}
+        EXISTS_NOT_NULL = {EXISTS: True, "$nin": [None, ""]}
```
dnil, line 525: I dunno, this looks like pleasing some silly code evaluation? It's not like mongo are going to change that function anytime soon?! And if so we could search-replace it just as easy. The rule is for different situations, like natural language to be translated, or more complex ones that could go wrong. "EXISTS_NOT_NULL" is good, and e.g. `EXISTS = {"$exists": True}` or whatnot would be ok in my book!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -10,13 +10,7 @@
...
 
 from scout.adapter.mongo.base import MongoAdapter
-from scout.constants import (
-    CASE_SEARCH_TERMS,
-    CASE_STATUSES,
-    DATE_DAY_FORMATTER,
-    ID_PROJECTION,
-    PHENOTYPE_GROUPS,
-)
+from scout.constants import CASE_STATUSES, DATE_DAY_FORMATTER, ID_PROJECTION, PHENOTYPE_GROUPS
```
dnil, line 13: 👍

scout/server/blueprints/institutes/controllers.py
```diff
@@ -435,20 +429,14 @@ def cases(store, request, institute_id):
...
-                request.args["search_term"].strip(),
-            ]
-        )
-    data["name_query"] = name_query
+
+    name_query = request.form
     limit = int(request.args.get("search_limit")) if request.args.get("search_limit") else 100
-    data["form"] = populate_case_filter_form(request.args)
+
+    data["form"] = CaseFilterForm(request.form)
```
dnil, line 439: 👍

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [4.87]
 ### Added
+- Advanced cases search to narrow down results using more than one search parameter
```
dnil, line 25: Careful with the changelog here! 😊

scout/adapter/mongo/case.py
northwestwitch: Mmm no, perhaps I intended to import it from some other place at one point, but I didn't do it. I'll revert!

scout/adapter/mongo/case.py
northwestwitch: I fixed it like this:

```
        if not or_options:
            return
        query.setdefault("$or", []).extend(or_options)
```

scout/adapter/mongo/case.py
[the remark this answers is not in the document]
northwestwitch: ChatGPT! 😆 I modified and removed the lambdas, but it wasn't so hard to follow I think. I mean, historically I'm the one no getting the lambda thing, so I reasoned that if I understood then everybody could. But sure, without lambdas is more readable!

scout/adapter/mongo/case.py
northwestwitch: Yes, of course it's pleasing that code checker, I would never think of such replacement myself. I like your suggestion but if you check in the code you'll find that the "exists" is used in slightly different situations, for instance:

- set_value={EXISTS: False}
- set_value={EXISTS: True, "$ne": []} # there are a bunch of these
- set_value={EXISTS: True, "$eq": True},
- EXISTS_NOT_NULL = {EXISTS: True, "$nin": [None, ""]}

That's why I decided to replace only "$exists" and not the whole dict. This way it is more easy to read and doesn't trigger the code checker. I'd like to leave it like this if it's not a problem for you

scout/adapter/mongo/case.py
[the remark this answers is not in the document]
dnil: I do like lambdas for functional stuff, like in sorting or for more map-reduce like operation, but mixed in for some functions in a large list, I think this is cleaner.

northwestwitch:
I'm ready for another review with this one. And I guess we can include it in the next minor, so I
'm not going to merge in the upcoming patch!

tests/adapter/mongo/test_case_handling.py
```diff
@@ -209,12 +248,14 @@ def test_get_research_case(real_adapter, case_obj, institute_obj):
     ## THEN assert that the case was inserted
     assert adapter.case_collection.find_one()
 
-    # THEN searching for reasearch cases should return one case
+    # THEN searching for research cases should return one case
```
dnil, line 251: 👌

dnil (commented):
Definitely getting there! Works for me, but I have a couple of comments regarding the handling and search feeling during testing, and also some small comment to the code. It's still a little bit back and forth with local functions and "suggested private" class methods as well, but good enough methinks, and a less strenuous read than the initial AI code.
- It can be a bit annoying that the advanced query accordion folds back in after search if you wanted to test a few different things.
- This is compounded by values staying populated - as they should. It is too easy to get confused why cases don't show up and realise you at one point typed something in an "advanced" field previously and forgot about it. Perhaps force the advanced tab open if anything is filled in, or indicate somewhere next to the case name search field that advanced options are in play?
- A "clear search" option would be useful with this many fields, but could be added later.

scout/server/blueprints/institutes/templates/overview/cases.html
```diff
@@ -22,49 +22,116 @@
...
 {% macro search_form() %}
-<form action="{{ url_for('overview.cases', institute_id=institute._id) }}" method="GET" accept-charset="utf-8">
-    <div class="row d-flex align-items-end">
-      <div class="col-3">
-        {{ form.search_type.label(class="control-label") }}
-        {{ form.search_type(class="form-control") }}
-      </div>
-      <div class="col-3">
-          <label class="control-label" for="search_term">Search cases</label>
-          <input class="form-control" id="search_term" name="search_term" value="{{form.search_term.data if form.search_term.data}}" placeholder="example:18201" type="text" pattern="[^\\\<\>\?\!\=\/]*" title="Characters \<>?!=/ are not allowed">
```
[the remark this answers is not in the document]
northwestwitch: Ah right, good catch!, I'll re-add it!

scout/adapter/mongo/case.py
northwestwitch: I feel that if I replace the cases above with 5 different constants then it will be much less readable. If you don't like the EXISTS = "$exists" replacement I can simply revert to how the code was before and we ignore the code checker warning instead

scout/server/blueprints/institutes/templates/overview/cases.html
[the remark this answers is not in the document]
northwestwitch: Fixed!

northwestwitch:
> Definitely getting there! Works for me, but I have a couple of comments regarding the handling and search feeling during testing, and also some small comment to the code. It's still a little bit back and forth with local functions and "suggested private" class methods as well, but good enough methinks, and a less strenuous read than the initial AI code.
> 
> * It can be a bit annoying that the advanced query accordion folds back in after search if you wanted to test a few different things.
> * This is compounded by values staying populated - as they should. It is too easy to get confused why cases don't show up and realise you at one point typed something in an "advanced" field previously and forgot about it. Perhaps force the advanced tab open if anything is filled in, or indicate somewhere next to the case name search field that advanced options are in play?
> * A "clear search" option would be useful with this many fields, but could be added later.

Good suggestions, I fixed it as you said:

image

- filters stays open when you perform an advanced search
- At the bottom it shows the keys/values used in the search
- Added a clear search button

scout/adapter/mongo/case.py
dnil: Hm, I'm probably slow today, but at least I'm looking.. 😆 It returns them in the order given, without sorting, and the intention appears to be that they should appear in the same order as the "case_id: $in" operation gave them previously. Are you saying this doesn't work anyway, or that the returned similarity values are the same so it doesn't work? 😊

scout/adapter/mongo/case.py
dnil: Whatever! It's just a constant string regardless, and the only reason to change should be to make it more readable and maintainable. Since there is a "EXISTS_NOT_NULL" already, adding the others as similar full constants would make sense to me, but its no big deal.

scout/server/blueprints/institutes/templates/overview/cases.html
```diff
@@ -27,7 +27,7 @@
     <div class="row">
         <div class="col-md-3 mb-3">
             {{ form.case.label(class="form-label") }}
-            {{ form.case(class="form-control", placeholder="example:18201") }}
+            {{ form.case(class="form-control", placeholder="example:18201", pattern="[^\\\<\>\?\!\=\/]*", title="Characters \<>?!=/ are not allowed") }}
```
dnil, line 40: Hm, I wonder if we needed it for the other query fields as well?

northwestwitch: Fix, thanks. Better to be safe than sorry!

northwestwitch:
Searching by similar case is actually broken, like you get an exception. I'll push a fix (not the fix to #4839, but a simple one)

dnil:
Ahhh, I was about to say it works on stage, but just tested on prod. It no longer (if ever) works since the empty match can become `None` instead of `[]`. Good catch.

northwestwitch:
> Searching by similar case is actually broken, like you get an exception. I'll push a fix (not the fix to #4839, but a simple one)

It happens when you look for cases similar to a case without phenotype. An existing bug I presume

northwestwitch:
This one should be ready also @dnil

dnil:
Hm, it still removes the old, arguably anyway partly broken order. Let's discuss that one - I would prefer to fix or remove it as part of #4882, but either fully or not here. I'm open to arguments though!

northwestwitch:
> Hm, it still removes the old, arguably anyway partly broken order. Let's discuss that one - I would prefer to fix or remove it as part of #4839 , but either fully or not here. I'm open to arguments though!

I don't want to fix that order thing on this PR. It's already quite long. I'll attempt a fix in a new PR

scout/adapter/mongo/case.py
```diff
@@ -108,13 +102,21 @@ def _set_similar_phenotype_query(
...
         else:  # similar HPO terms
-            hpo_terms = list(query_term.replace(" ", "").split(","))
+            hpo_terms = list(query_value.replace(" ", "").split(","))
             similar_cases = self.cases_by_phenotype(hpo_terms, institute_id, None)
 
-        query["_id"] = {"$in": [similar_case[0] for similar_case in similar_cases]}
+        similar_case_ids = [
+            similar_case[0] for similar_case in similar_cases
+        ]  # These are sorted by the most similar to the least
+        self._update_case_id_query(
```
northwestwitch, line 111: This advanced case search is quite delicate, because if you have other params that will affect the content of query["_id"]["$in"] you can mess up the order of the similar cases

scout/adapter/mongo/case.py
```diff
@@ -197,86 +194,149 @@ def _set_diagnosis_query(self, query: Dict[str, Any], query_term: str):
...
+                "panel",
+                "status",
+                "tags",
+                "track",
+                "pheno_group",
+                "cohort",
+                "pinned",
+                "causative",
+                "user",
+                "similar_case",  # In order to be able to sort results by phenotype similarity, keep this at the bottom
```
northwestwitch, line 331: I have moved these 2 at the bottom so whatever happens before setting the query _id values, the order by similarity will be maintained (hopefully)

northwestwitch:
I've included also the "search by case similarity" fix now. Tested with the combo search on cust003

{'cohort': 'hello', 'similar_case': 'holydassie', 'search_limit': 100}:

And it works, returning this case -> cout-stage.scilifelab.se/cust003/16062

dnil (approved):
Super, thank you, and especially for lifting the order fix part to the other PR! As before, this will be very useful, EXIST or not. 😁 Let's also make an issue for moving the dashboard search to this format, but no need to do in this PR. 😊

northwestwitch:
> Let's also make an issue for moving the dashboard search to this format, but no need to do in this PR

Exactly, and if we do that the code will look clearer!

PR #4722, opened 2024-07-26
Hovertip to gene panel names with associated genes on SV variant page

### LINKED ISSUE #2068: Highlight panel matching genes in variants gene column
>For some samples there are many genes in the variant list. It is very hard to distinguish if a variant is true or not if there >are so many genes within one column. I have discussed this with Hassan, and apparently there are some transcript errors >causing this. Is there any way to come around this problem? (ex: "bossponge variant nr 1126).

Fairly sure this happens already before Scout loading, and likely in VEP so hard for us to control. But, we could perhaps try to highlight the genes that were on the panel and caused the variant to show?

Current situation:
Screenshot 2020-09-03 at 16 41 25

How about either **boldfacing** or just showing matching genes plus an ellipsis, much like for the SV view?

### LINKED ISSUE #4280: Structural variant in genepanel?
When going through SV in Scout, I often end up wondering which of the genes in the SV that is in my panel. Is there an easy way of knowing this, or adding this information? 

Especially for larger INV or BND, it would also be great to know if the breakpoints are in a gene in my panel, or at all a gene of interest. Would it be possible to add the information of which gene (if any) is in the breakpoints? 

The GENS and the chromograph tools have really helped in evaluating DEL/DUP, but doesn´t help to dismiss the balanced SV.

northwestwitch:
- Closes #4280 ---> Display genes associated with gene panels when you hover on a gene panel name on SV variant page. 
- Closes #2068

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Deploy on stage
2. Go to a case runned with gene panels that has SVs
3. Go to a SV with multiple genes anche check on the `Gene panels:` section
4. Test with huge SVs to make sure that variant doesn't load too slowly

**Expected outcome**:

**Review:**
- [x] code approved by DN


dnil (commented):
How about we wait with this one as well? I think we will want to add a lookup for the genes only presented with ID, and consider that certain variants can have a huge number of genes on them.

northwestwitch:
> How about we wait with this one as well? I think we will want to add a lookup for the genes only presented with ID, and consider that certain variants can have a huge number of genes on them.

Ok, sounds like a safe plan!

dnil (commented):
I don't think the hgnc id list mixed with symbols flies, or how do you see it? Also the list  does not seem limited atm so could get very long? Perhaps only look up and show the ones matching panels? A slower options would be to look up all, but we would still likely have to limit for very large lists.

northwestwitch:
> Perhaps only look up and show the ones matching panels?

They are already matching panels, that's the point of the PR. But perhaps instead of an hovertip I can make it a small window with a link to the gene pages, so people click and the gene opens and it will be perhaps a bit easies? Otherwise it can be the list with 30 genes with ... ?

dnil:
> > Perhaps only look up and show the ones matching panels?
> 
> They are already matching panels, that's the point of the PR. But perhaps instead of an hovertip I can make it a small window with a link to the gene pages, so people click and the gene opens and it will be perhaps a bit easier? Otherwise it can be the list with 30 genes with ... ?

Oh, right, I see. But, but, there can be no need to see named genes for a whole chromosome event like this? 30 panel matching is silly much already. The issues must be for when one out of like tens in total match. And in that case showing the hgnc id is just slower for the user. Gene page links could help, but if one potentially has to click more than a couple it becomes a bother, doesn't it? I think we would probably either have to change the variant build to include primarily matching genes among the 30, or concede to do a lookup for the ids to get the symbols? In principle the symbols should also be available in a full length `hgnc_symobls` on the variant, but it is wise to do the matching on ids only.

northwestwitch:
> > > Perhaps only look up and show the ones matching panels?
> > 
> > 
> > They are already matching panels, that's the point of the PR. But perhaps instead of an hovertip I can make it a small window with a link to the gene pages, so people click and the gene opens and it will be perhaps a bit easier? Otherwise it can be the list with 30 genes with ... ?
> 
> Oh, right, I see. But, but, there can be no need to see named genes for a whole chromosome event like this? 30 panel matching is silly much already. The issues must be for when one out of like tens in total match. And in that case showing the hgnc id is just slower for the user. Gene page links could help, but if one potentially has to click more than a couple it becomes a bother, doesn't it? I think we would probably either have to change the variant build to include primarily matching genes among the 30, or concede to do a lookup for the ids to get the symbols? In principle the symbols should also be available in a full length `hgnc_symobls` on the variant, but it is wise to do the matching on ids only.

I think this is fixed now. It's deployed on stage ATM

image

northwestwitch:
Latest --> 

image

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -443,3 +443,37 @@ <h6>Phenotype terms (HPO):</h6>
...
+    {% for gene in variant.genes %}
+      {% if panel['panel_name'] in gene['associated_gene_panels'] %}
+        {% if gene.common %}
+          {{ gene.common.hgnc_symbol }}
+        {% else %}
+          {% set no_symbol_genes_counter.value = no_symbol_genes_counter.value + 1 %}
+        {% endif %}
+      {% endif %}
+    {% endfor %}
+    {% if no_symbol_genes_counter.value > 0 %}&nbsp; + {{no_symbol_genes_counter.value}} other genes {% endif %}
```
dnil, line 474:
```suggestion
    {% if no_symbol_genes_counter.value > 0 %}&nbsp; + {{no_symbol_genes_counter.value}} other genes on this panel {% endif %}
```

scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -443,3 +443,37 @@ <h6>Phenotype terms (HPO):</h6>
 	</tbody>
 </table>
 {% endmacro %}
+
+
+{% macro gene_panels(variant) %}
+<strong>Gene panels:</strong><br>
+{% for panel in variant.case_panels|sort(attribute='panel_name', case_sensitive=False)|rejectattr('removed') %}
+  {% set no_symbol_genes_counter = namespace(value=0) %}
```
dnil, line 451: 🥇

dnil (approved):
Super! This will be very informative for intermediate size events.

scout/server/blueprints/variant/templates/variant/utils.html
northwestwitch: I still have to fix for when that list has no genes with symbol so you'll have something like this, which is ugly:

 "blank" + 456 other genes

dnil:
If you have only like three you want to show, maybe you could also dare showing the symbols that we have on the panel in addition to the hgnc_id. They could certainly be outdated, but for the most part they would be right, and one might understand them as only slightly outdated aliases otherwise.

northwestwitch:
mmm collecting symbols from panels doesn't work yet and the code in that macro has become overly complicated. I'll reason a bit on this and fix it next week!

northwestwitch:
OK, I think I think this latest version is simpler and seems to work fine. How it works:

- Show max 30 gene symbols (collect symbols from genes in gene panel if the variant has no associated symbol for the genes)
- Show + nr other genes if genes are more than 30:

image

northwestwitch:
This one is also ready to be merged. Do you want to take another look to it @dnil?

dnil:
I had a Quick Look and think it looks very sensible now. I haven't dug into the details this time around though so shout if you want more input, otw I'm good.

northwestwitch:
> I had a Quick Look and think it looks very sensible now. I haven't dug into the details this time around though so shout if you want more input, otw I'm good.

Ok, let's merge. We have to test again before the new release anyway1

PR #4747, opened 2024-08-13
Parse and show CoLoRSdb frequencies

### LINKED ISSUE #4742: Add CoLoRSdb and gnomAD keys
Would it be a good idea to add allele frequencies from [Consortium of Long Read Sequencing Database (CoLoRSdb)](https://zenodo.org/records/13145123), preferably as `colorsdb_af`? 

Could I also get `gnomad_af` and `gnomad_popmax_af` added as gnomAD keys? 😃

northwestwitch:
- Closes #4742

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by FL
- [x] tests executed by CR, automatic tests

### DIFF
--- a/scout/parse/variant/frequency.py
+++ b/scout/parse/variant/frequency.py
@@ -75,6 +75,7 @@ def parse_frequencies(variant, transcripts):
     # These are SV-specific frequencies
     update_frequency_from_vcf(frequencies, variant, ["left_1000GAF"], "thousand_g_left")
     update_frequency_from_vcf(frequencies, variant, ["right_1000GAF"], "thousand_g_right")
+    update_frequency_from_vcf(frequencies, variant, ["colorsdb_af"], "colorsdb_af")
 
     # Search transcripts CSQ if not found in VCF INFO
     if not frequencies:
@@ -123,6 +124,7 @@ def parse_sv_frequencies(variant: cyvcf2.Variant) -> Dict:
     update_sv_frequency_from_vcf(sv_frequencies, variant, SWEGEN_KEYS, "swegen")
     update_sv_frequency_from_vcf(sv_frequencies, variant, DECIPHER_KEYS, "decipher")
     update_sv_frequency_from_vcf(sv_frequencies, variant, CG_KEYS, "clingen_mip")
+    update_sv_frequency_from_vcf(sv_frequencies, variant, ["colorsdb_af"], "colorsdb_af")
 
     return sv_frequencies
 
@@ -136,6 +138,7 @@ def parse_mei_frequencies(variant: cyvcf2.Variant) -> Dict:
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_HERV_KEYS, "swegen_herv")
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_L1_KEYS, "swegen_l1")
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_SVA_KEYS, "swegen_sva")
+    update_sv_frequency_from_vcf(mei_frequencies, variant, ["colorsdb_af"], "colorsdb_af")
 
     if any(mei_frequencies.values()):
         max_mei_frequency = max(mei_frequencies.values())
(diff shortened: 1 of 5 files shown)


northwestwitch:
Tested locally with the provided demo case:

image

scout/parse/variant/frequency.py
```diff
@@ -136,6 +138,7 @@ def parse_mei_frequencies(variant: cyvcf2.Variant) -> Dict:
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_HERV_KEYS, "swegen_herv")
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_L1_KEYS, "swegen_l1")
     update_sv_frequency_from_vcf(mei_frequencies, variant, SWEGEN_SVA_KEYS, "swegen_sva")
+    update_sv_frequency_from_vcf(mei_frequencies, variant, ["colorsdb_af"], "colorsdb_af")
```
northwestwitch, line 141: Perhaps MEI variants are not annotated on colorsdb and this is an overkill? Is it only SNV and SVs?

fellen31: It's only SNVs and SVs, at least for now.

scout/server/blueprints/variant/utils.py
```diff
@@ -333,6 +333,14 @@ def predictions(genes):
     return data
 
 
+def set_colorsdb_freq(variant_obj: dict, freqs: dict) -> None:
```
fellen31, line 336: Don't know the code well enough, is this it's own function because colorsdb will only exist in long-read samples?

fellen31 (commented):
Could/should it be added here as well? 

https://github.com/Clinical-Genomics/scout/blob/0cb3460abb81ef2de7b4534feab7d7829b4aa30e/scout/server/blueprints/variant/utils.py#L480-L486

scout/server/blueprints/variant/utils.py
northwestwitch: No, it's more like because there is a set of conditions [before it is called](https://github.com/Clinical-Genomics/scout/blob/cc3bf24ab84d2f738c5205836c1783d74b46633f/scout/server/blueprints/variant/utils.py#L350), like:

if snvs:
elif svs:
elif mei:

And in these conditions are created the dictionaries. Sometimes different variant categories gave the same keys so there are repetitions, sometimes the keys are specific to the variant type.

That needs refactoring I think. Instead of having to set these dictionaries like that I'd like to introduce a call to a specific function like I'm doing here. It would avoid a lot of repetitions. Like in this case I would have to add the same key value both in the "if snvs" and "if svs".

northwestwitch:
> ould/should it be added here as well?

Good idea, I'll add it!

scout/parse/variant/frequency.py
northwestwitch: Ok, I'll remove it from the MEIs

scout/server/blueprints/variant/utils.py
fellen31: Okay, thanks for clarifying :)

northwestwitch: I'll just refactor in this PR, so colorsdb frequencies don't have to be weird compared with the other frequencies

scout/server/blueprints/variant/utils.py
```diff
@@ -474,6 +483,7 @@ def frequency(variant_obj):
         variant_obj.get("gnomad_frequency") or 0,
         variant_obj.get("gnomad_mt_homoplasmic_frequency") or 0,
         variant_obj.get("swegen_mei_max") or 0,
+        variant_obj.get("colorsdb_af") or 0,
```
fellen31, line 486: Nice, I think there will be many long-read variants not present in gnomad, but only in colors_db.

fellen31 (approved):
Looks good! Thanks!

PR #4783, opened 2024-08-26
Fix #4782 - IGV does not remove chr str from non-autosomes

### LINKED ISSUE #4782: Issues to access IGV from variant page
When trying to access the IGV page from the variant page, the page doesn't load. The same experience from several users. 

Accessing the IGV page from the case page works fine.

dnil:

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. find a variant on X, Y, MT for a case with alignment files - e.g. https://scout-stage.scilifelab.se/cust000/15026-mip7/a173fd6d7d5cba5cdab9b3b8b2281585 and https://scout-stage.scilifelab.se/cust000/F0005756-mip7/a16b311a8140e6ce56772b764286afd9/igv.
2. visit the IGV.js alignment view from the variant page
3. notice no result and JS errors for main and 4.86.1 (any IGV.js >v3, most likely)
4. apply patch and notice it working

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/alignviewers/controllers.py
+++ b/scout/server/blueprints/alignviewers/controllers.py
@@ -284,7 +284,7 @@ def set_sample_tracks(display_obj, case_groups, chromosome):
     """
     sample_tracks = []
 
-    track_items = "mt_bams" if chromosome == "M" else "bam_files"
+    track_items = "mt_bams" if chromosome == "MT" else "bam_files"
     track_index_items = "mt_bais" if track_items == "mt_bams" else "bai_files"
 
     # Loop over a group of cases and add tracks for every individual of every case
--- a/scout/server/blueprints/alignviewers/controllers.py
+++ b/scout/server/blueprints/alignviewers/controllers.py
@@ -81,15 +81,18 @@ def make_igv_tracks(
         stop = stop or variant_obj["end"]
         chrom = chrom or variant_obj.get("chromosome")
 
-    if all([start, stop, chrom]):
-        chromosome = chrom
-        display_obj["locus"] = "{0}:{1}-{2}".format(chromosome, start, stop)
-
     # Set genome build for displaying alignments:
     if "38" in str(case_obj.get("genome_build", "37")) or chromosome == "MT":
         build = "38"
     else:
         build = "37"
+
+    if all([start, stop, chrom]):
+        chromosome = chrom
+        if build == "37" or chromosome == "MT":
+            display_obj["locus"] = "{0}:{1}-{2}".format(chromosome, start, stop)
+        else:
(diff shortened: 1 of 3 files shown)


dnil:
The behaviour from igv.js v3 is slightly inconsistent in that it is backwards compatible with autosomes with "chrN" style naming, but not non-autosomes. On the other hand, the locus construction was a bit strained as it was, and is in a way cleaner now.

dnil:
I think this is now good for hg19, and presumably liftover, but not entirely sure about native hg38. Let's test that before merging... 😅

northwestwitch:
Is this a bug of the newest version of igv?

dnil:
> Is this a bug of the newest version of igv? -- Ah I'm reading you comments above now

Yes, I think so, but it's a bug in a convenience function, stripping `chr` from contig names when they don't match. It has been convenient since we have the `chr` for hg38 but not hg19.

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -81,16 +81,19 @@ def make_igv_tracks(
...
     # Set genome build for displaying alignments:
-    if "38" in str(case_obj.get("genome_build", "37")) or chromosome == "M":
+    if "38" in str(case_obj.get("genome_build", "37")) or chromosome == "MT":
         build = "38"
     else:
         build = "37"
 
+    if all([start, stop, chrom]):
+        chromosome = chrom
+        if build == "37" or chromosome == "MT":
```
northwestwitch, line 92: I think we need to add an explanation to this code because if I am confused now imagine in one year time..

The problem is only X and Y chromosomes? What about chrom M from cases in build 38?

dnil:
Ah, actually it may be partial due to a feature: they have alias files (now?). I'll add those!

northwestwitch (commented):
I'm confused, sorry!

scout/server/blueprints/alignviewers/controllers.py
dnil: Sure, the hg38-for-hg19-MT move is way in the past for me too. The alias file option looks like The Preferred Way To Do It though. This works so perhaps we better save it; I'll make a separate PR to test the alias thing and see if it is really bugged, or it just didn't find the file for the 1kg_v37.

dnil: And no, the problem is with M/MT as well, but we already had the exception to use hg38 for M...

dnil:
Ill close this as the alias files seem to be working, and not risk changing anything.

PR #4815, opened 2024-09-03
Fix #4814 - actually use variant sample in displaying omics variants

### LINKED ISSUE #4814: Individual with omics variants arbitrarily set as first individual of the case
Check the code here --> https://github.com/Clinical-Genomics/scout/blob/34944ee153c33e14a57931f044f6f4bdf6ec74e8/scout/server/blueprints/omics_variants/templates/omics_variants/outliers.html#L112

What is the variant is not in the first individual of the list?

dnil:

Screenshot 2024-09-03 at 14 13 55

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN


northwestwitch:
assign me when you are ready with this one @dnil

scout/server/blueprints/omics_variants/templates/omics_variants/outliers.html
```diff
@@ -109,7 +109,12 @@
                 {% endif %}
               </td>
               <td class="text-end">{{ '%.3e' % variant.p_value }}</td>
-              <td>{{ case.individuals.0.display_name }}</td>
+              <td>{% for sample in variant.samples %}
+                  {% if sample.genotype_call != "./." %}
```
northwestwitch, line 113: Nice with this check, could we add it also in the variants export code? https://github.com/Clinical-Genomics/scout/blob/34944ee153c33e14a57931f044f6f4bdf6ec74e8/scout/server/blueprints/omics_variants/controllers.py#L105

dnil:
Thank you, will do - it looks a tad more complicated than I thought: we kind of wanted to be sure to match the right individual, but the individuals may go by another sample name if we are not so lucky that DNA and RNA happened to be sampled under the same clinical referral... I might just change this do display what we get instead of matching, but let me get some ☕ and see if we might not have the subject_id to arbitrate..

northwestwitch:
> I might just change this do display what we get instead of matching

This looks an easy solution as well..

dnil:
Ok, looks good on stage again:
Screenshot 2024-09-03 at 17 41 57

scout/server/blueprints/omics_variants/templates/omics_variants/outliers.html
dnil: Ah, right forgot - will check tomorrow!

scout/adapter/mongo/omics_variant.py
```diff
@@ -35,18 +35,42 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+                    sample_id = ind["individual_id"]
+                    display_name = ind["display_name"]
+                    break
+
+        sample = {
+            "sample_id": sample_id,
+            "display_name": display_name,
+            "genotype_call": "./1",
+        }
+        samples.append(sample)
```
northwestwitch, line 73: This is a list, but it will contain always one sample if I read the code above correctly, why?

northwestwitch (commented):
If I'm not misunderstanding the code you'll get only one individual in a case with the omics data associated. Is it so? Check also my comments/question in the code

dnil:
> If I'm not misunderstanding the code, you'll get only one individual in a case with the omics data associated. Is it so? Check also my comments/question in the code

Each variant is from one specific individual, yes! Each line in the omics tsv has one specific sample on it.

scout/adapter/mongo/omics_variant.py
dnil: Same reason, this function is called once for each variant, and each line is for one sample only. Honestly I dont think they have thought through trios fully yet, but the current idea is that there would be multiple lines if multiple individuals share the same variant.

scout/models/case/case_loading_models.py
```diff
@@ -221,6 +221,7 @@ class SampleLoader(BaseModel):
     rhocall_wig: Optional[str] = None
     rna_alignment_path: Optional[str] = None
     rna_coverage_bigwig: Optional[str] = None
+    rna_sample_id: Optional[str] = None
```
northwestwitch, line 224: I would not add this key right now in this patch PR. I'd wait and think about it a bit instead. Could as well go in the next minor no?

CHANGELOG.md
```diff
@@ -5,11 +5,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
+### Added
```
northwestwitch, line 8: I wouldn't add this new key in. the patch PR

scout/adapter/mongo/omics_variant.py
```diff
@@ -35,18 +35,42 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
-                    "sample_id": ind["individual_id"],
-                    "display_name": ind["display_name"],
-                    "genotype_call": "./1",
-                }
-                samples.append(sample)
+            if omics_model["sample_id"] == ind.get("rna_sample_id"):
+                sample_id = ind["rna_sample_id"]
+                display_name = ind["display_name"]
+
+        if not sample_id:
```
northwestwitch, line 55: What if we omit the comparison above and in this patch we just take into account the ind matching on sample_id == individual_id?

scout/adapter/mongo/omics_variant.py
```diff
@@ -35,18 +35,42 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+            for ind in case_obj.get("individuals"):
+                if omics_model["sample_id"] == ind.get("individual_id"):
+                    sample_id = ind["individual_id"]
+                    display_name = ind["display_name"]
+
+        if not sample_id:
+            for ind in case_obj.get("individuals"):
+                if ind.get("phenotype") in [2, "affected"]:
```
northwestwitch, line 63: This is quite arbitrary. What is there is more than one affected individual in the case?

scout/demo/643594.config.yaml
```diff
@@ -62,6 +63,7 @@ samples:
     mitodel_file: scout/demo/mitodel_test_files/slowlycivilbuck_lanes_1_sorted_md_mitodel.txt
     mt_bam: scout/demo/reduced_mt.bam
     vcf2cytosure: scout/demo/ADM1059A1.test.cgh
+    rna_sample_id: RNA1059A1
```
northwestwitch, line 66: I don't like this. It adds another level of complication to an already complex config file..

northwestwitch (commented):
I'm not very fond of the added `rna_sample_id` key to the config file in the individuals part of the yaml, because it feels completely disconnected from the omics section. 

I'd rather introduce a change in the [omics section](https://github.com/Clinical-Genomics/scout/blob/34944ee153c33e14a57931f044f6f4bdf6ec74e8/scout/demo/643594.config.yaml#L128) instead, so it becomes like this:

```
omics_files:
    fraser: scout/demo/drop/fraser_top_hits_clinical.tsv
    outrider: scout/demo/drop/outrider_top_hits_clinical.tsv
    sample_map: 
        ADM1059A1: RNA1059A1
        ADM1059A3: RNA1059A3
        ...
```

But I would definitely keep this thing out of the patch and fix it so that the sample id is matched against individual["sample_id"] or eventually phenotype for now

scout/adapter/mongo/omics_variant.py
```diff
@@ -35,18 +35,42 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+
         for ind in case_obj.get("individuals"):
-            if omics_model["sample_id"] == ind.get("individual_id"):
-                sample = {
-                    "sample_id": ind["individual_id"],
-                    "display_name": ind["display_name"],
-                    "genotype_call": "./1",
-                }
-                samples.append(sample)
+            if omics_model["sample_id"] == ind.get("rna_sample_id"):
```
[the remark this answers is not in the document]
dnil: Sure, I'll fix. I will just use the obnoxious "clean code" concept of making functions with docstrings, faking real object orientation here.

CHANGELOG.md
dnil: No, we can/should wait with merging this, but there is no real patch without it. I suppose I could just do all the fallbacks, but it makes relatively little sense?

scout/adapter/mongo/omics_variant.py
dnil: It would display.

scout/demo/643594.config.yaml
dnil: Ok, good! Counter suggestions?

northwestwitch: Adding a sample mapping in the omics section of the config file?

dnil: Like
```
omics_files:
    fraser: fraser.tsv
    outrider: outrider.tsv
    sample_identification:
        rna_sample_id: RNA1059A1
        sample_id: ACC1059A1
```
Possible, but it is a bit unfortunate with the files in "omics_files" then.

dnil: Did you see my refactor suggestion for laters as well? https://github.com/Clinical-Genomics/scout/discussions/4817

northwestwitch: > Did you see my refactor suggestion for laters as well? #4817

I'm reading it now. I see your point but I still think we should keep this PR as simple as possible. I agree that this PR will not be a solution to the sample naming, and that we should have a better way to connect samples and id samples, but this can be done in a dedicated PR. It feels a bit rushed to me right now and I'd rather reason on a permanent solution instead of having something that we'll need to change later

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+        """Select individual that matches omics model sample on rna_sample_id if these are provided."""
 
-        samples = []
+        match = {}
+        for ind in case_obj.get("individuals"):
+            if omics_model["sample_id"] == ind.get("rna_sample_id"):
+                match["sample_id"] = ind["rna_sample_id"]
+                match["display_name"] = ind["display_name"]
+                return match
+        return match
```
northwestwitch, line 45: You can just retun ind here

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
-                sample = {
-                    "sample_id": ind["individual_id"],
-                    "display_name": ind["display_name"],
-                    "genotype_call": "./1",
-                }
-                samples.append(sample)
+                match["sample_id"] = ind["individual_id"]
+                match["display_name"] = ind["display_name"]
+                return match
+        return match
```
northwestwitch, line 58: Here also, you can return ind

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+        Fall back to assigning the variants to an individual with affected status to have them display
+        on variantS queries.
+        """
+        match = {}
+        for ind in case_obj.get("individuals"):
+            if ind.get("phenotype") in [2, "affected"]:
+                match["sample_id"] = ind["individual_id"]
+                match["display_name"] = ind["display_name"]
+                return match
+        return match
```
northwestwitch, line 71: Same

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+        Fall back to assigning the variants to any one individual to have them display
+        on variantS queries.
+        """
+        match = {}
+
+        for ind in case_obj.get("individuals"):
+            match["sample_id"] = ind["individual_id"]
+            match["display_name"] = ind["display_name"]
+            return match
+        return match
```
northwestwitch, line 84: Same

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+        for ind in case_obj.get("individuals"):
+            if ind.get("phenotype") in [2, "affected"]:
+                match["sample_id"] = ind["individual_id"]
+                match["display_name"] = ind["display_name"]
+                return match
+        return match
+
+    def _get_first_individual(self, case_obj: dict) -> dict:
+        """
+        Fall back to assigning the variants to any one individual to have them display
```
northwestwitch, line 75: 🤣

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+
+        match = (
+            self.get_matching_rna_sample_id(case_obj, omics_model)
+            or self.get_matching_sample_id(case_obj, omics_model)
+            or self._get_affected_individual(case_obj)
+            or self._get_first_individual(case_obj)
+        )
+
+        sample = {
+            "sample_id": match["sample_id"],
```
northwestwitch, line 105:
```suggestion
            "sample_id": match.get("rna_sample_id",match["sample_id]),
```

if you return the individual in the matching functions

northwestwitch (approved):
This fixes the issue, so OK

scout/server/blueprints/omics_variants/templates/omics_variants/outliers.html
dnil: A GT check would be a placeholder as it stands now, with no GTs delivered from the pipeline. It will likely not be forgotten if or when we get such GTs.

scout/adapter/mongo/omics_variant.py
```diff
@@ -33,20 +33,80 @@ def delete_omics_variants(self, case_id: str, file_type: str):
...
+        match = {}
         for ind in case_obj.get("individuals"):
             if omics_model["sample_id"] == ind.get("individual_id"):
-                sample = {
-                    "sample_id": ind["individual_id"],
-                    "display_name": ind["display_name"],
-                    "genotype_call": "./1",
-                }
-                samples.append(sample)
+                match["sample_id"] = ind["individual_id"]
```
northwestwitch, line 55: Here you can return ind

scout/adapter/mongo/omics_variant.py
northwestwitch, line 71:
```suggestion
               return ind
```

scout/adapter/mongo/omics_variant.py
northwestwitch, line 84: return case_obj.get("individuals")[0]

scout/adapter/mongo/omics_variant.py
dnil: I agree it's possible to pass individual, but think this in particular will be difficult to understand without reading the details in the functions above as well? Also the individual is a bit larger data structure. Not so that it wouldn't work, but...

scout/adapter/mongo/omics_variant.py
dnil: Lets make that explicit at least!

scout/adapter/mongo/omics_variant.py
dnil: You are right, it would only return the name for the data, not all the data. And its potato-potato with reading - it might just reinforce the idea that the `rna_sample_id` is taken first and the `individual_id`after.

dnil: Plus, hey, it does save lines and memory then!

docs/admin-guide/load-config.md
```diff
@@ -74,6 +74,7 @@ Below are available configuration parameters for a Scout case. Names marked with
       - **reference** _String_ Path or URL for REViewer to reference sequence for the individual STR alignment
     - **rna_alignment_path** _String_ Path to RNA alignment file (BAM/CRAM)
     - **rna_coverage_bigwig** _String_ Path to coverage islands file generated
+    - **rna_sample_id** _String_ Sample ID for RNA, as in outliers files
```
northwestwitch, line 77: Last doubt:

not `omics_sample_id` ?

northwestwitch (commented):
I have a final doubt, sorry 😆

docs/admin-guide/load-config.md
dnil: Yes, I like that!

dnil:
✅ Tested on stage:
Screenshot 2024-09-04 at 18 24 11

https://scout-stage.scilifelab.se/cust002/F0057227

PR #4826, opened 2024-09-05
Display WTS outliers again

### LINKED ISSUE #4825: GT check excludes all variants from the WTS outliers page
See [this comment](https://github.com/Clinical-Genomics/scout/pull/4824#issuecomment-2330712526) and the one after. Reproducible also on the demo version

northwestwitch:
- Fix #4825 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -83,7 +83,7 @@ class OmicsVariantHandler:
 
         sample = {
             "sample_id": match["individual_id"],
-            "display_name": match.get("omics_sample_id", match["display_name"]),
+            "display_name": match.get("omics_sample_id", omics_model["display_name"]),
             "genotype_call": "./1",
         }
         samples.append(sample)
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -83,7 +83,7 @@ class OmicsVariantHandler:
 
         sample = {
             "sample_id": match["individual_id"],
-            "display_name": match.get("omics_sample_id") or omics_model["display_name"],
+            "display_name": match.get("omics_sample_id") or omics_model["sample_id"],
             "genotype_call": "./1",
         }
         samples.append(sample)
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -83,7 +83,7 @@ class OmicsVariantHandler:
 
         sample = {
             "sample_id": match["individual_id"],
-            "display_name": match.get("omics_sample_id") or omics_model["sample_id"],
+            "display_name": match["display_name"],
             "genotype_call": "./1",
         }
         samples.append(sample)
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -83,7 +83,7 @@ class OmicsVariantHandler:
 
         sample = {
             "sample_id": match["individual_id"],
-            "display_name": match.get("omics_sample_id", omics_model["display_name"]),
+            "display_name": match.get("omics_sample_id") or omics_model["display_name"],
             "genotype_call": "./1",
         }
         samples.append(sample)
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -82,8 +82,8 @@ class OmicsVariantHandler:
         )
 
         sample = {
-            "sample_id": match.get("omics_sample_id", match["individual_id"]),
-            "display_name": match["display_name"],
+            "sample_id": match["individual_id"],
+            "display_name": match.get("omics_sample_id", match["display_name"]),
             "genotype_call": "./1",
         }
         samples.append(sample)
--- a/scout/models/omics_variant.py
+++ b/scout/models/omics_variant.py
@@ -234,9 +234,7 @@ class OmicsVariantLoader(BaseModel):
     @model_validator(mode="before")
     def set_sample_display_name(cls, values) -> "OmicsVariantLoader":
         """Set a display name."""
-        values["display_name"] = values.get(
-            "display_name", values.get("sample_name", values.get("individual_id"))
-        )
+        values["display_name"] = values.get("sample_id", values.get("sampleID"))
         return values
 
 
--- a/scout/models/omics_variant.py
+++ b/scout/models/omics_variant.py
@@ -231,12 +231,6 @@ class OmicsVariantLoader(BaseModel):
         )
         return values
 
-    @model_validator(mode="before")
-    def set_sample_display_name(cls, values) -> "OmicsVariantLoader":
-        """Set a display name."""
-        values["display_name"] = values.get("sample_id", values.get("sampleID"))
-        return values
-
 
 def get_qualification(values: dict) -> str:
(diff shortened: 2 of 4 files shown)


northwestwitch:
Looks like this is fixing the issue:

image

scout/models/omics_variant.py
```diff
@@ -231,14 +231,6 @@ def set_omics_variant_id(cls, values) -> "OmicsVariantLoader":
         )
         return values
 
-    @model_validator(mode="before")
-    def set_sample_display_name(cls, values) -> "OmicsVariantLoader":
-        """Set a display name."""
-        values["display_name"] = values.get(
-            "display_name", values.get("sample_name", values.get("individual_id"))
```
northwestwitch, line 238: display_name is the display_name of the variant, no? I don't think this check is needed

scout/adapter/mongo/omics_variant.py
```diff
@@ -82,8 +82,8 @@ def set_samples(self, case_obj: dict, omics_model: dict):
         )
 
         sample = {
-            "sample_id": match.get("omics_sample_id", match["individual_id"]),
-            "display_name": match["display_name"],
+            "sample_id": match["individual_id"],
```
northwestwitch, line 85: The variant query is using individual_id to collect all variants to display on the WTS variants page, so this has to be like this.

scout/adapter/mongo/omics_variant.py
```diff
@@ -82,8 +82,8 @@ def set_samples(self, case_obj: dict, omics_model: dict):
...
+            "display_name": match.get("omics_sample_id") or omics_model["sample_id"],
```
northwestwitch, line 86: Here one can display the name provided as sampleID in the TSV file tho

northwestwitch:
After installing this branch on stage and reloading case orientedsheepdog: 

image

scout/models/omics_variant.py
dnil: Right, but there is no display name in the tsv file, and it's a mandatory field on variants. This was a placeholder. Let's fill it with something else; perhaps echo the variant_id, or make a light version of that more prominently focusing on coords or the variant qualification.

scout/adapter/mongo/omics_variant.py
dnil: Did it end up on the wrong line after all those changes? 🙄Well spotted. But loading a case on stage worked just fine?! Good with many fallbacks I guess. I'll test through a few other scenarios.

scout/adapter/mongo/omics_variant.py
dnil: Yes, precisely.

scout/adapter/mongo/omics_variant.py
northwestwitch: Loading a case on stage did work, but then no vars were shown

dnil: Im guessing your case had a different setup, and either hit a different fallback?

scout/models/omics_variant.py
northwestwitch: There's this other check that sets the display_name of the variant --> https://github.com/Clinical-Genomics/scout/blob/0611bcec0b1e8d18f0e939c9c77643412ed861fc/scout/models/omics_variant.py#L199

scout/adapter/mongo/omics_variant.py
northwestwitch: Yes, probably!

northwestwitch: But it's replicable also with the demo instance, try

scout/models/omics_variant.py
dnil: 😂 Good, that would be what I intended. I agree, let's remove the placeholder then! 😁

scout/adapter/mongo/omics_variant.py
dnil: Quite, but there are apparently more moving parts here.

scout/adapter/mongo/omics_variant.py
dnil: Or wait, no, stop leading me on! 😂 That is the CG internal ID. It will have no meaning to the customers - we are much better off displaying the matched sample id.

dnil: display_name to display_name, sample_id to sample_id. I can update if you want. But there is still at least one more issue hiding..

northwestwitch: Sure, push. What's hiding?

dnil: It worked right before merging on hasta-stage. I've rechecked the case in db, and it does not seem to have used any fallback, which makes me think there is a compensatory "error" later in the chain.

northwestwitch:
Mm there is also thing of the omics_sample_id key added to the samples of all cases, regardless of their value. For instance this is a sample from the demo cancer case:

image

Let's fix so the key is not added when it's not present!

dnil:
> After installing this branch on stage and reloading case [orientedsheepdog](https://scout-stage.scilifelab.se/cust003/24045-Tomte) the variants are visible:

Ahh, I might be starting to understand: this example was without `omics_sample_id` and with `sample_ids` matching the actual ids in the file, like a pure Tomte upload without any MIP data to base it off?!

dnil:
> Let's fix so the key is not added when it's not present!

We can do that two ways, either the old check-and-prune-each key, or setting the whole class to prune all null values. I dunno, I would prefer the latter but do we dare?

northwestwitch:
> We can do that two ways, either the old check-and-prune-each key, or setting the whole class to prune all null values. I dunno, I would prefer the latter but do we dare?

I pushed a conservative solution. with the if:  I can't handle the implications of the prune whatever thing right now 😆

northwestwitch:
> I pushed a conservative solution. with the if: I can't handle the implications of the prune whatever thing right now 😆

There are some similar ifs just [upstream](https://github.com/Clinical-Genomics/scout/blob/d5648d65ed8a549ce68c0ef48964d4bcf400d306/scout/build/individual.py#L119)

dnil:
> > We can do that two ways, either the old check-and-prune-each key, or setting the whole class to prune all null values. I dunno, I would prefer the latter but do we dare?
> 
> I pushed a conservative solution. with the if: I can't handle the implications of the prune whatever thing right now 😆

I like that, it solves the problem, and there is prior art.

dnil:
- tested vanilla MIP + tomte WITH correctly set `omics_sample_id`: https://scout-stage.scilifelab.se/cust002/F0057227/omics_variants/outliers?variant_type=clinical
- tested a stand alone tomte trio, with `sample_id` present in the outlier files: https://scout-stage.scilifelab.se/cust003/24045-Tomte
- tested fallback (locally) removing `omics_sample_id` from the demo case for a case with mismatching ids

dnil (approved):
✅ Great - this now tests out correctly for at least the three most common use cases. There will also be reason to return to this in refactoring later anyway.

dnil (approved):

PR #4830, opened 2024-09-06
A bit more advanced users' activity logging

### LINKED ISSUE #4758: Add user sending requests to log
### Description
As part of the move into a hospital setting they request more thorough logging of who has accessed patient data. Currently all requests send by the scout front-end are logged, however the user performing the action is not included.

I would like scout to write both which endpoint was accessed and by who to the log.

northwestwitch:
- Closes #4758 - The logging to file is independent from the existing system logging

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Deploy this branch on cg-vm1
2. Deploy also this branch of servers ---> https://github.com/Clinical-Genomics/servers/pull/1448
3. Restart scout service stage
4. Go inside the scout container on cg-vm1 (`podman  exec -it <container-id> /bin/bash` ) and make sure logs have been written to this following file --> /home/worker/scout_users_activity.log

**Expected outcome**:
- Your user should be logged out and you must agree to be monitored before being able to log in again

**Review:**
- [x] code approved by DN


northwestwitch:
> Weird, all tests pass on my local 🤔

Ah no, I lied!

northwestwitch:
While this seems to work OK locally (haven't tested n stage yet), It feels like I need to add the setting howto on the documentation. I think there should also be some notification to the users asking if they agree that their browsing will be recorded. For instance if the scout config contains the specific settings. log out all users and make them log in again after agreeing to the thing?

dnil:
> While this seems to work OK locally (haven't tested n stage yet), It feels like I need to add the setting howto on the documentation. I think there should also be some notification to the users asking if they agree that their browsing will be recorded. For instance if the scout config contains the specific settings. log out all users and make them log in again after agreeing to the thing?

Yes, for most hospitals/users the event log, representing our user action audit trail, will be the important part anyway. But this is straightforward enough, and with an option to enable it should be all good. Info to the users is a Good Thing. We can add an info on the landing page, describing this, the event audit log, web server logs and the user land cookies used to keep the user logged in, link the user to their google Oauth2 identification, and temporarily store internal file permission requests like alignment files and tracks!

northwestwitch:
Ok, this is working now on stage with Google Oauth2. I've decided to keep the checkbox because that way I can set a session param that will help understand if the user consents or not. This is in turn used to understand if the user has to be logged out or not when the functionality is in place (has the agree). Fixed also the dark mode thing now. I'll open for review now

scout/server/app.py
```diff
@@ -88,6 +108,22 @@ def check_user():
                 login_url = url_for("public.index", next=next_url)
                 return redirect(login_url)
 
+        if USERS_LOGGER_PATH_PARAM in app.config:
+            if current_user.is_authenticated and session.get("consent_given") is None:
```
northwestwitch, line 112: Since this code is found in before_request, it would log out users every time they do anything. That's why I need that consent given session variable to be in place!

scout/server/blueprints/login/views.py
```diff
@@ -66,20 +72,19 @@ def login():
             flash("User not authorized by LDAP server", "warning")
             return redirect(url_for("public.index"))
 
-    if current_app.config.get("GOOGLE"):
+    elif current_app.config.get("GOOGLE"):
```
northwestwitch, line 75: They become mutually exclusive. I think it gives more control to the thing

scout/server/blueprints/public/templates/public/index.html
```diff
@@ -20,12 +20,12 @@ <h1 class="display-4">Scout</h1>
         </a>
       </p>
     {% else %}
-      {% if config.GOOGLE %}
-        <a class="btn btn-primary btn-lg text-white" href="{{ url_for('login.login') }}" role="button">
-          Login with Google
-        </a>
-      {% elif config.LDAP_HOST or config.LDAP_SERVER %}
-        <form class="row" method="POST" action="{{ url_for('login.login') }}">
+      <form class="row" method="POST" action="{{ url_for('login.login') }}">
```
northwestwitch, line 23: This was a get request before, but I need everything to be in a form in order to submit also the `consent_checkbox`  with the form. So it becomes a POST request in all cases: LDAP/Google and simple login

northwestwitch:
Ah wait, for some reason it's not printing to log. It's a draft again!

northwestwitch:
Mmm I can't seem to start the scout service with this branch on cg-vm1. Pretty sure it's a matter of writing permissions on the log file specified in the [servers PR](https://github.com/Clinical-Genomics/servers/pull/1448).. I need to reason on this

dnil:
> Mmm I can't seem to start the scout service with this branch on cg-vm1. Pretty sure it's a matter of writing permissions on the log file specified in the [servers PR](https://github.com/Clinical-Genomics/servers/pull/1448).. I need to reason on this

I remember seeing a (temp?) PR with writing to the container but no mounting of the fs onto there yet - could that be this one?

northwestwitch:
Ok now at least it starts on stage but still doesn't write to the log file 🙄

northwestwitch:
Ok the log file is created correctly, but when I enter into the container it stays empty -->

image

northwestwitch:
AHA! Looks like the problem is not permissions over files in Docker (any more). I tried to run a local instance via gunicorn and the log file gets created but not written (same behavior as on stage). So the reason must be how the logs are handled by gunicorn instead! 🤔

northwestwitch:
I'm still having issue with the permissions to write to the host file system (specifically `/tmp/tmp_scout_users_activity.log`, but I think for the sake of this PR it is enough if I show that a file is created inside the container (under `/home/worker/scout_users_activity.log`) and the file contains user's info. When I execute the container and I check that file I see this -->

image


Marking as ready for review!

scout/server/blueprints/login/views.py
dnil: Yes: I had to reread because I thought we could rearrange and have the common returns first, but I agree with this.

scout/server/blueprints/public/templates/public/index.html
```diff
@@ -46,8 +44,18 @@ <h1 class="display-4">Scout</h1>
               <button type="submit" class="btn btn-primary form-control text-white">Login</button>
             </div>
           </div>
-        </form>
-      {% endif %}
+        {% endif %}
+        <!-- Consent Banner -->
```
dnil, line 48: Shouldn't we always show a notice saying something like
"Scout uses session cookies. They are necessary for your usage experience."

scout/server/app.py
```diff
@@ -35,6 +35,40 @@
 )
 
 LOG = logging.getLogger(__name__)
+USERS_LOGGER_PATH_PARAM = "USERS_ACTIVITY_LOG_PATH"
```
dnil, line 38: 😂 These no repeated strings things make the python code read like assembler!

scout/server/app.py
```diff
@@ -35,6 +35,40 @@
...
+    "favicon",
+    "GET /",
+    "Closing",
+    "Enable",
+    "enabled",
+    "Collecting IGV tracks",
+]  # Substrings used when filtering messages to show if users activity log is on
+
+
+class ActivityLogFilter(logging.Filter):
```
dnil, line 55: 🥇

scout/server/app.py
```diff
@@ -35,6 +35,40 @@
...
+    "custom_images",
+    "autozygous_images",
+    "coverage_images",
+    "upd_regions_images",
+    "favicon",
+    "GET /",
+    "Closing",
+    "Enable",
+    "enabled",
+    "Collecting IGV tracks",
```
dnil, line 51: I dunno, some of these aren't really sub URLs. They are, shall we say, filter ignore triggers?

scout/server/app.py
```diff
@@ -35,6 +35,40 @@
...
+
+    def filter(self, record):
+        return any(sub_url in record.getMessage() for sub_url in SUB_URL_IGNORE_LIST) is False
+
+
+def set_activity_log(app):
+    """Log users' activity to a file, if specified in the scout config."""
+    if USERS_LOGGER_PATH_PARAM not in app.config:
+        return
+    app.logger.setLevel("INFO")
```
dnil, line 67: This sets the main log streamhandler to INFO, regardless of the log level asked?

scout/server/app.py
```diff
@@ -35,6 +35,40 @@
...
+
+class ActivityLogFilter(logging.Filter):
+    """When monitoring users activity, log only navigation on main pages.
+    - Do not log messages that contain the substrings specified in SUB_URL_IGNORE_LIST"""
+
+    def filter(self, record):
+        return any(sub_url in record.getMessage() for sub_url in SUB_URL_IGNORE_LIST) is False
+
+
+def set_activity_log(app):
```
dnil, line 63: Could we reuse the log/log/init_log() or if not move this there as a helper/special case?

dnil (changes requested):
Maybe this is just me, logging can be odd, but I seem to loose DEBUG output on the main console stderr stream when applying this? lets check!

scout/server/app.py
northwestwitch: yes

dnil: Ahem, maybe not. This is tricky! 😅

scout/server/app.py
northwestwitch: Is that good or bad?

dnil:
> Maybe this is just me, logging can be odd, but I seem to loose DEBUG output on the main console stderr stream when applying this? lets check!

No, sorry, it's just me being silly - it was the same old Gunicorn logger stream thing. We could actually apply https://stackoverflow.com/questions/26578733/why-is-flask-application-not-creating-any-logs-when-hosted-by-gunicorn 
permanently in code as well since we always use and recommend it. But let's do that in a separate PR, and remove some useless debug messages when we do.

We currently agree the new log file and its filter would make sense in the scout log module. let me know if I missed a reason it shouldn't be there. 😊

scout/server/blueprints/public/templates/public/index.html
northwestwitch: Yeah we can add it, but cookies are a different thing and I think that type of banner should be shown always. Let's have on a different PR, to keep this one more simple.

scout/server/app.py
northwestwitch: Ah we can rename it, no problem!

scout/server/blueprints/public/templates/public/index.html
dnil: Yes, it should always be there! I was thinking it would just be one line of text that would anyway sit together with this new one. But we could do the EU-cookie banner in a separate banner as well.

dnil (approved):
The new feature works well! The old logging at least through tornado, is intact, and the interaction with Gunicorn wasnt super before anyway. The logging module in scout is apparently unused since quite a while back. Let's revisit this in a refactor PR, fixing the Gunicorn logger extension as well.

scout/server/app.py
dnil: Ha, good question! I kind of liked assembler, especially compared to the "high" level languages like BASIC that were the early alternatives for me. But I have not felt the need to go back to that style since I learned C and C++. 😂

scout/server/app.py
```diff
@@ -36,7 +36,7 @@
 
 LOG = logging.getLogger(__name__)
 USERS_LOGGER_PATH_PARAM = "USERS_ACTIVITY_LOG_PATH"
-SUB_URL_IGNORE_LIST = [
+ACTIVITY_LOG_IGNORE_TRIGGERS = [
```
dnil, line 39: 💯

PR #4903, opened 2024-10-01
Fix gene alias missing true HGNC ID when symbol is outdated

northwestwitch:
This PR adds a functionality or fixes a bug.
- Tentative fix #4902 

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the Pax software available at [https://pax.scilifelab.se/](https://pax.scilifelab.se). The resource you are going to call dibs on is `scout-stage` and the server is `cg-vm1`.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, repeat procedure at [https://pax.scilifelab.se/](https://pax.scilifelab.se), which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

<details>
<summary>Testing on hasta server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. `ssh <USER.NAME>@hasta.scilifelab.se`
1. Book your testing time using the Pax software. `us; paxa -u <user> -s hasta -r scout-stage`. You can also use the WSGI Pax app available at [https://pax.scilifelab.se/](https://pax.scilifelab.se).
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `conda activate S_scout; pip freeze | grep scout-browser`
1. Deploy the branch to test: `bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_scout -t scout -b <this_branch>`
1. Make sure the branch is deployed: `us; scout --version`
1. After testing is done, repeat the `paxa` procedure, which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [ ] code approved by
- [ ] tests executed by

PR #4904, opened 2024-10-01
Fix #4902 - use unambiguos alias if no true hgnc id is known

### LINKED ISSUE #4902: OMIM gene missing HGNC id?
**Describe the bug**

Screenshot 2024-10-01 at 10 22 07

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

dnil:

- Fix #4902 

Nice with #4903 - thank you, and sorry I didn't see it in time. I think this is a safer solution. Compare how the same data structure is handled in other places e.g. here https://github.com/Clinical-Genomics/scout/blob/07ad1e26150cf9c0092af3154254b00a6b050bf8/scout/adapter/mongo/hgnc.py#L436. 

We could and should probably refactor it though - seems overly complex, and the generating code is a bit dusty - but if its ok with you I'd prefer not to mess with it since it when it is in use in other places?

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR


dnil:
Did we get the same gene count? 😊 
Screenshot 2024-10-01 at 15 46 35

scout/parse/panel.py
```diff
@@ -424,6 +424,19 @@ def get_omim_panel_genes(genemap2_lines, mim2gene_lines, alias_genes):
                 break
 
         if hgnc_id_info:
-            yield {"hgnc_id": hgnc_id_info["true"], "hgnc_symbol": hgnc_symbol}
+            if hgnc_id_info["true"]:
+                hgnc_id = hgnc_id_info["true"]
+            elif hgnc_id_info["ids"] and len(hgnc_id_info["ids"]) == 1:
```
northwestwitch, line 429:
```suggestion
            elif len(hgnc_id_info.get("ids", ())) == 1:
```

northwestwitch: All aliases should have hgnc_id_info["ids"] in theory tho

northwestwitch (approved):
Looks good! 👍🏻

scout/parse/panel.py
northwestwitch: I guess one could also skip the first if and just use the `if len(hgnc_id_info.get("ids", ())) == 1`

northwestwitch:
> Did we get the same gene count? 😊

No, I got one more

dnil:
> > Did we get the same gene count? 😊
> 
> No, I got one more

Right, but that was from when I updated only the panel, and you had both new genes and panel, right! Or did I miss something? After updating genes I got that dubious ATXN8 gene in, so we both ended up with 4630. Importantly the patch seems to fix the issue with POPDC1 cleanly, and be safe enough by explicitly checking for aliases.

Screenshot 2024-10-02 at 09 31 12

Screenshot 2024-10-02 at 09 44 08

PR #4931, opened 2024-10-10
Fix updating genes in a panel with custom info

### LINKED ISSUE #4925: Updating a panel gene, sets Reduced penetrance and Mosaicism fields to False
For instance I've added a comment to a gene in the panel:

image

When I export it looks like this, these 2 booleans and no comment:

image

northwestwitch:
- Fix #4925

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Locally - go on main branch and run `scout setup demo`
2. Go to panel1 page and edit the gene with symbol MT-TF (it's easier to test because it's the first gene contained in the list of genes in the panel document in the database). You can add for instance a comment:

image

3. Save the pending action and check the panel version 2.0 in the database: the update was messy --->

image

4. Switch to this branch and repeat from point 1. 
5. The updated gene will be like this instead:

image

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -507,8 +507,8 @@ class PanelHandler:
                 continue
 
             if action == "edit":
-                for key, value in current_update["info"].items():
-                    gene[key] = value
+                gene["info"] = current_update["info"]
+                LOG.warning(gene)
                 new_genes.append(gene)
 
         new_panel["genes"] = new_genes
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -470,71 +470,47 @@ class PanelHandler:
             inserted_id(str): id of updated panel or the new one
         """
 
-        updates = {}
         new_panel = deepcopy(panel_obj)
         new_panel["pending"] = []
         new_panel["date"] = dt.datetime.now()
         new_genes = []
 
-        for update in panel_obj.get("pending", []):
-            hgnc_id = update["hgnc_id"]
+        # Process 'add' actions and gather updates
+        updates = {u["hgnc_id"]: u for u in panel_obj.get("pending", []) if u["action"] != "add"}
+        new_genes += [
+            {"hgnc_id": u["hgnc_id"], "symbol": u["symbol"], **u.get("info", {})}
+            for u in panel_obj.get("pending", [])
+            if u["action"] == "add"
+        ]
 
-            # If action is add we create a new gene object
-            if update["action"] != "add":
-                updates[hgnc_id] = update
-                continue
-
-            info = update.get("info", {})
-            gene_obj = {"hgnc_id": hgnc_id, "symbol": update["symbol"]}
-
-            for field in info:
-                gene_obj[field] = info[field]
-
-            new_genes.append(gene_obj)
-
+        # Process existing genes
         for gene in panel_obj.get("genes", []):
             hgnc_id = gene["hgnc_id"]
+            update = updates.get(hgnc_id)
 
-            if hgnc_id not in updates:
+            if not update:  # No update, keep the gene
                 new_genes.append(gene)
-                continue
-
-            current_update = updates[hgnc_id]
-            action = current_update["action"]
-
-            # If action is delete we do not add the gene to new genes
-            if action == "delete":
-                continue
-
-            if action == "edit":
-                # reset all fields except hgnc and symbol:
+            elif update["action"] == "edit":  # Edit gene fields
                 for key in EXPORT_PANEL_FIELDS[2:]:
-                    gene.pop(key[1], None)
-                # And update keys with the pending changes
-                for key, value in current_update["info"].items():
-                    gene[key] = value
+                    gene.pop(key[1], None)  # Reset all fields except hgnc and symbol
+                gene.update(update["info"])
                 new_genes.append(gene)
+            # Skip 'delete' actions
 
         new_panel["genes"] = new_genes
         new_panel["version"] = float(version)
 
-        # if the same version of the panel should be updated
-        if new_panel["version"] == panel_obj["version"]:
-            # replace panel_obj with new_panel
+        # Update same version or create new version
+        LOG.warning(f"version is {version} -- type: {type(version)}")
+        if version == panel_obj["version"]:
             result = self.panel_collection.find_one_and_replace(
-                {"_id": panel_obj["_id"]},
-                new_panel,
-                return_document=pymongo.ReturnDocument.AFTER,
+                {"_id": panel_obj["_id"]}, new_panel, return_document=pymongo.ReturnDocument.AFTER
             )
             inserted_id = result["_id"]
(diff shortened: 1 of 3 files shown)


scout/server/blueprints/panels/views.py
```diff
@@ -384,12 +384,14 @@ def gene_edit(panel_id, hgnc_id):
...
     if form.validate_on_submit():
         action = "edit" if panel_gene else "add"
-        info_data = form.data.copy()
-        if "csrf_token" in info_data:
-            del info_data["csrf_token"]
-        if info_data["custom_inheritance_models"] != "":
-            info_data["custom_inheritance_models"] = info_data["custom_inheritance_models"].split(
-                ","
+        info_data: dict = {
+            k: v for k, v in form.data.copy().items() if v
```
northwestwitch, line 388: This is the fix

dnil: But, we have booleans there? What f the user sets a value of e.g. mosaicism to `False`?

northwestwitch: Yes, it's a checkbox (boolean). So on or off. One could refactor and allow it to be reset to None, but I don't want to do it in this PR. I think it's best to assume on is True (so mosaicism or reduced penetrance) and off is those things unknown. For now at least. These checkbox are not really used much I think, otherwise somebody would have complained already

dnil: Sorry, I'm writing too tersely.

The thing with that is you can in the main version unset a `Yes` once you have it set (perhaps mistakenly, perhaps since literature changed). This PR will not allow it:
Screenshot 2024-10-11 at 08 19 00
Screenshot 2024-10-11 at 08 22 17
Screenshot 2024-10-11 at 08 22 21
Screenshot 2024-10-11 at 08 21 53
Screenshot 2024-10-11 at 08 19 28
since a False or a None value in the update is treated the same way.

dnil (changes requested):
We kind of need to differentiate between freshly set `False` values and unset/`None` value options, unless we save everything...

northwestwitch:
> We kind of need to differentiate between freshly set `False` values and unset/`None` value options, unless we save everything...

If you save everything and uncheck the checkbox the in will be set to False - so it's correct that there is no way to differentiate between if a user wants to set it as false or wants to reset it. This needs refactoring of the interface. I'll try to add a simple way

dnil:
> > We kind of need to differentiate between freshly set `False` values and unset/`None` value options, unless we save everything...
> 
> If you save everything and uncheck the checkbox then in will be set to False - so it's correct that there is no way to differentiate between if a user wants to set it as false or wants to reset it. This needs refactoring of the interface. I'll try to add a simple way

One way of thinking is it is not so much an issue with the interface as with the db update method. We could introduce a check on the update action here instead of cleaning the pending object.
https://github.com/Clinical-Genomics/scout/blob/24c7216af0da58796709781b57267ba9f0b25191/scout/adapter/mongo/panel.py#L505

northwestwitch:
> One way of thinking is it is not so much an issue with the interface as with the db update method. We could introduce a check on the update action here instead of cleaning the pending object.
> 
> https://github.com/Clinical-Genomics/scout/blob/24c7216af0da58796709781b57267ba9f0b25191/scout/adapter/mongo/panel.py#L505

I don't understand, sorry.  Say you have this checkbox: 

image

How would you distinguish between the 2 options:
- Set reduced penetrance as False
- Set reduced penetrance as unknown

dnil:
> How would you distinguish between the 2 options:
> 
> * Set reduced penetrance as False
> * Set reduced penetrance as unknown

You probably can't, unless you have like a three-state button, but the thing is you dont quite need to? When you go to do the db update, you have access to the currently set values, and can opt to update only if the key exists and needs updating. E.g. if it is set True on the panel, and the pending change contains a False. The user must then actively have unselected it for the pending object to look like that.

northwestwitch:
> You probably can't, unless you have like a three-state button, but the thing is you dont quite need to? When you go to do the db update, you have access to the currently set values, and can opt to update only if the key exists and needs updating. E.g. if it is set True on the panel, and the pending change contains a False. The user must then actively have unselected it for the pending object to look like that.

If the matter is not about distinguish between None and False then I still don't understand. If you have reduced penentrance set to yes:

image

And you want to unset it using this branch, you actually can?

Tested locally:

1. Edit the gene and uncheck the checkbox ->
image

2. Apply action by clicking on "Update Panel"
image

northwestwitch:
AHHHHHH NO, sorry I was on another branch, I get it now! I'll fix!

scout/adapter/mongo/panel.py
```diff
@@ -512,22 +501,23 @@ def apply_pending(self, panel_obj, version):
...
             # If action is delete we do not add the gene to new genes
             if action == "delete":
                 continue
 
             if action == "edit":
-                for field in info_fields:
-                    if field in info:
-                        gene[field] = info[field]
+                # reset all fields except hgnc and symbol:
+                for key in EXPORT_PANEL_FIELDS[2:]:
```
northwestwitch, line 511: This should fix the resetting values issue

northwestwitch:
Mmm it had a complexity of 20 now. I'll try to refactor and simplify the function after lunch!

northwestwitch:
This is ready for another review when you have time DN. No rush though

scout/adapter/mongo/panel.py
```diff
@@ -462,89 +463,53 @@ def reset_pending(self, panel_obj):
         )
         return updated_panel
 
-    def apply_pending(self, panel_obj, version):
+    def apply_pending(self, panel_obj: dict, version: float) -> str:
```
dnil, line 466: 🥇

scout/adapter/mongo/panel.py
```diff
@@ -462,89 +463,53 @@ def reset_pending(self, panel_obj):
...
-                continue
-            info = update.get("info", {})
-            gene_obj = {"hgnc_id": hgnc_id, "symbol": update["symbol"]}
-
-            for field in info_fields:
-                if field in info:
-                    gene_obj[field] = info[field]
-            new_genes.append(gene_obj)
+        # Process 'add' actions and gather updates
+        updates = {u["hgnc_id"]: u for u in panel_obj.get("pending", []) if u["action"] != "add"}
```
dnil, line 479: This is quite compact, but looks good to me. Let's hope we don't need to change it in a few months.

scout/adapter/mongo/panel.py
dnil: 👍 Looks good; we trust the update info to have and set any true values and the rest are unset.

dnil (approved):
💯

dnil:
Tested on local; extra false values no longer show up, and true values can be set.

PR #4934, opened 2024-10-11
Dissmisive variantS page

### LINKED ISSUE #4922: VariantS page easier to use for dismissal
Here is a nice suggestion and request from @alieden.

When going through the variantS list to dismiss variants in bulk, a few key facts are not evident from the list but require hover over. While good, it still requires a bit of scrolling. In particular OMIM gene inheritance patterns, and local frequency are common dismissals.

At the same time, expanding them will take up real page real estate. It would be possible to adjust the columns dynamically based on the case type (e.g. the case "inheritance model" column is not as relevant for singletons). But the simplest solution is probably to 1) merge the Gene/Region annotation and Functional annotation. The exonic/splicing/intronic/ncRNA region call is often evident from the functional annotation and could eg be in a hover on it.

Expand the OMIM inheritance pattern and local frequency.

At the same time, expanding them will take up real page real estate. The simplest solution is probably to merge the Gene/Region annotation and Functional annotation. The exonic/splicing/intronic/ncRNA region call is often evident from the functional annotation and could eg be in a hover on it. Also the inheritance pattern and local frequency may not require separate columns; badges with inheritance models next to the genes for OMIM and the local counts/freq in the Local badge and not in hover would only take so many % of the page width.

**Describe alternatives you've considered**
It would be possible to adjust the columns dynamically based on the case type (e.g. the case "inheritance model" column is not as relevant for singletons). Or allow individual settings, but that is a whole different can of worms.

dnil:

- Fix #4922 

Screenshot 2024-10-11 at 12 20 23

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN


dnil:
Next week obviously! No rush whatsoever - it's QoL prio!

dnil:
Since it is still here, I collapsed the other functional field on the other variantS pages as well. It is a bit more atomic this way. It looks like many files changed, but is. basically the same thing happening on each page: 
- genes get an inheritance model bade, 
- "local" badges get counts
- the region field disappears and moves into a tooltip on the function field

dnil:
I've got one weird feedback item: one gene that lost its inheritance model, compared to the current db/OMIM release. I'm going keep my fingers crossed that was also an issue with the OMIM file release, but I'm going to try updating that case to be sure.

Screenshot 2024-10-14 at 17 32 51
Screenshot 2024-10-14 at 17 33 02
Screenshot 2024-10-14 at 17 33 19

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -272,21 +290,20 @@
...
     <span class="badge bg-secondary">MITOMAP</span>
-    <br>
   {% endif %}
   {% if variant.clingen_cgh_pathogenic %}
-    <span class="badge bg-secondary">CGH Pathogenic</span><br>
+    <span class="badge bg-secondary">CGH Pathogenic</span>
   {% endif %}
   {% if variant.local_obs_old or variant.local_obs_cancer_germline_old or variant.local_obs_cancer_somatic_old %}
-    <span class="badge bg-secondary">Local</span>
+    <span class="badge bg-secondary position-relative">Local{% if variant.category in ['snv', 'sv'] %}<span class="badge bg-info rounded-pill position-absolute top-0 start-100 translate-middle">
```
dnil, line 305: counts...

northwestwitch: I like this style but I'm afraid the count numbers are not very visible?

image

With dark therme it is slightly better but perhaps add them inside the "Local" badge itself inside a parenthesis?

northwestwitch (approved):
Very nice! I'm sure users will appreciate these changes. Regarding the local observations count I'm not that sure, for visibility reasons - check my comment!

scout/server/blueprints/variants/templates/variants/components.html
dnil: Hm, yes, the info color is really a bit in-between on whether to have light or dark color text on it, contrast wise. Especially with all the white around on light mode, it becomes a bit hard to see. It is still in the tooltip I guess, and it is pretty obvious what order of magnitude the number is. But for magnitude we could do different colors, stars or whatnot instead of fuzzy numbers. Lets see...

dnil: How about this:
Screenshot 2024-10-15 at 08 35 41
Screenshot 2024-10-15 at 08 35 47
When they changed to automatic pairing of bg and text color helpers in B 5.2, they paired dark text with the info badge: https://getbootstrap.com/docs/5.2/components/badge/#background-colors

northwestwitch: Better! 👍🏻

dnil: Or maybe the dark is a little worse. Let's go with `text-body`.

northwestwitch: > Or maybe the dark is a little worse. Let's go with `text-body`.

It was better on light mode tho

dnil: Yeah, so with body keeping the dark on light, and light on dark we should be good, right?

northwestwitch: > Yeah, so with body keeping the dark on light, and light on dark we should be good, right?

Or just put it between parentheses in the other badge?

dnil: Ok, let's make it a badge-in-the-badge then. It's essentially like parenthesis, but without the extra characters.
Screenshot 2024-10-15 at 10 41 10
Screenshot 2024-10-15 at 10 40 54

PR #5025, opened 2024-11-14
Refactor PanelApp panels loading to use the correct PanelApp API

### LINKED ISSUE #4988: Different number of PanelApp panels in /api and /WebService paths
I am investigating starting to run Scout's built in method to update the PanelApp Green panel, similarly to how we are updating the OMIM-AUTO panel.

When looking closer, I found that although most genes are shared between our panels, a significant number differ. We have a few hundred additional genes in our local panel, while missing a handful. (Both end up ~4000).

Looking closer, the difference seems to be where we access the PanelApp services. 

Scout it seems is using this one end-point to grab panels (yielding 175 panels):

* https://panelapp.genomicsengland.co.uk/WebServices/list_panels/

While in Lund we are using the `/api` end points. The panel ones here yields 400+ panels:

* https://panelapp.genomicsengland.co.uk/api/v1/panels/

Looking at the names, the 400+ is mostly (except one panel) a superset of the 175 ones.

Are you guys familiar with this difference? Any reason for why you are getting the data from the `/WebServices` URL rather than the `/api`? Unsure whether I am overlooking something fundamental here 🤔 Would like to start using the Scout-based PanelApp Green update tool, but would of course need to understand where all the panels and genes go first ...

northwestwitch:

### OBS: the "new" API has json files with slightly different key names/values, that's why I'm using new demo files in the tests and I had to touch the panelapp parsing code

- Closes #4988

Allows you to choose which type of panels your PanelApp GREEN panel is based on:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
0. Make sure your local instance has all genes
1. Try updating the PanelAPP GREEN panel --> `scout --demo update panelapp-green --institute cust000 (--signed-off)` (perhaps select the option to include only 100K RD type, to better compare with the panel we create using main branch?)
2. Try creating one gene panel based on one specific panelapp panel, example: `scout --demo load panel --panel-app  --panel-app-confidence green --panel-id 522`
`
**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
I'll just park this one here. I have to fix the tests' fixtures to return a format that looks like the correct API. But I think I'll start doing the testing of the new release now, it's more urgent.

scout/constants/panels.py
```diff
@@ -9,3 +9,9 @@
     ("custom_inheritance_models", "custom_inheritance_models"),
     ("comment", "comment"),
 ]
+
+PANELAPP_CONFIDENCE_EXCLUDE = {
```
northwestwitch, line 13: This has changed, and I'm moving it here from gene_tags, because it's panel-related

scout/commands/update/panelapp.py
```diff
@@ -19,14 +19,20 @@
     default="cust002",
     show_default=True,
 )
+@click.option(
```
northwestwitch, line 22: New option

scout/load/panelapp.py
```diff
@@ -0,0 +1,132 @@
...
+    for number, type in enumerate(available_types, 1):
+        LOG.info(f"{number}: {type}")
+    types_filter: str = input(
+        "Please provide a comma-separated list of types you'd like to use to build your panel (leave blank to use all types):  "
+    )
+    types_filter: List[str] = parse_types_filter(
+        types_filter=types_filter, available_types=available_types
+    )
+    LOG.info(f"Collecting green genes from panels of type: {types_filter}")
+    green_panel["description"] = (
```
northwestwitch, line 116: Description will contain the specifics of the filters applied (signed off or not, and selected types), for reference. Description was blank before.

scout/parse/panel.py
```diff
@@ -266,114 +260,6 @@ def parse_gene_panel(
     return gene_panel
 
 
-def parse_panel_app_gene(
```
northwestwitch, line 269: Moved to parse/panelapp.py

scout/load/panel.py
```diff
@@ -111,140 +106,6 @@ def load_panel(panel_path, adapter, **kwargs):
         raise err
 
 
-def _panelapp_panel_ids() -> List[str]:
```
northwestwitch, line 114: Moved to load/panelapp.py

northwestwitch:
This looks promising! If I use this branch to create the PanelApp GREEN panel only using the 100K RD type of panels, I get one panel that contains 3771 genes:

image

Exactly the same number as the genes I get when I create the panel using main branch:

image

I think I'm ready for a review now!

scout/demo/panelapp_panel.json
```diff
@@ -0,0 +1,463 @@
+{
```
dnil, line 1: 🥇

scout/constants/panels.py
dnil: Yikes - well spotted! Only on the new API I hope?

tests/commands/update/test_update_panelapp_green.py
```diff
@@ -32,42 +32,46 @@ def test_update_panelapp_green_non_existing_institute(empty_mock_app):
...
+    assert runner
 
     # THEN command to create the PANELAPP-GREEN panel should be successful
-    result = runner.invoke(cli, ["update", "panelapp-green", "-i", "cust000"])
+    result = runner.invoke(cli, ["update", "panelapp-green", "-i", "cust000"], input="1")
+
     assert result.exit_code == 0
 
-    # AND the panel should have been saved in databsse
+    # AND the panel should have been saved in database
```
dnil, line 73: ✅ 😄

scout/load/panelapp.py
```diff
@@ -0,0 +1,132 @@
...
+        "version": float(math.floor(old_panel["version"]) + 1) if old_panel else 1.0,
+        "date": datetime.now(),
+    }
+
+    LOG.info("Fetching all PanelApp panels")
+
+    panel_ids = panelapp.get_panel_ids(signed_off=signed_off)
+    LOG.info(f"\n\nQuery returned {len(panel_ids)} panels\n")
+    LOG.info("Panels have the following associated types:")
+    available_types: List[str] = panelapp.get_panel_types()
```
dnil, line 106: Could we do this on the CLI as well? I don't mind an `input()` fallback, but it is Good Practice to be able to run things in batch without having to rely on manual input. We could even have the panel types fetched for the `--help` output or a separate subcommand if you feel they are liable to change.

tests/commands/update/test_update_panelapp_green.py
```diff
@@ -32,42 +32,46 @@ def test_update_panelapp_green_non_existing_institute(empty_mock_app):
...
     )
 
-    # GIVEN an app CLI
+    # GIVEN a cli runner
     runner = mock_app.test_cli_runner()
+    assert runner
 
     # THEN command to create the PANELAPP-GREEN panel should be successful
-    result = runner.invoke(cli, ["update", "panelapp-green", "-i", "cust000"])
+    result = runner.invoke(cli, ["update", "panelapp-green", "-i", "cust000"], input="1")
```
dnil, line 69: Ah, cool - I didn't know that one! Still, I think a named option for the panel types would not hurt!

CHANGELOG.md
```diff
@@ -55,6 +56,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - VariantS page local observation badges have counts visible also without hover
 - On Matchmaker page, show number of matches together with matching attempt date
 - Display all custom inheritance models, both standard and non-standard, as gathered from the gene panel information on the variant page
+- Moved PanelApp-related code to distinct modules/extension
```
dnil, line 59: 👍

CHANGELOG.md
```diff
@@ -24,6 +25,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Increased visibility of variant callers's "Pass" or "Filtered" on the following pages: SNV variants (cancer cases), SV variants (both RD and cancer cases)
 - Names on IGV buttons, including an overview level IGV MT button
 - Cases query no longer accepts strings for the `name_query` parameter, only ImmutableMultiDict (form data)
+- Refactor the loading of PanelApp panels to use the maintained API - Customised PanelApp GREEN panels
```
dnil, line 28: 👍

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - MANE badges on list of variant's Genes/Transcripts/Proteins table, this way also SVs will display MANE annotations
 - Export variant type and callers-related info fields when exporting variants from variants pages
 - Cases advanced search on the dashboard page
+- Possibility to use only signed off panels when building the PanelApp GREEN panel
```
dnil, line 13: Not sure we will actually use that one, but seems very reasonable and prudent! Certification level stuff. 😁

dnil (changes requested):
Looks really good to me! I will do a little testing as well, but I already know I am not a big fan of CLIs that suddenly demand user input in one of who knows how many subcommands with multiple options. 😊 Could we just add that panel type input as an option as well, perhaps together with having an option to display the categories, on this commands help, or on a separate helper subcommand.

scout/load/panelapp.py
northwestwitch: We could, but I had this design in mind for this reason: 
If we had a list of possible panel types, say in the constants, we could use them as options in the cli at the beginning of the command, as you suggest. Butwhat if they introduce a new model type? Then we wouldn't be aware for quite some time (because we don't follow these things on PanelApp). 
Instead panel types get collected in real time from panel data and the options presented are always going to be up-to-date.

One thing I'd do for simplifying the options is perhaps ordering them so that the options will follow always the same order when they are presented to the user 🤔

scout/constants/panels.py
northwestwitch: Yes, only in the API!

northwestwitch:
> Looks really good to me! I will do a little testing as well, but I already know I am not a big fan of CLIs that suddenly demand user input in one of who knows how many subcommands with multiple options. 😊 Could we just add that panel type input as an option as well, perhaps together with having an option to display the categories, on this commands help, or on a separate helper subcommand.

I could change it like you suggested (I was going to to it like this when I started working on this feature), but then I reasoned that panel types may be mutable and that we don't have much control over them, so perhaps it is a good idea to have a system that is agnostic of panel type and collects this info as the script go. Checl also my other comment

tests/commands/update/test_update_panelapp_green.py
northwestwitch: Took me a while to remember how it was done, but we do have the same thing also in other tests:

image

scout/load/panelapp.py
dnil: Ordering is good! How about trying for dynamically checking the choice list anyway? It would not likely change between a call to retrieve the list and a second call. Having the user listing clear text category names on the command line/option would also add some safety.

northwestwitch: mmm I don't like much the idea of having the user to write text, since it can be a lot of categories and if you start to add more than one it can be really cumbersone. There are already two checks:

After the input when you write for instance 1,3,5, the script will list the chosen categories. The second check will be the description of the panel, where it clearly states the types of panel that were used for collecting the genes

tests/commands/update/test_update_panelapp_green.py
dnil: Right, I do vaguely remember now. 😁 But nice anyway!

scout/load/panelapp.py
dnil: Mm, in particular any options like this should be possible to add to config. How about that then, a dict with (cleartext) options in the config.py to configure what will happen by default. If none is given, or it is overridden, you could use the number input thing?

northwestwitch: Seems like an ovecomplication to me. Most facilities won't use this functionality. Lund will just press enter lo use all panels. When it comes to us, if we want to keep using the same settings as before we will insert one number. Otherwise a list of numbers, like 3 or 4 perhaps? And if we sort the options it will be likely the same sumber all over. Config file is already complicated as it is, and don't forget we are deprecating the config file currently used by the cli

northwestwitch:
After last change:

image

northwestwitch:
Latest latest:

image

northwestwitch:
Tested with the preselected panels shown in the pic above. The **GREEN panel contains now 4075 genes** (instead of 3771) so they're more, but not so many more

dnil:
> Tested with the preselected panels shown in the pic above. The **GREEN panel contains now 4075 genes** (instead of 3771) so they're more, but not so many more

Nice! A simplistic diff says about 80 of those are actually new as compared to our bulk OMIM-AUTO morbid genes, so well worth the effort, though I also don't think we need to rage too much about not having them in there earlier.

Skipping the prep, and possibly not 100% consistency between all genes in db etc, but basically:
```bash
 ➜  Downloads diff OMIM-AUTO_5.0_twocol.txt PANELAPP_15_twocol.txt |grep -c \>
229
 ➜  Downloads diff OMIM-AUTO_5.0_twocol.txt PANELAPP-GREEN_2.0_2024-11-18_twocol.txt |grep -c \>
309
```
Some are new (*ARHGAP35*), a lot are actually more border-lining issues with susceptibility in OMIM: *APOL1, AQP1, BMP6, FOXD3, GSTZ1,* ...

But hey, they include *CTNND2* now - we guessed that many years back (https://pubmed.ncbi.nlm.nih.gov/25473103/), so that absolutely has to be a solid list! 😉 

Hm, but for others again I beg to differ, as in *AGTR2*. That one may have been on some green lists, but should now really be deprecated? Is really?
Screenshot 2024-11-18 at 15 58 01
Let's just double check. Maybe it comes down to some db issue or timing there - it has been incorrectly described as causative before.

scout/constants/panels.py
```diff
@@ -9,3 +9,19 @@
...
+    "green": ["0", "1", "2"],
```
dnil, line 14: I suspect we have this in prod as well: good to have it fixed!

dnil (approved):
Really nice - this does clearly belong in an extension, and in the you end fix both the API-change and that "0" confidence issue, which probably was around with the old API as well.

I like the new constant of preselected panels. I would like it a lot better if there was a way to churn through the update without having to have a manual input, e.g. a `--yes` flag if you feel it is important to have it require input by default. I think we will likely do it your way anyway, but I've spent too much time automating input sequences for things like this in cronjobs to like it. 😁 

The code looks good to me, save perhaps the ambivalence on member variables or member function level variables - compare `page` vs `panel_ids`. If you have a functional reason, ignore it, but otherwise pick either way.

scout/server/extensions/panelapp_extension.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self):
+        self.panels_page = 1
+        self.panel_types = set()
+
+    def get_panel_types(self) -> list:
+        """Returns available panel types, collected from processed panels"""
+        return sorted(list(self.panel_types))
+
+    def get_panels(self, page: int, signed_off: bool = False) -> Optional[dict]:
```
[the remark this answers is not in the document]
northwestwitch: This way you could invoke the method of the extension to get a specific page without having to retrieve all the previous pages before - not that we are using the functionality but it's more flexible I think

scout/server/extensions/panelapp_extension.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_panel_ids(self, signed_off: bool) -> list[int]:
+        """Returns a list of panel ids contained in a json document with gene panels data."""
+
+        panel_ids = []
+
+        def get_ids(json_panels):
+            LOG.info(f"Retrieving IDs from API page {self.panels_page}")
+            for panel in json_panels.get("results", []):
+                panel_ids.append(panel["id"])
+            self.panels_page += 1
```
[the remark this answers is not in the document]
northwestwitch: I can add panel_ids to the constructor, good suggestion! 👍🏻

scout/server/extensions/panelapp_extension.py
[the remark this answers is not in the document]
dnil: Ok, good, then it is premeditated - it was just that the functionality was missing so far. All good!

PR #5092, opened 2024-12-02
Expand observation number

### LINKED ISSUE #5080: Allow expanding to show all local observations

On the variant page there is a box "local observations" with a summary of both the number of previously seen cases having the variant and a shortlist of max ten of these.

local_observations

This means that in this case 3 cases are hidden.

One industrious CLG here in Lund has mentioned a few times that he sometimes would like to be able to see cases beyond these 10, for instance if wanting to check out the final three in the example above.

One neat solution to this might be an "expand to show all" button. Then the box could default to showing max 10 as now, but would allow a user to dig deeper if interested.

Jakob37:
Close #5080 

Still new to the Scout code base, so this is a good practice. Don't hesitate giving feedbacks.

I have a draft solution which is behaving as expected in my test instance:

Before click:

pr_1_1

After click:

pr_1_2
pr_1_3

Thought I'd get your input on this draft before I wrap it up.

* Is the approach generally in line with the "Scout design" ?
* There are many hard-coded 10's here. It was hard-coded before as well, but would you prefer putting this in constants / configs somewhere?
* There is a `if len(obs_cases) == 10:` in `get_loqusdb_obs_cases` which confuses me a bit. Would this not mean that max 10 are retrieved? And that the "more ..." would not appear. (Clearly it is working fine though, so I guess I missed something). I used max 50 in my testing.
* Right now this is solved for one observation panel. Guess if an OK approach it should be implemented across other locations as well (Harder for me to test though)

Thanks 🙏 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
I'm trying to test the code locally, but I'm having a hard time because our stage loqus doesn't have so many cases apparently. But code-wise it looks OK I think.

I agree with you that number 10, or also 50, is really an arbitrary amount. One thing that one could do is allowing the institute's admins to change thins number in the institute settings: In these settings we also have the multiselect with the loqusdb instances that are used to collect the local frequency data:

image


Perhaps we could add there a input where one could specify the number of cases to show or alternatively it could be 10?

Jakob37:
> Perhaps we could add there a input where one could specify the number of cases to show or alternatively the default number could be 10?

Yes, that sounds like a nice idea.

> I'm trying to test the code locally, but I'm having a hard time because our stage loqus doesn't have so many cases apparently. But code-wise it looks OK I think.

OK I see! Then I'll look into wrapping up the actual PR.

Jakob37:
Hmm. Looking a bit more into the institute settings. Searching for other settings, it seems they are used in ~50 places in the code. Probably the default max display is too small of a thing to warrant touching the code so much.

I'd say let's keep it with 10 as the default (as before, just slightly more exposed). With the added ability to expand more if desired.

Jakob37:
Remaining points:

- [ ] Deal with that `max_display` just appends more `max_display=20` `max_display=30` after each other at the moment, update the current value if present instead
- [ ] More?

Jakob37:
Had an issue before with that the "max_display" parameter was simply appended to the URL, instead of updating the actual value.

Trying a different approach now. Seems to work fine for all pages, testing with some dummy data.

testing_obs_cancer
testing_obs_snv
testing_obs_sv

Will try this out on our "real" HG002 data as well.

Thinking points are:

* I removed the max-10 threshold when retrieving cases from db. Might impact performance? Does not seem to during testing though.
* Some calculations in the Jinja template now. Not sure how desired, but also not sure how to do it otherwise.

scout/server/blueprints/variant/controllers.py
```diff
@@ -442,8 +442,6 @@ def get_loqusdb_obs_cases(
     obs_cases = []
     user_institutes_ids = set([inst["_id"] for inst in user_institutes(store, current_user)])
     for i, case_id in enumerate(obs_families):
-        if len(obs_cases) == 10:
```
northwestwitch, line 445: This could be a bit dangerous.. what if you have 5000 observations?

Jakob37: Hmm, I see! Yes, you might be right. Let's bring back a slightly higher fixed max number.

Jakob37:
Testing it out on a real case

real_case

Jakob37:
Ready for review, when you have the time @northwestwitch 

I felt a bit uncertain on how to best set up the "Show more" URL, i.e. the changes in `variant_details.html` culminating in the line below. Feedbacks there are extra welcome!

```
<a href="{{ request.script_root }}{{ request.path }}?{{ args|urlencode }}">Show more</a>
```

northwestwitch (commented):
I'm having an hard time testing this PR because of our stage setup, sorry. Even in cases with many local observations I have a situation like this, which is not helping:

image

Anyway, I think if we don't go for a more complicated approach (the one I've suggested yesterday), which probably is an overkill, perhaps it is fine to collect 50 variants and show them all? I'm not very fond of your solution (requests and passing around that number 10). So I'd rather display 50 or display 10 and a button or link like; `show more ` which opens a collapsible div showing the rest? 
Check here on how to do it-> https://getbootstrap.com/docs/5.0/components/collapse/

northwestwitch:
> Yes, I agree - your collapse approach seems both nicer for the user and much less intrusive code-wise 🙂 I'll give it a go. Anyway, it has been a good journey into Flask and Scout.

Yes, it's a very good introduction :)

Jakob37:
OK, here is a go at the collapse approach. Some variant of this is clearly the way to go I think. I didn't realize bootstrap could do this by itself - I thought you needed custom JS (or HTMX if you are familiar with that).

collapse_0

collapse_1

It runs nicely. A minor itch is that I didn't find a HTML/bootstrap only way to toggle the button text, so now it always appears as "Show more/less". And as there may be many of these panels, a vanilla-JS approach would be quite messy (then I would need some pointer on where to put the JS). 

I would be happy to give it a go, but I would also be happy to leave it like this 😄

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -251,8 +251,14 @@
...
                   {% endfor %}
-                  {% if obs.cases|length > 10 %}
-                     + more
+                  {% if obs.cases|length > max_display %}
+                    {% set current_max_display = request.args.get('max_display', default=max_display) | int %}
+                    {% set updated_max_display = current_max_display + 10 %}
+                    {% set args = request.args.to_dict() %}
+                    {% do args.update({'max_display': updated_max_display}) %}
+                    <div>
+                      <a href="{{ request.script_root }}{{ request.path }}?{{ args|urlencode }}">Show more</a>
```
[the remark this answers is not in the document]
Jakob37: Yes, I think 50 would be enough. I think it would work for us to just assign that. Or use the toggle. Either way!

scout/server/blueprints/variant/templates/variant/variant_details.html
```diff
@@ -240,19 +254,23 @@
...
-                       <a class="badge rounded-pill bg-light text-dark" target="_blank" href="{{ url_for('variant.variant', institute_id=data.case.owner, case_name=data.case.display_name, variant_id=data.variant._id) }}">{{ data.case.display_name }}</a>
-                     {% elif data.variant and data.variant.category == "sv"%}
-                       <a class="badge rounded-pill bg-light text-dark" target="_blank" href="{{ url_for('variant.sv_variant', institute_id=data.case.owner, case_name=data.case.display_name, variant_id=data.variant._id) }}">{{ data.case.display_name }}</a>
-                     {% else %}
-                       <span data-bs-toggle="tooltip" title="Missing link, this might be caused by variants not loaded after a rerun or inexact SV matching.">
-                         <span class="ml-3 badge rounded-pill bg-light text-dark">{{ data.case.display_name }}</span>
-                       </span>
-                     {% endif %}
+                  {% for observed_case in obs.cases %}
+                    {% if loop.index <= 10 %}
```
northwestwitch, line 258: Nice update! just for the sake of testing with real data, do you mind if I lower this number to 1 or 2. It should work on our stage server

Jakob37: Of course, go ahead!

northwestwitch:
Ok almost there I would say.

I have this situation with 3 occurrences, 1 is a case without a link:

image

When I click on show more/less I get this (it basically repeats the 2 cases with the link: 

image

northwestwitch:
Other example:

image

After I click to expand:

image

Jakob37:
Or, I think you need to update here as well for it to be a fair test:

```
   {% if loop.index <= 10 %}
```

northwestwitch:
It works great! I'll revert so we can merge! 💯

Jakob37:
> It works great! I'll revert so we can merge! 💯

Nice, thanks for testing!

scout/server/blueprints/variant/controllers.py
```diff
@@ -462,7 +462,8 @@ def get_loqusdb_obs_cases(
...
+        max_observations = 50
+        if len(obs_cases) == max_observations:
```
northwestwitch, line 466:
```suggestion
        if len(obs_cases) == 50:
```

Without adding a new variable seems easier, no?

northwestwitch (approved):
Looks great 💯

scout/server/blueprints/variant/controllers.py
Jakob37: Matter of style I guess, I tend to be a bit more verbose and use named variables to make it clear what the purpose of each variable is.

But let's stick with your preferred approach here. I'll update!

PR #5093, opened 2024-12-02
Contain custom image size and add a link to show them full size in another tab

### LINKED ISSUE #5086: Wide image destroys the case page layout
**Describe the bug**

When the image is wide enough, it pushes itself to a new row, leaving the sidebar behind.

**To Reproduce**

Upload a wide enough image with a case and toggle it open on the case page.

**Expected behavior**

Don't break the layout. Probably a scroll-box would be my recommended approach.

**Screenshots**

large_image_destroys_layout

**Additional context**

I would be happy to provide a PR for this, if you agree with this issue (as it is easier for me to test, having the image already available).

northwestwitch:
- Closes #5086

## Image gets resized with ratio in scout:

image

## New tab with the full size image:

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Add a very big custom image to case page or a variant page
2. Make sure the page doesn't break
3. Click on the image and check that the image gets displayed much bigger on a new tab

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
I have a PR to fix that image size issue @Jakob37, you could perhaps review?

Jakob37:
> I have a PR to fix that image size issue @Jakob37, you could perhaps review?

Yes, I'll check!

Jakob37:
Hmm. I think this would work nicely on a large display, but less so on my laptop (where I am testing it out right now). I suspect it is a very normal size for a laptop (1920 x 1080, 150% scaled, both at recommended-by-windows settings).

Confirming that the image indeed is resized to 1200px:

wide_image_4

Still shifted to a new row:

wide_image_3

Checking with the inspector, it looks like I have 992px space (inside the padding) in the area right of the left bar.

A basic solution would maybe be to reduce the width even further to 950px or so.

A more correct solution would probably be to see if you could make the image fit whatever size is available beyond the left sidebar. This would require more CSS-digging though. Not sure whether that currently is worth it or not for you. Up to you :)

scout/server/static/custom_images.js
```diff
@@ -9,15 +17,24 @@
 		 .then(blob => {
 				 const img = document.createElement("img");
 				 img.src = URL.createObjectURL(blob);
+				 if (width > imgMaxWidth || height> imgMaxHeight) {
```
Jakob37, line 20:
```suggestion
				 if (width > imgMaxWidth || height > imgMaxHeight) {
```

scout/server/static/custom_images.js
```diff
@@ -9,15 +17,24 @@
...
+				 		var resized_pixels = calculateAspectRatioFit(width, height, imgMaxWidth, imgMaxHeight);
```
Jakob37, line 21:
```suggestion
				 		const resized_pixels = calculateAspectRatioFit(width, height, imgMaxWidth, imgMaxHeight);
```

I would stick with the block scoped `let` and `const` of you don't have specific reasons not to. Here the value is not mutated, so `const` would be the correct one.

Jakob37 (commented):
It does not solve it yet on a small laptop screen.

The ideal would be to solve this without hard-coded widths and heights, using CSS. But that is more intrusive and maybe beyond the scope of the current work.

Jakob37:
Opening in new window on click works nicely, that will be convenient!

northwestwitch:
> It does not solve it yet on a small laptop screen (see comment above).
> 
> The ideal would be to solve this without hard-coded widths and heights, using CSS. But that is more intrusive and maybe beyond the scope of the current work. Up to you!

I'll see if these max width and height can be set automatically given the browser's window size. I agree that that way is better 👍🏻

northwestwitch:
I've pushed a possible fix, when you have some time, check how it works for you!

Jakob37 (approved):
Looks good to me!

northwestwitch:
Thanks for the review @Jakob37. I'm slightly modified the code to replace one of the 2 jpeg images (small images, same size) of the demo case with a bigger one. Merging when tests are passing!

PR #5158, opened 2025-01-08
Fix odd hgnc gene inheritance

dnil:

It's not easy when it's not easy I guess. But I'm tempted to say "trivial" post hoc. Old aliases could contain lower case letters in the HGNC `previous_symbols`, which would then not match gene symbols (that are now mandated to be upper case).
❌  Before:
Screenshot 2025-01-08 at 11 47 12

✅ After:
Screenshot 2025-01-08 at 14 33 48

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. check the entries for eg gene 714 (*ARSB*) and 18746 (*SLURP1*) and perhaps verify that ARSB variants show no inheritance badges, pLi constraints etc.
2. apply patch
3. `scout update genes`
4. check the same entries and variants, noticing good content for both

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/scout/utils/link.py
+++ b/scout/utils/link.py
@@ -36,15 +36,16 @@ def genes_by_alias(hgnc_genes):
         hgnc_symbol = gene["hgnc_symbol"]
 
         for alias in gene["previous_symbols"]:
+            alias = alias.upper()
             true_id = None
             if alias == hgnc_symbol:
                 true_id = hgnc_id
             if alias in alias_genes:
-                alias_genes[alias.upper()]["ids"].add(hgnc_id)
+                alias_genes[alias]["ids"].add(hgnc_id)
                 if true_id:
-                    alias_genes[alias.upper()]["true"] = hgnc_id
+                    alias_genes[alias]["true"] = hgnc_id
             else:
-                alias_genes[alias.upper()] = {"true": true_id, "ids": set([hgnc_id])}
+                alias_genes[alias] = {"true": true_id, "ids": set([hgnc_id])}
 
     return alias_genes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
+## [unreleased]
+### Fixed
+- Gene aliases with lower case letters mixed would not match gene symbols when adding inheritance, constraint etc
+
 ## [4.94.1]
 ### Fixed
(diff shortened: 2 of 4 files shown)


dnil:
For good measure I also tried reloading a case to ensure that the inheritance badges would return to *ARSB* and others (upon reload - only the variant page is live against the gene collection, variantS use what we have on load).

Before:
Screenshot 2025-01-08 at 14 52 45

After:
Screenshot 2025-01-08 at 15 12 08

northwestwitch:
> It's not easy when it's not easy I guess. But I'm tempted to say "trivial" post hoc. Old aliases could contain lower case letters in the HGNC `previous_symbols`, which would then not match gene symbols (that are now mandated to be upper case). ❌ Before: Screenshot 2025-01-08 at 11 47 12
> 

What where the old aliases for these gene? The screenshots doesn't show them

northwestwitch:
> > It's not easy when it's not easy I guess. But I'm tempted to say "trivial" post hoc. Old aliases could contain lower case letters in the HGNC `previous_symbols`, which would then not match gene symbols (that are now mandated to be upper case). ❌ Before: Screenshot 2025-01-08 at 11 47 12
> 
> What where the old aliases for these gene? The screenshots doesn't show them

Answering to myself: The aliases in the gene document are the same. I see now that the difference that you wanted to show is the presence of the right inheritance model. The fact that aliases array field was expanded made me think that there was a change in the names of the aliases saved as well!

scout/utils/link.py
```diff
@@ -36,15 +36,16 @@ def genes_by_alias(hgnc_genes):
         hgnc_symbol = gene["hgnc_symbol"]
 
         for alias in gene["previous_symbols"]:
+            alias = alias.upper()
```
northwestwitch, line 39: 💯

northwestwitch (approved):
Neat! 🥇

dnil:
> > What where the old aliases for these gene? The screenshots doesn't show them
> 
> Answering to myself: the aliases in the gene document are the same. I see now that the difference that you wanted to show is the presence of the right inheritance model. The fact that aliases array field was expanded made me think that there was a change in the names of the aliases saved as well!

Yes, all that is already saved is fine, only that due to the potential for case mismatches on a couple of comparisons out of a large number we could miss populating inheritance patterns and constraints on the variant document. For cases like this, where there was a really old alias from the time when gene symbols could use lower case...
Screenshot 2025-01-09 at 12 21 33

Issue #5170, opened 2025-01-13
Collect and save inheritance models for PanelApp panels correctly


northwestwitch:
Inheritance models are provided for genes, but don't get saved in the panels created on scout.

northwestwitch:
Fix also missing penetrance info!

PR #5171, opened 2025-01-13
Save also reduced penetrance = false (together with true) when loading specific/all PanelApp panels

northwestwitch:
- Fixes partially #5170 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Saving gene inheritance models when loading/updating specific/all PanelApp panels (doesn't apply to the `PanelApp Green Genes panel`)
 - Save also complete penetrance status (in addition to incomplete) if available when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
 
-
 ## [4.94.1]
 ### Fixed
 - Temporary directory generation for MT reports and pedigree file for case general report
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Missing inheritance, constraint info for genes with symbols matching other genes previous aliases with some lower case letters
 - Loading of all PanelApp panels from command line
 - Saving gene inheritance models when loading/updating specific/all PanelApp panels (doesn't apply to the `PanelApp Green Genes panel`)
-- Save also reduced penetrance = false (together with true) when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
+- Save also complete penetrance status (in addition to incomplete) if available when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
 
 
 ## [4.94.1]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - ACMG temperature on case general report should respect term modifiers
 - Missing inheritance, constraint info for genes with symbols matching other genes previous aliases with some lower case letters
 - Loading of all PanelApp panels from command line
+- Saving gene penetrance info when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
 
 ## [4.94.1]
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - ACMG temperature on case general report should respect term modifiers
 - Missing inheritance, constraint info for genes with symbols matching other genes previous aliases with some lower case letters
 - Loading of all PanelApp panels from command line
-- Saving gene penetrance info when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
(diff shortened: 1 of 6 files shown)


northwestwitch:
### Main branch:

image


### This branch:

image

scout/server/blueprints/panels/templates/panels/panel.html
```diff
@@ -178,7 +178,7 @@
               <td>{{ gene.hgnc_id }}</td>
 
               <td>{{ gene.disease_associated_transcripts|join(', ') }}</td>
-              <td>{{ 'Yes' if gene.reduced_penetrance }}</td>
+              <td>{{ gene.reduced_penetrance }}</td>
```
northwestwitch, line 181: I'm not 100% sure about the changes in this PR. I think it's good to include also reduced penetrance = False, because it could be important info when creating the panels from PanelApp. But I'm concerned about an old bug when creating gene panels. The bug affected all genes of a panel and the effect was that reduced penetrance was set to False (instead of the key being missing) whenever the user didn't specify the field for the gene. Introducing this change we might have a situation like this, which is extremely confusing 🤔 

image

scout/constants/gene_tags.py
```diff
@@ -33,7 +33,7 @@
     "other": {"bgcolor": "bg-light", "text_color": "text-dark"},
 }
 
-INCOMPLETE_PENETRANCE_MAP = {"unknown": None, "Complete": None, "Incomplete": True}
+INCOMPLETE_PENETRANCE_MAP = {"unknown": None, "None": None, "Complete": False, "Incomplete": True}
```
dnil, line 36: 👍

scout/build/panel.py
```diff
@@ -42,7 +42,7 @@ def build_gene(gene_info: dict, adapter) -> dict:
 
     # Add boolean flags
     gene_obj.update(
-        {key: True for key in ["reduced_penetrance", "mosaicism"] if gene_info.get(key)}
+        {key: gene_info.get(key) for key in ["reduced_penetrance", "mosaicism"] if key in gene_info}
```
dnil, line 45: Ah, yes. This can potentially be painful. I dunno, I think I would have played it safe and also added a check against `None` values, although they should in theory not be there now. But, we can always fix if it breaks.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Missing inheritance, constraint info for genes with symbols matching other genes previous aliases with some lower case letters
 - Loading of all PanelApp panels from command line
 - Saving gene inheritance models when loading/updating specific/all PanelApp panels (doesn't apply to the `PanelApp Green Genes panel`)
+- Save also reduced penetrance = false (together with true) when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
```
dnil, line 21:
```suggestion
- Save also complete penetrance status (in addition to incomplete) if available when loading specific/all PanelApp panels (does not apply to the `PanelApp Green Genes panel`)
```

scout/server/blueprints/panels/templates/panels/panel_pdf_simple.html
```diff
@@ -67,7 +67,7 @@ <h4>Scout - Gene panel report</h4> - created on:&nbsp;<strong>{{report_created_a
             <a style="text-decoration:none;" href="https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:{{gene.hgnc_id}}" target="_blank">{{gene.hgnc_id}}</a>
           </td>
           <td>{{ gene.disease_associated_transcripts|join(', ') }}</td>
-          <td>{{ 'Reduced penetrance' if gene.reduced_penetrance }}</td>
+          <td>{{ gene.reduced_penetrance }}</td>
```
dnil, line 70: This is simple enough, but looking at eg mosaicism the columns will be less easy to read. Perhaps a three way comparison, or even using "Reduced" and "Complete"?

dnil (approved):

scout/server/blueprints/panels/templates/panels/panel_pdf_simple.html
northwestwitch: I'd rather keep it as it is (showing only reduced penentrance or  `<td>{{ gene.reduced_penetrance }}</td>`

That's because of the old bug I mentioned before, see a comment above. If you check for instance the document relative to [this](https://scout-stage.scilifelab.se/panels/61dd2cb5aa60e01391e78beb) panel, you see that all genes have False penentrance and False mosaicism:

image


Which is only due to the old bug. If I turn all the genes with penentrance False on these panels affected by the bug into a "Complete" penentrance in the html/pdf page then we are stating something that is wrong (we actually don't know if the penetrance is complete or not).. Perhaps it is safer to just keep the code as it is and show only the reduced penetrance = true?

dnil: Ah, I see - good, I thought we cleaned that up, but you are right - not for the ones that were changed during that period. Yes, I like the keeping strategy then: I think the redundant words vs the header are useful when the values are sparse. Right?

northwestwitch: I'll just revert to how it was then, thanks!

Issue #5204, opened 2025-01-28
MT-TP gene is missing


dnil:
**Is your feature request related to a problem in the current program to new available techology or software? Please describe and add links/citations if appropriate.**
Certain RNA genes are implicated in disease, but we are failing load of many of them. Let's investigate and try to fix well-behaved ones. It is clear that for certain groups of rRNA genes etc it will not be possible to assign proper located genes and symbols, but the disease mapped ones are more unique.

northwestwitch:
I've started to look into this matter using the MT-TP gene. The gene is present in the downloaded file hgnc.txt:

<img width="1336" alt="Image" src="https://github.com/user-attachments/assets/3340a454-d148-4862-9397-7d383ca919e6" />

But is missing from the Ensembl files, both builds.

Good thing that it is reported also in the phenotype_to_genes.txt file


```
(py39) chiararasi@n159-p41 scout-downloads % grep MT-TP phenotype_to_genes.txt
HP:0031704	Abnormal ear physiology	4571	MT-TP	OMIM:545000
HP:0031704	Abnormal ear physiology	4571	MT-TP	ORPHA:551
HP:0001510	Growth delay	4571	MT-TP	ORPHA:551
HP:0004322	Short stature	4571	MT-TP	ORPHA:551
HP:0002197	Generalized-onset seizure	4571	MT-TP	ORPHA:551
HP:0002197	Generalized-onset seizure	4571	MT-TP	OMIM:545000
HP:0001250	Seizure	4571	MT-TP	OMIM:545000
HP:0000005	Mode of inheritance	4571	MT-TP	OMIM:545000
HP:0100543	Cognitive impairment	4571	MT-TP	ORPHA:551
HP:0002493	Upper motor neuron dysfunction	4571	MT-TP	OMIM:545000
HP:0001507	Growth abnormality	4571	MT-TP	ORPHA:551
HP:0001324	Muscle weakness	4571	MT-TP	OMIM:545000
HP:0001251	Ataxia	4571	MT-TP	OMIM:545000
HP:0001251	Ataxia	4571	MT-TP	ORPHA:551
HP:0001336	Myoclonus	4571	MT-TP	OMIM:545000
HP:0020219	Motor seizure	4571	MT-TP	ORPHA:551
HP:0020219	Motor seizure	4571	MT-TP	OMIM:545000
HP:0002664	Neoplasm	4571	MT-TP	ORPHA:551
[...]

PR #5218, opened 2025-02-05
Fix parsing of entire Ensembl documents - including MT-TP gene

northwestwitch:
- Fix parsing of Ensembl resources. While working on #5204, I've realised that the only missing gene after the parsing is MT-TP - I'll include comments in the PR code, to explain why I think this happens

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Install the branch locally or on stage
2. Download all resources and update the database
3. Compare with the number of genes available on prod

**Expected outcome**:
- The only missing gene on prod is MT-TP!!

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/utils/ensembl_biomart_clients.py
+++ b/scout/utils/ensembl_biomart_clients.py
@@ -4,7 +4,7 @@ from typing import Dict, Iterator
 import requests
 
 LOG = logging.getLogger(__name__)
-SCHUG_BASE = "https://schug.scilifelab.se"
+SCHUG_BASE = "https://schug-stage.scilifelab.se"
 
 BUILDS: Dict[str, str] = {"37": "GRCh37", "38": "GRCh38"}
 
@@ -29,15 +29,6 @@ class EnsemblBiomartHandler:
     def stream_resource(self, interval_type: str) -> Iterator[str]:
         """Use schug web to fetch genes, transcripts or exons from a remote Ensembl biomart in the right genome build and save them to file."""
 
-        def yield_resource_lines(iterable) -> str:
-            """Removes the last element from an iterator."""
-            it = iter(iterable)
-            current = next(it)
-            for i in it:
-                yield current
-                current = i
-
         shug_url: str = f"{SCHUG_BASE}{SCHUG_RESOURCE_URL[interval_type]}{self.build}"
 
-        # return all lines except the last, which contains the "[success]" string
-        return yield_resource_lines(self.stream_get(shug_url))
+        return self.stream_get(shug_url)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Do not show overlapping gene panels badge on variants from cases runned without gene panels
 ### Fixed
 - Don't save any "-1", "." or "0" frequency values for SNVs - same as for SVs
+- Downloading and parsing of genes from ensembl (including MT-TP)
 
 ## [4.96]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Do not show overlapping gene panels badge on variants from cases runned without gene panels
 ### Fixed
 - Don't save any "-1", "." or "0" frequency values for SNVs - same as for SVs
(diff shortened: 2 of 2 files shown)


scout/utils/ensembl_biomart_clients.py
```diff
@@ -4,7 +4,7 @@
 import requests
 
 LOG = logging.getLogger(__name__)
-SCHUG_BASE = "https://schug.scilifelab.se"
+SCHUG_BASE = "https://schug-stage.scilifelab.se"
```
northwestwitch, line 7: To be reverted before merging. This is only for testing

scout/utils/ensembl_biomart_clients.py
```diff
@@ -29,15 +29,6 @@ def stream_get(self, url: str) -> Iterator:
     def stream_resource(self, interval_type: str) -> Iterator[str]:
         """Use schug web to fetch genes, transcripts or exons from a remote Ensembl biomart in the right genome build and save them to file."""
 
-        def yield_resource_lines(iterable) -> str:
```
northwestwitch, line 32: I remember introducing this code in order to remove the last line of the streamed Ensembl files, which was the following:
`[success]`

In the files downloaded using the [latest schug](https://github.com/Clinical-Genomics/schug/pull/99) - deployed now on stage - this line is missing. The last line of the downloaded file is instead the one relative to MT-TP. By removing these lines we make sure to parse the entire file, including all genes

northwestwitch:
### Genes on stage

image

No MT-TP:

image

### Locally, using this branch:

image

MT-TP is there:

image

scout/utils/ensembl_biomart_clients.py
northwestwitch: This change goes together with [this recent change in schug](https://github.com/Clinical-Genomics/schug/pull/87/files)

dnil:
Heh, an obi-one! But let's see, werent they in like blocks with one `[success]` per chromosome? Are we missing like 20 genes then?

northwestwitch:
> Heh, an obi-one! But let's see, werent they in like blocks with one `[success]` per chromosome? Are we missing like 20 genes then?

The parsing didn't work at all before, see https://github.com/Clinical-Genomics/schug/issues/82#issue-2792845990. 

Then the error was fixed in schug, where all `[success]` lines where removed. Afterwards we updated the genes in production database. The only problem left is in scout. because scout is still removing the last line assuming it is a `[success]` line. Instead it is a gene

dnil:
Agreed, but wouldn't there have been one success statement per chr before?

northwestwitch:
> Agreed, but wouldn't there have been one success statement per chr before?

Ok before the bug in schug, that introduced one [success] line per chromosome (so yes, you are right) there was only one [success] line, at the end, and the scout code worked just fine, since it removed the last line.

Then I fixed the error in schug, removing ALL [success] lines. Since we didn't fix scout accordingly, the software is currently removing the last line, even if it is a gene (MT-TP). And that is why we have one less gene in production now.

This PR fixes scout so we parse the last line instead of removing it.

So to answer one of your questions, no, no difference of 20 genes. just one. Does this answer your questions?

northwestwitch:
> Ok, and during the interim with multiple [success] we would have had a crash on an intervening malformatted line. I think I'm in the clear. All good then! 👍🏻

Exactly, the parsing did not work when we had multiple [success] lines. 👍🏻

scout/utils/ensembl_biomart_clients.py
```diff
@@ -29,15 +29,6 @@ def stream_get(self, url: str) -> Iterator:
...
-            current = next(it)
-            for i in it:
-                yield current
-                current = i
-
         shug_url: str = f"{SCHUG_BASE}{SCHUG_RESOURCE_URL[interval_type]}{self.build}"
 
-        # return all lines except the last, which contains the "[success]" string
-        return yield_resource_lines(self.stream_get(shug_url))
+        return self.stream_get(shug_url)
```
dnil, line 34: Ahem, I feel very convinced I approved this first time around? But, I guess it was a busy day. 😊

dnil (approved):
💯

scout/utils/ensembl_biomart_clients.py
northwestwitch: It's OK, I am not in a rush

PR #5309, opened 2025-03-12
Institute settings reset

### LINKED ISSUE #5306: Clear soft filters doesn't clear them
Institute settings, soft filters

northwestwitch:
- Fixes #5306 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Try updating institute settings, both using the CLI or the web page

**Expected outcome**:
- It should be possible to clear soft filters
- Additionally, make sure that when a key has null values, it gets removed from the institute document

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/institute.py
+++ b/scout/adapter/mongo/institute.py
@@ -65,7 +65,7 @@ class InstituteHandler(object):
         if not institute_obj:
             raise IntegrityError("Institute {} does not exist in database".format(internal_id))
 
-        updates = {"$set": {}}
+        updates = {"$set": {}, "$pull": {}, "$unset": {}, "$push": {}}
         updated_institute = institute_obj
 
         if sanger_recipient:
@@ -78,7 +78,7 @@ class InstituteHandler(object):
                     internal_id, sanger_recipient
                 )
             )
-            updates["$push"] = {"sanger_recipients": sanger_recipient}
+            updates["$push"]["sanger_recipients"] = sanger_recipient
 
         if remove_sanger:
             LOG.info(
@@ -86,24 +86,7 @@ class InstituteHandler(object):
                     remove_sanger, internal_id
                 )
             )
-            updates["$pull"] = {"sanger_recipients": remove_sanger}
-
-        # Set a number of items
-        GENERAL_SETTINGS = {
-            "cohorts": cohorts,
-            "collaborators": sharing_institutes,
-            "coverage_cutoff": coverage_cutoff,
-            "display_name": display_name,
-            "frequency_cutoff": frequency_cutoff,
-            "gene_panels": gene_panels,
-            "gene_panels_matching": gene_panels_matching,
-            "loqusdb_id": loqusdb_ids,
-            "sanger_recipients": sanger_recipients,
-            "clinvar_submitters": clinvar_submitters,
-        }
-        for key, value in GENERAL_SETTINGS.items():
-            if value not in [None, ""]:
-                updates["$set"][key] = value
+            updates["$pull"]["sanger_recipients"] = remove_sanger
 
         if phenotype_groups is not None:
             if group_abbreviations:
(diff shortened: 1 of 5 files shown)


scout/adapter/mongo/institute.py
```diff
@@ -65,7 +65,7 @@ def update_institute(
         if not institute_obj:
             raise IntegrityError("Institute {} does not exist in database".format(internal_id))
 
-        updates = {"$set": {}}
+        updates = {"$set": {}, "$pull": {}, "$unset": {}, "$push": {}}
```
northwestwitch, line 68: $unset is missing, this allows you to remove the key completely if the value is empty, I'll run some checks by resetting everything and see what happens

scout/adapter/mongo/institute.py
```diff
@@ -121,29 +93,39 @@ def update_institute(
...
+            "soft_filters": soft_filters,  # Admin setting
+            "check_show_all_vars": check_show_all_vars is not None,
         }
-        for key, value in ADMIN_SETTINGS.items():
-            if value not in [None, "", []]:
+        for key, value in UPDATE_SETTINGS.items():
+            if bool(value) is True:
                 updates["$set"][key] = value
+            else:
+                updates["$unset"][key] = ""  # Remove the key from the institute document
```
dnil, line 120: Let's see - the whole issue with separating admin from user was that for ordinary users, the keys would not be set on the form, but that would not mean we want them unset in db. But since they are really tested the same way now, we should be fine!

scout/server/blueprints/institutes/controllers.py
```diff
@@ -391,8 +391,16 @@ def update_institute_settings(store: MongoAdapter, institute_obj: Dict, form: Mu
     updated_institute = store.update_institute(
         internal_id=institute_obj["_id"],
         sanger_recipients=get_sanger_recipients(form),
-        coverage_cutoff=int(form.get("coverage_cutoff")),
-        frequency_cutoff=float(form.get("frequency_cutoff")),
+        coverage_cutoff=(
```
dnil, line 394: 👍

scout/adapter/mongo/institute.py
```diff
@@ -65,50 +65,22 @@ def update_institute(
...
             raise IntegrityError("Institute {} does not exist in database".format(internal_id))
 
-        updates = {"$set": {}}
+        updates = {"$set": {}, "$unset": {}}
         updated_institute = institute_obj
 
         if sanger_recipient:
-            user_obj = self.user(sanger_recipient)
-            if not user_obj:
-                raise IntegrityError("user {} does not exist in database".format(sanger_recipient))
```
dnil, line 74: I think I like this, but it is a bit of a change. Obviously nice to be able to redirect to accounts that are not "users" yourself, but on the minus side of this you dont get a check that the email is correct. If you like, you could break out a small function for Sanger updates and document the intention in a docstring. But it is smallish already, so all good in the current function docstring. 😊

dnil (approved):
Base fix, super, no issues! Refactor (hehe) all good I think. Then there are a couple of function changes (eg for non-user recepients) and if I follow them correctly corresponding test changes. I do approve, but maybe write a couple of words about it the docstring, or even break out a couple of small functions with docstrings of their own.

scout/adapter/mongo/institute.py
northwestwitch: The thing is, you are forced to provide the email of an existing user if you are adding a recipient using the CLI, but if you add a recipient from the web page of the institute settings, then you can add whichever email you like (works like this on current main). I figured this check is not important at all then in the CLI

scout/adapter/mongo/institute.py
```diff
@@ -65,50 +65,22 @@ def update_institute(
...
 
         if remove_sanger:
-            LOG.info(
-                "Removing sanger recipient {0} from institute: {1}".format(
-                    remove_sanger, internal_id
-                )
+            sanger_recipients = list(
+                set(institute_obj.get("sanger_recipients", [])) - set([remove_sanger])
             )
-            updates["$pull"] = {"sanger_recipients": remove_sanger}
```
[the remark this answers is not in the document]
northwestwitch: Yes, a little more complex. I like this way better

scout/adapter/mongo/institute.py
```diff
@@ -60,36 +60,51 @@ def update_institute(
...
 
-        # Set a number of items
-        GENERAL_SETTINGS = {
+        UPDATE_SETTINGS = {
+            "alamut_institution": alamut_institution,  # Admin setting
```
northwestwitch, line 103: Ordered these alphabetically now

scout/adapter/mongo/institute.py
```diff
@@ -60,36 +60,51 @@ def update_institute(
     ) -> Union[dict, str]:
         """Update the information for an institute."""
 
-        add_groups = add_groups or False
+        def get_phenotype_groups() -> dict:
```
northwestwitch, line 63: Moved this to its own function

scout/adapter/mongo/institute.py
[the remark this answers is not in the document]
dnil: To be fair, it is only more complex with the new joint loop, otw it needed an extra set operation. But all good! 👍

scout/adapter/mongo/institute.py
dnil: No, agreed, the risk of misspelling should be about the same! 😁 Just saying that could be worth a couple of doc words somewhere.

northwestwitch: I think it is fine. I wouldn't document a feature that is not there any more. And even when it was there it wasn't used if you added the sanger user via web form

northwestwitch: Unless we want to add the check that the user exist also when you add it using the web form?

PR #5353, opened 2025-03-25
Ensembl data integrity check

### LINKED ISSUE #5352: Introduce a check for downloaded Ensembl data
Ensembl downloads are problematic and often result in incomplete files available when we have to update genes and exons. 
It would be good to add a check to make sure the files are complete before attempting the update of the database.

One way to do this would be to count the number of [success] lines present in the files. It should be one for each chromosome - see also this [PR on schug](https://github.com/Clinical-Genomics/schug/pull/105).

One could also check that all chromosomes are available in the files, but the method above is more secure, because what if the file is interrupted before the very end and lacks just a few lines of chr MT?

northwestwitch:
- Closes #5352 

Tested locally:
- Genes 37 🆗 
- Genes 38 🆗 
- Transcripts 37 🆗 
- Transcripts 38 ❌ - incomplete file

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/commands/download/ensembl.py
+++ b/scout/commands/download/ensembl.py
@@ -18,9 +18,9 @@ LOG = logging.getLogger(__name__)
 def integrity_check(nr_chromosomes_in_file: int):
     if nr_chromosomes_in_file < NR_EXPECTED_CHROMS:
         raise BufferError(
-            f"Ensembl resource does not seem to be complete. Please retry downloading the file."
+            "Ensembl resource does not seem to be complete. Please retry downloading the file."
         )
-    LOG.info(f"Integrity check OK.")
+    LOG.info("Integrity check OK.")
 
 
 def print_ensembl(
--- a/scout/commands/download/ensembl.py
+++ b/scout/commands/download/ensembl.py
@@ -2,7 +2,6 @@
 
 import logging
 import pathlib
-import time
 from typing import List, Optional
 
 import click
--- a/scout/commands/download/ensembl.py
+++ b/scout/commands/download/ensembl.py
@@ -40,6 +40,8 @@ def print_ensembl(
 
         LOG.info(f"{file_name} file saved to disk")
 
+        ensembl_client.check_integrity(file_path)
+
 
 @click.command("ensembl", help="Download files with ensembl info")
 @click.option("-o", "--out-dir", default="./", show_default=True)
--- a/scout/commands/download/ensembl.py
+++ b/scout/commands/download/ensembl.py
@@ -2,13 +2,25 @@
 
 import logging
 import pathlib
+import time
 from typing import List, Optional
 
 import click
 
 from scout.utils.ensembl_biomart_clients import EnsemblBiomartHandler
 
+CHROM_SEPARATOR = "[success]"
+NR_EXPECTED_CHROMS = 24
+
 LOG = logging.getLogger(__name__)
+
+
+def integrity_check(nr_chromosomes_in_file: int):
+    if nr_chromosomes_in_file < NR_EXPECTED_CHROMS:
+        raise BufferError(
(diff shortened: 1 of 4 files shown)


scout/utils/ensembl_biomart_clients.py
```diff
@@ -32,3 +32,19 @@ def stream_resource(self, interval_type: str) -> Iterator[str]:
...
+        """Counts the number of lines containing '[success]'. One such line per chromosome is expected for the file to be considered intact."""
+
+        def count_line_occurrences(file_path, target_line):
+            with open(file_path, "r", encoding="utf-8") as file:
+                return sum(1 for line in file if line.strip() == target_line)
+
+        nr_chromosomes_in_file = count_line_occurrences(file_path, "[success]")
+        nr_chromosomes = 24
+        if nr_chromosomes_in_file < nr_chromosomes:
+            raise BufferError(
```
northwestwitch, line 46: This will stop the downloads (all downloads if your command is `scout download everything`), but it's fine, since the Ensembl resources are the last resources downloaded by that command, see [here](https://github.com/Clinical-Genomics/scout/blob/50de25163c351ef6cddb6d422635cd19df07ad7f/scout/commands/download/everything.py#L40)

scout/commands/download/ensembl.py
```diff
@@ -8,9 +8,20 @@
...
+
 LOG = logging.getLogger(__name__)
 
 
+def integrity_check(nr_chromosomes_in_file: int):
+    if nr_chromosomes_in_file < NR_EXPECTED_CHROMS:
+        raise BufferError(
+            "Ensembl resource does not seem to be complete. Please retry downloading the file."
+        )
+    LOG.info("Integrity check OK.")
```
dnil, line 22:
```suggestion
    LOG.info("Ensembl download integrity check OK.")
```
or sth to identify what the integrity is for. Maybe not even needed? 😊

scout/commands/download/ensembl.py
```diff
@@ -8,9 +8,20 @@
...
+CHROM_SEPARATOR = "[success]"
+NR_EXPECTED_CHROMS = 24
+
 LOG = logging.getLogger(__name__)
 
 
+def integrity_check(nr_chromosomes_in_file: int):
+    if nr_chromosomes_in_file < NR_EXPECTED_CHROMS:
+        raise BufferError(
+            "Ensembl resource does not seem to be complete. Please retry downloading the file."
```
dnil, line 20:
```suggestion
            "Ensembl resource does not seem to be complete. Please retry downloading genes/transcripts."
```
Perhaps pass the file type, or catch the exception in the processing function where the resource type is known.

dnil (approved):
Straightforward solution; this should cover teh the most important - making sure one does not just update with a broken file. Then the next step would be to add a retry/partial result caching solution, but perhaps better implemented on the schug side.

scout/commands/download/ensembl.py
northwestwitch: It's referred to the file that is printed immediately before. I don't think it's needed

scout/commands/download/ensembl.py
northwestwitch: File name is printed immediately before, it should be clear from the context, I think

northwestwitch:
I'm testing the downloads again on hasta stage. Build 37 gives no problems at all. Build 38 is a tragedy: many failed attempts, both downloading genes and transcripts. I have the feeling that Biomart is not reliable as it was any more 😞

PR #5365, opened 2025-03-31
Keep phenotype-related info on case reupload with --keep-actions tag

### LINKED ISSUE #5333: Retain HPO gene panel on case --update

When cases have been updated using `scout load case --update`, the same HPO terms and / or dynamic gene panels should typically be assigned. These are reset though, meaning that the user needs to readd them for each case.

In instances where we have needed to reprocess many samples, this leads to a considerable amount of work and is also a potential source of errors.

Retaining the HPO gene panel on update (or adding a command line flag to keep it `--keep-hpo`).

Would that make sense? :)

I.e. added HPO terms / panel here.

Image

northwestwitch:
- Closes #5333 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -875,7 +875,7 @@ class CaseHandler(object):
 
         if keep_actions:
             # collect eventual phenotype-related data already existing for the case
-            update_case_phenotypes(old_case=existing_case, new_case=case_obj)
+            self.update_case_phenotypes(old_case=existing_case, new_case=case_obj)
 
             # collect all variants with user actions for this case
             return list(self.evaluated_variants(case_obj["_id"], institute_obj["_id"]))
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -834,7 +834,7 @@ class CaseHandler(object):
     def check_existing_data(
         self, case_obj: dict, existing_case: dict, institute_obj: dict, update, keep_actions: bool
     ) -> Optional[List[dict]]:
-        """Make sure data from case to be loaded/re-uploaded conforms to case data already saved in database.
+        """Make sure data from a case to be loaded/re-uploaded conforms to any case data already saved in the database.
         Return eventual evaluated variants to be propagated to the updated case if keep_actions is True.
         """
 
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -831,20 +831,11 @@ class CaseHandler(object):
         result = self.case_collection.delete_one(query)
         return result
 
-    def check_existing_data(self, case_obj, existing_case, institute_obj, update, keep_actions):
-        """Make sure data from case to be loaded/reuploaded conforms to case data already saved in database.
-           Return eventual evaluated variants to be propagated to the updated case if keep_actions is True
-
-        Args:
-            case_obj(dict): case dictionary to be loaded/reuploaded
-            existing_case(dict): a case with same _id or same display_name and institute_id as case_obj
-            institute_obj(dict): institute dictionary
-            update(bool): If existing case should be updated
-            keep_actions(bool): If old evaluated variants should be kept when case is updated
-
-        Returns:
-            previous_evaluated_variants(list): list of variants evaluated in previous case
-                or None if case is not already present in the database.
+    def check_existing_data(
+        self, case_obj: dict, existing_case: dict, institute_obj: dict, update, keep_actions: bool
+    ) -> Optional[List[dict]]:
+        """Make sure data from case to be loaded/re-uploaded conforms to case data already saved in database.
+        Return eventual evaluated variants to be propagated to the updated case if keep_actions is True.
         """
 
         if existing_case is None:
@@ -883,8 +874,24 @@ class CaseHandler(object):
             )
 
         if keep_actions:
+            # collect eventual phenotype-related data already existing for the case
+            update_case_phenotypes(old_case=existing_case, new_case=case_obj)
+
             # collect all variants with user actions for this case
             return list(self.evaluated_variants(case_obj["_id"], institute_obj["_id"]))
+
+    def update_case_phenotypes(self, old_case: dict, new_case: dict):
+        """If case is re-runned/re-uploaded, remember phenotype-related settings from the old case, including assigned diseases, HPO terms, phenotype groups and HPO panels."""
+        for key in [
+            "phenotype_terms",
+            "phenotype_groups",
+            "diagnosis_phenotypes",
+            "dynamic_panel_phenotypes",
+            "dynamic_gene_list",
+            "dynamic_gene_list_edited",
+        ]:
+            if key in old_case:
+                new_case[key] = old_case[key]
 
     def update_case_data_sharing(self, old_case: dict, new_case: dict):
         """Update data sharing info for a case that is re-runned/re-uploaded."""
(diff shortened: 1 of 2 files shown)


northwestwitch:
### Given a case with edited diseases/HPO terms and assigned HPO panel:

image


### When re-upload is done using master branch:

`scout --demo load case scout/demo/643594.config.yaml --update --keep-actions`

- Edited phenotypes and phenotype groups are not remembered:


image



### If you use this branch instead

The above phenotype-related info is maintained

-  Edited phenotypes and phenotype groups are remembered

scout/adapter/mongo/case.py
```diff
@@ -831,20 +831,11 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
-            update(bool): If existing case should be updated
-            keep_actions(bool): If old evaluated variants should be kept when case is updated
-
-        Returns:
-            previous_evaluated_variants(list): list of variants evaluated in previous case
-                or None if case is not already present in the database.
+    def check_existing_data(
+        self, case_obj: dict, existing_case: dict, institute_obj: dict, update, keep_actions: bool
+    ) -> Optional[List[dict]]:
+        """Make sure data from case to be loaded/re-uploaded conforms to case data already saved in database.
```
dnil, line 837:
```suggestion
        """Make sure data from a case to be loaded/re-uploaded conforms to any case data already saved in the database.
```

scout/adapter/mongo/case.py
```diff
@@ -831,20 +831,11 @@ def delete_case(self, case_id=None, institute_id=None, display_name=None):
...
+        Return eventual evaluated variants to be propagated to the updated case if keep_actions is True.
```
dnil, line 838:
```suggestion
        If keep_actions is True, return any evaluated variants to be propagated to the updated case. The flag indicates that old evaluated variants should be kept when the case is updated.
```

scout/adapter/mongo/case.py
```diff
@@ -883,9 +874,25 @@ def check_existing_data(self, case_obj, existing_case, institute_obj, update, ke
...
 
         if keep_actions:
+            # collect eventual phenotype-related data already existing for the case
+            self.update_case_phenotypes(old_case=existing_case, new_case=case_obj)
+
             # collect all variants with user actions for this case
             return list(self.evaluated_variants(case_obj["_id"], institute_obj["_id"]))
 
+    def update_case_phenotypes(self, old_case: dict, new_case: dict):
+        """If case is re-runned/re-uploaded, remember phenotype-related settings from the old case, including assigned diseases, HPO terms, phenotype groups and HPO panels."""
```
dnil, line 884:
```suggestion
        """If case has been re-run/re-uploaded, remember phenotype-related settings from the old case, including assigned diseases, HPO terms, phenotype groups and HPO panels."""
```

scout/adapter/mongo/case.py
```diff
@@ -883,9 +874,25 @@ def check_existing_data(self, case_obj, existing_case, institute_obj, update, ke
             )
 
         if keep_actions:
+            # collect eventual phenotype-related data already existing for the case
+            self.update_case_phenotypes(old_case=existing_case, new_case=case_obj)
```
dnil, line 878: Wouldn't this be better called over here?
https://github.com/Clinical-Genomics/scout/blob/6a68ef2636622cfc1ef14e117b85f523007f50f3/scout/adapter/mongo/case.py#L1010

scout/adapter/mongo/case.py
```diff
@@ -883,9 +874,25 @@ def check_existing_data(self, case_obj, existing_case, institute_obj, update, ke
...
+        for key in [
+            "phenotype_terms",
+            "phenotype_groups",
+            "diagnosis_phenotypes",
+            "dynamic_panel_phenotypes",
+            "dynamic_gene_list",
+            "dynamic_gene_list_edited",
+        ]:
+            if key in old_case:
+                new_case[key] = old_case[key]
```
dnil, line 894: Hm, it's a little bit more complicated than that, isn't it? Case orders can also have phenotype_groups and _terms set from order, which should probably have precedence, though that can be argued. From experience, users are a little more prone to edit these in Scout than in their LIMS or the Orderportal initially. I'm honestly not 100% certain they can edit everything in a rerun order in the orderportal right now, but from a scout perspective, there is a new case load config to compare to the case on db.

There are a couple more fields like this on the case: e.g `synopsis` and `cohorts` from a quick check. I would suggest adding them to the list, and to act on any new info from the case. Perhaps adding to lists e.g. phenotype_terms and _groups concatenating synopsis if set etc?



Obviously the dynamic ones are already perfectly handled here! We also have case `groups` that would fit nicely on that list.

You already deal with some like this when updating the data sharing; e.g. `collaborators` would fit snugly there.

There is also the individual settings, really; like sampling age, tissue, and tumor type that could be manually edited. Could be another issue though, if you wish!

dnil (changes requested):
Nice! Should work really well for the dynamic ones. I have some worries about the ones you can set on a load config though, plus some suggestions for additional keys to save and a suggestion to move the case save checking out of the variant getter function and onto the case controller. If you are worried about the complexity there, maybe even group a couple of similar functions like the data sharing preservation one and make a new joint one?

scout/adapter/mongo/case.py
northwestwitch: I wouldn't mix data-sharing and phenotype-related keys if the phenotype-related ones have to be checked against, and eventually be integrated with keys present on the config file. It would be more messy I think.

northwestwitch: Ah OK, but I think you din't meant, joining the keys of the two functions into a unique function. Then yes, why not, I'll move the call to the function in the finally, as you suggest!

scout/adapter/mongo/case.py
```diff
@@ -886,6 +878,29 @@ def check_existing_data(self, case_obj, existing_case, institute_obj, update, ke
...
+        for key in [
+            "phenotype_terms",
+            "phenotype_groups",
+            "diagnosis_phenotypes",
+        ]:  # Remember key/values from old case and integrate with info provided on case config file
+            if key not in old_case:
+                continue
+            new_case[key] = list(
+                {
+                    json.dumps(d, sort_keys=True): d for d in new_case.get(key, []) + old_case[key]
```
northwestwitch, line 900: These are lists of dictionaries so to merge them and avoid duplicates I can't do `set(list1) + set(list2)`, that's whyit is like this

scout/adapter/mongo/case.py
```diff
@@ -886,6 +878,29 @@ def check_existing_data(self, case_obj, existing_case, institute_obj, update, ke
             # collect all variants with user actions for this case
             return list(self.evaluated_variants(case_obj["_id"], institute_obj["_id"]))
 
+    def update_case_phenotypes(self, old_case: dict, new_case: dict):
+        """If case has been re-run/re-uploaded, remember phenotype-related settings from the old case, including assigned diseases, HPO terms, phenotype groups and HPO panels."""
+        for key in [
```
dnil, line 883: ✅ Looks good, those should always be kept, no discussion.

scout/adapter/mongo/case.py
dnil: Looks solid; I even asked ChatGPT about a simplification: 😆 
>Ah yes, that one's got "ChatGPT wrote this at 2AM to deduplicate a list of dicts" energy. 😅 

But it really just came up with variations on a theme, so let's go with this. I can read it after an initial while of staring. 😁  If we do it more than once, it would sure not hurt putting it in a function with a docstring.

dnil (approved):
Thank you, this solves the issue at hand. The situation with some other the other values I mentioned, like synopsis, still isn't great, but let's just merge this and make a new issue for that.

PR #5366, opened 2025-03-31
Fix #5361 - attempt at least, with js load after window complete

### LINKED ISSUE #5361: Variant page load is slow when Chanjo2 is active
**Describe the bug**

When updating Scout to v4.99 I found that variant loading page had become very slow due to the new coverage indicator. Too slow to be released to our users I think.

As we are not yet actively using Chanjo2 (hopefully we will eventually 🙏 ), I decided to disable it for now. But then the variant page crashes:

```
scout_dev  |   File "/home/worker/app/scout/server/blueprints/variant/views.py", line 106, in variant
scout_dev  |     data = variant_controller(
scout_dev  |            ^^^^^^^^^^^^^^^^^^^
scout_dev  |   File "/home/worker/app/scout/server/blueprints/variant/controllers.py", line 409, in variant
scout_dev  |     "gene_has_full_coverage": get_gene_has_full_coverage(
scout_dev  |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
scout_dev  |   File "/home/worker/app/scout/server/blueprints/variant/controllers.py", line 430, in get_gene_has_full_coverage
scout_dev  |     hgnc_id: chanjo2.get_gene_complete_coverage(
scout_dev  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
scout_dev  |   File "/home/worker/app/scout/server/extensions/chanjo2_extension.py", line 78, in get_gene_complete_coverage
scout_dev  |     gene_cov = resp.json()
scout_dev  |                ^^^^^^^^^^^
scout_dev  |   File "/home/worker/app/.venv/lib/python3.12/site-packages/requests/models.py", line 978, in json
scout_dev  |     raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
scout_dev  | requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Looking into the code, it seems that it now expects Chanjo2 to be present here:

https://github.com/Clinical-Genomics/scout/blob/921751e46e6256d991f1b628dda6531953c82241/scout/server/blueprints/variant/controllers.py#L419-L438

From a brief reading it seems that Scout does not check for whether chanjo2 is available before calling this function.

**To Reproduce**

1. Disable Chanjo2
2. Open the variant page

**Expected behavior**

If Chanjo2 is not available - don't calculate the coverage.

If Chanjo2 is available - make it configurable if this should be calculated automatically. An alternative would be to let the user actively choose to calculate it on a button click.

dnil:

- Fix #5361 - use js call to api, to chanjo2 extension, to set indicator after window load completion

Screenshot 2025-03-31 at 13 46 43
Screenshot 2025-03-31 at 13 43 20
Screenshot 2025-03-31 at 13 45 12

This is about what I meant. Looks tentatively ok, but I'm a little worried about the test server load still with this. We will be spamming the chanjo2 service, and if d4tools is going to take some time each call.

Anyway, looks like the tests are also bugging with the latest and greatest in the lock file. Presumably some library has updated with some significance, but one thing at a time.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


northwestwitch:
> Looks like a good solution to me. I can't really say if it makes a difference because loading of the button is too fast on our stage server, so perhaps @Jakob37 could give it a try instead?
> 
> Could something be done for the icons? They don't look super clear I think:
> 
> image

Perhaps just removing the second button and having just one green button that says "complete coverage" or a yellow one with "incomplete coverage"?

Jakob37:
I'll give it a go!

Jakob37:
Animation

Works fine! Async is a good choice here I think. Worked fine opening a bunch of variants as well.

It means a lot of data transfer though. Unsure whether an issue when server are under more load here. Could there be an institute-option to disable these async calculations?

Finally, no request, just a user question - What about a mini-spinner instead of a static question mark when waiting to signal things are happening?

Nice job 💪 👌

dnil:
Screenshot 2025-04-23 at 16 07 50
Screenshot 2025-04-23 at 16 08 18


Ok, this should be ready for another check:
- added a button-looking span to get some more visible status color: perhaps we could merge that with the report button itself to save space at some point, but I didn't want to refactor that function as well now. This also looks a little better when the window is narrow, and the elements end up stacked over one another.
- added a fadein-fadeout pulse to the "Checking" question mark. It's kind of pretty, but with any luck no-one will ever notice it. 😄 

We could absolutely make this on demand, either by expecting a click to check, or by institute settings, but I feel we will need to sort any performance issues with the chanjo2-api regardless? The users may just as well click each variant coverage report, or multiple HPO term reports.

dnil:
Ok like this for you both? See anything odd or is all ok still?

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -161,12 +161,8 @@
...
-          <span class="bg-success fa-solid fa-circle-check" data-bs-toggle="tooltip" data-bs-placement="bottom" title="Chanjo2 coverage is at 100% completeness."></span>
-        {% else %}
-          <span class="bg-warning fa-solid fa-triangle-exclamation" data-bs-toggle="tooltip" data-bs-placement="bottom" title="Note that Chanjo2 coverage is below 100% completeness."></span>
-        {% endif %}
+        {{ chanjo2_report_form(institute, case, gene.hgnc_symbol if gene.hgnc_symbol else gene.hgnc_id, 'overview', gene.hgnc_id, "Gene coverage " ~ (gene.hgnc_symbol if gene.hgnc_symbol else gene.hgnc_id), "btn btn-sm btn-secondary text-white") }} <!--chanjo2 genes overview -->
+        <span class="button complete-coverage-indicator btn-info text-white btn-sm" style="display:flex" id="coverage-indicator-{{gene.hgnc_id}}-button" data-bs-toggle="tooltip" data-bs-placement="bottom" data-loading-text="Checking completeness..." title="Checking completeness..."><span id="coverage-indicator-{{gene.hgnc_id}}-icon" class='blink_me fa-solid fa-circle-question'></span><span id="coverage-indicator-{{gene.hgnc_id}}-text">Checking</span>&nbsp;{{gene.hgnc_symbol}}</span>
```
northwestwitch, line 165: This works very nicely. About the looks of this button, but I noticed that sometimes it's aligned with the coverage button and sometimes it isn't:

image

image


Turning these 2 buttons into a button group could help here. Is it doable? If this is doable then you can also omit the name of the gene in the second button

Also the icon inside the button is misaligned with the text:

image

northwestwitch (approved):
Very nice with the loading! I have a suggestion for a couple of improvements to the aesthetics of the button, check my comment.

scout/server/blueprints/variant/templates/variant/components.html
dnil: Sure, lets try at least! It would be less repetitive, but slightly wider by default without the stacking.

Jakob37:
I like the pulsing - it looks pleasing and clearly shows that something is going on 😄 

I agree with the styling comments already mentioned by Chiara. Nothing else from me.

dnil:
Ok, so, I moved the coverage buttons down to their own list item row to make the display a little less crowded. The local obs column is usually larger anyway, so it's not like it is wasting vertical space. I think I managed to coerce all spans to be proper flex style, perhaps overly so, but now they align pretty well. The aestethics when there are multiple genes for the same variant could be better, but the remainder will be a little complicated to fix with different length gene names.

Narrow display, two genes:
Screenshot 2025-04-28 at 15 32 49

Wide display, two genes:
Screenshot 2025-04-28 at 15 33 06

One gene: 
Screenshot 2025-04-28 at 15 33 49

I also moved the tooltips a little so we only get one for chanjo, one for chanjo2 coverage.

northwestwitch:
It looks very nice, thanks!

Issue #4842, opened 2024-09-12
Variant searches hangs indefinitely


Jakob37:
Hello from Lund!

We have been hunting down mysterious performance drops in Scout causing indefinite high disk usage by Mongo, bogging things down until restart.

Realizing today that it probably is related to the variant search. This exact search hangs (or at least seems to be running indefinitely) while using a lot of disk, halting other activity on the server.

![search_variants](https://github.com/user-attachments/assets/a52d6290-925f-4ee6-97de-b5a289d24e07)

It seems it might be due to our large variant collection of 80M+ variants.

Is it possible to speed up the search you think? Or somehow prevent / warn when a user is going to do a huge search. 

Also wondering whether this is an issue for you in Stockholm doing a simple gene search.

I guess we either way also should consider trimming down our db, as most likely not all of those variants are relevant 🤔

northwestwitch:
Hi @Jakob37, I've just tried on our prod server, same gene search and it gets through:

<img width="1453" alt="image" src="https://github.com/user-attachments/assets/62e21f7f-7e1e-4502-b3e3-3286da9a45ec">

We have something like 465.7M variants in pur database. Could it be a missing index?

Checking on compass we have the following indexes in place:

<img width="1194" alt="image" src="https://github.com/user-attachments/assets/bd1da546-5e5d-40ef-8533-d572d735cd24">

Jakob37:
OK, interesting! That is good to know. Will look over the index again.

Can I ask, roughly how long time does it take for you to do this search? Seconds or more?

northwestwitch:
> Can I ask, roughly how long time does it take for you to do this search? Seconds or more?

9 seconds! :)

dnil:
What does your partial index criteria look like @Jakob37 - thats really the key thing here. We could not sustain searches against the whole variantS collection either, it's really in how we limit the partial index. Maybe if your db server is a bit less powerful, you could try setting the rank score cutoff for searches a tad higher? Every point at that low level shaves a huge number of variants. https://github.com/Clinical-Genomics/scout/blob/6a52248f53dbecc322dd18e9943af8cd04dacfe2/scout/constants/indexes.py#L76

Jakob37:
OK, great tip @dnil ! We will try this out

ehre:
@Jakob37 Just as a background, I do these types of searches now and then, and in my experience it wasn't a big problem until you moved the servers yesterday. Before, I would get a result in under 30 seconds for sure.

[...]

PR #4538, opened 2024-03-28
reintroduce and slightly modify variant rankscore sort index

dnil:
This PR adds a functionality or fixes a bug.

- Fix https://github.com/Clinical-Genomics/scout/issues/4533

<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the Pax software available at [https://pax.scilifelab.se/](https://pax.scilifelab.se). The resource you are going to call dibs on is `scout-stage` and the server is `cg-vm1`.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, repeat procedure at [https://pax.scilifelab.se/](https://pax.scilifelab.se), which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

<details>
<summary>Testing on hasta server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. `ssh <USER.NAME>@hasta.scilifelab.se`
1. Book your testing time using the Pax software. `us; paxa -u <user> -s hasta -r scout-stage`. You can also use the WSGI Pax app available at [https://pax.scilifelab.se/](https://pax.scilifelab.se).
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `conda activate S_scout; pip freeze | grep scout-browser`
1. Deploy the branch to test: `bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_scout -t scout -b <this_branch>`
1. Make sure the branch is deployed: `us; scout --version`
1. After testing is done, repeat the `paxa` procedure, which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

PR #5413, opened 2025-04-24
Fix #5412 - speed up SV matching SNV

### LINKED ISSUE #5412: Large loading times on the SV view for cases with excessive numbers of SNVs
**Describe the bug**

Loading times out on cases with a large number of SNVs. This is not intended use, but still feels paradoxical to have the SV page break, when the SNV page works fine.

**To Reproduce**
1. Go to e.g. https://scout-stage.scilifelab.se/cust002/F0057291/sv/variants?variant_type=clinical
2. See error
 
**Expected behavior**
Lets fix the gene overlapping function - likely it caps nr of matching variants late in the process (or maybe not at all with SNVs?).

**Additional context**
One could also consider to in addition to fixing the error, also improving the loading experience by "async"-ing the "gene overlapping" variants, either loading all the tooltips late, or possibly doing them on demand (hover), though I suspect that would feel really sluggish.

dnil:

- Fix #5412 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/constants/indexes.py
```diff
@@ -100,15 +100,16 @@
                 ("case_id", ASCENDING),
                 ("variant_type", ASCENDING),
                 ("rank_score", ASCENDING),
+                ("hgnc_ids", ASCENDING),
```
dnil, line 102: This is the actual speedup, allowing to use this index with hgnc_id filtering, with score for sorting.

scout/constants/indexes.py
```diff
@@ -57,12 +57,12 @@
...
             # Filterish index
             [
-                ("case_id", ASCENDING),
                 ("variant_type", ASCENDING),
+                ("case_id", ASCENDING),
                 ("sub_category", ASCENDING),
                 ("hgnc_ids", ASCENDING),
             ],
-            name="case_id_variant_type_sub_category_hgnc_ids",
+            name="variant_type_case_id_sub_category_hgnc_ids",
```
dnil, line 65: This is an old thing from a consultant meeting, making it a little bit easier on the query scheduler, and making the indices a little less redundant.

scout/constants/indexes.py
```diff
@@ -100,15 +100,16 @@
...
                 ("variant_type", ASCENDING),
                 ("rank_score", ASCENDING),
+                ("hgnc_ids", ASCENDING),
             ],
-            name="category_caseid_varianttype_rankscore",
+            name="category_caseid_varianttype_rankscore-hgncids",
         ),
         IndexModel(
             [
+                ("chromosome", ASCENDING),
```
dnil, line 109: This is also the same as the first, just making it a bit easier on the scheduler.

dnil:
Not including the variant_id in the search, and adding it as a filter after, should help MongoDB planning the query - there is no index with it so it will do some slow fetch-limit on it. 

However, even with that, the query scheduler actually does this:
Screenshot 2025-04-24 at 12 04 08

Screenshot 2025-04-24 at 12 01 29

The key reason being that there is no index that has all of the relevant fields. We have a nice one for hgnc_ids filter, but it has variant_rank instead of score so sorting would be slow. Attempting to fix this with a new index:
Screenshot 2025-04-24 at 11 35 34
I'll let it build and we can evaluate after.

dnil:
Hm, that test fail is actually a tad worrying. Hopefully nothing, but it might suggest that that clinical filter test previously tested ok due to the variant id of an overlapping variant being found on the html response, but perhaps not as a full variant. Lets see.

dnil:
A wait, it's probably just a different db when switching to the app context!

scout/constants/indexes.py
northwestwitch: Nice! 🚀

dnil:
Ok, the index is up to a working level; looks as expected:
Screenshot 2025-04-24 at 15 52 54
Note the difference to the steps of 96 ms and 3.7 s before. 😆

dnil:
🎬 45 index keys checked casually for 45 documents, compared to 200k keys checked and 200k docs fetched on the above.

northwestwitch (approved):
Super! 💯

dnil:
If you have time @Jakob37 it would be nice to make sure this does not repeat the issue you guys had with slow queries here last year (https://github.com/Clinical-Genomics/scout/pull/4538 I think?). 😊

Jakob37:
> If you have time @Jakob37 it would be nice to make sure this does not repeat the issue you guys had with slow queries here last year (#4538 I think?). 😊

Hmm. The memory is faint now, but I see from the issue that I had memory overload when loading a case in a demo environment. Is that it? I can test that out during this week.

Looking back in time I also saw that we had slow search queries (#4842), but that seems to have been due to us not having reindexed after update.

dnil:
Tested on stage:
Screenshot 2025-04-29 at 11 26 21

PR #5460, opened 2025-05-15
Show assignees in case list when user ID is different from email

maehler:
This PR fixes an issue where the assignees for a case wouldn't be visible in the list of cases. This is exactly the same fix as for #3979, but here it wasn't as apparent that something was missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fix long STR variant pinned display on case page (#5455)
 - Variant page crashing when Loqusdb instance is chosen on institute settings but is not found at the given URL (#5447)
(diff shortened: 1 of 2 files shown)


maehler:
I guess I should reword the PR title and the changelog though since it's not an LDAP issue per se. It's rather an issue when the user ID is different from the email.

northwestwitch:
> I guess I should reword the PR title and the changelog though since it's not an LDAP issue per se. It's rather an issue when the user ID is different from the email.

That works as well, and should cover other types of authentication. Nice!

maehler:
> Nicely done, thank you! You are a collaborator on this repo as well, if it works for you, you could open the PR directly on this repo to get all automation working out of the box. But in this case it made perfect sense: our stage setup doesn't use LDAP by default anyway! 😁

That makes sense, and I'll keep it in mind for the future!

PR #5507, opened 2025-05-28
Avoid variant events duplication on case re-upload

### LINKED ISSUE #5501: Automatic new activity when case is reloaded
When a case is reloaded, previous dismiss activity appear with a new date under activity. I don't see the point of this but it's not annoying. Example of when a dismissal from 2022 is automatically listed again as new activity when the rerun is loaded in 2025:

Image

However, for another variant where Sanger sequencing was ordered and then cancelled, it appears again as ordered after the rerun: 

Image

This is not a big deal, it says Sanger pending in Scout which is a bit confusing when the Sanger order request was cancelled. But no email was sent so it didn't cause any actual problem. Would it be possible to stop reloading dismiss and Sanger order when a rerun is loaded?

northwestwitch:
- Closes #5501

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
Seems to work as expected locally so I'm marking it as ready for review!

scout/adapter/mongo/case.py
```diff
@@ -1391,131 +1386,43 @@ def update_variant_actions(self, institute_obj, case_obj, old_eval_variants):
...
-            "manual_rank": [],
-            "dismiss_variant": [],
-            "mosaic_tags": [],
-            "cancer_tier": [],
-            "acmg_classification": [],
-            "ccv_classification": [],
-            "is_commented": [],
-        }
-
-        update_action_map = {
```
dnil, line 1404: 👍

scout/adapter/mongo/case.py
```diff
@@ -1391,131 +1386,43 @@ def update_variant_actions(self, institute_obj, case_obj, old_eval_variants):
...
+        ACTION_KEYS = [
+            "manual_rank",
+            "dismiss_variant",
+            "mosaic_tags",
+            "cancer_tier",
+            "acmg_classification",
+            "ccv_classification",
+            "is_commented",
+        ]
+        updated_variants = {action: [] for action in ACTION_KEYS}
```
dnil, line 1398: 💯

scout/adapter/mongo/case.py
```diff
@@ -1391,131 +1386,43 @@ def update_variant_actions(self, institute_obj, case_obj, old_eval_variants):
...
-                    continue
-
-                user_obj = self.user(old_event["user_id"])
-                if user_obj is None:
-                    continue
-
-                updated_variant = None
-
-                if action == "is_commented":
-                    updated_comments = self.comments_reupload(
```
dnil, line 1468: I believe that may have been the only call for that function, so we can probably retire it as well?

tests/adapter/mongo/test_case_handling.py
```diff
@@ -1221,12 +1215,11 @@ def test_keep_variant_comments_after_reupload(
...
-        institute_obj=institute_obj,
         case_obj=case_obj,
         old_eval_variants=[old_variant],
     )
 
     assert updated_new_vars["is_commented"] == [new_variant["_id"]]
 
-    # and 2 new comments should be created in the database
-    assert sum(1 for _ in adapter.event_collection.find()) == 4
+    # and np new comments should be created in the database
```
dnil:
```suggestion
    # and no new comments should be created in the database
```

scout/adapter/mongo/case.py
```diff
@@ -1372,15 +1370,12 @@ def case_sanger_variants(self, case_id):
 
         return case_verif_variants
 
-    def update_variant_actions(self, institute_obj, case_obj, old_eval_variants):
+    def update_variant_actions(
```
dnil, line 1373: 🥇

dnil (approved):
This seems sound. The events are not removed upon a case delete anyway, and we have no other reliable way of reconnecting them if they don't already match well enough.

Makes me feel all the more we should move from marking anything in the variant documents to having that state in a separate collection, but that is another story.

Note the `comments_reupload()` function - I'm pretty sure we can retire it as well! I was first thinking maybe we had no "is_commented" properties, but there appear to be. It was perhaps there only to deal with case level comments?

scout/adapter/mongo/case.py
northwestwitch: I think you are right!

northwestwitch: Removed

PR #5509, opened 2025-05-28
Fix #5506 - dim more cases if no specific case query given

### LINKED ISSUE #5506: Highlighting of cases
Try to be as clear and precise as possible, adding sufficient detail and context.

Hi 

In my cases list (cust004), in the list of the prioritized cases (8 at the moment) I see the last 5 highlighted in grey when there is no search query active. From what I understood from the last update, the grey highlighting should be only on cases that do not match a specific search string (for example if I search for a specific case number among the prioritized ones (8 -in my case-) I get 7 highlighted in grey and only the one matching the search string "not grey"). 

Could you clarify how this work? Thanks!

/Marco

dnil:

- Fix #5506

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. go to an institute, note the cases displayed by default. 
2. ensure you have some prioritised case
3. if no prioritised case is highlighted, select a default case from the list below and prioritise it
4. If no prioritised case is dimmed, increase the limit slightly from the default (100 -> 101) to get an extra case. Mark that case prioritised.
5. go back to the normal limit (100). You should now have one prioritised case dimmed, the others "highlighted" - i e displaying as per usual, showing the issue
6. apply patch
7. note that all prio cases now appear dimmed

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
-###
+### Added
 - Add cancer SNVs to Oncogenicity ClinVar submissions (downloadable json document only) (#5449)
+### Fixed
+- More clearly dim cases for empty queries (#5507)
 
 ## [4.102]
 ### Added
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -603,11 +603,14 @@ def get_and_set_cases_by_status(
             case_groups[status].append(case_obj)
             nr_cases_showall_statuses += 1
 
-    def get_specific_query(request):
+    def get_specific_query(request: request) -> bool:
         """Check if only (non-)specific query terms were used in query.
         If so we assume this is a default query, and dim all cases
         that match the "show_all_cases_status", highlighting the
         (max limit number) other cases that were returned.
+
+        If this is a specific query, cases returned by this query part will be highlighted even
+        if they have a status that matches "show_all_cases_status".
         """
         cases_query = store.cases(
             collaborator=institute_obj["_id"],
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -621,7 +621,7 @@ def get_and_set_cases_by_status(
         )
         LOG.warning(f"Case query was {cases_query}")
         for key, value in cases_query.items():
-            if key not in ["collaborators"] and value not in [None, ""]:
+            if key not in NONSPECIFIC_QUERY_TERMS and value not in [None, ""]:
                 return False
         return True
 
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -620,7 +620,7 @@ def get_and_set_cases_by_status(
             yield_query=True,
         )
         LOG.warning(f"Case query was {cases_query}")
-        for key, value in cases_query():
+        for key, value in cases_query.items():
             if key not in ["collaborators"] and value not in [None, ""]:
                 return False
         return True
--- a/scout/server/blueprints/institutes/controllers.py
+++ b/scout/server/blueprints/institutes/controllers.py
@@ -571,17 +571,6 @@ def get_cases_by_query(
         projection=ALL_CASES_PROJECTION,
     )
 
-    cases_query = store.cases(
-        collaborator=institute_id,
-        name_query=name_query,
-        skip_assigned=request.form.get("skip_assigned"),
-        is_research=request.form.get("is_research"),
-        has_rna_data=request.form.get("has_rna"),
-        verification_pending=request.form.get("validation_ordered"),
-        has_clinvar_submission=request.form.get("clinvar_submitted"),
-        yield_query=True,
-    )
-    LOG.warning(f"Case query was {cases_query}")
     return all_cases
 
 
@@ -620,7 +609,18 @@ def get_and_set_cases_by_status(
         that match the "show_all_cases_status", highlighting the
         (max limit number) other cases that were return.
         """
-        for key, value in form.items():
+        cases_query = store.cases(
+            collaborator=institute_obj["_id"],
+            name_query=request.form,
+            skip_assigned=request.form.get("skip_assigned"),
+            is_research=request.form.get("is_research"),
+            has_rna_data=request.form.get("has_rna"),
+            verification_pending=request.form.get("validation_ordered"),
+            has_clinvar_submission=request.form.get("clinvar_submitted"),
+            yield_query=True,
+        )
+        LOG.warning(f"Case query was {cases_query}")
+        for key, value in cases_query():
             if key not in ["collaborators"] and value not in [None, ""]:
                 return False
         return True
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,8 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
-###
+### Added
```
northwestwitch: 🥇

scout/server/blueprints/institutes/controllers.py
```diff
@@ -598,13 +603,37 @@ def get_and_set_cases_by_status(
             case_groups[status].append(case_obj)
             nr_cases_showall_statuses += 1
 
+    def get_specific_query(request):
```
northwestwitch: Could you add type hints?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -598,13 +603,37 @@ def get_and_set_cases_by_status(
...
+            verification_pending=request.form.get("validation_ordered"),
+            has_clinvar_submission=request.form.get("clinvar_submitted"),
+            yield_query=True,
+        )
+        for key, value in cases_query.items():
+            if key not in NONSPECIFIC_QUERY_TERMS and value not in [None, ""]:
+                return True
+        return False
+
+    specific_query_asked = get_specific_query(request)
```
northwestwitch, line 633: Type hint here would help also

scout/server/blueprints/institutes/controllers.py
```diff
@@ -598,13 +603,37 @@ def get_and_set_cases_by_status(
...
+            collaborator=institute_obj["_id"],
+            name_query=request.form,
+            skip_assigned=request.form.get("skip_assigned"),
+            is_research=request.form.get("is_research"),
+            has_rna_data=request.form.get("has_rna"),
+            verification_pending=request.form.get("validation_ordered"),
+            has_clinvar_submission=request.form.get("clinvar_submitted"),
+            yield_query=True,
+        )
+        for key, value in cases_query.items():
```
northwestwitch, line 622: What is happening here, is this checking that cases key are not in ["collaborators", " ", ""]. Why?

northwestwitch (commented):
Sorry, perhaps it's Monday morning, but I feel I need a bit more context for reviewing this PR. I understand from your comment on #5506 that the issue is the following:

The shadowed high prio cases are grey because they are returned after the default number of 100 default cases is reached. I understand this, I think. But I don't get this solution? Could you comment on it?

dnil:
> Sorry, perhaps it's Monday morning, but I feel I need a bit more context for reviewing this PR. I understand from your comment on #5506 that the issue is the following:
>
> The shadowed high prio cases are grey because they are returned after the default number of 100 default cases is reached. I understand this, I think. But I don't get this solution? Could you comment on it?

Sure, it is indeed Monday however, so don't expect too much! The docstring for the new function was supposed to do just this! 😂 

The original PR worked well in regards to highlighting query results, and differentiates well enough between those returned by the direct query and extra prio ("show all") cases. The user confounding edge cases seen in the issue are from the "default" query (name query empty, no extra boxes checked - and `collaborators` set to current institute) that the cases page does whenever you land there without doing an explicit search.

The solution here introduces an intermediate step that checks if the query is just a default query, or a more specific one. It then uses this information to decide if specific, use the previous behaviour (highlight selectively depending on query) or if non-specific blindly keep all the cases matching the "show all" categories dimmed. Hence no odd mix of highlighted and dimmed prio cases on a blank query.

scout/server/blueprints/institutes/controllers.py
```diff
@@ -598,13 +603,37 @@ def get_and_set_cases_by_status(
             case_groups[status].append(case_obj)
             nr_cases_showall_statuses += 1
 
+    def get_specific_query(request):
+        """Check if only (non-)specific query terms were used in query.
+        If so we assume this is a default query, and dim all cases
+        that match the "show_all_cases_status", highlighting the
+        (max limit number) other cases that were returned.
+        """
+        cases_query = store.cases(
```
northwestwitch: Why are you running this query again, you should already have all cases returned by the exact same query, see [here](https://github.com/Clinical-Genomics/scout/blob/be77892e1e2ce15ac095affeb484a6d917b1fabe/scout/server/blueprints/institutes/controllers.py#L638)

dnil: Ah, I am not running the query again, just returning the query `yield_query=True`. (Another option would be to save and ask the cursor for the query, but we only return case lists as it is.)

northwestwitch: Ahhh OK!

scout/server/blueprints/institutes/controllers.py
dnil: Added at the function call. Should be enough for most editors methinks?

scout/server/blueprints/institutes/controllers.py
northwestwitch: This was the source of my confusion. Could we add at least a type hint here.  A comment saying that this variable doesn't contain the cases, but the query itself could be useful for future maintenance

scout/server/blueprints/institutes/controllers.py
```diff
@@ -598,13 +603,40 @@ def get_and_set_cases_by_status(
...
+            name_query=request.form,
+            skip_assigned=request.form.get("skip_assigned"),
+            is_research=request.form.get("is_research"),
+            has_rna_data=request.form.get("has_rna"),
+            verification_pending=request.form.get("validation_ordered"),
+            has_clinvar_submission=request.form.get("clinvar_submitted"),
+            yield_query=True,
+        )
+        for key, value in cases_query.items():
+            if key not in NONSPECIFIC_QUERY_TERMS and value not in [None, ""]:
```
northwestwitch, line 629:
```suggestion
            if key != "collaborators" and value not in [None, ""]:
```

We can fix in the future if we have something else to add to NONSPECIFIC_QUERY_TERMS?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -51,6 +51,10 @@
     "cancel_sanger",
 ]
 
+NONSPECIFIC_QUERY_TERMS = [
```
northwestwitch, line 55: I would remove this variable if it has only one element. See also other comment down

dnil: I was honestly expecting to find more examples, so sure. I do feel this is an opportunity to name something to remember why later, rather than add a comment, but it doesn't matter to me today.

scout/server/blueprints/institutes/controllers.py
dnil: Hm, I'll check the docstring for the function. I'm not a fan of the current `adapter.cases()` construction, but then that is a pretty complex beast anyway. As it is it can return three different types (a query dict, a sorted list, or a sorted Cursor). If it was just me, I would refactor refactor to have one function for the sorted list, one for the Cursor and then both calling the yield query one, which could also be called separately if need be. And before we drop 3.9 we can't use `|` for multiple return type hints. 😞 But you are right, we can add a type hint to the call at least!

PR #5513, opened 2025-06-04
On variant SNVs search, display multiple HGVS for multigenic variants

### LINKED ISSUE #5512: Variant search, display annotation for all transcripts
Hi,

When searching for variants in a particular gene using the "Search SNVs and SVs" function, variant annotation is displayed in the table under 'Var' and in the downloaded file under 'HGVS'. However, the annotation is only displayed for one transcript. If the variant is annotated to two different genes, variant information is only displayed for one gene - the first gene, not necessarily the gene where the variant is in coding sequence. Example:

Image
In this example, we see one variant that is only in _PEX6_ and the annotation is displayed. This is followed by two variants that are annotated to both _GNMT_ and _PEX6_ and for those only the _GNMT_ annotation is shown. 

Would it be possible to add the annotation for one transcript per gene?

Thank you!

northwestwitch:
- Closes #5512 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/app.py
+++ b/scout/server/app.py
@@ -282,6 +282,31 @@ def register_filters(app):
         return cosmicId
 
     @app.template_filter()
+    def format_variant_canonical_transcripts(variant):
+        """Formats canonical transcripts for all genes in a variant."""
+
+        def truncate_string(s, length):
+            if s and len(s) > length:
+                return s[: length - 3] + "..."
+            return s
+
+        output_lines = []
+
+        for gene in variant.get("genes", []):
+            for tx in gene.get("transcripts", []):
+                if tx.get("is_canonical"):
+                    line = f"{gene.get('canonical_transcript', '')} ({gene.get('hgnc_symbol', '')})"
+                    hgvs = gene.get("hgvs_identifier")
+                    if hgvs:
+                        line += " " + truncate_string(hgvs, 20)
+                    protein = tx.get("protein_sequence_name")
+                    if protein:
+                        line += " " + truncate_string(protein, 20)
+                    output_lines.append(line)
+
+        return output_lines
+
+    @app.template_filter()
     def upper_na(string):
         """
         Uppercase ocurences of "dna" and "rna" for nice display.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ## [unreleased]
 ### Added
 - Add cancer SNVs to Oncogenicity ClinVar submissions (downloadable json document only) (#5449)
+### Changed
+- On `Search SNVs and SVs` page, display multiple HGVS descriptors when variant has mpre than one gene ()
 ### Fixed
 - Instance badge class and config option documentation (#5500)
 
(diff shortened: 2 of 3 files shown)


northwestwitch:
### Before:

image

### After:

image

scout/server/app.py
```diff
@@ -281,6 +284,33 @@ def cosmic_prefix(cosmicId):
...
+
+        lines = set()
+        genes = variant.get("genes") or []
+
+        for gene in genes:
+            transcripts = gene.get("transcripts") or []
+            for tx in transcripts:
+                if not tx.get("is_canonical"):
+                    continue
+                canonical_tx = tx.get("transcript_id")
```
northwestwitch, line 324: I am collecting canonical transcripts from the list of transcripts and not from the gene at the level above because of this bug -> #5515

scout/server/app.py
```diff
@@ -281,6 +284,33 @@ def cosmic_prefix(cosmicId):
...
+            if hgvs:
+                line_components.append(hgvs)
+            if protein:
+                line_components.append(protein)
+
+            lines.add(" ".join(line_components))
+
+        return list(lines)
+
+    app.custom_filters.format_variant_canonical_transcripts = format_variant_canonical_transcripts
```
northwestwitch, line 337: This is to be able to reuse the function also in the python code

dnil: Right, clever! It is a slightly complex solution. I suppose there was something blocking not moving the filter code to a more internally native module or class that could be called from the custom filter function?

dnil (approved):
Approving since it is working, but the solution is a bit complex. I don't quite see the reason for not picking an easier solution, but I also have not been trying to write it.

scout/server/app.py
northwestwitch: It might look complex, I agree, but it's a way to be able to invoke the function from both the python classes and the jinja2. We have a bunch of places where we have duplicated code in jinja and python, so I was envisioning that we could just do the same and add to app.custom_filters - given that we remember that the option exists! :)

scout/server/blueprints/institutes/controllers.py
```diff
@@ -865,7 +865,9 @@ def generate(header, lines):
         )  # CADD score
         variant_line.append(" | ".join(variant.get("region_annotations", [])))  # Region
         variant_line.append(" | ".join(variant.get("functional_annotations", [])))  # Function
-        variant_line.append(variant.get("hgvs", "-"))  # HGVS
+        variant_line.append(
+            " | ".join(current_app.custom_filters.format_variant_canonical_transcripts(variant))
```
northwestwitch, line 869: See, here I am reusing the filter but in the python controller

scout/server/app.py
northwestwitch: For instance on the variant page, we have the list of transcripts which might even be bugged at the moment. We could reuse the code in there possibly

PR #5602, opened 2025-07-25
Treat panel analyses as WES when getting chanjo2 stats

northwestwitch:
- Panel data doesn't have associated panels at the moment and when d4 files are provided for these cases the chanjo2 stats are computed as if they were WGS cases, resulting in very low coverage stats. Treating these cases as WES (still incorrect) would help improving the stats 
- This PR has to be used in a combo with a fix for this issue on chanjo2 -> https://github.com/Clinical-Genomics/chanjo2/issues/488, otherwise the stats won't change. But I guess it can be fixed in scout anyway because if you provide a gene list in the request then it would work perfectly

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Test with [this case](https://scout-stage.scilifelab.se/cust087/F0043044-2)

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Parsing variant's`local_obs_cancer_somatic_panel_old` and `local_obs_cancer_somatic_panel_old_freq`from `Cancer_Somatic_Panel_Obs` and `Cancer_Somatic_Panel_Frq` INFO keys respectively (#5594)
 - Filter cancer variants by archived number of cancer somatic panel observations (#5598)
+### Changed
+- Compute chanjo2 coverage on exons only when at least case individual has analysis_type=panel (#5601)
 ### Fixed
 - Treat -1 values as None values when parsing archived LoqusDB frequencies (#5591)
 
--- a/scout/server/blueprints/cases/templates/cases/chanjo2_form.html
+++ b/scout/server/blueprints/cases/templates/cases/chanjo2_form.html
@@ -23,7 +23,7 @@
     {% endif %}
   {% endfor %}
 
-  {% if "wes" in analysis_types %}
+  {% if "wes" in analysis_types or "panel" in analysis_types %}
     {% set interval_type = "exons" %}
   {% elif "wts" in analysis_types %}
     {% set interval_type = "transcripts" %}


northwestwitch:
FIY @mathiasbio

mathiasbio:
Nice! 🌟 This intuitively sounds like a good solution from my perspective. It is at least better than interpreting the panels as WGS. But I have not discussed this feature with users at all, and I don't know what's possible from chanjo2. I still think this is an upgrade and should be fine to deploy 🤔 but maybe we can discuss it a bit more.

I will make an educated guess that a feature that the users may want is for instance if they have a case where they can't find any interesting variant, to ensure that the regions that are covered in their panel are not below some certain coverage threshold. For instance they might search for certain genes and see that a region of geneX where they typically see a variant isn't covered above 100X, which may prompt them to ask for a sequencing top up or something.

Maybe ideally the regions that are checked for coverage should also be covered by the panel, because sometimes not all exons may be included from a gene and the same regions would pop up all the time. 

Is there anything like this that one can get from chanjo2? I actually think I see something like that here: https://github.com/Clinical-Genomics/chanjo2?tab=readme-ov-file maybe it's in fact a feature that would be activated if it was treated as WES 🤔

northwestwitch:
> Maybe ideally the regions that are checked for coverage should also be covered by the panel, because sometimes not all exons may be included from a gene and the same regions would pop up all the time.
> 
> Is there anything like this that one can get from chanjo2?

There is something like this already in chanjo2 but it doesn't create a report as the one that you see from scout, here's the documentation: https://clinical-genomics.github.io/chanjo2/usage/sample-coverage/#direct-coverage-query-over-the-intervals-present-in-a-bed-file. Instead of a web page you get a json-like coverage and completeness stats. If there is interest it could be converted into a web report, I guess.

dnil (approved):
Sure; why not. It isn't really a given that a panel is for exons only - quite a few are for full transcripts as well. It would be ideal to track both the actual genes used (e.g. add gene panels to the "panel" type case, perhaps derived from the actual panel kit used - the LIMS knows) and information on wether the targets are exon only or whole gene.

northwestwitch:
> Sure; why not. It isn't really a given that a panel is for exons only - quite a few are for full transcripts as well. It would be ideal to track both the actual genes used (e.g. add gene panels to the "panel" type case, perhaps derived from the actual panel kit used - the LIMS knows) and information on wether the targets are exon only or whole gene.

I know, it's far from ideal as it is

Issue #5601, opened 2025-07-25
Refactor how loqusdb instances are used on variantS filter


northwestwitch:
Instead of hard coding the instances used on variantS filters, collect them from the settings and create the form fields interactively. That would allow to filter for customised instances of loqus, not necessarily named as the instances in Solna.

PR #5622, opened 2025-08-19
Add flexibility to the parsing of custom archived loqusDB instances

northwestwitch:
- A partial fix to #5601
- Removes hard coded archived loqusDB instance names from the parsing module, allowing to define custom keys
- Next step (another PR) would be using these keys from the config file to define custom filters on the variantS pages

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Display inheritance models on structural variantS pages (#6015)
 - Analysis types wgs-lr and panel-lr (#6029)
 - Option to specify database name for the loqusdb API in the scout config file (#6028)
+- Allowing filtering cases by type of analysis (advanced options) (#6045)
 ### Changed
 - Institutes are now ordered alphabetically by display name on gene panels search (#5965)
 - Display all available individual/sample IDs for a case (display_name, individual_id and subject_id) directly on cases page (#5966)
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -162,6 +162,10 @@ class CaseHandler(object):
         case_ids = [case["_id"] for case in cases_with_gene_doc]
         self._update_case_id_query(query, case_ids)
 
+    def set_analysis_types_query(self, query: Dict[str, Any], query_value: str):
+        """Set individual.analysis_type for case query."""
+        query["individuals.analysis_type"] = {"$in": query_value.split(",")}
+
     def _set_case_name_query(self, query: Dict[str, Any], query_value: str):
         """Set case query to reg exp search in case and individual display names for parts of the name query."""
 
@@ -231,6 +235,7 @@ class CaseHandler(object):
         def set_case_item_query(query: dict, query_field: str, query_value: str):
             # Mapping query fields to their corresponding handling functions
             handlers = {
+                "analysis_type": self.set_analysis_types_query,
                 "case": self._set_case_name_query,
                 "exact_pheno": self._set_case_phenotype_query,
                 "exact_dia": self._set_diagnosis_query,
@@ -249,6 +254,7 @@ class CaseHandler(object):
             }
 
             handler = handlers.get(query_field)
+
             if handler:
                 handler(query, query_value)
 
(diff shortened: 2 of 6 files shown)


scout/server/config.py
```diff
@@ -87,6 +87,22 @@
...
+# Comment out if no archived LoqusDB instance is annotated in your VCF files
+LOQUSDB_ARCHIVE_VCF_INFO_FIELDS = {
+    # single keys (map VCF INFO key → target field in parsed_variant)
+    "Obs": {"field": "local_obs_old", "type": int},
+    "Hom": {"field": "local_obs_hom_old", "type": int},
+    "Frq": {"field": "local_obs_old_freq", "type": float},
+    "clinical_genomics_loqusObs": {"field": "local_obs_old", "type": int},
+    "clinical_genomics_loqusFrq": {"field": "local_obs_old_freq", "type": float},
+    "clin_obs": {"field": "local_obs_old", "type": int},
+    # prefixes → auto-expand into *_Obs, *_Hom, *_Frq
```
northwestwitch: Is this too complex to have in the config file? Would it be better with a long list of single keys instead?

dnil: Hm, yes, it is rather complex isn't it? I like the idea though. Perhaps we could leave the old ones in running code, or as a similar structure from constants, and allow adding only the systematic prefix ones from config?

northwestwitch: I like the idea of removing hard coded keys as much as possible. Perhaps we could try to just have a list (no matter how long) of keys (containing prefixes and suffixes). That would be very clear

northwestwitch:
I want to revise this PR and make sure I access the config values after the config file has been parsed, otherwise I might be skipping the situation where the config is the yaml file used for instance by our cli. Still i progress!

scout/server/config.py
```diff
@@ -87,6 +87,16 @@
 #    "loqus_api" : {"api_url": "http://127.0.0.1:9000"},
 # }
 
+# Define here the VCF keys you want to be parsed for eventual custom archived LoqusDB instances.
+# Comment out if no archived LoqusDB instance is annotated in your VCF files
+LOQUSDB_ARCHIVE_VCF_INFO_FIELDS = {
```
dnil, line 92: How about including this, mirroring the current behaviour, as a fallback in constants? And then override with this from the config file if it exists. That way the whole system doesn't break from missing to add this dict.

scout/parse/variant/variant.py
```diff
@@ -215,44 +216,34 @@ def set_sv_specific_annotations(parsed_variant: dict, variant: dict):
...
+    These keys are mostly used for different instances of Cancer (Balsamic) Germline and Somatic loqus.
+    RD germline (for MIP and Balsamic) SNVs contain INFO field Obs, SVs contain clinical_genomics_loqusObs.
     """
 
     def safe_val(val):
         """Convert -1 to None, leave other values unchanged."""
         return None if val == -1 else val
 
     info = variant.INFO
+    VCF_INFO_FIELDS = getattr(config, "LOQUSDB_ARCHIVE_VCF_INFO_FIELDS", {})
```
dnil, line 239:
```suggestion
    VCF_INFO_FIELDS = getattr(config, "LOQUSDB_ARCHIVE_VCF_INFO_FIELDS", LOQUSDB_ARCHIVE_VCF_INFO_FIELDS_DEFAULT)
```

scout/server/config.py
```diff
@@ -87,6 +87,16 @@
...
+    "Obs": {"field": "local_obs_old", "type": int},
+    "Hom": {"field": "local_obs_hom_old", "type": int},
+    "Frq": {"field": "local_obs_old_freq", "type": float},
+    "clinical_genomics_loqusObs": {"field": "local_obs_old", "type": int},
+    "clinical_genomics_loqusFrq": {"field": "local_obs_old_freq", "type": float},
```
dnil, line 97: Plus the Cancer_Germline, etc!

northwestwitch:
> I'm not entirely sure about the wisdom in making the config more complex, but given a fallback constant equivalent to what we do currently I guess it would be ok.

Never mind this PR, It needs some more thinking and should definitely NOT be included in the upcoming release. Sorry, I should have converted it into a draft!

northwestwitch:
I'm closing this one. I'm fed up to see it around!

PR #5623, opened 2025-08-19
Add GT check to overlapping genes search

### LINKED ISSUE #5621: Parent variant in SV page "Gene overlapping variants" box
Thank you for taking the time to contribute an issue!

**Describe the bug**

The gene overlapping variants box on the SV variant page seems to be able to show variants from parents in trios.

**To Reproduce**

Hard to reproduce as this was initially found in a clinical dataset. In that case, four overlapping variants showed up, where only two was present in proband. The highest scoring was present in one of the parents.

**Expected behavior**

Only variants from the proband to show up.

**Additional info**

I cross-checked with the original VCF. The call is indeed only present in one of the parents.

northwestwitch:
- Closes #5621 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -702,12 +702,24 @@ class VariantHandler(VariantLoader):
         if not limit:
             limit = 30 if variant_obj["category"] == "snv" else 45
 
+        variant_carriers = [
+            s["sample_id"] for s in samples if re.search(CARRIER, s["genotype_call"])
+        ]
+
         query = {
             "$and": [
                 {"case_id": variant_obj["case_id"]},
                 {"category": category},
                 {"variant_type": variant_type},
                 {"hgnc_ids": {"$in": hgnc_ids}},
+                {
+                    "samples": {
+                        "$elemMatch": {
+                            "sample_id": {"$in": variant_carriers},
+                            "genotype_call": {"$regex": CARRIER},
+                        }
+                    }
+                },
             ]
         }
         sort_key = [("rank_score", pymongo.DESCENDING)]
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -703,7 +703,7 @@ class VariantHandler(VariantLoader):
             limit = 30 if variant_obj["category"] == "snv" else 45
 
         variant_carriers = [
-            s["sample_id"] for s in samples if re.search(CARRIER, s["genotype_call"])
+            s["sample_id"] for s in variant_obj["samples"] if re.search(CARRIER, s["genotype_call"])
         ]
 
         query = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Exporting managed variants from the command line with the `--collaborator` option will return variants from the specified institute plus those not assigned to any institute (#5607)
 - Safer redirect to previous page for variants views (#5599)
 - Make whole caseS row clickable link for case page (#5620)
+- Restrict gene-overlapping variants search to variants affecting the same individuals as the original variant (#5623)
 ### Fixed
 - Treat -1 values as None values when parsing archived LoqusDB frequencies (#5591)
 - Links to SNVs and SVs from SMN CN page (#5600)


scout/adapter/mongo/variant.py
```diff
@@ -689,6 +689,12 @@ def delete_variants(self, case_id, variant_type, category=None):
         result = self.variant_collection.delete_many(query)
         LOG.info("{0} variants deleted".format(result.deleted_count))
 
+    def get_variant_carriers(self, variant_obj: dict) -> List[str]:
```
dnil: Hm, I was impressed with the idea, but slowly starting to doubt when we will really benefit from this? If the idea is to mask variants from unaffected individuals, we are worse off than just asking for affected carriers. If the idea is that we should be able to start from an unaffected carrier and ask for other variants in the same gene, again hits to the phenotypically affected seem like the most relevant ones. The remaining cases would be looking for variants that are poorly genotyped, but then we basically need the original query without restrictions anyway?

scout/adapter/mongo/variant.py
```diff
@@ -702,12 +708,22 @@ def get_variants_hgnc_overlapping(
...
 
+        variant_carriers = self.get_variant_carriers(variant_obj)
+
         query = {
             "$and": [
                 {"case_id": variant_obj["case_id"]},
                 {"category": category},
                 {"variant_type": variant_type},
                 {"hgnc_ids": {"$in": hgnc_ids}},
+                {
```
dnil, line 719: Liking this better right now: https://github.com/Clinical-Genomics/scout/blob/0bfb415f6de01a8bea2f355e63830f03aade1348/scout/adapter/mongo/variant.py#L494

dnil (changes requested):
Hm, please argue the case, but I feel if we are to restrict matching based on gt, we are better off only showing carriership for affected individuals.

scout/adapter/mongo/variant.py
northwestwitch: I don't know, what you suggest seems quite restrictive to me. We are not even restricting the variantS lists to the affected only (you can see variants that are only in the unaffected if you select `Include variants present only in unaffected`). In that case we need to add an explanation why if I start from an unaffected sample I just return variants in genes from affected individuals.
What if an affected for quality reasons doesn't have the call? Or perhaps the sample was marked as not affected but later turns out is affected? But I don't know, I'm not a user. Any suggestion @Jakob37

dnil: I also do not know. Pragmatically I think either would be good enough to solve the issue while retaining operation, but let's try to do the best we can when changing it.

Let's reason. You present three cases. I'll start with the only easy one.
-  If the affected status changes, we need to recompute a lot of things, so this to me is absolutely fine to assume correct for the use case.
- The missing call is a real and present issue, and part of why we were very permissive from the get-go. But it is also seems hard to cover for while at the same time not showing potential carriership for other genes overlapping in e.g. a large structural variant. Still, if it is not there, it is somewhat simple to explain why it is not there. And one could use the same argument against your version with first checking the carrier status as called for the originating variant - what about the `./.` calls there, perhaps they are actually carriers and should be shown.
- As for starting from an unaffected carrier, I kind of agree, but also feel it would be somewhat out of the scope of the investigation to look for other variants for the same gene. The reasonable explanation again is suspecting missing calls from the affected individuals. Or a very non-standard investigation, which we should then probably think through more explicitly. Regardless, starting from the same carrier profile as the original variant doesn't seem like the solution there either? I guess it would mostly do the same as affecteds for the typical culprit large SVs: they will normally only be found in one individual. But for say compounds to an SNV, it seems a bit odd to limit to exactly the same pattern as the variant, eg ones only found inherited from the mother? One would rather have the compound MEI from the father listed. (And again, if the call is missing from the affected, well, fine, it is missing then.)

northwestwitch: Ok, I don't mind changing this to return only hits in the affected. But then we need to fix the web pages to clearly explain that I think, with a tooltip or something. Let's see..

northwestwitch:
Ready for another round as well!

scout/adapter/mongo/variant.py
```diff
@@ -495,7 +496,7 @@ def match_affected_gt(self, case_obj, positional_variant_ids, limit_genes):
             {"samples": {"$size": 1}},  # Condition for samples with exactly one element
             {
                 "samples": {
-                    "$elemMatch": {  # Condition for samples with more than one element: individual/sample should be carrier
+                    ELEM_MATCH: {  # Condition for samples with more than one element: individual/sample should be carrier
```
dnil, line 499: 🤣 It is just the linter not understanding the context...

tests/adapter/mongo/test_query.py
```diff
@@ -961,32 +961,50 @@ def test_query_svs_by_coordinates_bnds(adapter, case_obj):
     assert list(results)
 
 
-def test_get_overlapping_variant(real_variant_database, case_obj, variant_obj, sv_variant_obj):
+def test_get_overlapping_variant(real_variant_database, variant_obj, sv_variant_obj):
```
dnil, line 964: 🎖️

scout/server/blueprints/variants/templates/variants/components.html
```diff
@@ -442,7 +442,7 @@
       data-bs-content="<div>Gene overlapping non-SNVs</div>
       {{ overlapping_tooltip_table(institute, case, variant.overlapping[:20]) }}">Gene overlapping (DNA)</a>
     {% else %}
-      data-bs-content="<div>Gene overlapping variants</div>
+      data-bs-content="<div>Gene overlapping variants in affected individuals</div>
```
dnil, line 445: 💯

scout/server/blueprints/variant/controllers.py
```diff
@@ -372,7 +372,9 @@ def variant(
 
     overlapping_variants, overlapping_outliers = [], []
     if get_overlapping:
-        overlapping_variants, overlapping_outliers = map(list, store.hgnc_overlapping(variant_obj))
+        overlapping_variants, overlapping_outliers = map(
```
dnil, line 375: Right, the outliers: we need to rig `get_omics_variants_hgnc_overlapping()` the same way, don't we?

scout/adapter/mongo/variant.py
dnil: Yes, thank you - this makes sense to me given that we are to impose limits here. I guess gene list matching might have been nice as well, but it should be ok to leave that. They carry the variant of origin anyway.

dnil (commented):
Nice! But I just realised we have the omics variants overlapping in a separate function.. 😊

northwestwitch:
> Nice! But I just realised we have the omics variants overlapping in a separate function.. 😊

I know, do these variants contain exactly the same sample IDs as the DNA ones? to do the matching..

scout/server/blueprints/variant/controllers.py
northwestwitch: Let's do it in a separate PR, if possible

northwestwitch:
> > Nice! But I just realised we have the omics variants overlapping in a separate function.. 😊
> 
> I know, do these variants contain exactly the same sample IDs as the DNA ones? to do the matching..

Ah there was this key to do the linking:

image

Let's have this check on another PR, it's easier

scout/server/blueprints/variant/controllers.py
dnil: 🤷 sure!

PR #5634, opened 2025-08-22
Fix #5586 - TRGT_ID loci gets late genes.hgnc_symbol population, and hence strange reload behaviour

### LINKED ISSUE #5586: Weird sorting of STRs first 2 times that the page loads
To reproduce:
 
1. Go to a nallo case (build 38)
2. Go to STRs
3. Notice that some gene names are even missing 
4. Reload, notice that the sorting is messy
5. Reload and all is good, and sorting is fine

dnil:
- Fix #5586 

Ok, so three-four steps here:
- [x] Firstly, we fix the display, by changing the fallback order for the symbol. `variant.hgnc_ids` and `variant.hgnc_symbols` are properly parsed for loci that are annotated with an `HNGCId` by `Stranger`, which is the case for current `Nallo` runs. This will be useful for new cases that are already uploaded as well.
Now `variant.genes` appears repopulated later when discovered missing on variants page load, and not used until later.
- [x] So, secondly, we allow the update function to work a bit more aggressively and do more updates in one reload
- [ ] Third, we should really do the "update" already at the parsing step. No reason to wait other than perhaps not to have an extra function for the same thing. Here https://github.com/Clinical-Genomics/scout/blob/9af4dbc1e7febfa9115e96ea3171d0e5861ed7a1/scout/parse/variant/variant.py#L140 or maybe better here https://github.com/Clinical-Genomics/scout/blob/9af4dbc1e7febfa9115e96ea3171d0e5861ed7a1/scout/parse/variant/variant.py#L628-L627.
- [ ] So fourth, maybe, maybe, we could refactor to merge at least a couple of these functions.

For now this kind of solves the issue already with steps 1 and 2 anyone will be hard pressed to find a use case where the variants will not simply be displayed/populated in time. But they will arguably not be fully "finished" until the second reload (compared to fourth before), so we should probably fix it. Just not Friday evening. 😝 

Here is testing with a local case reload, first on main:
❌ no symbols on the loci (will resolve after reloading, converging after four)
Screenshot 2025-08-22 at 17 48 18
✅ with this PR, looks fine from the start, though behind the scenes the db objects will not be fully updated until the second reload (or all variants have been individually visited)
Screenshot 2025-08-22 at 17 47 11

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -551,7 +551,7 @@ def update_variant_genes(store, variant_obj, genome_build):
             variant_obj["genes"].append(variant_gene)
         variant_genes = variant_obj.get(
             "genes"
-        )  # we need update variant_genes here, to get all fixes in one call to the funcion
+        )  # we need update variant_genes here, to get all fixes in one call to the function
 
     if variant_genes is not None:
         for gene_obj in variant_genes:
--- a/scout/server/blueprints/variants/controllers.py
+++ b/scout/server/blueprints/variants/controllers.py
@@ -549,6 +549,9 @@ def update_variant_genes(store, variant_obj, genome_build):
             has_changed = True
             variant_gene = {"hgnc_id": hgnc_id}
             variant_obj["genes"].append(variant_gene)
+        variant_genes = variant_obj.get(
+            "genes"
+        )  # we need update variant_genes here, to get all fixes in one call to the funcion
 
     if variant_genes is not None:
         for gene_obj in variant_genes:
--- a/scout/server/blueprints/variants/templates/variants/str-variants.html
+++ b/scout/server/blueprints/variants/templates/variants/str-variants.html
@@ -245,8 +245,12 @@
     {%endif%}
   </div>"
   title="">{% if variant.str_repid %} {{ variant.str_repid }}
+  {% elif variant.hgnc_symbols %}
+    {% for hgnc_symbol in variant.hgnc_symbols %} {{ hgnc_symbol }} {% endfor %}
   {% elif variant.genes %}
     {% for gene in variant.genes %} {{ gene.symbol or gene.hgnc_symbol }} {% endfor %}
+  {% elif variant.hgnc_ids %}
+    {% for hgnc_id in variant.hgnc_ids %} {{ hgnc_id }} {% endfor %}
   {% else %}
     {{ variant.str_trid }}
   {% endif %}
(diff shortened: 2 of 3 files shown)


scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -245,8 +245,12 @@
     {%endif%}
   </div>"
   title="">{% if variant.str_repid %} {{ variant.str_repid }}
+  {% elif variant.hgnc_symbols %}
+    {% for hgnc_symbol in variant.hgnc_symbols %} {{ hgnc_symbol }} {% endfor %}
   {% elif variant.genes %}
     {% for gene in variant.genes %} {{ gene.symbol or gene.hgnc_symbol }} {% endfor %}
+  {% elif variant.hgnc_ids %}
```
dnil, line 252: Just in case.

scout/server/blueprints/variants/templates/variants/str-variants.html
```diff
@@ -245,8 +245,12 @@
     {%endif%}
   </div>"
   title="">{% if variant.str_repid %} {{ variant.str_repid }}
+  {% elif variant.hgnc_symbols %}
+    {% for hgnc_symbol in variant.hgnc_symbols %} {{ hgnc_symbol }} {% endfor %}
```
dnil, line 249: This will solve the main issue, nothing being shown for trgt loci, which will have an empty variant.genes... Just checking for empty variant.genes would have been nice, but this is better.

scout/server/blueprints/variants/controllers.py
```diff
@@ -549,6 +549,9 @@ def update_variant_genes(store, variant_obj, genome_build):
             has_changed = True
             variant_gene = {"hgnc_id": hgnc_id}
             variant_obj["genes"].append(variant_gene)
+        variant_genes = variant_obj.get(
+            "genes"
+        )  # we need update variant_genes here, to get all fixes in one call to the funcion
```
dnil: and yes, here is the slightly more agressive fix-them-in-the-same-go.

scout/server/blueprints/variants/controllers.py
```diff
@@ -549,6 +549,9 @@ def update_variant_genes(store, variant_obj, genome_build):
             has_changed = True
             variant_gene = {"hgnc_id": hgnc_id}
             variant_obj["genes"].append(variant_gene)
+        variant_genes = variant_obj.get(
+            "genes"
+        )  # we need update variant_genes here, to get all fixes in one call to the function
```
dnil, line 554: And here is the slightly more aggressive fix-them-in-the-same-go. No reason not to.

PR #5681, opened 2025-09-04
Fix #5679 - update case owner on `scout load case --update`

### LINKED ISSUE #5679: Pinned variant linkout to incorrect institute?!
**Describe the bug**
When linking to a pinned variant on `civilsole` on `stage`, we are somehow pointing to the wrong institute. The event link is fine, and the variant can be found on `cust000`. However the link is 

Image

`cust059` is on `collaborators`, but not the `owner`. Let's look at the link generation; suspect this is an admin only "feature"?

dnil:
OR

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
+## [unreleased]
+
+### Fixed
+- Allow updating case owner on `scout load case -u`
+
 ## [4.104]
 ### Added
 - Parsing variant's`local_obs_cancer_somatic_panel_old` and `local_obs_cancer_somatic_panel_old_freq`from `Cancer_Somatic_Panel_Obs` and `Cancer_Somatic_Panel_Frq` INFO keys respectively (#5594)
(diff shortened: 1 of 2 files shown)


dnil:
> This fixes the link but you end up with the same case on both institutes, only from the initial institute all links to variants point to the case on inst2.
> 
> I wonder if it wouldn't be better to remove the case from the initial institute then?

I think what you mean is that the updated case still has the original institute as a collaborator? We have previously opted to only add collaborators for a case with the update from file. This can be practical, as it does not force users to repeat WUI addition of collaborators on a case update. 

I feel we should leave this as-is, but do think we should flag more clearly when a case is "only" shared, not owned for an institute. This in turn is a tad complicated, since users can have access to multiple institutes, and navigate to the same case from two different routes..

northwestwitch:
> I think what you mean is that the updated case still has the original institute as a collaborator? We have previously opted to only add collaborators for a case with the update from file. This can be practical, as it does not force users to repeat WUI addition of collaborators on a case update.
> 
> I feel we should leave this as-is, but do think we should flag more clearly when a case is "only" shared, not owned for an institute. This in turn is a tad complicated, since users can have access to multiple institutes, and navigate to the same case from two different routes..

Right! I didn't go that far with this reasoning and didn't remember that we had a discussion about this. But then I think this PR is fine to merge! 👍🏻

PR #5691, opened 2025-09-10
Parse all ClinVar annotations when VEP has been run with `--check_existing`

### LINKED ISSUE #5690: Missing ClinVar information when running VEP with `--check_existing`
**Describe the bug**

Not entirely sure whether this is a bug or if it's something that I've overlooked when annotating variants with VEP. The issue arises when I annotate a VCF with VEP (113) and use the `--existing_variation` flag together with a custom ClinVar annotation:

```bash
vep \
  --check_existing \
  --custom file=clinvar_20250831.vcf.gz,short_name=CLINVAR,format=vcf,type=exact,coords=0,fields=CLNSIG%CLNREVSTAT
# other arguments omitted
```

This results in a VCF that has the annotations from existing known variation (dbSNP IDs, COSMIC, etc) and the custom ClinVar annotations (`CSQ` keys `CLINVAR` (accession), `CLINVAR_CLNSIG`, and `CLINVAR_CLNREVSTAT`). The problem is that using `--existing_variation` also adds the `CSQ` key `CLIN_SIG` which contains the ClinVar clinical significance from the VEP cache. This key seems to be prioritised when parsing, with an early return in `parse_clnsig`:

https://github.com/Clinical-Genomics/scout/blob/843ca0f651c5ab0e8b4eedee1e636e4030b3fd76/scout/parse/variant/clnsig.py#L81-L97

The `"clnsig"` key in the transcript ultimately comes from the [`CLIN_SIG` field in `CSQ`](https://github.com/Clinical-Genomics/scout/blob/843ca0f651c5ab0e8b4eedee1e636e4030b3fd76/scout/parse/variant/transcript.py#L124-L126). This then results in entries where I only have the clinical significance and not the accession number nor the review status, even though that information is present in the VCF.

Skipping `--check_existing` doesn't seem to be a viable option since this will result in losing annotations that we want.

**To Reproduce**

1. Annotate a VCF with VEP using `--check_existing` and a `--custom` ClinVar annotation
2. Load the case/VCF into Scout
3. Navigate to a variant that has ClinVar information
4. See that the ClinVar accession number and review status is missing

Image

**Expected behavior**

Full Clinvar information should be present on the variant page if the variant has these annotations in the VCF. I would expect the example above to look something along these lines:

Image

Alternatively something like this:

Image

It looks like the second example could be achieved by simply removing the early return in `parse_clnsig`.

maehler:

This PR addresses an issue that arises when running VEP with the `--check_existing` flag that affects ClinVar annotations. See #5690 for details.

What has been done is to remove the early return from `parse_clnsig` in [scout/parse/clnsig.py](https://github.com/Clinical-Genomics/scout/blob/843ca0f651c5ab0e8b4eedee1e636e4030b3fd76/scout/parse/variant/clnsig.py#L58). In order to support ClinVar annotations using an unaltered ClinVar VCF, the transcript parsing has also been modified to look both for `CLINVAR_CLNVID` as well as just `CLINVAR` among the VEP annotations.

With a VCF file that has been annotated with both `--check_existing` and an unmodified ClinVar VCF file, there will be at least one entry among the ClinVar annotations that has an accession number and review status, if the variant has those annotations. If there are annotations from the VEP cache in the key `CLIN_SIG`, then these will be presented as well, but not have any accession number nor review status.

Fixes #5690.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**Review:**
- [x] code approved by CR
- [x] tests executed by maehler

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1202,7 +1202,6 @@ def one_vep104_annotated_variant(request, vep_104_annotated_variant_clinical_fil
 
 @pytest.fixture(scope="function")
 def one_vep113_existing_variation_variant(request, vep_113_existing_variation_file):
-    LOG.info("Return one parsed variant")
     variant_parser = VCF(vep_113_existing_variation_file)
 
     variant = next(variant_parser)


maehler:
I have to run now, so I'll wrap this up tomorrow (i.e. at least fixing the changelog). Let me know if there's any issues with this or if I need to clarify anything.

maehler:
I can confirm that this results in what I expect on the variant page:
image

northwestwitch:
> I can confirm that this results in what I expect on the variant page:

Nice! I'll review

tests/conftest.py
```diff
@@ -1199,6 +1200,15 @@ def one_vep104_annotated_variant(request, vep_104_annotated_variant_clinical_fil
     return variant
 
 
+@pytest.fixture(scope="function")
+def one_vep113_existing_variation_variant(request, vep_113_existing_variation_file):
+    LOG.info("Return one parsed variant")
```
northwestwitch:
```suggestion
```

maehler:
It was a while ago since I contributed anything, so I don't remember what your merging strategy is. Do you want me to squash it?

northwestwitch:
> It was a while ago since I contributed anything, so I don't remember what your merging strategy is. Do you want me to squash it?

Yes, go ahead! :)

PR #5832, opened 2025-11-03
Fix #5827 - WTS outliers clinical filter not active

### LINKED ISSUE #5826: Tomte results table does not contain adjusted p-value
Hi!
We realized just now that for the recent Tomte runs, our results files for fraser and outrider do not contain the adjusted p-value anymore (it is just 1.0 for all the calls). Not sure since when this occured, but I can dig into it if necessary.
Kind regards,
Malin

### LINKED ISSUE #5827: Clinical filter on WTS outliers not working properly?
See e.g. `enormoustoucan`. A quick filter for padjust or p_adjust_gene less than 1 returns 3+3 variants, whereas clicking clinical filter returns >1000 variants.

dnil:

- Fix #5827
- Fix #5826 - at least the Scout side of the issue

Actually added them to the form as well:
Screenshot 2025-11-04 at 10 44 29

Screenshot 2025-11-04 at 10 42 33

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - GnomAD links for the STR variant page (#5823)
 - `get_gene_complete_coverage` breaking for genes with no intervals on chanjo2, which is returning 'NA' instead of numerical stats (#5830)
 - Cleaner managed variant VCF export and detailed error messages for variants which are not included (#5813)
+- Chromosome order for hg38 in IGV.js view (#5838)
 - Clinical filter for outlier variants (#5832)
 - Decipher query was treated as a primary - now secondary again (#5832)
 
--- a/scout/constants/igv_tracks.py
+++ b/scout/constants/igv_tracks.py
@@ -48,6 +48,7 @@ HUMAN_REFERENCE_38 = {
     "indexURL": HG38REF_INDEX_URL,
     "cytobandURL": HG38CYTOBAND_URL,
     "aliasURL": HG38ALIAS_URL,
+    "chromosomeOrder": "chr1, chr2, chr3, chr4, chr5, chr6, chr7, chr8, chr9, chr10, chr11, chr12, chr13, chr14, chr15, chr16, chr17, chr18, chr19, chr20, chr21, chr22, chrX, chrY, chrM",
 }
 
 # igv.js track settings common for all users and all cases. Selectable by users
--- a/scout/server/blueprints/alignviewers/controllers.py
+++ b/scout/server/blueprints/alignviewers/controllers.py
@@ -396,6 +396,7 @@ def set_common_tracks(display_obj, build):
 
     # Set up IGV tracks that are common for all cases:
     display_obj["reference_track"] = HUMAN_REFERENCE[build]  # Human reference is always present
+    display_obj["chromosomeOrder"] = HUMAN_REFERENCE.get("chromosomeOrder")
 
     # if user settings for igv tracks exist -> use these settings, otherwise display default tracks ---> Genes, ClinVar and ClinVar CNVs
     custom_tracks_names = (
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_sashimi_viewer.html
@@ -50,7 +50,8 @@
                         fastaURL: "{{ reference_track.fastaURL }}",
                         indexURL: "{{ reference_track.indexURL }}",
                         cytobandURL: "{{ reference_track.cytobandURL }}",
-                        aliasURL: "{{ reference_track.aliasURL }}"
+                        aliasURL: "{{ reference_track.aliasURL }}",
+                      {% if reference_track.chromosomeOrder %}chromosomeOrder: "{{ reference_track.chromosomeOrder }}",{% endif %}
                     },
                 locus: '{{ locus }}',
                 tracks:
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
@@ -49,7 +49,8 @@
                         fastaURL: "{{ reference_track.fastaURL }}",
                         indexURL: "{{ reference_track.indexURL }}",
                         cytobandURL: "{{ reference_track.cytobandURL }}",
-                        aliasURL: "{{ reference_track.aliasURL }}"
+                        aliasURL: "{{ reference_track.aliasURL }}",
+                        {% if reference_track.chromosomeOrder %}chromosomeOrder: "{{ reference_track.chromosomeOrder }}",{% endif %}
                     },
                     locus: "{{locus}}",
                     tracks: [


scout/constants/filters.py
```diff
@@ -33,23 +33,8 @@
 }
 
 CLINICAL_FILTER_BASE_OUTLIER = {
-    "$and": [
```
dnil, line 36: This never worked. Perhaps we should allow a passthrough to a direct db query, but as it is the clinical filter uses the build query function. That allows it to be a bit more controlled, and similar regardless of how the query options are set. But also quite a bit more complex.

scout/adapter/mongo/query.py
```diff
@@ -206,86 +206,80 @@ def build_query(
         # set up the fundamental query params: case_id, category, type and
         # restrict to list of variants (if var list is provided)
         for criterion in FUNDAMENTAL_CRITERIA:
-            if criterion == "case_id":
```
dnil, line 209: For the most part just switching (pun intended) to match-case, and removing some debug statements. Maybe a table of function calls would be even better, but one thing at a time.

scout/adapter/mongo/query.py
```diff
@@ -206,86 +206,80 @@ def build_query(
...
-                coordinate_query = None
-                if category in ["snv", "cancer"]:
-                    mongo_query["chromosome"] = query.get("chrom")
-                    if query.get("start") and query.get("end"):
-                        self.coordinate_filter(query, mongo_query)
-                else:  # sv
-                    coordinate_query = [self.sv_coordinate_query(query)]
-                continue
-
-            if criterion == "variant_ids" and variant_ids:
```
dnil, line 265: we only need to do this once. 😄

scout/adapter/mongo/query.py
```diff
@@ -206,86 +206,80 @@ def build_query(
...
-                LOG.debug("Adding variant_ids %s to query" % ", ".join(variant_ids))
-                mongo_query["variant_id"] = {"$in": variant_ids}
-                continue
-
-            # Do not retrieve dismissed variants if hide_dismissed checkbox is checked in filter form
-            if criterion == "hide_dismissed" and query.get(criterion) is True:
-                mongo_query["dismiss_variant"] = {"$in": [None, []]}
-
-            gt_query = _get_query_genotype(query)
```
dnil, line 274: Use walrus, avoiding extra calls when it's not the time for the gt criteria.

scout/adapter/mongo/query.py
```diff
@@ -308,7 +302,7 @@ def build_query(
         # such as a Pathogenic ClinSig.
 
         if secondary_terms is True:
-            secondary_filter = self.secondary_query(query, mongo_query)
+            secondary_filter = self.secondary_query(query)
```
dnil, line 300: So decipher was handled a bit awkwardly. It is in the list of secondary criteria, but was added to the db query on more or less a primary level if it ever got there.

This function had mongo_query precisely for making decipher more like a primary criterion. It should be just fine as a secondary, where it is also listed. This is arguably a change worth considering though - we could also add decipher among the primary, with clinvar, where I believe it was originally. It is not used much I believe - e.g. not in clinical sv filter.

scout/adapter/mongo/query.py
```diff
@@ -572,289 +566,326 @@ def gene_filter(self, query, build="37"):
 
         return list(hgnc_ids)
 
-    def secondary_query(self, query, mongo_query, secondary_filter=None):
```
dnil, line 575: secondary_filter was never used.

scout/adapter/mongo/query.py
```diff
@@ -572,289 +566,326 @@ def gene_filter(self, query, build="37"):
...
 
+        We first set the sample genotype call fiter if the somewhat mixed fundamental criterion "show unaffected"
+        is given.
         """
         LOG.debug("Creating a query object with secondary parameters")
 
         mongo_secondary_query = []
 
-        # loop over secondary query criteria
+        if query.get("show_unaffected") and (gt_query := _get_query_genotype(query)):
```
dnil: Moved this up before the loop, instead of running it every iteration.

scout/adapter/mongo/query.py
```diff
@@ -572,289 +566,326 @@ def gene_filter(self, query, build="37"):
...
-                gnomad = query.get("gnomad_frequency")
-                mongo_secondary_query.append(
-                    {
-                        "$or": [
-                            {"gnomad_frequency": {"$lt": float(gnomad)}},
-                            {"gnomad_frequency": NOT_EXISTS},
-                        ]
-                    }
-                )
+            match criterion:
```
dnil, line 593: match-case here as well

scout/adapter/mongo/query.py
```diff
@@ -572,289 +566,326 @@ def gene_filter(self, query, build="37"):
...
-
-                size_query = {
-                    "$or": [
-                        {"$expr": {size_selector: [{"$abs": "$length"}, size]}},
-                        {"length": NOT_EXISTS},  # Include documents where 'length' is missing
-                    ]
-                }
-
-                mongo_secondary_query.append(size_query)
+                case "decipher":
```
dnil, line 714: Here is the decipher change

scout/adapter/mongo/query.py
```diff
@@ -572,289 +566,326 @@ def gene_filter(self, query, build="37"):
...
-            if criterion == "p_value":
-                p_value = query.get("p_value")
-                mongo_secondary_query.append({"p_value": {"$lt": p_value}})
+                case "p_value":
+                    p_value = query.get("p_value")
+                    mongo_secondary_query.append(
+                        {"p_value": {"$lt": p_value}},
+                    )
+
+                case "padjust" | "p_adjust_gene":
```
dnil: And here are the new ones finally.

tests/adapter/mongo/test_query.py
```diff
@@ -590,8 +590,7 @@ def test_build_has_cosmic_ids(
 
     mongo_query = adapter.build_query(case_id, query=query)
 
-    assert {"cosmic_ids": {"$exists": True}} in mongo_query["$and"]
-    assert {"cosmic_ids": {"$ne": None}} in mongo_query["$and"]
+    assert {"cosmic_ids": {"$exists": True, "$ne": None}} in mongo_query["$and"]
```
dnil, line 593: merged the two and items - implicit anyway

scout/adapter/mongo/query.py
```diff
@@ -206,86 +206,75 @@ def build_query(
...
-            if criterion == "hide_dismissed" and query.get(criterion) is True:
-                mongo_query["dismiss_variant"] = {"$in": [None, []]}
-
-            gt_query = _get_query_genotype(query)
-            if criterion == "show_unaffected" and query.get(criterion) is False:
-                self.affected_inds_query(mongo_query, case_id, gt_query)
-
-            if criterion == "show_soft_filtered" and query.get(criterion) is False:
-                self.soft_filters_query(query=query, mongo_query=mongo_query)
+            match criterion:
```
northwestwitch, line 209: That is familiar and nice!

scout/adapter/mongo/query.py
northwestwitch: Fine by me

dnil:
Regarding Decipher, I checked the MIP/raredisease references, and the file with these annotations is one with the deletions from DDD, here https://www.deciphergenomics.org/about/downloads/data. It is a population frequency reference, so nothing to flag up as if a true disease cause. We might consider having a count filter instead, if at all - the hg19 version hasn't been updated since 2015, and was old already when we converted it to svdb and started using it in 2017.

PR #5897, opened 2025-11-25
Add soft filters modulation

### LINKED ISSUE #5740: Allow a sub-selection of soft filters to apply to the variants
Image

Close to this checkbox, allow a selection of soft filter to apply.

northwestwitch:
- Closes #5740 - enable selection of single soft filters

### Before
- You don't get to choose which filters to exclude and they're all selected by default. Checkbox is not checked:

image

### After
- The exclude checkbox is pre-selected:
 
image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - On cases page, print a `RNA` label whenever WGS or WES cases have associated RNA data (#5886)
 - Save and display maintainer/institutes for managed variants (#5890)
 - Link to VuTR at the transcript level from the variant page (#5893)
+- Filter germline ClinVar submission IDs by ClinVar ID (#5907)
 ### Changed
 - A case becomes inactive whenever there is a change to any VCF or RNA-associated file (#5888)
 - Enhance variants filters by allowing selection of individual soft filters (#5897)
--- a/scout/adapter/mongo/clinvar.py
+++ b/scout/adapter/mongo/clinvar.py
@@ -262,14 +262,35 @@ class ClinVarHandler(object):
             "updated_at": result.get("updated_at"),
         }
 
-    def get_clinvar_onc_submissions(self, institute_id: str) -> pymongo.cursor.Cursor:
-        """Collect all open and closed ClinVar oncogenocity submissions for an institute."""
-        query = {"institute_id": institute_id, "type": "oncogenicity"}
-        return self.clinvar_submission_collection.find(query).sort("updated_at", pymongo.DESCENDING)
-
-    def get_clinvar_germline_submissions(self, institute_id: str) -> List[dict]:
+    def get_clinvar_germline_submissions(
+        self, institute_id: str, clinvar_id_filter: Optional[str] = None
+    ) -> List[dict]:
         """Collect all open and closed ClinVar germline submissions for an institute."""
+
+        def populate_cases_from_variant_data(variant_data, institute_id):
+            cases = {}
+            for var_info in variant_data:
+                case_id = var_info["_id"].rsplit("_", 1)[0]
+                CASE_CLINVAR_SUBMISSION_PROJECTION = {"display_name": 1}
+                var_info["added_by"] = self.clinvar_variant_submitter(
+                    institute_id=institute_id,
+                    case_id=case_id,
+                    variant_id=var_info["local_id"],
+                )
+                case_obj = self.case(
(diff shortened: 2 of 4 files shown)


scout/server/blueprints/variants/controllers.py
```diff
@@ -114,10 +114,11 @@ def populate_chrom_choices(form, case_obj):
     form.chrom.choices = [(chrom, chrom) for chrom in chromosomes]
 
 
-def populate_institute_soft_filters(form, institute_obj):
-    """Populate the hidden field 'institute_soft_filters' with a string containing all institute's soft filters."""
-    if institute_obj.get("soft_filters"):
-        form.institute_soft_filters.data = ",".join(institute_obj["soft_filters"])
+def populate_institute_soft_filters(form, institute_obj: dict):
```
dnil, line 117: 💯

scout/adapter/mongo/query.py
```diff
@@ -264,8 +264,8 @@ def build_query(
                     if query.get(criterion) is False:
                         self.affected_inds_query(mongo_query, case_id, _get_genotype_query(query))
 
-                case "show_soft_filtered":
-                    if query.get(criterion) is False:
+                case "exclude_soft_filtered":
```
dnil, line 267: We don't save this to filters, right?

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -117,16 +117,37 @@
...
+      <div class="d-flex align-items-center gap-3">
+        <!-- Checkbox -->
+        <div class="form-check flex-shrink-0 mb-0">
+            {{ form.exclude_soft_filtered(class="form-check-input") }}
+            {{ form.exclude_soft_filtered.label(
+                class="form-check-label",
+                style="white-space: nowrap;",
+                data_bs_toggle="tooltip",
+                title="Filters are defined by an admin in the institute settings. Current filters are: " ~
+                      form.institute_soft_filters.data|join(', ')|safe
```
dnil, line 129: Did you mean to show the current selection (is what happens now, note that the contents of data changes during user checking) or the selection that said admin has entered (original choices)?

scout/adapter/mongo/query.py
northwestwitch: You could. For instance I've tested locally:

Say you are on main branch - prod behaviour - and you saved a filter like this:

Image

Then you would find it on prod with the same query, but I didn't:

Image

When you try switching to this branch, locally, even if you load the old filter with "show_soft_filtered" it doesn't make the page crash. So we should be fine

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: It's fine I think. All filters are shown in the dropdown anyway. This tooltip showing just the selected ones is fine.

scout/adapter/mongo/query.py
dnil: Hm, I'm not sure I follow you fully? No-one seems to have saved `show_soft_filtered` on our end, and I guess it probably hasn't had time to catch on in other centers. But in principle then we would need a migration script or a fallback for others not to make this function breaking, even if it is not crashing?

scout/server/blueprints/variants/templates/variants/utils.html
dnil: As long as you know. 😸

scout/adapter/mongo/query.py
northwestwitch: > Hm, I'm not sure I follow you fully? No-one seems to have saved `show_soft_filtered` on our end, and I guess it probably hasn't had time to catch on in other centers. But in principle then we would need a migration script or a fallback for others not to make this function breaking, even if it is not crashing?

I just didn't want to over-complicate a code (the filters) that is already very complicated and hard to follow. I think it will be enough if we explain how this filter has changed in the next release email - saying also that pre-saved filter might be affected. I don't think many people are using these filters anyway, apart from our cancer people. And the main functionality - if you didn't save it within a filter - hasn't changed al all: it's just that before your filtered if you didn't check the checkobox, while now the checkbox is already checked and all filters are selected. Let's keep simple the things that are simple

PR #5903, opened 2025-11-27
Outliers query inclusive of missing fields

### LINKED ISSUE #5900: P adj filtering on wts page is non-inclusive
When filtering only with P adj for fraser, existence is required for inclusion, so that e.g. outrider hits are not shown. This is in contrast to much other functionality we have, where e.g. frequencies etc are filtered in if not existing. We already have a category filter that can choose only Expression. Let's try to make them absence-inclusive for consistency, even if it means writing the clinical filter query might be a bit more difficult.

northwestwitch:
- Fix #5900 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Refactored code for collecting and displaying STR variants, which now shares same routines as SV and MEI variants (#5895)
 - Deprecate reupload_sv (#5933)
 - Dual locus IGV view for BNDs with breaks on different chromosomes (#5939)
+- Lazier checking of variant load criteria: most costly query last (#5924)
+- Caching variant load criteria queries for considerable speedup for large variant files (#5924)
 ### Fixed
 - Fix PanelApp fetch failures caused by 429 responses - Retry-After support (#5864)
 - Parsing of `REVEL rank score` field for LRS from all transcript of a LRS variant (#5860)
--- a/scout/adapter/mongo/variant_loader.py
+++ b/scout/adapter/mongo/variant_loader.py
@@ -392,30 +392,37 @@ class VariantLoader(object):
         bulk = {}
         current_region = None
 
+        managed_variants_cache = self._cache_managed_variants(category, build)
+        causative_variants_cache = self._cache_causative_other_cases()
+
         LOG.info(f"Number of variants present on the VCF file:{nr_variants}")
         with progressbar(
             variants, label="Loading variants", length=nr_variants, file=sys.stdout
         ) as bar:
             for idx, variant in enumerate(bar):
                 # All MT variants are loaded
-                mt_variant = variant.CHROM in ["M", "MT"]
                 rank_score = parse_rank_score(variant.INFO.get("RankScore"), case_obj["_id"])
-                pathogenic = is_pathogenic(variant)
-                managed = self._is_managed(variant, category)
-                causative = self._is_causative_other_cases(variant, category)
 
                 # Check if the variant should be loaded at all
                 # if rank score is None means there are no rank scores annotated, all variants will be loaded
-                # Otherwise we load all variants above a rank score treshold
+                # Otherwise we load all variants above a rank score threshold
                 # Except for MT variants where we load all variants
                 if (
                     (rank_score is None)
                     or (rank_score > rank_threshold)
-                    or mt_variant
-                    or pathogenic
-                    or causative
-                    or managed
+                    or variant.CHROM in ["M", "MT"]
                     or category in ["str"]
+                    or is_pathogenic(variant)
+                    or self._is_managed(
+                        variant=variant,
+                        category=category,
+                        managed_variants_cache=managed_variants_cache,
+                    )
+                    or self._is_causative_other_cases(
+                        variant=variant,
+                        category=category,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Parsing of `REVEL rank score` field for LRS from all transcript of a LRS variant (#5860)
 - Tests for cli command to update VCF files not running (#5888)
 - Gene in panels search results broken layout - overflow when a panel has many panel versions containing the gene (#5899)
+- Include outliers with missing p-value fields in filtering logic ()
```
dnil:
```suggestion
- Include outliers with missing p-value fields in filtering logic (#5903)
```

scout/adapter/mongo/query.py
```diff
@@ -867,7 +867,7 @@ def _get_outlier_query(query: dict) -> dict:
                 }
 
             outlier_padjust_query["$or"].append(pval_struct)
-
+        outlier_padjust_query["$or"].append({pval_name: {"$exists": False}})
```
dnil: Yes, you are on the track. This solves the issue, but breaks the clinical filter.

Image

We will need to test for a condition and/or cross terms as well to make this work with the clinical filter (or in general filtering for both categories, as they will be mutually exclusive.

scout/adapter/mongo/query.py
```diff
@@ -866,7 +870,12 @@ def _get_outlier_query(query: dict) -> dict:
                     ]
                 }
 
-            outlier_padjust_query["$or"].append(pval_struct)
+            outlier_padjust_query["$or"].append(
+                {"$and": [pval_struct, {mutual_exclusion[pval_name]: {"$exists": False}}]}
```
northwestwitch, line 875: Enforce mutual exclusivity of the values: whenever a key/value is provided -> the other should't be existing, but if a key/value is not provided then the field might be missing - see line 878. Something like this @dnil? This is complicated 🤯

dnil: Yes, I recall we felt a bit lucky with the clinical filter, where missing values could be handled implicitly.

dnil: Better, but still not quite! Bugs with using just one or the other p value - on account of the `$and` they actually exclude, also when they should not (on single value query):
Image

dnil (changes requested):
Better, but still not quite! Bugs with using just one or the other p value.

northwestwitch:
@dnil it will take a while before I can check this one out. Please go ahead
and fix it if you want. Also my other PRs, I don't want to block anything 😄

dnil:
> @dnil it will take a while before I can check this one out. Please go ahead and fix it if you want. Also my other PRs, I don't want to block anything

No worries, and thanks, will do if we really need to get a release out before the holidays. I would prefer not to. The only potential blocker I see is the REVEL and similar weird CSQ entries with transcript number of entries per transcript. 🙄 But there is also a workaround on the pipeline side, using an option to the VEP plugin that ought to be best practice anyway.

northwestwitch:
@dnil I've been thinking about a fix to this PR - I totally understand the will to make it consistent with other queries - but thinking about maintenance (very hard query at the moment) and future applications of the WTS (#5932) I'm not really sure this is the direction we should pursue. We already have a nicely working clinical filter which is inclusive of all the missing fields, so why not keeping as it is, like you query for a fraser pvalue and you only get fraser variants?

dnil:
> @dnil I've been thinking about a fix to this PR - I totally understand the will to make it consistent with other queries - but thinking about maintenance (very hard query at the moment) and future applications of the WTS (#5932) I'm not really sure this is the direction we should pursue. We already have a nicely working clinical filter which is inclusive of all the missing fields, so why not keeping as it is, like you query for a fraser pvalue and you only get fraser variants?

Yes, mathematically terse, clever solutions like the one in the clinical filter tend to be difficult to represent generally.

If we can make it clear from the interface I would be fine with not being able to search for everything at once. (Although I feel this may slow down development of e.g. new limits for the clinical filter a little, it should still be possible to think that through with just an extra step.)

Question is how to visually note it. Maybe like auto selecting the corresponding outlier type in the dropdown, or disabling the opposite fields while something is in the other. But still make it open and work with the clinical filter.

northwestwitch:
> Yes, mathematically terse, clever solutions like the one in the clinical filter tend to be difficult to represent generally.
> 
> If we can make it clear from the interface I would be fine with not being able to search for everything at once. (Although I feel this may slow down development of e.g. new limits for the clinical filter a little, it should still be possible to think that through with just an extra step.)
> 
> Question is how to visually note it. Maybe like auto selecting the corresponding outlier type in the dropdown, or disabling the opposite fields while something is in the other. But still make it open and work with the clinical filter.

I'm not sure either. Perhaps yes, deactivate the fraser pvalue if you select an outrider one, and perhaps modify the hover over the filter specifying that only variants with these specific field will be returned

northwestwitch:
I think I'm going to close this PR for now then!

Issue #5792, opened 2025-10-20
User request: highlight case when it is updated with RNA data


northwestwitch:
"We often miss when the RNAseq has been uploaded to the case, as the original case stays archived. would it be possible for the case be re-activated and maybe add a tag WGT to the seq column?"

dnil:
Marking this QoL as we should always be getting delivery messages from the order via tickets, but also quite agree it is a large enough change to a case to warrant a status update to ”inactive”. We would surely have if it was updated via case load config.

northwestwitch:
This is a duplicate of #5870, closing as already completed

PR #5933, opened 2025-12-19
Fix #5885 - deprecate reupload sv

### LINKED ISSUE #5885: Deprecate reupload_svs
The `reupload_sv` functionality on the `scout update case` is as a bit of an oddball, purely there for historical reasons as far as I can tell. Let's deprecate it, remove it with The Next Major™️, and possibly refactor to make sure we have the same functionality for all variants with `update`/`load` combos.

dnil:

- Fix #5885 

```
  --reupload-sv                  Remove all SVs and re upload from existing
                                 files(DEPRECATED)
```

Screenshot 2025-12-19 at 15 04 15

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/commands/update/case.py
+++ b/scout/commands/update/case.py
@@ -202,9 +202,7 @@ def case(
         store.update_case_cli(case_obj, institute_obj)
 
     if reupload_sv:
-        LOG.info(
-            "Reupload-sv is deprecated and will be removed in Scout v5.0 in favor of separate needs check marking and generic variant type re-uploads."
-        )
+        LOG.info("The reupload-sv option is deprecated and will be removed in Scout v5.")
         LOG.info("Set needs_check to True for case %s", case_id)
         updates = {"needs_check": True}
         if sv_rankmodel_version:


scout/commands/update/case.py
```diff
@@ -201,6 +202,9 @@ def case(
         store.update_case_cli(case_obj, institute_obj)
 
     if reupload_sv:
+        LOG.info(
+            "Reupload-sv is deprecated and will be removed in Scout v5.0 in favor of separate needs check marking and generic variant type re-uploads."
+        )
```
northwestwitch:
```suggestion
        LOG.info(
           'Reupload-sv is deprecated and will be removed in Scout v5.0. Use "scout load variants --sv" instead'
        )
```

northwestwitch: Clearer, I think? You could also simply just expand the help of the command option like this: 
```
help=(
        "Deprecated. Remove all SVs and re-upload from existing files. "
        "Use 'scout load variants --sv' to achieve the same result."
    )
```

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Update GitHub workflows (#5935)
 - IGV.js to version 3.7.1 (#5931)
 - Refactored code for collecting and displaying STR variants, which now shares same routines as SV and MEI variants (#5895)
+- Deprecate reupload_sv (#5932)
```
northwestwitch:
```suggestion
- Deprecate reupload_sv (#5933)
```

scout/commands/update/case.py
dnil: Mm, thing is this is a special functionality with that needs_checking marking for cases. It was used for altering to and tracking progress in rechecking reuploaded variants after an error.

northwestwitch: > Mm, thing is this is a special functionality with that needs_checking marking for cases. It was used for altering to and tracking progress in rechecking reuploaded variants after an error. 

Ok. Would make the case inactive do, just like requested in #5792?

dnil: That will be a good generic. The particular functionality was to allow for a single specific additional thing to be checked - in this case some missed SVs. But if we keep it at all, better make it possible to mark cases separately. A combination of case status and case tags could probably do it: like make status inactive with a "technical attention", or automate adding a case comment. But all this is potential. How about I just skip the reasons in the message, and we decide on that when actually changing it?

northwestwitch: > How about I just skip the reasons in the message, and we decide on that when actually changing it?

Agreed! 👍🏻

PR #5939, opened 2026-01-07
Fix #5937 - allow IGV display of both BND ends if on different chr

### LINKED ISSUE #5937: Clicking a BND Length IGV button brings you to the wrong location
Thank you for taking the time to contribute an issue!

**Describe the bug**
Clicking a BND Length IGV button brings you to the wrong location.

**To Reproduce**
1. Go to 'a BND'
2. Click on 'Length IGV button'

**Expected behavior**
Is a split view of both breakpoints possible in igv.js? Then that would be nice. Otherwise, perhaps make the button unclickable for BNDs?

dnil:

- Fix #5937 
I was about to just hide the button again, since we can't easily trigger mate display. But, we can show multiple loci in IGV, so here goes. Opens multiple loci if a BND has a different `end_chrom`.

Looks ok on stage:
Screenshot 2026-01-07 at 16 51 14

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/scout/server/blueprints/alignviewers/controllers.py
+++ b/scout/server/blueprints/alignviewers/controllers.py
@@ -197,7 +197,7 @@ def make_igv_tracks(
                 f"chr{end_chromosome}:{stop}",
             ]
         else:
-            display_obj["locus"] = f"chr{chromosome}:{start}-{stop}"
+            display_obj["loci"] = [f"chr{chromosome}:{start}-{stop}"]
 
     # Set genome build for displaying alignments:
     if get_case_genome_build(case_obj) == "38" or chromosome == "M":
--- a/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
+++ b/scout/server/blueprints/alignviewers/templates/alignviewers/igv_viewer.html
@@ -52,7 +52,7 @@
                         aliasURL: "{{ reference_track.aliasURL }}",
                         {% if reference_track.chromosomeOrder %}chromosomeOrder: "{{ reference_track.chromosomeOrder }}",{% endif %}
                     },
-                    locus: {% if length(loci)>1 %}[{% for locus in loci %}"{{locus}}",{% endfor %}]{% else %} "{{ loci[0] }}"{% endif %},
+                    locus: {% if loci|length > 1 %}[{% for locus in loci %}"{{locus}}",{% endfor %}]{% else %} "{{ loci[0] }}"{% endif %},
                     tracks: [
                       {% for custTrack in custom_tracks %}
                         {
--- a/tests/server/blueprints/alignviewers/test_alignviewers_controllers.py
+++ b/tests/server/blueprints/alignviewers/test_alignviewers_controllers.py
@@ -116,7 +116,7 @@ def test_make_igv_tracks(app, case_obj, variant_obj):
         display_obj = controllers.make_igv_tracks(case_obj, variant_obj["_id"], "MT", 100, 101)
 
         # The function should return a track list with the expected tracks
-        assert display_obj["locus"] == "chrM:100-101"
+        assert display_obj["loci"] == ["chrM:100-101"]
         assert display_obj["display_center_guide"] == True
         assert display_obj["reference_track"]
         assert display_obj["custom_tracks"]


scout/server/blueprints/variant/templates/variant/utils.html
```diff
@@ -383,6 +383,15 @@ <h6>Phenotype terms (HPO):</h6>
             disabled" title="Alignment file(s) missing" data-bs-toggle="tooltip" aria-disabled="true"><span class="fa fa-times-circle fa-fw me-1"></span>IGV gDNA
           </a>
         </span>
+      {% elif variant.length == 100000000000 %}
```
northwestwitch:
```suggestion
      {% elif variant.chromosome != variant.end_chrom %}
```

If variant.end_chrom is always available, seems more professional? It doesn't really matter since the length thing is all over the place anyway..

scout/server/blueprints/alignviewers/controllers.py
```diff
@@ -187,7 +189,15 @@ def make_igv_tracks(
...
-        display_obj["locus"] = "chr{0}:{1}-{2}".format(chromosome, start, stop)
+
+        if end_chrom:
+            end_chromosome = end_chrom.replace("MT", "M")
+            display_obj["loci"] = [
+                f"chr{chromosome}:{start}",
+                f"chr{end_chromosome}:{stop}",
+            ]
+        else:
+            display_obj["locus"] = f"chr{chromosome}:{start}-{stop}"
```
northwestwitch: Perhaps for consistency this can become a loci array with just one element?

dnil:
> Looks good to me, test and codewise. One could argue that displaying 2 loci at the same time makes both windows narrower and more difficult to inspect, but we could try and see the feedback we get?

Yes. It is very similar to the default mode of analysis if you discover one of these en passant for SR. You would try to open the "mate" of a bridinging read and puzzle them together in parallel. For LRS we don't have mates, and IGV.js doesn't yet have "open supplemental aln" or whatever desktop IGV calls it when you do the same for long reads.

EDIT: it's these that are still missing.
Screenshot 2026-01-09 at 11 09 38


> The double button on variant page now becomes redundant:
> But since there are also the UCSC and genes links it's not like you can remove either, and I don't have any alternative solution to suggest.

Yes, kind of, but the same could be said for the length at all. How about we change the name on the button from length to "Split View" or something when this is available?

scout/server/blueprints/alignviewers/controllers.py
dnil: Sure, I can test - not sure how IGV.js handles it!

scout/server/blueprints/variant/templates/variant/utils.html
dnil: No, I thought the same, but opted for having the same test as a few lines above. I can change both as well!

dnil:
I guess it was already on their internal roadmap, but I noticed there was no issue for it so I asked the IGV team as well: https://github.com/igvteam/igv.js/issues/2062.

northwestwitch:
> Yes, kind of, but the same could be said for the length at all. How about we change the name on the button from length to "Split View" or something when this is available?

Ok!

scout/server/blueprints/alignviewers/controllers.py
dnil: Perfectly fine by the looks of it! Let's go with that.

dnil: Hm, wait, it did something with the colors of the mates.. Eh, probably not important, but I would like to keep it like this in case does something I don't quite follow right now. I changed to always pass an array named `loci` instead.

dnil:
> > Yes, kind of, but the same could be said for the length at all. How about we change the name on the button from length to "Split View" or something when this is available?
> 
> Ok!

Screenshot 2026-01-09 at 13 45 18

Something like this?

northwestwitch:
> 
> Screenshot 2026-01-09 at 13 45 18
> Something like this?

You mean you keep the 2 links to the gDNA the same and you include the new split button? I like it!

PR #5993, opened 2026-01-27
Metbat outliers

### LINKED ISSUE #5932: Add methbat methylation region calls as outliers
[MethBat](https://github.com/PacificBiosciences/MethBat) calls are soon in Nallo, as a tsv kind of similar to DROP RNA outliers. Will need a little research still what columns to save and display, but should be somewhat obvious from https://github.com/PacificBiosciences/MethBat/blob/main/docs/profile_guide.md#rare-methylation-analysis. 

Two kinds of outlier regions so far: `promoter` and imprinting (`nanoimprint`?), part of the `cpg_label`. 

Ideally there will be an HGNC id column, but if not the HGNC id and gene symbol can be parsed from the same label:
```
chr6	144007487	144008719	nanoimprint_PLAGL1_HGNC:9046
chr1	64419	65418	promoter_OR4F5_HGNC:14825
```

dnil:

- Fix #5932 

Screenshot 2026-01-29 at 15 49 24

Screenshot 2026-01-29 at 17 39 43

For a start, this PR presumes we will have `sample_id` as a column in https://github.com/Clinical-Genomics/MTP-NALLO/issues/88. It is intended to be agnostic of https://github.com/Clinical-Genomics/MTP-NALLO/issues/89, at least with regard to HGNC id being present as a separate column, or as part of the `cpg_label` - which is the current state.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN (local tests; files not in prod yet)

### DIFF
--- a/scout/adapter/mongo/omics_variant.py
+++ b/scout/adapter/mongo/omics_variant.py
@@ -182,10 +182,10 @@ class OmicsVariantHandler:
         case_id: str,
         query: Optional[dict] = None,
         category: Optional[str] = "outlier",
-        nr_of_variants=50,
-        skip=0,
+        nr_of_variants: Optional[int] = 50,
+        skip: Optional[int] = 0,
         projection: Optional[Dict] = None,
-        build: str = "37",
+        build: Optional[str] = "37",
     ):
         """Return omics variants for a case, of a particular type (clinical, research) and category (outlier, ...)."""
 
@@ -212,17 +212,14 @@ class OmicsVariantHandler:
         )
 
     def count_omics_variants(
-        self, case_id, query, variant_ids=None, category="outlier", build="37"
-    ):
-        """Returns number of variants
-
-        Arguments:
-            case_id(str): A string that represents the case
-            query(dict): A query dictionary
-
-        Returns:
-             integer
-        """
+        self,
+        case_id: str,
+        query: dict,
+        variant_ids: List[str] = None,
+        category: Optional[str] = "outlier",
+        build: Optional[str] = "37",
+    ) -> int:
+        """Returns the number of OMICs variants returned for a given query."""
 
         query = self.build_query(
             case_id, query=query, variant_ids=variant_ids, category=category, build=build
--- a/scout/models/omics_variant.py
+++ b/scout/models/omics_variant.py
@@ -220,16 +220,18 @@ class OmicsVariantLoader(BaseModel):
         """HGNC ids and gene symbols are found one on each line in DROP tsvs.
         Convert to a list with a single member in omics_variants for storage.
 
-        In case of MethBat the CPG_label is formed as nanoimprint_PLAGL1_HGNC:9046.
+        In case of MethBat we intend to have the same, but early versions will carry the gene symbol and HGNC id in
+        the CPG_label joined by underscore as nanoimprint_PLAGL1_HGNC:9046.
         """
+
+        if "hgncSymbol" in values:
+            if not isinstance(values.get("hgncSymbol"), list):
+                values["hgncSymbol"] = [str(values.get("hgncSymbol"))]
 
         if "hgncId" in values:
             values["hgncId"] = [int(values.get("hgncId"))]
         elif "hgnc_id" in values:
             values["hgncId"] = [int(values.get("hgnc_id"))]
-        elif "hgncSymbol" in values:
-            if not isinstance(values.get("hgncSymbol"), list):
-                values["hgncSymbol"] = [str(values.get("hgncSymbol"))]
         elif "cpg_label" in values:
             cpg_label = values.get("cpg_label").split("_")
             values["hgncSymbol"] = [cpg_label[1]]


scout/models/omics_variant.py
```diff
@@ -183,13 +215,19 @@ def ensure_end(cls, values):
...
 
         if "hgncId" in values:
             values["hgncId"] = [int(values.get("hgncId"))]
         elif "hgnc_id" in values:
             values["hgncId"] = [int(values.get("hgnc_id"))]
-
+        elif "cpg_label" in values:
+            cpg_label = values.get("cpg_label").split("_")
+            values["hgncSymbol"] = [cpg_label[1]]
+            values["hgncId"] = [cpg_label[2].split(":")[1]]
```
fellen31: Seems like it might have been a good thing to do this in the pipeline. 

Especially if we would like to be able to filter on clinical/research HGNCs before upload, or run with a GoS background file that might have e.g. `CpG:_104` as `cpg_label`.

dnil: Absolutely! I got the impression from Ine that was as far as you got? It would also be prudent to have the sample ID in the file, not just in the filename.

fellen31: Yes, this is indeed the current state. We can add it to the Nallo TODO.

fellen31: > It would also be prudent to have the sample ID in the file, not just in the filename.

Are you happy to have these files per sample, or would you rather have them per case if we add sample ID as a column?

dnil: Both are ok. We would be very used to joint case VCF files, but many of these "extras" are provided on a per sample basis anyway. I guess conceptually I would like more of a per-sample single analysis "outcome" column. I have a feeling if we want to pass more of the values for full cases, we will be straining the tsv format and perhaps better go to a VCF.

dnil: The DROP files went single sample per line. That is one way to avoid having a "FORMAT" type column.

dnil: Hm, and yes, they are per-case files, as are the other variant files, so it would be convenient.

scout/models/case/case_loading_models.py
```diff
@@ -444,7 +446,6 @@ class CaseLoader(BaseModel):
     somalier_ancestry: Optional[str] = None
     somalier_pairs: Optional[str] = None
     somalier_samples: Optional[str] = None
-    exe_ver: Optional[str] = None
```
dnil, line 447: Just duplicated, likely from an older sort. 🤷‍♂️

dnil:
I'll mark this ready for feedback at least. Feel free to check in @fellen31. I've only tested locally: demo files are included with the PR. It's no rush really, the files are still a little off from Nallo (missing sample id) and not stored and delivered through Hermes/cg. Does not have to be in this release, but could also safely be regardless of the current Nallo version in prod.

scout/adapter/mongo/omics_variant.py
```diff
@@ -180,12 +180,12 @@ def omics_variant(self, variant_id: str, projection: Optional[Dict] = None):
     def omics_variants(
         self,
         case_id: str,
-        query=None,
-        category: str = "outlier",
+        query: Optional[dict] = None,
+        category: Optional[str] = "outlier",
         nr_of_variants=50,
         skip=0,
```
northwestwitch: Add type hint to these ones as well, since all the other vars have it?

scout/commands/update/case.py
```diff
@@ -33,6 +33,14 @@
     "--outrider-research",
     help="Path to research WTS OMICS outlier OUTRIDER TSV file to be added - NB variants are NOT loaded",
 )
+@click.option(
+    "--methbat",
+    help="Path to clinical LRS OMICS outlier MethBat TSV file to be added - NB variants are NOT loaded",
```
northwestwitch: What are NB variants?

scout/commands/update/case.py
```diff
@@ -190,6 +202,10 @@ def case(
 
         case_obj["omics_files"][key_name] = key
         case_obj["has_outliers"] = True
+
+        if "meth" in key_name:
```
northwestwitch, line 206: Image

scout/models/omics_variant.py
```diff
@@ -241,4 +281,11 @@ def get_qualification(values: dict) -> str:
         qualification = "up" if float(values.get("zScore", 0)) > 0 else "down"
     if values.get("sub_category") == "splicing":
```
northwestwitch: What about having a bunch of elifs here?

scout/commands/update/case.py
fellen31: ✝️

dnil: 😄 Nota Bene variants! Someone has been away from Italy for a long time. 😉 I can rephrase - e.g. "please note" sounds better to me - but most likely it's the same on the other options, so I'd do all.

scout/models/omics_variant.py
dnil: 💯 Or better yet a `match` `case`!

scout/commands/update/case.py
northwestwitch: Right, it does ring a bell 🤣 But in my defense I would have immediately understood if written like this "nota bene:" , with colon, as to start introducing what you have no note well

dnil:
I hear a lot of text; changing to arrows and symbols for summary and comparison output, with a hover to show the text. The CPG label is useful to separate known imprinting region from promoter type calls currently. For the prior, one should expect ASM, but hyper/hypo is still a potential deviation to be considered. Most of the well established diagnostic methylation sites fall in the imprinting category, but it is of course interesting to know if we have something promoter related as well.

Screenshot 2026-02-23 at 12 04 41

northwestwitch:
> I hear a lot of text; changing to arrows and symbols for summary and comparison output, with a hover to show the text. The CPG label is useful to separate known imprinting region from promoter type calls currently. For the prior, one should expect ASM, but hyper/hypo is still a potential deviation to be considered. Most of the well established diagnostic methylation sites fall in the imprinting category, but it is of course interesting to know if we have something promoter related as well.
> 
> Screenshot 2026-02-23 at 12 04 41

Nice! What's the arrow to the right?

dnil:
> Nice! What's the arrow to the right?

Unchanged / data present, but not enough to warrant a clear up or down call. I'm very open to changing it if you have an idea...

northwestwitch:
> Unchanged / data present, but not enough to warrant a clear up or down call. I'm very open to changing it if you have an idea...

How about "=" or "==" ?

dnil:
> > Unchanged / data present, but not enough to warrant a clear up or down call. I'm very open to changing it if you have an idea...
> 
> How about "=" or "==" ?

Mm, yes, maybe &asymp; could be good?

northwestwitch:
> Mm, yes, maybe ≈ could be good?

Yes, I like it!

dnil:
Ok, I think I'll merge this as is now, and we can update from here when we start getting some files from new versions of Nallo.

PR #6027, opened 2026-02-09
Fix transcript overview, sometimes missing RefSeq ID and links

### LINKED ISSUE #6026: RefSeq transcript now shown as refseq tx and is missing links
On variant page, TX RefSeq transcripts (table on top).

This is a general problem it seems:
- RefSeq official transcripts are missing values in the `Refseq IDs` cell
- Same row is also missing transcript links

It looks like this:

Image

northwestwitch:
- Fix #6026 

### Main branch

image

### This branch

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Support for additional variant scoring systems: `MivmirScore`, `GicamScore` (#5970)
 - Parse and display `MivmirExplanation` key/values on variant page, when available (#6011)
 - Display inheritance models on structural variantS pages (#6015)
+- Analysis types wgs-lr and panel-lr (#6029)
 ### Changed
 - Institutes are now ordered alphabetically by display name on gene panels search (#5965)
 - Display all available individual/sample IDs for a case (display_name, individual_id and subject_id) directly on cases page (#5966)
--- a/docs/admin-guide/load-config.md
+++ b/docs/admin-guide/load-config.md
@@ -52,7 +52,7 @@ Below are available configuration parameters for a Scout case. Names marked with
 - **rna_delivery_report** _String_: Path to HTML RNA delivery report.
 - **samples** _List_ List of samples included in the case:
     - **alignment_path** _String_ Path to BAM/CRAM file to view alignments.
-    - **analysis_type** _String_ Specifies the analysis type for the sample. Options: {wgs, wes, panel, unknown, external}.
+    - **analysis_type** _String_ Specifies the analysis type for the sample. Options: {wgs, wgs-lr, wes, wts, panel, panel-lr, panel-umi, ogm, unknown, external}.
     - **assembly_alignment_path** _String_ Path to BAM/CRAM file to view de novo assembly alignments.
     - **bam_file** _String_ Path to BAM/CRAM file to view alignments **WARNING:** Soon to be deprecated, use *alignment_path*.
     - **bam_path** _String_ Path to BAM/CRAM file to view alignments **WARNING:** Soon to be deprecated, use *alignment_path*.
--- a/scout/constants/case_tags.py
+++ b/scout/constants/case_tags.py
@@ -1,4 +1,16 @@
-ANALYSIS_TYPES = ("external", "mixed", "ogm", "panel", "panel-umi", "unknown", "wes", "wgs", "wts")
+ANALYSIS_TYPES = (
+    "external",
+    "mixed",
+    "ogm",
+    "panel",
+    "panel-lr",
+    "panel-umi",
+    "unknown",
+    "wes",
+    "wgs",
+    "wgs-lr",
+    "wts",
+)
 
 # keys are used in the load report cli, while key_name is saved in case database documents
 CUSTOM_CASE_REPORTS = {


scout/server/blueprints/variant/controllers.py
```diff
@@ -98,20 +99,23 @@ def tx_overview(variant_obj: dict):
             ovw_tx["mane"] = tx.get("mane_select_transcript", "")
             ovw_tx["mane_plus"] = tx.get("mane_plus_clinical_transcript", "")
 
-            ovw_tx["decorated_refseq_ids"] = []
-            ovw_tx["muted_refseq_ids"] = []
+            ovw_tx["decorated_refseq_ids"] = set()
+            ovw_tx["muted_refseq_ids"] = set()
+
+            if not tx.get("transcript_id", "").startswith("ENST"):
```
northwestwitch: This is the fix to the RefSeq ID missing in the RefSeq IDs cell

scout/server/blueprints/variant/controllers.py
```diff
@@ -98,20 +99,23 @@ def tx_overview(variant_obj: dict):
...
                 if ovw_tx["mane"] and ovw_tx["mane"].startswith(refseq_id):
                     decorated_tx = ovw_tx["mane"]
                 elif ovw_tx["mane_plus"] and ovw_tx["mane_plus"].startswith(refseq_id):
                     decorated_tx = ovw_tx["mane_plus"]
                 elif refseq_id.startswith("XM"):
-                    ovw_tx["muted_refseq_ids"].append(refseq_id)
+                    ovw_tx["muted_refseq_ids"].add(refseq_id)
                     continue
-                else:
+                elif not tx.get("transcript_id", "").startswith("ENST"):
```
northwestwitch: Not adding an Ensembl ID in the RefSeq IDs cell any more

scout/server/blueprints/variant/utils.py
```diff
@@ -156,9 +156,6 @@ def update_transcripts_information(
         if hgnc_transcript.get("is_primary"):
             transcript["is_primary"] = True
 
-        # Add the transcript links
-        add_tx_links(transcript, genome_build, hgnc_symbol)
```
northwestwitch, line 160: This was adding links only to transcripts with transcript_id that are Ensembl IDs

scout/server/blueprints/variant/controllers.py
```diff
@@ -126,13 +130,18 @@ def tx_overview(variant_obj: dict):
             ovw_tx["coding_sequence_name"] = tx.get("coding_sequence_name")
             ovw_tx["protein_sequence_name"] = tx.get("protein_sequence_name")
 
+            # Add transcript links
+            add_tx_links(ovw_tx, genome_build, ovw_tx["hgnc_symbol"])
```
northwestwitch: Instead add links to any transcript that ends up in that transcript overview table

scout/server/blueprints/variant/controllers.py
```diff
@@ -98,20 +98,23 @@ def tx_overview(variant_obj: dict):
             ovw_tx["mane"] = tx.get("mane_select_transcript", "")
             ovw_tx["mane_plus"] = tx.get("mane_plus_clinical_transcript", "")
 
-            ovw_tx["decorated_refseq_ids"] = []
-            ovw_tx["muted_refseq_ids"] = []
+            ovw_tx["decorated_refseq_ids"] = set()
+            ovw_tx["muted_refseq_ids"] = set()
```
dnil: 💯

scout/server/blueprints/variant/controllers.py
dnil: 💯

scout/server/blueprints/variant/controllers.py
```diff
@@ -402,7 +411,7 @@ def variant(
         for disease in gene.get("disease_terms", []):
             disease["disease_link"] = disease_link(disease_id=disease["_id"])
 
-    tx_overview(variant_obj)
+    tx_overview(variant_obj=variant_obj, genome_build=genome_build)
```
dnil, line 407: 👍

scout/server/blueprints/variant/controllers.py
```diff
@@ -126,13 +129,19 @@ def tx_overview(variant_obj: dict):
             ovw_tx["coding_sequence_name"] = tx.get("coding_sequence_name")
             ovw_tx["protein_sequence_name"] = tx.get("protein_sequence_name")
 
+            # Add transcript links
+            add_tx_links(tx, genome_build, ovw_tx["hgnc_symbol"])
```
dnil: That now makes sense!

scout/server/blueprints/variant/utils.py
dnil: Here I'm a bit confused - I had a recollection the reason it was called here was that we have links that are a bit of both gene and variant links? But, maybe we get to it anyway - and skimming the links, while a few "variant" links actually use the gene id, that should not be blocking. So all good I guess!

scout/server/blueprints/variant/controllers.py
```diff
@@ -98,20 +98,23 @@ def tx_overview(variant_obj: dict):
...
                     decorated_tx = ovw_tx["mane_plus"]
                 elif refseq_id.startswith("XM"):
-                    ovw_tx["muted_refseq_ids"].append(refseq_id)
+                    ovw_tx["muted_refseq_ids"].add(refseq_id)
                     continue
-                else:
+                elif not tx.get("transcript_id", "").startswith("ENST"):
                     decorated_tx = refseq_id
-                ovw_tx["decorated_refseq_ids"].append(decorated_tx)
+                ovw_tx["decorated_refseq_ids"].add(decorated_tx)
```
dnil: This should fail when e.g. the refseq transcript starts with "ENST"? Maybe just put it under an if, for simplicity. Or we could rework the entire if-continue-else construction, flattening it a bit. 

```suggestion
                 if decorated_tx:
                     ovw_tx["decorated_refseq_ids"].add(decorated_tx)
```

scout/server/blueprints/variant/controllers.py
```diff
@@ -98,20 +98,23 @@ def tx_overview(variant_obj: dict):
...
 
-            ovw_tx["decorated_refseq_ids"] = []
-            ovw_tx["muted_refseq_ids"] = []
+            ovw_tx["decorated_refseq_ids"] = set()
+            ovw_tx["muted_refseq_ids"] = set()
+
+            if not tx.get("transcript_id", "").startswith("ENST"):
+                ovw_tx["decorated_refseq_ids"].add(tx["transcript_id"])
 
             for refseq_id in tx.get("refseq_identifiers", []):
```
dnil, line 111:
```suggestion
            for refseq_id in tx.get("refseq_identifiers", []):
                decorated_tx = None
```

scout/server/blueprints/variant/controllers.py
dnil: Yes, see crash on the demo rd case.

dnil (changes requested):
Good realisation, and good start! The table is more complete with the decorated transcript name, though I seem to recall we had a reason for asking ppl to check for the extended set below. Note the refseq id is in the "undecorated" id column already. There are a couple of issues: first a crash for genes with that ENST that now attempts to add an undefined var to a set. Then you astutely note that there are links missing, but they still are after fixing this. The core issue there I believe is that the primarily refseq transcripts don't seem to have a separate `refseq_id` set anyway to make the links from. See e.g. the mutalyzer and varsome links (missing from your example picture) or the refseq links themselves.

scout/server/blueprints/variant/utils.py
northwestwitch: Yes, it should be fine. Transcript links are not shown for all transcripts, but only for those in the transcripts overview on top!

scout/server/blueprints/variant/controllers.py
northwestwitch: Right! I didn't test with POT1 this time! :)

northwestwitch:
> There are a couple of issues: first a crash for genes with that ENST that now attempts to add an undefined var to a set. Then you astutely note that there are links missing, but they still are after fixing this. The core issue there I believe is that the primarily refseq transcripts don't seem to have a separate `refseq_id` set anyway to make the links from. See e.g. the mutalyzer and varsome links (missing from your example picture) or the refseq links themselves.

Thanks for the useful review! Both things should be fixed now. Tested OK locally:

### BRAF variant from before now shows links in that transcript:

image

### POT1 variant doesn't crash:

image

scout/server/links.py
```diff
@@ -357,20 +357,21 @@ def add_tx_links(tx_obj, build=37, hgnc_symbol=None):
         tx_obj["ensembl_link"] = ensembl_38_link
 
     refseq_links = []
-    if tx_obj.get("refseq_id"):
-        refseq_id = tx_obj["refseq_id"]
+    refseq_id = tx_obj.get("refseq_id") or (
+        tx_obj.get("transcript_id")
+        if not tx_obj.get("transcript_id", "").startswith("ENST")
```
northwestwitch, line 362: This is the fix to the links still missing

scout/server/blueprints/variant/controllers.py
dnil: It has served us well...

scout/server/blueprints/variant/controllers.py
```diff
@@ -67,80 +67,82 @@
...
         return
-    for gene in variant_obj.get("genes", []):
-        for tx in gene.get("transcripts", []):
-            ovw_tx = {}
 
-            if "refseq_identifiers" not in tx and tx.get("is_canonical", False) is False:
-                continue  # collect only RefSeq or canonical transcripts
+    for gene in genes:
+        common = gene.get("common", {})
+        hgnc_symbol = common.get("hgnc_symbol", gene.get("hgnc_id"))
```
dnil, line 87: And there would typically be a `gene.hgnc_symbol` as well. If there was a class for it we would know for sure... 😁

scout/server/blueprints/variant/controllers.py
```diff
@@ -67,80 +67,82 @@
...
-            ovw_tx["hgnc_id"] = gene.get("hgnc_id")
-
-            # ---- create content for the Refseq IDs column -----#
-            ovw_tx["mane"] = tx.get("mane_select_transcript", "")
-            ovw_tx["mane_plus"] = tx.get("mane_plus_clinical_transcript", "")
+        for tx in gene.get("transcripts", []):
+            if "refseq_identifiers" not in tx and not tx.get("is_canonical", False):
+                continue
+
+            ovw_tx = {
```
dnil, line 93: 💯

scout/server/links.py
dnil: Nice, yes!

scout/server/blueprints/variant/utils.py
dnil: Mm, the original use case was for the transcripts table below (the genes/transcripts/proteins accordion). But I think it works well anyway.

PR #6028, opened 2026-02-09
Add loqusdb api db name

muabnezor:
This PR adds a functionality 
**feature**: Adds support for specifying a database name (*db_name*) when sending requests to the loqusdb API. This allows scout to connect to specific loqusdb databases when multiple database exists at the same instance. 

Changes
- Added db_name parameter support in loqus_extension.py
- Added unit tests

User can now specify a database name together with an api url. E.g.

```python
LOQUSDB_SETTINGS = {
    "default": {
        "api_url": "http://loqusdb-api:8000",
        "db_name": "a_database"  # <-- NEW
    },
    "id1": {
        "api_url": "http://loqusdb-api:8000",
        "db_name": "another_database"  # <-- NEW
    }
}
```

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. Prepare a loqusdb API Instance with more than one frequency databases
2. Configure scout to use different *db_name* for different loqusdb IDs. *api_url* should be the same
3. Open a case with variants

**Expected outcome**:
Variant observations should load correctly from the specified database

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/tests/server/extensions/test_loqusdb_api_extension.py
+++ b/tests/server/extensions/test_loqusdb_api_extension.py
@@ -117,7 +117,7 @@ def test_loqusdb_api_sv_variant(loqus_api_app, monkeypatch, loqus_api_variant):
 def test_loqusdb_api_snv_variant_with_db_name(loqus_api_app, monkeypatch, loqus_api_variant):
     """Test fetching a SNV variant info from loqusdb API with custom database name"""
 
-    captured_url = None
+    captured_url = ""
 
     def mockapi(url):
         nonlocal captured_url
@@ -140,7 +140,7 @@ def test_loqusdb_api_snv_variant_with_db_name(loqus_api_app, monkeypatch, loqus_
 def test_loqusdb_api_sv_variant_with_db_name(loqus_api_app, monkeypatch, loqus_api_variant):
     """Test fetching a SV variant info from loqusdb API with custom database name"""
 
-    captured_url = None
+    captured_url = ""
 
     def mockapi(url):
         nonlocal captured_url


dnil:
Ah, you know more than me here @muabnezor! I assumed the db name setting was read when starting the loqusdb service, not for each api call. Have you tested on a local setup where you have the same URI, but two different db names?  
Also, don't forget to fill in the changelog!

muabnezor:
> Ah, you know more than me here @muabnezor! I assumed the db name setting was read when starting the loqusdb service, not for each api call. Have you tested on a local setup where you have the same URI, but two different db names? Also, don't forget to fill in the changelog!

Hi @dnil! Yes, that's my understanding. Each API call to loqusdbapi creates a database adapter with `uri` and `db_name` parameters. Our current issue: we have a loqusdbapi instance backed by a mongodb with two separate databases, but Scout has no way to specify which database to query. It always defaults to the database configured in loqusdbapi's settings. This PR solves that by allowing database selection from the scout config.

I'll test this PR today and update the changelog!

dnil:
> > Ah, you know more than me here @muabnezor! I assumed the db name setting was read when starting the loqusdb service, not for each api call. Have you tested on a local setup where you have the same URI, but two different db names? Also, don't forget to fill in the changelog!
> 
> Hi @dnil! Yes, that's my understanding. Each API call to loqusdbapi creates a database adapter with `uri` and `db_name` parameters. Our current issue: we have a loqusdbapi instance backed by a mongodb with two separate databases, but Scout has no way to specify which database to query. It always defaults to the database configured in loqusdbapi's settings. This PR solves that by allowing database selection from the scout config.
> 
> I'll test this PR today and update the changelog!

Cool, I'm not versed enough in FastAPI I guess - I thought that `Dependency` injection meant one instance would share a db connector between API calls, and `Settings` parsed on startup/first db use. We solve the same issue by firing up multiple instances of loqusdbapi, on different ports. I suppose one would anyway likely have to increase the worker number to sustain a presumably increased query load, so performance wise it should be somewhat similar (if the db does not have to reconnect each call) but there is for sure an administrative penalty with many services.

muabnezor:
I think as the code in loqsdbapi is now, a db adapter is created for each API call. Anyway, we thought about setting up several loqusdbapi instances, but figured this would be a good solution as well. 

Tested this locally, and it seem to work just fine :)

scout/server/extensions/loqus_extension.py
```diff
@@ -196,27 +197,29 @@ def get_variant(self, variant_info: dict, loqusdb_id: str = "default") -> Option
             return self.get_exec_loqus_variant(loqus_instance, variant_info)
 
         # Loqus instance is a REST API
-        return self.get_api_loqus_variant(loqus_instance.get(API_URL), variant_info)
+        return self.get_api_loqus_variant(loqus_instance, variant_info)
 
     @staticmethod
-    def get_api_loqus_variant(api_url, variant_info) -> dict:
+    def get_api_loqus_variant(loqusdb_instance, variant_info) -> dict:
```
northwestwitch:
```suggestion
    def get_api_loqus_variant(loqusdb_instance:dict, variant_info:dict) -> dict:
```

scout/server/extensions/loqus_extension.py
```diff
@@ -196,27 +197,29 @@ def get_variant(self, variant_info: dict, loqusdb_id: str = "default") -> Option
...
         SNV/INDELS can be queried in loqus by defining a simple id. For SVs we need to call them
         with coordinates.
 
         Args:
-            api_url(str): query url for the Loqus API
+            loqusdb_instance(dict): dictionary containing LoqusDB instance information
             variant_info(dict): dictionary containing variant coordinates
 
         Returns:
             res(dict)
```
northwestwitch: This is a somewhat old function, which is why the docstring is written this way. We are gradually updating it to describe the variables using type hints instead. This could be removed

scout/server/extensions/loqus_extension.py
```diff
@@ -225,7 +228,8 @@ def get_api_loqus_variant(api_url, variant_info) -> dict:
 
             sv_type = variant_info["variant_type"]
             search_url = f"{search_url}/?chrom={chrom}&end_chrom={end_chrom}&pos={pos}&end={end}&sv_type={sv_type}"
-
+            if loqusdb_instance.get(DB_NAME) is not None:
```
northwestwitch, line 231: This code is duplicated here and at line 221. I'd just add it outside these if/else, before line 233

scout/server/extensions/loqus_extension.py
northwestwitch, line 214: This is a somewhat old function, which is why the docstring is written this way. We are gradually updating it to describe the variables using type hints instead. The whole part Args and Returns could be removed

scout/server/extensions/loqus_extension.py
muabnezor: Actually not duplicated :) 
if snv we append`?db_name={loqusdb_instance.get(DB_NAME)}`
if svs we append `&db_name={loqusdb_instance.get(DB_NAME)}` since it already follows other parameters after `?`

northwestwitch: Right! I missed that! Never mind then! 

image

muabnezor:
Fixed the tests. Can I go ahead and merge @northwestwitch and  @dnil?

northwestwitch:
> Fixed the tests. Can I go ahead and merge @northwestwitch and @dnil?

Yes, go ahead! 👍🏻

dnil:
All good; I am now quite suspicious about the performance of loqusdbapi, but this is quite apparently the way the API was designed:
Screenshot 2026-02-11 at 14 50 10

No wonder it takes a long time to make multiple queries to it if goes through a full db adapter setup and teardown with each query. 🤔

muabnezor:
> All good; I am now quite suspicious about the performance of loqusdbapi, but this is quite apparently the way the API was designed: Screenshot 2026-02-11 at 14 50 10
> 
> No wonder it takes a long time to make multiple queries to it if goes through a full db adapter setup and teardown with each query. 🤔

I think this should be fixable. The database() function in  loqusdbapy/main.py indeed seem to create a new client, and MongoAdapter each time called. Another concern might be, are these clients, and mongoadapters piled up without proper cleanup?

```python
def database(uri: str = None, db_name: str = None) -> MongoAdapter:
    uri = uri or settings.uri
    db_name = db_name or settings.db_name
    try:
        client = get_client(
            uri=uri,
        )
    except DB_Error:
        raise HTTPException(
            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail="Could not connect to database"
        )

    return MongoAdapter(client, db_name=db_name)
```

The solution would be to initialize it once at startup, and reuse it across requests. If you are experiencing performance issues using scout, maybe I can have a look at loqusdbapi maybe next week. 😄 

maybe something like this would fix it:

```python
@asynccontextmanager
async def lifespan(app: FastAPI):
    try:
        app.state.client = get_client(uri=settings.uri)
    except DB_Error:
       ...
    yield
    app.state.client.close()

app = FastAPI(lifespan=lifespan)

def database(db_name) -> MongoAdapter:
    return MongoAdapter(app.state.client, db_name=db_name)

...
```

PR #5377, opened 2025-04-08
Fix general report to show all dismissed variants (not pinned or causatives) only among the dismissed

northwestwitch:
This PR adds a functionality or fixes a bug.
- Closes #5374
 
<details>
<summary>Testing on cg-vm1 server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. Make sure the PR is pushed and available on [Docker Hub](https://hub.docker.com/repository/docker/clinicalgenomics/scout-server-stage)
1. Fist book your testing time using the Pax software available at [https://pax.scilifelab.se/](https://pax.scilifelab.se). The resource you are going to call dibs on is `scout-stage` and the server is `cg-vm1`.
1. `ssh <USER.NAME>@cg-vm1.scilifelab.se`
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `podman ps`
1. Stop the service with current deployed branch: `systemctl --user stop scout.target`
1. Start the scout service with the branch to test: `systemctl --user start scout@<this_branch>`
1. Make sure the branch is deployed: `systemctl --user status scout.target`
1. After testing is done, repeat procedure at [https://pax.scilifelab.se/](https://pax.scilifelab.se), which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

<details>
<summary>Testing on hasta server (Clinical Genomics Stockholm)</summary>

**Prepare for testing**
1. `ssh <USER.NAME>@hasta.scilifelab.se`
1. Book your testing time using the Pax software. `us; paxa -u <user> -s hasta -r scout-stage`. You can also use the WSGI Pax app available at [https://pax.scilifelab.se/](https://pax.scilifelab.se).
1. (optional) Find out which scout branch is currently deployed on cg-vm1: `conda activate S_scout; pip freeze | grep scout-browser`
1. Deploy the branch to test: `bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_scout -t scout -b <this_branch>`
1. Make sure the branch is deployed: `us; scout --version`
1. After testing is done, repeat the `paxa` procedure, which will release the allocated resource (`scout-stage`) to be used for testing by other users.
</details>

**How to test**:
1. how to test it, possibly with real cases/data

**Expected outcome**:
The functionality should be working
Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

Issue #5374, opened 2025-04-07
Case report can contain same variant both dismissed and as ACMG classification


dnil:
**Describe the bug**
Variants can end up in multiple categories on the case general report. This is perhaps due to the accounting logic? Reported by cust002, and reproduced on local in a quick test.

**To Reproduce**
1.  Classify a variant with ACMG and dismiss it. 
2. Probably have some more variants on the report?
3. See variant both in ACMG-other and as dismissed.

**Expected behavior**
The suggested action from the reporter at cust002 was to not display variants that only had an ACMG classification. For the record I think this is the wrong approach - we will quickly end up in the other ditch where someone has classified a variant LP but forgot to pin it, and then pendulum shift. And it will be hard to explain. 

Better to patch, either setting dismiss priority higher (and showing a little more additional info on the dismissed ones, including ACMG classification if available) or simply make it graphically very clear that the variant was dismissed where it first appears (eg as a LB in ACMG). 

**Additional context**
A more future safe approach might be to split the ACMG-other variants into three categories. One P/LP appearing between causative and pinned. One VUS appearing around commented. And one B/LB appearing with the dismissed variants with terse information. This is from a quick thought applicable to all the other classification systems as well - ie it would make sense to treat a manual "Pathogenic" very much like an ACMG P/LP, or a Benign CVC as an effective dismiss.

northwestwitch:
I wouldn't call it a bug, because variant is shown anyway among the dismissed. But I agree that it's more useful to just have it among the dismissed, otherwise it's a waste of time

dnil:
I agree, it’s borderline: not super harmful, more than a time thief, but also not quite the intended function - we do try to make sure variants appear only once. But also a change, since the time benefit part would be having such variants clearly dismissed only.

PR #6035, opened 2026-02-10
Fix report timeout

### LINKED ISSUE #6009: General report timeouts
See https://scout.scilifelab.se/cust002/F0039928-4

Observations include several stage reanalysis and lots of dismissed variants/interactions. If it is just the dismissals perhaps we should consider not showing any details for them over a certain number, only say have a total count.  It's not like any md is ever going to dive into details about the seventy-fourth dismissed variants.

northwestwitch:
- Closes #6009

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/variant.py
+++ b/scout/adapter/mongo/variant.py
@@ -766,19 +766,26 @@ class VariantHandler(VariantLoader):
 
     def evaluated_true_dismissed(self, case_id: str, institute_id: str) -> Set[str]:
         """Returns a unique set of variants which have been dismissed for a case."""
-        variant_ids_dismissed = self.evaluated_variant_ids_from_events(
-            case_id, institute_id, include_verbs=["dismiss_variant"]
-        )  # Can contain duplicates
-        variant_ids_reset_dismissed = self.evaluated_variant_ids_from_events(
-            case_id, institute_id, include_verbs=["reset_dismiss_variant"]
-        )  # Can contain duplicates
-
-        # Remove reset dismissed from list of variant_ids_dismissed
-        for vid in variant_ids_reset_dismissed:
-            if vid in variant_ids_dismissed:
-                variant_ids_dismissed.remove(vid)
-
-        return set(variant_ids_dismissed)
+        pipeline = [
+            {
+                "$match": {
+                    "case": case_id,
+                    "institute": institute_id,
+                    "category": "variant",
+                    "verb": {"$in": ["dismiss_variant", "reset_dismiss_variant"]},
+                }
+            },
+            # Newest events first for each variant
+            {"$sort": {"variant_id": 1, "created_at": -1}},
+            # Group so the first (newest) event per variant is captured
+            {"$group": {"_id": "$variant_id", "latestVerb": {"$first": "$verb"}}},
+            # Keep only those where the latest event is a dismissal
+            {"$match": {"latestVerb": "dismiss_variant"}},
+            # Project only the variant ID
+            {"$project": {"_id": 0, "variant_id": "$_id"}},
+        ]
+        cursor = self.event_collection.aggregate(pipeline)
+        return {doc["variant_id"] for doc in cursor}
 
     def evaluated_variant_ids_from_events(
         self,


scout/adapter/mongo/variant.py
```diff
@@ -802,27 +786,39 @@ def evaluated_variant_ids_from_events(self, case_id, institute_id):
...
+            "ccv",
+            "manual_rank",
+            "cancer_tier",
+            "escat_tier",
+            "mosaic_tags",
+        ]
+        variant_ids_non_dismissed = self.evaluated_variant_ids_from_events(
+            case_id, institute_id, eval_verbs=EVAL_VERBS
+        )
+        variant_ids_dismissed = self.evaluated_variant_ids_from_events(
```
northwestwitch: This is no elegant fix, I know

dnil: It retrieves the dismissed variants separately, which is a good start.

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -804,7 +804,7 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 {% macro dismissed_panel() %}
   <div class="card border-danger mb-3" style="border-width: 5px; display: block;">
     <div class="card-header bg-danger text-black" style="border-top-left-radius: 0; border-top-right-radius: 0;">
-      <p id="dismissedDesc"><a id="dismissed_variants"><strong>Dismissed Variants</strong></a></p>
+      <p id="dismissedDesc"><a id="dismissed_variants"><strong>Dismissed Variants (displaying only the first 15 results)</strong></a></p>
```
dnil: It would be nice to pass the constant.

scout/adapter/mongo/variant.py
```diff
@@ -802,27 +786,39 @@ def evaluated_variant_ids_from_events(self, case_id, institute_id):
...
+            case_id, institute_id, eval_verbs=["dismiss_variant"]
+        )
 
+        LIMIT_DISMISSED = 15
+        variant_ids = (
+            list(variant_ids_non_dismissed) + list(variant_ids_dismissed)[:LIMIT_DISMISSED]
+        )
```
dnil, line 845: The downstream queries are slow (partly unindexed or indexed with retrieving single variants in mind). But slashing the number must help.

scout/adapter/mongo/variant.py
```diff
@@ -802,27 +786,39 @@ def evaluated_variant_ids_from_events(self, case_id, institute_id):
...
+            "mosaic_tags",
+        ]
+        variant_ids_non_dismissed = self.evaluated_variant_ids_from_events(
+            case_id, institute_id, eval_verbs=EVAL_VERBS
+        )
+        variant_ids_dismissed = self.evaluated_variant_ids_from_events(
+            case_id, institute_id, eval_verbs=["dismiss_variant"]
+        )
 
+        LIMIT_DISMISSED = 15
```
dnil: It would be nice to display the actual count on the report.

dnil (changes requested):
Well, this should help towards the original problem. It would really be natural to display the number of dismissed variants before truncating the list as well. 

More complicated is that I think this will break the solution in #5377, resurrecting #5374. We would then need to add a bit of logic fixing that issue, perhaps by by checking that the classified variants are not dismissed in the controller instead of doing that with dict membership in the template.

northwestwitch:
> Well, this should help towards the original problem. It would really be natural to display the number of dismissed variants before truncating the list as well.

There is a problem with this. You get all the dismissed variants from the events table (quick). But in order to provide the precise number of dismissed vars, you must check that those variants you selected are still dismissed. And here the second query is super slow, whenever the variant_ids list is very long. I'll see if it can be done in a smart way without slowing the whole thing too much, othewerwise we might as well keep the code as it is..

dnil:
> > Well, this should help towards the original problem. It would really be natural to display the number of dismissed variants before truncating the list as well.
> 
> There is a problem with this. You get all the dismissed variants from the events table (quick). But in order to provide the precise number of dismissed vars, you must check that those variants you selected are still dismissed. And here the second query is super slow, whenever the variant_ids list is very long. I'll see if it can be done in a smart way without slowing the whole thing too much, othewerwise we might as well keep the code as it is..

Gotcha. Then the "showing also the first 15" should probably be rephrased, right? The first 15 may just have been from a batch of dismiss events that the user later reset, so none are shown at all? Giving the impression there were none, when there might have been any number.

scout/server/blueprints/cases/templates/cases/case_report.html
northwestwitch: Fixed

scout/adapter/mongo/variant.py
northwestwitch: Fixed:

image

scout/server/blueprints/cases/templates/cases/case_report.html
```diff
@@ -804,7 +804,7 @@ <h4>Scout case analysis report</h4> - created on:&nbsp;<strong>{{report_created_
 {% macro dismissed_panel() %}
   <div class="card border-danger mb-3" style="border-width: 5px; display: block;">
     <div class="card-header bg-danger text-black" style="border-top-left-radius: 0; border-top-right-radius: 0;">
-      <p id="dismissedDesc"><a id="dismissed_variants"><strong>Dismissed Variants</strong></a></p>
+      <p id="dismissedDesc"><a id="dismissed_variants"><strong>Dismissed Variants (displaying {{limit_dismissed}} results out of {{n_dismissed_variants}})</strong></a></p>
```
dnil: Displaying `min(limit_dismissed, n_dismissed_variants)`. 😊 This looks a little off:

Image

scout/adapter/mongo/variant.py
```diff
@@ -764,68 +764,91 @@ def hgnc_overlapping(
...
-        evaluation_verbs = [
-            "acmg",
-            "ccv",
-            "manual_rank",
-            "cancer_tier",
-            "dismiss_variant",
-            "escat_tier",
-            "mosaic_tags",
-        ]
+        # Remove reset dismissed from list of variant_ids_dismissed
```
dnil: Hm? But they could have been before and after? Like, dismissed, then reset and finally dismissed again?

Yes, tested by doing exactly that. The variant is no longer on the dismissed list as long as it has a single reset somewhere in the history.

scout/adapter/mongo/variant.py
```diff
@@ -764,68 +764,91 @@ def hgnc_overlapping(
...
-        Returns:
-            variants(iterable(Variant))
-        """
+    def evaluated_true_dismissed(self, case_id: str, institute_id: str) -> Set[str]:
+        """Returns a unique set of variants which have been dismissed for a case."""
+        variant_ids_dismissed = set(
+            self.evaluated_variant_ids_from_events(
+                case_id, institute_id, include_verbs=["dismiss_variant"]
+            )
+        )  # Can contain duplicates
```
dnil: Mm, cannot really contain duplicates if it is a set? Or do you mean that the returned list from the adapter can? True!

scout/adapter/mongo/variant.py
```diff
@@ -764,68 +764,91 @@ def hgnc_overlapping(
...
-        query = {
-            "category": "variant",
-            "institute": institute_id,
-            "case": case_id,
-            "verb": {"$in": evaluation_verbs},
-        }
-        evaluation_events = self.event_collection.find(query)
-        if evaluation_events is None:
-            return []
+        return set(variant_ids_dismissed)  # Return unique variant_ids
```
dnil: Is already a set?

scout/adapter/mongo/variant.py
```diff
@@ -764,68 +764,91 @@ def hgnc_overlapping(
...
         Return only if the variants still exist and still have the assessment.
-        Variants can be removed on reanalyis, and assessments can be cleared.
-
-        Args:
-            case_id(str)
-            institute_id(str)
-
-        Returns:
-            variants(iterable(Variant))
+        Variants can be removed on re-analysis, and assessments can be cleared.
```
dnil, line 829: It would be great to explain the reasoning about dismissed variants, and the returned number_of_unique_dismissed int in the docstring as well! I would personally rather have a decent length story here, than the comments interleaved later that tell that story bit by bit in parallel with the rest. 😊

scout/adapter/mongo/variant.py
northwestwitch: Yes, the second!

scout/adapter/mongo/variant.py
northwestwitch: Yes, I see your point. I'll revisit that `evaluated_true_dismissed` function. What a headache!

dnil: I know, and the moment one starts mass querying variants it gets slow. Checking a few should be fine, but it is easy to imagine worst case scenarios for progressive fetching of more dismissed variants. Consider the indices again perhaps?

scout/adapter/mongo/variant.py
dnil: Maybe describe in the docstring instead to avoid any confusion?

northwestwitch: You know I think if I just avoid transforming variant_ids_dismissed to a set in this step I might solve the problem of the dismissed, then reset then re-dismissed altogether.. I'll try

northwestwitch: I think that's what I wanted to do and that's why that comment is there

scout/adapter/mongo/variant.py
northwestwitch: Needed now, since it's no longer a set.

scout/adapter/mongo/variant.py
dnil: Yes, that is a lot better, but perhaps still not quite there? We don't have an "update dismissed reasons" event, so a user can actually dismiss one variant several times for changing the reasons, then reset. Since these are not ordered/timed (or anchored in the "semaphore" on the variants collection) it would currently look like the variant was still dismissed, although it was actually reset, wouldn't it?

How about sorting events on creation time, using the latest?

northwestwitch: > How about sorting events on creation time, using the latest?

I'll give it a try!

northwestwitch: What's the the "semaphore" on the variants collection?

dnil: I mean that we have that array with dismiss reasons on the variants collection, which is the finally say on if the variant is dismissed or not. As such it acts like a semaphore - there can be any number of events in either direction, and the final dismiss state is still stored there.

dnil: Hm, maybe it's also not the end of the world if we show one less or more. I'll fetch some coffee - complexity is nearing the 🤯 level. 😸

scout/adapter/mongo/variant.py
```diff
@@ -764,68 +764,97 @@ def hgnc_overlapping(
...
-    def evaluated_variant_ids_from_events(self, case_id, institute_id):
-        """Returns variant ids for variants that have been evaluated
-           Return all variants, snvs/indels and svs from case case_id
-            which have an event entry for 'acmg_classification', 'ccv_classification', 'manual_rank', 'dismiss_variant',
-            'cancer_tier', 'escat_tier', 'mosaic_tags'.
-        Args:
-            case_id(str)
+    def evaluated_true_dismissed(self, case_id: str, institute_id: str) -> Set[str]:
+        """Returns a unique set of variants which have been dismissed for a case."""
+        pipeline = [
```
northwestwitch, line 769: @dnil check out this fantastic solution I came up to ALL BY MYSELF!

dnil: Are we channeling some state-of-the-union here? 😆 

`$first` was a really nice one - I don't think I ever used it!
Ahem, but perhaps we can rework those comments into the docstring?

northwestwitch: > Ahem, but perhaps we can rework those comments into the docstring?

I knew you were going to ask this, but this time I wouldn’t like that. In general, I find it much clearer when comments explain exactly what a complicated line does—I’m not much of a docstring reader. I really feel that, in this case, the comments are helping to understand precisely what each line does. One year from now, we might even be thankful that these comments are exactly there.

dnil: It messes with my code reading buffer; I never know whether I'm supposed to read both the comments and the code, and spend effort verifying that the code and the comment are in agreement. We are all different I guess; as long as the docstrings are accurate, its fine naturally.

northwestwitch: Feel free to move the comments in the docstring then. In general I never read docstrings (my brain is like that) so it's like removing the comments entirely for me. But whatever that works. It's the code that is important

northwestwitch:
> Nice, and well fought with the complexity. I still feel docstrings beat comments for documentation and ease of reading, but as long as the code is sound and there is some documentation, I guess it's ok.

You could move those in the docstring, seriously! :)

> 
> When reading, the code now looks logical!
> 
> I am getting some strange results on counts of variants included in the report when dismissing large chunks of variants from my local, but who knows what strangeness is in the db. Perhaps some existing but unset keys on the variants? Anyway, when tested on a newish case on stage, it works well as far as I can tell.

Could be that it's dismissed that also fall in the other categories of evaluated or something?

PR #6094, opened 2026-03-04
Speedup and refactor remove variants cmd

### LINKED ISSUE #6092: Speed up variant cleanup
See https://github.com/Clinical-Genomics/IT-issues/issues/955#issuecomment-3989780289.

In particular, 
- don't keep cases cursor open - fetch to a list - otherwise it can time out during operation
- consider avoiding queries with multiple in variant_id selections, and instead doing those in python as well.

### LINKED ISSUE #6104: Delete variants command does not allow to specifically remove research variants
If you were to select research variants as remove category you can't. Same for keep category:

Image

northwestwitch:
- Refactor, speedup and add a progress bar to `scout delete variants` cmd -> closes #6092 
- Speedup affects only the dry-run, which by itself was crashing when tested on stage.
- The command now accepts an optional `--out-file` where it prints a detailed report of the deletion process - real or simulated (no number of removed vars in this case)
- Take into account also research variants in removal categories - fix #6104

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR


northwestwitch:
Ok, now the dry run is quite quick: less than 1m23s (it took hours before crashing last time)

Tested with the following command for instance: `time scout delete variants -u chiara.rasi@scilifelab.se --rank-threshold 5 --variants-threshold 50000  --dry-run`

image

It created a tab-separated list of cases 113 with more than 50K variants which can be used in the following step (delete without dry run):


**NOTE THAT THE NON DRY-RUN COMMAND IS STILL SLOW** but I'm reasoning that we don't care that much because we are processing the removal by batches of cases after having confirmed with the clinicians that those cases are fine to purge - at least we did like this last time. Note that the script accepts also a list of case IDs for this reason.

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+    Handle variant removal for a case.
+
+    If dry_run is True, counts how many variants *would be removed* without deleting.
+    Returns a tuple: (number_deleted_variants, number_deleted_omics_variants)
+    """
+
+    remove_n_variants = 0
+    remove_n_omics_variants = 0
+
+    if not dry_run:
```
northwestwitch, line 48: before if it was dry-run it was counting  exactly how many vars would be deleted. I figured WHO CARES: what we are interested in is a list of cases with more than a number of variants and their specifics, to be be processed afterwards

scout/commands/delete/variants.py
```diff
@@ -141,104 +338,36 @@ def variants(
...
-                    case_id,
-                    case.get("track", ""),
-                    str(case["analysis_date"]),
-                    case.get("status", ""),
-                    str(case.get("is_research", "")),
-                    str(case_n_variants),
-                    str(remove_n_variants + remove_n_omics_variants),
-                ]
-            )
+            f"Total {items_name}: {delete_stats['deleted_variant_counter'] + delete_stats['deleted_outlier_counter']}",
```
northwestwitch: Probably do not print this line if it's dry-run, no?

dnil:
Nice! I never thought I'd see the words "speedup" and "add progressbar" in the same PR! 😁

scout/commands/delete/variants.py
dnil: Yes, though it isn't perhaps a full dry-run: maybe we should have another name for it.

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+    store: MongoAdapter,
+    remove_ctg: List[str],
+    outlier_remove_ctg: List[str],
+    dry_run: bool,
+    variants_query: dict,
+) -> Tuple[int, int]:
+    """
+    Handle variant removal for a case.
+
+    If dry_run is True, counts how many variants *would be removed* without deleting.
```
dnil:
```suggestion
    If dry_run is True, do not delete. Also do not count again, as this needlessly takes a long time and does not do exactly the same thing.
```

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+        "causatives": 1,
+        "status": 1,
+        "owner": 1,
+        "analysis_date": 1,
+        "individuals": 1,
+    }
+
+    cases_cursor = store.case_collection.find(match, projection)
+    total_cases = store.case_collection.count_documents(match)
+    with click.progressbar(cases_cursor, length=total_cases, label="Processing cases") as cases_bar:
```
dnil: This should be susceptible to the same issue as before ie a timeout of the cursor?

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+        "suspects": 1,
+        "causatives": 1,
+        "status": 1,
+        "owner": 1,
+        "analysis_date": 1,
+        "individuals": 1,
+    }
+
+    cases_cursor = store.case_collection.find(match, projection)
+    total_cases = store.case_collection.count_documents(match)
```
dnil, line 135: 👍

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+        {"case_id": case["_id"]}
+    ) + store.omics_variant_collection.count_documents({"case_id": case["_id"]})
+
+    if variants_threshold is not None and variant_count < variants_threshold:
+        return
+    delete_stats["case_counter"] += 1
+    case_evaluated, _ = store.evaluated_variants(
+        case_id=case["_id"], institute_id=case["owner"], limit_dismissed=5
+    )
+    evaluated_not_dismissed = [v["_id"] for v in case_evaluated if "dismiss_variant" not in v]
```
dnil:
```suggestion
    evaluated_not_dismissed = [variant["_id"] for variant in case_evaluated if "dismiss_variant" not in variant]
```

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,227 @@
...
+    """
+    variant_count = store.variant_collection.count_documents(
+        {"case_id": case["_id"]}
+    ) + store.omics_variant_collection.count_documents({"case_id": case["_id"]})
+
+    if variants_threshold is not None and variant_count < variants_threshold:
+        return
+    delete_stats["case_counter"] += 1
+    case_evaluated, _ = store.evaluated_variants(
+        case_id=case["_id"], institute_id=case["owner"], limit_dismissed=5
```
dnil, line 177: Any reason for the 5? 🤔

scout/commands/delete/variants.py
```diff
@@ -141,104 +338,36 @@ def variants(
...
+                "Variants are going to be deleted from database. Continue?",
+                abort=True,
+            )
+            items_name = "deleted variants"
 
-        remove_n_variants, remove_n_omics_variants = handle_delete_variants(
-            store=store, keep_ctg=keep_ctg, dry_run=dry_run, variants_query=variants_query
+        remove_ctg, outlier_remove_ctg = _set_remove_ctg(keep_ctg=keep_ctg, rm_ctg=rm_ctg)
+        click.echo("\t".join(DELETE_VARIANTS_HEADER), file=output_file)
+        _process_cases(
```
dnil, line 362: 👍

scout/commands/delete/variants.py
```diff
@@ -50,27 +235,32 @@ def get_case_ids(case_file: Optional[str], case_id: List[str]) -> List[str]:
...
+    """Define the categories of variants that should be removed.
+
+    Returns a tuple (remove_ctg, outlier_remove_ctg).
+    """
+
+    if keep_ctg and rm_ctg:
+        raise click.UsageError("Please use either '--keep-ctg' or '--rm-ctg', not both.")
+
+    if rm_ctg:
+        remove_ctg = [ctg for ctg in rm_ctg if ctg not in OUTLIERS_CTG]
```
dnil, line 253: 👍

scout/adapter/mongo/query.py
```diff
@@ -52,31 +52,30 @@ def build_case_query(
...
+
+        query: dict = {
+            "case_id": case_id,
+            "$or": [
+                {"rank_score": {"$lt": min_rank_threshold}},
+                {"rank_score": {"$exists": False}},
+            ],
+        }
+        if remove_ctg:
+            query["category"] = {"$in": remove_ctg}
```
dnil, line 75: Hm, we never split the CTGs right? So these would still have those `whatever_research` on them? But it would never propagate to a ```{ category: 'whatever', variant_type: "research" }```

scout/commands/delete/variants.py
northwestwitch: We can have a list, I agree!

scout/commands/delete/variants.py
northwestwitch: It would be perhaps more appropriate, but I don't think it's worth because it would be a major change

scout/commands/delete/variants.py
northwestwitch: Just to speed it up

scout/adapter/mongo/query.py
northwestwitch: Ah right, now that you tell me I remember. I'll fix, or perhaps I'll just leave it as it was

northwestwitch: Reverted now! Very good point!

northwestwitch:
When you have time I'm ready for another round on this one @dnil. Thanks!

scout/commands/delete/variants.py
dnil: Absolutely, I do think it is correct: just wondering if you saw it somehow safer than say 2, or less so than 15?

scout/commands/delete/variants.py
dnil: Ah, yes. We could add it to that list perhaps! Or split into two, a true dry-run, and a case variants count option. But lets not overdo it - its a use-once-a-year kind of thing.

scout/commands/delete/variants.py
northwestwitch: No, I haven't tested. We can have 15 if you prefer, doesn't matter to me

dnil: I'm also fine with 0 as long as we understand what is happening?

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,226 @@
...
-    "Removed variants",
+    "DNA variants",
+    "Removed DNA variants",
+    "Removed Outlier variants",
 ]
-VARIANT_CATEGORIES = list(VARIANTS_TARGET_FROM_CATEGORY.keys())
+VARIANT_CTG = sorted(VARIANTS_TARGET_FROM_CATEGORY)
+OUTLIERS_CTG = ["outlier"]
+TIMESTAMP = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+OUTFILE = f"variant_cleanup_report_{TIMESTAMP}.tsv"
```
dnil: Could we have the timestamps set at runtime instead? It's fine to set it early if you want start rather than finish time.

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,226 @@
...
+        match["owner"] = institute
+    if status:
+        match["status"] = {"$in": list(status)}
+    if older_than:
+        older_than_date = datetime.now() - timedelta(weeks=older_than * 4)
+        match["analysis_date"] = {"$lt": older_than_date}
+    if analysis_type:
+        match["individuals.analysis_type"] = {"$in": analysis_type}
 
+    projection = {
```
dnil, line 121: Good idea to keep the doc size down - you don't need images etc. But doesn't this get a bit narrow for the downstream still?

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,226 @@
...
+        "_id": 1,
+        "display_name": 1,
+        "suspects": 1,
+        "causatives": 1,
+        "status": 1,
+        "owner": 1,
```
dnil, line 127:
```suggestion
        "owner": 1,
        "track": 1,
        "is_research": 1
```
There might be more...

scout/commands/delete/variants.py
```diff
@@ -1,42 +1,226 @@
...
 from flask import current_app, url_for
 from flask.cli import with_appcontext
 
-from scout.adapter.mongo import MongoAdapter
+from scout.adapter import MongoAdapter
 from scout.constants import ANALYSIS_TYPES, CASE_STATUSES, VARIANTS_TARGET_FROM_CATEGORY
 from scout.server.extensions import store
 
-BYTES_IN_ONE_GIGABYTE = 1073741824
+delete_stats = {"case_counter": 0, "deleted_variant_counter": 0, "deleted_outlier_counter": 0}
```
dnil: This makes my head spin. I guess as long as it is single thread CLI it is fine, perhaps? But consider putting in a more controlled scope, in the command function, and just pass it to the counting function, or return it even. Modules scope can be finicky. 😊

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Rank model URLs can now be filenames, if not starting with "http" (#6150)
 - Methbat outliers significance filtering extended with "AlleleSpecificMethylation" and "imprint" (#6156)
 - Custom IGV tracks settings shown on variant don't depend on the specific variant genome build (#6165)
+- Refactor, speedup (dry-run only) and add a progress bar to `scout delete variants` cmd (#6094)
```
dnil: Don't forget to move these!

tests/commands/delete/test_delete_variants.py
```diff
@@ -123,27 +151,37 @@ def test_delete_outlier_variants(mock_app, case_obj, user_obj):
...
+
     # WHEN variants are removed using the command line
     cmd_params = [
         "delete",
         "variants",
         "-u",
         user_obj["email"],
         "--rank-threshold",
-        0,
+        RANK_THRESHOLD,
```
dnil, line 164: 👍

scout/commands/delete/variants.py
dnil: This still bugs me a little, but I guess I will accept it. How about making it a constant, and add a comment for it describing the implications? Like at most this many random random dismissals for each case will be kept in order to speed the process up. You have actually evaluated and deemed those variants irrelevant anyway, so probably safer than a lot of stuff, but will affect say case report generation.

scout/commands/delete/variants.py
northwestwitch: Moved the timestamp inside the function itself, so it's set when the function actually runs

scout/commands/delete/variants.py
northwestwitch: It's essential for the fields that get printed in the outfile/terminal - they correspond to the header. We can add more fields if we want to show them as well, but otherwise I think it's fine

scout/commands/delete/variants.py
northwestwitch: Ah OK, you want to show also track and id research. Sure, I'll also add those to the header!

northwestwitch: Ah now I get it. These 2 keys are in the header but are not shown in the file/terminal. Good observation!

scout/commands/delete/variants.py
northwestwitch: Fixed!

scout/commands/delete/variants.py
northwestwitch: Fixed

PR #6150, opened 2026-03-23
Fix #6139 - allow reading rank model from a file on the server filesystem

### LINKED ISSUE #6139: Allow loading rank model from file

Rank model (snv and sv) yaml files contain version keys. These are stored on the db collection eg. It would sometimes be convenient if the rank model version could be retrieved from within the file, as the pipeline/orchestator does not necessarily have the version number handy (currently, it appears the version is a constant in CG for Nallo https://github.com/Clinical-Genomics/cg/blob/78226f711b743adb0b1c4991fe429e93fc842fd9/cg/constants/scout.py#L192-L193, is parsed from the manifest file for raredisease etc).

Parse rank model version for a case based on the rank model urls. If keys are also given directly on the case, use these as a redundancy check.

dnil:

- Fix #6139 - allow reading rank model from file, and warn when different Version given in model vs on the case

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] local tests executed by DN

### DIFF
--- a/scout/server/blueprints/variant/controllers.py
+++ b/scout/server/blueprints/variant/controllers.py
@@ -559,7 +559,7 @@ def variant_rank_scores(store: MongoAdapter, case_obj: dict, variant_obj: dict) 
         return rank_score_results
 
     rank_model_version = case_obj.get(
-        _get_category_string(case_obj.get("category"), "rank_model_version")
+        _get_category_string(variant_obj.get("category"), "rank_model_version")
     )
 
     if rank_model := store.rank_model_from_url(rank_model_url):


scout/adapter/mongo/rank_model.py
```diff
@@ -10,7 +11,7 @@
 
 
 class RankModelHandler(object):
-    def fetch_rank_model(self, rank_model_url):
+    def fetch_rank_model(self, rank_model_url: str) -> StringIO:
```
northwestwitch: Perhaps rename this function something like `fetch_rank_model_lines_from_url`?

scout/adapter/mongo/rank_model.py
```diff
@@ -25,34 +26,34 @@ def fetch_rank_model(self, rank_model_url):
         except Exception as ex:
             LOG.warning(ex)
 
-    def parse_rank_model(self, stringio):
-        """Use configobj lib to extract RankModel key/values and return them in a dictionary
+    def read_rank_model(self, rank_model_url) -> StringIO:
```
northwestwitch: Rename to something line `fetch_rank_model_lines_from_file`?

scout/adapter/mongo/rank_model.py
```diff
@@ -25,34 +26,34 @@ def fetch_rank_model(self, rank_model_url):
...
-        Args:
-            stringio(StringIO): Content of model from a file as a StringIO
+        with open(rank_model_url, "r") as rank_model_file:
+            rank_model_lines = rank_model_file.read()
 
-        Returns:
-            ConfigObj.dict(dictionary): dictionary with variant rank model key/values
+        return StringIO(rank_model_lines)
+
+    def parse_rank_model(self, stringio: StringIO):
```
northwestwitch, line 37: Rename to `parse_rank_model_lines` or something?

scout/server/blueprints/variant/controllers.py
```diff
@@ -488,47 +488,94 @@ def set_edge_genes(store: MongoAdapter, case_obj: dict, variant_obj: dict):
...
-            rank_model_link_prefix=rm_link_prefix,
+            rank_model_link_prefix=rank_model_link_prefix,
             rank_model_version=rank_model_version,
-            rank_model_file_extension=rm_file_extension,
+            rank_model_file_extension=rank_model_file_extension,
         )
+    return rank_model_url
+
+
+def variant_rank_scores(store: MongoAdapter, case_obj: dict, variant_obj: dict) -> list:
```
northwestwitch:
```suggestion
def variant_rank_scores(store: MongoAdapter, case_obj: dict, variant_obj: dict) -> List[dict]:
```

scout/adapter/mongo/rank_model.py
dnil: Sure!

scout/adapter/mongo/rank_model.py
dnil: Sure!

scout/adapter/mongo/rank_model.py
dnil: Hm, there is no other parsing going on here?

northwestwitch: It's fine also as it is

dnil:
> Looks good to me. 👍🏻 Do we expect any flash messages where models disagree?

Mm, perhaps not. Part of the reason for this is that Nallo has apparently been running on a permanent v1.0 since release, and the rank model has been tweaked and updated in place. This in itself would not be flagged, without like a full model checksum, but maybe we can catch similar ones. Next time there is a bump, we can expect a change in the version file, but probably not in the filename. This is what we might be able to catch here. 

Since keeping the same name is very practical for development, core-dev and Felix are looking at storing rank model files to housekeeper instead for all cases.

What I think we should do eventually is start parsing rank models already at load. I don't really see why we should not. The current solution at variant display time does address all cases, old and new, but also runs all the time. At load, a version warning like this could instead be cause to halt loading, or at least scream about it.

northwestwitch:
> What I think we should do eventually is start parsing rank models already at load. I don't really see why we should not. The current solution at variant display time does address all cases, old and new, but also runs all the time. At load, a version warning like this could instead be cause to halt loading, or at least scream about it.

Yes, it would be nice

Issue #6098, opened 2026-03-04
Causatives liftover to build 38


northwestwitch:
- Lift over of SNVs only?
- Make them available as a parsed field on the VCF?
- Add them to managed variants with a comment like "causative in build 37"?

Take care of the issue that users have access to a limited list of institutes

jemten:
I'm a proponent of doing SNVs only. 

Proposed strategy is:
`export -> liftOver -> annotate in vcf with specific tag -> load in scout and display strategy for causatives`

What do you think about using a flag like this in the VCF?
`##INFO=<ID=Scout37_causative,Number=0,Type=Flag,Description="Marked as causative in Scout 37">`
 
Edit: 
For reference this is the command used by Annick to lift over variants 
```
picard -Xmx16g LiftoverVcf I=/home/proj/stage/workflows/raredisease/references/v2.0_grch38/mutacc_20251212_causative_variants.vcf.gz  \
    O=/home/proj/stage/workflows/raredisease/references/v2.0_grch38/mutacc_20251212_causative_variants_hg38.vcf.gz   \
    REJECT=/home/proj/stage/workflows/raredisease/references/v2.0_grch38/mutacc_20251212_causative_variants_hg38_reject.vcf.gz \
    CHAIN=/home/proj/stage/workflows/raredisease/references/v2.0_grch38/hg19ToHg38_nochr.over.chain \
    R=/home/proj/stage/workflows/raredisease/references/v2.0_grch38/hg38.p14.fa.gz \
    VERBOSITY=DEBUG WARN_ON_MISSING_CONTIG=true LIFTOVER_MIN_MATCH=0.95
```

northwestwitch:
> Proposed strategy is: `export -> liftOver -> annotate in vcf with specific tag -> load in scout and display strategy for causatives`
> 
> What do you think about using a flag like this in the VCF? `##INFO=<ID=Scout37_causative,Number=0,Type=Flag,Description="Marked as causative in Scout 37">`

I would handle this in Scout only instead:
1. Fix managed so they can be optionally assigned to one or more customers
2. Add the lifted causatives to managed, so they can eventually be matched to old causatives if user has access to the customer(s) associated with the managed:

<img width="397" height="49" alt="Image" src="https://github.com/user-attachments/assets/d845b62d-d30b-48a6-aeab-ec01824b8518" />

3. The user sees that the managed is a causative on build 37 and checks in which cases inside scout 37


It seems easier to me..

[...]

PR #6205, opened 2026-04-15
Export causatives and verified variants to managed variants infile

### LINKED ISSUE #6207: Some managed variants fields in upload file are parsed incorrectly
While working on another PR I've been trying specifying the fields in the exact order reported in the guide:

Image

My file:

```
#chromosome;position;end;reference;alternative;category;sub_category;build;description;maintainer;institutes
17;7577559;7577559;G;A;cancer;snv;37;causatives 37 cust000;;cust000
7;124491972;124491972;C;A;snv;snv;37;causatives 37 cust000;;cust000
```

Result:

Image

### Cancer variant is loaded but not displayed at all. Saved document has some errors:

-  cancer should be cancer_snv
- Institute not parsed

Image

### SNV institutes field is not parsed

- Institute not parsed

Image

northwestwitch:
- Fixes partially #6098
- Fix #6207 

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,37 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+            category = "cancer_snv"
+
+        chromosome = variant.get("chromosome")
+        position = variant.get("position")
+        end = variant.get("end", position)
+        ref = variant.get("reference")
+        alt = variant.get("alternative")
+        sub_category = variant.get("sub_category")
+        build = variant.get("case_obj", {}).get("genome_build", "37")
+        description = f"{type} {build} {institute_id}"
```
northwestwitch, line 187: Or something else? Any suggestion is welcome!

scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
```diff
@@ -105,8 +105,8 @@ <h5>
           <td>{{ variant.chromosome }}</td>
           <td>{{ variant.position }}</td>
           <td>{{ variant.end }}</td>
-          <td>{{ variant.reference }}</td>
-          <td>{{ variant.alternative }}</td>
+          <td>{{ variant.reference | truncate(20, True, '...') }}</td>
+          <td>{{ variant.alternative | truncate(20, True, '...') }}</td>
```
northwestwitch, line 109: SVs can mess up the whole table if the have long strings here

northwestwitch:
This now more or less works. 

It still warns on weird variants like this one:

image


But in general does load most of them. I don't know if it's worth to modify the code to include the parsing of these weird ones or one could also modify the outfile for the few cases that fail?

scout/build/managed_variant.py
```diff
@@ -29,7 +29,7 @@ def build_managed_variant(managed_variant_info):
             alternative=managed_variant_info["alternative"],
             build=managed_variant_info.get("build", "37"),
             maintainer=managed_variant_info.get("maintainer", []),
-            institute=managed_variant_info.get("institute", ""),
+            institute=managed_variant_info.get("institutes", ""),
```
northwestwitch, line 32: This is the fix for #6207

docs/user-guide/variants.md
```diff
@@ -151,12 +151,12 @@ The columns that will be used by Scout are the following.
 - **end(int)** End position. *Optional*
 - **reference(str)** Reference sequence *Mandatory* Use for SVs is only exact.
 - **alternative(str)** Alternative sequence *Mandatory* Use for SVs is only exact.
-- **category(str)** One of "snv", "sv", "cancer" or "cancer_sv" *Mandatory*
+- **category(str)** One of "snv", "sv", "cancer_snv" or "cancer_sv" *Mandatory*
```
northwestwitch, line 154: This is a fix for #6207

docs/user-guide/variants.md
```diff
@@ -151,12 +151,12 @@ The columns that will be used by Scout are the following.
...
 - **sub_category(str)** Sub category of variant. For "snv" and "cancer" this is either "snv" or "indel". For "sv" and "cancer_sv" this is one of "ins", "del", "dup", "cnv", "inv" or "bnd". *Mandatory*
 - **build(str)** Defaults to "37". *Optional*
 - **description(str)** Free text description of the variant. Classification status, evidence and references are appropriate, and if applicable affected gene(s), protein consequences and alternative variant names. *Optional*
 - **maintainer(str)** Placeholder, replaced by the current uploading user. *Optional placeholder*
-- **institutes(list(str))** Placeholder, replaced with the current uploading user institutes. *Optional placeholder*
+- **institutes(str)** Comma-separated list of institutes. Replaced with the current uploading user's institute if empty. *Optional*
```
northwestwitch, line 159: No longer a placeholder

dnil: How about replaced with current user, unless admin? It feels strange to be able to change this on load when you can't with manual edit.

docs/user-guide/variants.md
dnil: 💯 
We should refactor this to have the same category name as the regular variants at some point. But I do like this one better. 🤔

scout/server/blueprints/institutes/static/variants_list_scripts.js
```diff
@@ -1,20 +1,28 @@
...
+      },
       buttons: [
           'copyHtml5',
           'excelHtml5',
+          {
+              text: 'Managed Variants infile',
+              className: 'btn btn-primary',
+              action: function () {
+              		$(globalThis).off('beforeunload');
+									globalThis.location.href = exportAsManagedUrl;
```
dnil, line 21: Nice! I realise I do not know that much about js encapsulation / namespaces. 😁

scout/server/blueprints/institutes/controllers.py
dnil:
```suggestion
        description = f"{type} {build} {chromosome}:{position}-{end} {ref}>{alt} {sub_category} {institute_id} "
```
Having the original build coordinates in the comment will be helpful when looking for old vars. Institute_id is a tiny bit of a new info leak, but we already show that it is for another institute, and these should be variants that we have anyway cleared to send to ClinVar, at which point the institute would also be known anyway. And so far, users appear fine with having this information. 
Maybe actually consider some pretty print function, as this will be subject to all the usual ref and alt too long issues etc.

scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
dnil: 💯

dnil (commented):
I'm a little worried the users will think they should start to export and import their own lifts, but maybe that is fine. We would just see it as a duplicate later. 🤷 

How about we make arbitrary institute ownership setting an admin level feature?
And let's find a good way to represent the variant in the comments, perhaps with a pretty print call or similar.

scout/server/blueprints/institutes/static/variants_list_scripts.js
northwestwitch: Good thing is ChatGPT knows more than us. Who knows if "globalThat" is also a thing?! 😆

scout/server/blueprints/institutes/controllers.py
northwestwitch: I don't think we should add complexity to this field. A short and clear description like:

`causatives 37 cust000` will make why the variant is there evident, no need for pretty print or other.

Moreover:
- Coordinates ref/alt and sub category are all fields that are already displayed on that same line.
- To keep the description consistent for all exported causatives, 37 is beneficial I think, as a list where all entries share the same description indicates that an export action has been performed.
- If we want to share more information about the matching causatives on the case page, a more complex description does not add value.

northwestwitch:
> I'm a little worried the users will think they should start to export and import their own lifts, but maybe that is fine. We would just see it as a duplicate later. 🤷

I think this is fine. Also, I don't think you'd be able to insert exactly the same managed variant twice, it's an upsert, no? See code: https://github.com/Clinical-Genomics/scout/blob/a9102585ba9990c380d2245961b92a103dc9ecf0/scout/server/blueprints/managed_variants/controllers.py#L154

> How about we make arbitrary institute ownership setting an admin level feature?

I don't understand. Could make a practical example?

> And let's find a good way to represent the variant in the comments, perhaps with a pretty print call or similar.

I don't agree that we should make a complicated description, see my other comment

docs/user-guide/variants.md
northwestwitch: Sure!

northwestwitch:
> > I'm a little worried the users will think they should start to export and import their own lifts, but maybe that is fine. We would just see it as a duplicate later. 🤷
> 
> I think this is fine. Also, I don't think you'd be able to insert exactly the same managed variant twice, it's an upsert, no? See code:
> 
> https://github.com/Clinical-Genomics/scout/blob/a9102585ba9990c380d2245961b92a103dc9ecf0/scout/server/blueprints/managed_variants/controllers.py#L154
> 

Yes, it's an upsert. It doesn't add any more new variants if you try to load the exported causatives from file twice:

image

docs/user-guide/variants.md
northwestwitch: To make sure I understood correctly:

- If the managed variants files that has to be imported has a list of institutes already, use that list
- If the field is missing, use user institutes (if user is not admin)
- if field is missing, do not save anything under institutes?

dnil: > To make sure I understood correctly:
> 
> * If the managed variants files that has to be imported has a list of institutes already, use that list
> * If the field is missing, use user institutes (if user is not admin)
> * if field is missing, do not save anything under institutes?

I would make that
* If the managed variants files that has to be imported has a list of institutes already, use that list but only set to institutes the user has access to. It's ok to leave institutes blank if none match. Admins can so import setting to any institute they like. 
* if field is missing, do not save anything under institutes

dnil:
> > How about we make arbitrary institute ownership setting an admin level feature?
> 
> I don't understand. Could make a practical example?
> 

E.g. user from cust001 tries to load managed variants set to have institute cust002 as owner. This makes very little sense to me. It's already fine to load the variants without institute.

docs/user-guide/variants.md
northwestwitch: > * if field is missing, do not save anything under institutes

Ah OK! Sure, simpler, and respects the content of the file. I like it

dnil:
> > And let's find a good way to represent the variant in the comments, perhaps with a pretty print call or similar.

> I don't agree that we should make a complicated description, see my other comment

Where was that? I'm just saying that it would be helpful to have the original build coordinates referenced when looking for the variant in the old instance. And that exactly what to print differs a little depending on the variant type e.g. the ref, alt and subtype of SVs, size of STRs etc.

northwestwitch:
> > > And let's find a good way to represent the variant in the comments, perhaps with a pretty print call or similar.
> 
> > I don't agree that we should make a complicated description, see my other comment
> 
> Where was that? I'm just saying that it would be helpful to have the original build coordinates referenced when looking for the variant in the old instance. And that exactly what to print differs a little depending on the variant type e.g. the ref, alt and subtype of SVs, size of STRs etc.

But then the link to the specific variant would be even more useful no?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,50 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+            variant_id=variant["_id"],
+        )
+        link = f'<a target="blank" rel="noopener noreferrer" href="{variant_href}">{variant["display_name"][0:15]}</a>'
+        description = f"{type} {build} {institute_id} {link}"
+
+        managed_lines.append(
+            f"{chromosome};{position};{end};{ref};{alt};"
+            f"{category};{sub_category};{build};{description};;{institute_id}"
+        )
+
```
northwestwitch, line 208: Description renders like this now - I'm not 100% convinced

Image

dnil:
> > Where was that? I'm just saying that it would be helpful to have the original build coordinates referenced when looking for the variant in the old instance. And that exactly what to print differs a little depending on the variant type e.g. the ref, alt and subtype of SVs, size of STRs etc.
> 
> But then the link to the specific variant would be even more useful no?

As per other conversation: yes, but more brittle. It would be one more thing we need to edit in case of an URL change. But doable, and as long as it is working, absolutely more useful!

northwestwitch:
> > > Where was that? I'm just saying that it would be helpful to have the original build coordinates referenced when looking for the variant in the old instance. And that exactly what to print differs a little depending on the variant type e.g. the ref, alt and subtype of SVs, size of STRs etc.
> > 
> > 
> > But then the link to the specific variant would be even more useful no?
> 
> As per other conversation: yes, but more brittle. It would be one more thing we need to edit in case of an URL change. But doable, and as long as it is working, absolutely more useful!

I've added the link. Give it a try when you have time

scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,50 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+        "cancer_sv": "variant.sv_variant",
+    }
+    managed_lines = [MANAGED_VARIANTS_INFILE_HEADER]
+
+    for variant in variants:
+        category = variant.get("category")
+        if category not in valid_categories:
+            continue
+
+        if category == "cancer":
```
dnil: Good to convert, but this has to happen later - or pehaps even keep them both alive as say 
`managed_variant_category` and `variant_category` to tell them apart at a glance?

scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,50 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+        chromosome = variant.get("chromosome")
+        position = variant.get("position")
+        end = variant.get("end", position)
+        ref = variant.get("reference")
+        alt = variant.get("alternative")
+        sub_category = variant.get("sub_category")
+        build = variant.get("case_obj", {}).get("genome_build", "37")
+
+        variant_href = url_for(
+            valid_categories[category],
```
dnil, line 191: This will break for `cancer_snv` now. 😊

scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,50 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+            institute_id=institute_id,
+            case_name=variant["case_obj"]["display_name"],
+            variant_id=variant["_id"],
+        )
+        link = f'<a target="blank" rel="noopener noreferrer" href="{variant_href}">{variant["display_name"][0:15]}</a>'
```
dnil: No `pretty_link` variant here... 😁 Well, I guess this works, but some truncated ones will look a little weird. The activity log has 1:30 fwiw.

scout/server/blueprints/institutes/controllers.py
dnil: The link makes searching for variants a lot faster for the users. It saves a bit of code complexity, and adds some other, including a bit of technical debt. Any way goes. Did you try it out on the case- and variant pages on matching?

scout/server/blueprints/institutes/views.py
```diff
@@ -102,6 +105,37 @@ def causatives(institute_id):
...
 
 
+@blueprint.route("/<institute_id>/<type>/get_managed_infile", methods=["GET"])
+def get_managed_infile(institute_id: str, type: str):
+    """Downloads a list of variants (causatives or verified) in the right format to be imported into the managed variants list."""
+    institute_obj = institute_and_case(store, institute_id)
+    if type == "causatives":
+        variants = controllers.causatives(institute_obj, request)
+    elif type == "verified":
+        variants = controllers.verified_vars(institute_obj)
```
dnil:
```suggestion
        variants = controllers.verified_vars(institute_obj["_id"])
```
This is a an unfortunate dissimilarity of the controller definitions. Took me a long time to understand. 😅

dnil (changes requested):
Yes, nice! I still don't see it working locally, but I think it is pretty near. One thing is the cancer/cancer_snv conversions that keeps bugging a little, the other export from `verified` which just gives me a header. Turns out the controller needs the institute_id. But it could do with a little testing also after those! 😊

scout/server/blueprints/institutes/controllers.py
northwestwitch: Yes, case page doesn't need any fix:

image


Variant page needed fixing and became like this:

image

scout/server/blueprints/institutes/views.py
northwestwitch: Thank you! 🙏🏻

scout/server/blueprints/institutes/controllers.py
northwestwitch: Right! I'm fixing it

northwestwitch: Done

scout/server/blueprints/institutes/controllers.py
northwestwitch: Well, actually the link on case page will wrongly point to the same instance (38). I need to fix it

northwestwitch:
1. Ok, latest changes lead to this type of description on Managed Variants, which I like:

image


2. I've just realised that on case page I don't have to do anything because the matching managed come from the 38 database, not the 37, as it's those in that case that will match the managed variants which we will lift to 38, so all good.

I'm ready for another review on this one

scout/server/blueprints/institutes/controllers.py
```diff
@@ -162,6 +163,52 @@ def decorate_institute_variant(variant_obj: dict) -> Optional[dict]:
...
+        build = variant.get("case_obj", {}).get("genome_build", "37")
+
+        variant_href = url_for(
+            valid_categories[category],
+            institute_id=institute_id,
+            case_name=variant["case_obj"]["display_name"],
+            variant_id=variant["_id"],
+            _external=True,
+        )
+        pretty_variant_name = current_app.jinja_env.filters["pretty_variant"](variant)
```
dnil, line 197: Sure, or cleaner, we add it to the python exported CustomFilters? I can do that in the other PR!

tests/server/blueprints/managed_variants/test_managed_variants_controllers.py
```diff
@@ -6,11 +6,8 @@
 def test_upload_managed_variants(adapter, managed_variants_lines):
     # when calling upload_managed_variants for these lines
     test_user = "noone"
-    test_institutes = ["cust000"]
 
-    result = controllers.upload_managed_variants(
-        adapter, managed_variants_lines, test_institutes, test_user
-    )
+    result = controllers.upload_managed_variants(adapter, managed_variants_lines, test_user)
```
dnil, line 10: 👍

tests/server/blueprints/institutes/test_institute_views.py
```diff
@@ -501,6 +501,64 @@ def test_causatives(app, user_obj, institute_obj, case_obj):
         assert var2_id not in str(resp.data)
 
 
+def test_get_managed_infile(app, user_obj, institute_obj, case_obj):
+    """Test the endpoint that download causative variants into a managed variants infile."""
+
+    # GIVEN a causative variant
+    VAR_ID = "4c7d5c70d955875504db72ef8e1abe77"
```
dnil: It's for sure better than pulling just the first variant and relying on it to be something specific. But couldn't we be a bit more precise about why this variant? Perhaps even add a fixture that retrieves this variant, but named and just one place to change if we ever update the files?

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -342,7 +342,7 @@
             {{ managed_variant.display_id }}
           </div>
           <div class="col-10">
-            {{ managed_variant.description }}
+            {{ managed_variant.description|safe }}
```
dnil, line 345: Ah. Interesting. So now the users would have a really complicated way to phish each other for passwords etc? But how solve with liftover links kept intact? 😭

dnil (approved):
Great, seems to work just fine! I note we will be making the user security disservice by adding the option to inject clickable links into the the managed variants collection, but we will perhaps just have to accept that for now, weighing it against practical use. Classic safety vs ease of use issue, perhaps?

tests/server/blueprints/institutes/test_institute_views.py
northwestwitch: It's the POT1 variant. It is also used in a test immediately above. Since it is only used on these 2 tests and it could literally be the _id of any SNV or SV, I will make it a constant on top of the page. It's kind of an overkill to turn it to a fixture.

scout/server/blueprints/institutes/controllers.py
northwestwitch: Thanks! We can fix this code later

tests/server/blueprints/institutes/test_institute_views.py
dnil: That's something at least! I just remember the pain of tests under changing load order; we might have that  soon again with new rank orders. Stuff like this would be painful if we move to another test file. Oh, well, it's not alone, and POT1 is great. 😆

PR #6209, opened 2026-04-16
Fix #6208 - singleton exception

### LINKED ISSUE #6208: Many questions about "Include variants present on unaffected" on singletons
While each question has had a reasonable conclusion (the variants in question were really called with GT ./.) it is surprising to have so many in a short period of time. I wonder if some refactoring has touched the singleton exception?

dnil:

- Fix #6208

This seems likely to have been around since 4.107, 260113.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:
1. On a singleton SV variants view, reset filter, uncheck panels. 
2. note the number of variants found with "Filter Include variants present only in unaffected" checked - should be all
3. uncheck "Filter Include variants present only in unaffected" notice the count go down with a handful of odd variants
4. apply patch
5. note all variants filtered in again, even with the toggle unchecked.

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -360,10 +360,10 @@ class QueryHandler(object):
         case_obj = self.case(case_id=case_id, projection=CASE_AFFECTED_INDS_PROJECTION)
         case_inds = case_obj.get("individuals", [])
 
-        gt_query = gt_query or {"$nin": ["0/0", "./.", "./0", "0/."]}
-
         if len(case_inds) == 1 and not gt_query:
             return
+
+        gt_query = gt_query or {"$nin": ["0/0", "./.", "./0", "0/."]}
 
         affected_query = {
             "$elemMatch": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - By default, display STR loci from VCF also when no call is made (#6197)
 ### Fixed
 - Formatting of a list on managed variant export documentation (#6203)
+- Exception for singletons from the "Include variants present only in unaffected" filter (#6209)
 
 ## [4.109.3]
 ### Fixed
--- a/tests/adapter/mongo/test_query.py
+++ b/tests/adapter/mongo/test_query.py
@@ -72,6 +72,20 @@ def test_build_query_hide_not_in_affected(adapter, case_obj):
     }
 
 
+def test_build_query_show_unaffected_singleton(adapter, singleton_case):
+    """Test variants query build with show_unaffected parameter for singletons"""
+
+    # GIVEN a singleton case (only one sample)
+    adapter.case_collection.insert_one(singleton_case)
+
+    # WHEN show_unaffected = True param is provided to the query builder
+    query = {"show_unaffected": False}
+    mongo_query = adapter.build_query(singleton_case["_id"], query=query)
+
+    # Then the variant query should not restrain sample id or genotype
+    assert "samples" not in mongo_query
+
+
 def test_build_query_hide_dismissed(adapter, case_obj):
     """Test variants query with hide_dismissed parameter"""
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -688,6 +688,15 @@ def case_obj(request, parsed_case):
 
 
 @pytest.fixture(scope="function")
+def singleton_case(case_obj):
+    """A fixture that returns a singleton object. Useful for testing that the same object is returned across multiple calls."""
+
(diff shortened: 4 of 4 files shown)


scout/adapter/mongo/query.py
```diff
@@ -360,11 +360,11 @@ def affected_inds_query(self, mongo_query, case_id, gt_query):
         case_obj = self.case(case_id=case_id, projection=CASE_AFFECTED_INDS_PROJECTION)
         case_inds = case_obj.get("individuals", [])
 
-        gt_query = gt_query or {"$nin": ["0/0", "./.", "./0", "0/."]}
```
northwestwitch, line 363: I don't doubt this solution works but I'd like to understand why.

Which genotypes these weird cases have and where in the code is the cause of the bug?

I'm looking at the code and I see that gt_query is set in [here](https://github.com/Clinical-Genomics/scout/blob/65216baa06b01eb23a468252f3cf36bb3eec691f/scout/adapter/mongo/query.py#L901)

Could you explain a bit?

northwestwitch (commented):
What is exactly causing the bug? See my comment

scout/adapter/mongo/query.py
dnil: The general idea of the hiding feature is to avoid accidentally seeing carrier status for variants in unaffected parents when the variant is not also inherited by the affected child. This is a rather useless check for a singleton though, and will only cause issues there, such as hiding poorly genotyped SVs or STRs.

We have two filter options that want to constrain sample genotype. This one, and then also the search-for-a-specific genotype filter option.

The idea of L363 (365 now) is to check if we have a singleton case, and if so ignore requiring the affected individual not being genotyped for the variant, but only if there is not a specific GT query.

Before the last bug fix to the gt_query filter, the check was fine (kind of, though it failed GT searches if the Include unaffected button was unchecked). But then the fallback for `gt_query` appeared BEFORE the singleton check L364, and ensuring that `gt_query` would always be `true`. The fix addresses that, restoring the check for singletons by filling in the unaffected filter at L366 instead.

northwestwitch: > The general idea of the hiding feature is to avoid accidentally seeing carrier status for variants in unaffected parents when the variant is not also inherited by the affected child. This is a rather useless check for a singleton though, and will only cause issues there, such as hiding poorly genotyped SVs or STRs.

That I understood, and that's why we had 

```
if len(case_inds) == 1 and not gt_query:
   return
```

> The idea of L363 (365 now) is to check if we have a singleton case, and if so ignore requiring the affected individual not being genotyped for the variant, but only if there is not a specific GT query.

Ahhh, this PR: https://github.com/Clinical-Genomics/scout/commit/2090430f661b7d47a690dc14636a2ad170507bca

I see why before it was working. OK, thanks!

scout/adapter/mongo/query.py
```diff
@@ -360,11 +360,11 @@ def affected_inds_query(self, mongo_query, case_id, gt_query):
...
-
         if len(case_inds) == 1 and not gt_query:
```
northwestwitch, line 363: Why not:

`if len(case_inds) == 1:`

??

northwestwitch (commented):
If you have only one individual, why even checking the value of the gt_query?

scout/adapter/mongo/query.py
dnil: Then we would be back to before the December bug fix, with the gt_filter being broken.
Again, two filter functions that want to add genotype query options. If the user selects to filter for "1/1" variants, this should work even if we are in a singleton.

northwestwitch: Right! Ok, you convinced me. And have you tested it on stage?

northwestwitch (approved):
Looks good! 👍🏻

scout/adapter/mongo/query.py
dnil: I tested with a couple of real cases, perhaps not on stage as such though. Will do.

dnil: Before this PR, the Include unaffected toggle would be needed to see all variants in a stage singleton.
Screenshots without Include unaffected untoggled (false).

:x: Before:
Screenshot 2026-04-17 at 08 56 43

✅ After:
Screenshot 2026-04-17 at 08 59 24

✅ The toggle still works for trios:
Screenshot 2026-04-17 at 09 03 25

✅ The GT filter works for singletons
Screenshot 2026-04-17 at 09 02 47

✅ The GT filter works for trios
Screenshot 2026-04-17 at 09 46 54

PR #6260, opened 2026-05-05
Hide empty categories of cases on caseS page

### LINKED ISSUE #5866: Respect cases searches on cases page
On cases page, do not show prio cases regardless

northwestwitch:
- Fix #5866 - Institutes with no prioritised cases will not see the category this way

### Main branch

image

### This branch

image

 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/institutes/templates/overview/cases.html
+++ b/scout/server/blueprints/institutes/templates/overview/cases.html
@@ -227,7 +227,7 @@
         {% set ordered_statuses = ['prioritized', 'inactive', 'active', 'archived', 'solved', 'ignored'] -%}
         {% for status in ordered_statuses %}
           {% for group_name, case_group in cases %}
-            {% if status == group_name and case_group%}
+            {% if status == group_name and case_group %}
               <div class="table-responsive">{{ cases_table(group_name, case_group) }}</div>
             {% endif %}
           {% endfor %}
--- a/scout/server/blueprints/institutes/templates/overview/cases.html
+++ b/scout/server/blueprints/institutes/templates/overview/cases.html
@@ -227,7 +227,7 @@
         {% set ordered_statuses = ['prioritized', 'inactive', 'active', 'archived', 'solved', 'ignored'] -%}
         {% for status in ordered_statuses %}
           {% for group_name, case_group in cases %}
-            {% if status == group_name %}
+            {% if status == group_name and case_group%}
               <div class="table-responsive">{{ cases_table(group_name, case_group) }}</div>
             {% endif %}
           {% endfor %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Default gnomAD linkout for hg38 to non_uk_biobank subset (#6247)
 - Handle conflicts and duplicates when uploading managed variants (#6256, #6258, #6259)
+- On caseS page, dynamically hide empty categories of cases (#6260)
 ### Fixed
 - Saltshaker report update command (#6246)
 - Managed variants counter, wrongly defaulting to build 37 (#6252)


scout/adapter/mongo/institute.py
```diff
@@ -99,11 +100,14 @@ def get_phenotype_groups() -> dict:
                 set(institute_obj.get("sanger_recipients", [])) - set([remove_sanger])
             )
 
+        if not show_all_cases_status:
+            show_all_cases_status = STATUS
+
```
northwestwitch: No choice == all statuses

dnil:
Im not sure I follow? This is kind of the opposite to the issue, no? Did you have some more discussions and reach another conclusion?

northwestwitch:
> Im not sure I follow? This is kind of the opposite to the issue, no? Did you have some more discussions and reach another conclusion?

No, it doesn't default any more to prioritised. And you can deselect everything now - which you couldn't do before

scout/adapter/mongo/institute.py
northwestwitch: I'm setting this to all statuses here because you can't have None or [] in settings parameters, because otherwise the key gets removed [here](https://github.com/Clinical-Genomics/scout/blob/a8291e2b74e0b9b09defb22ace1017d27ec10a4d/scout/adapter/mongo/institute.py#L125). When the key is missing it gets a default value to ['prioritised'] in the institute controllers and we're back to the initial issue. It's cumbersome to make a specific case just for that parameter to have it empty. And no selection is equivalent to all statuses selected anyway

northwestwitch: I should have perhaps specified that if you don't want to see prioritised among the statuses shown, you need to deselect it now, and select the rest of the statuses

dnil:
> > Im not sure I follow? This is kind of the opposite to the issue, no? Did you have some more discussions and reach another conclusion?
> 
> No, it doesn't default any more to prioritised. And you can deselect everything now - which you couldn't do before

But this would just mean prioritised are always shown anyway, along with everything else, and all will be flagged/dimmed as included outside query?

northwestwitch:
> But this would just mean prioritised are always shown anyway, along with everything else, and all will be flagged/dimmed as included outside query?

Ok, some questions/observations

1) We have to default to "prioritised" when institute doesn't have the key/values (most of our institutes are like this) because our users want to see the category - in fact the issue comes from some user outside Solna.

2) If you deselect everything you intrinsically accept that all categories are shown no? Because in case you don't want to see the prio cases then you select everything except that category

So, in order to fix the issue you might just do as in 2) without code changes - we could just close this PR and for me this is fine.

This PR just clarifies that if you don't select anything in that select and update the institute settings then you are going to see all categories (by selecting them all and updating your institute document accordingly)

Is there anything I am missing? Like weird stuff happening in the cases query, like selecting all prio cases even if the category is deselected on the settings? 

I'm accepting any suggestion here, included "leave it as it is and instruct to select everything except prioritised" 🤔

northwestwitch:
What could be done that would make a real difference, is that if you deselect one or more categories then you don't see at all the dedicated section on caseS page. That would be enough perhaps

northwestwitch:
@dnil see if you like it this way: a quick fix for institutes that don't have prioritised  cases and don't want to see the category is to hide it if empty. Simplest solution I think. The aesthetics of caseS improve as well because the other empty categories and not shown

dnil:
Yes, this is more reasonable to me. It reduces clutter at a slight expense of consistency. I have a feeling we will see an increase in dedicated searches for specific categories to ensure there were no cases there.

Also not quite sure it will solve the original issue, but we can try like this and ask back perhaps?

northwestwitch:
> Also not quite sure it will solve the original issue, but we can try like this and ask back perhaps?

👍🏻 If I remember correctly they are not using the category so it was annoying to always see it on top. I really think it might solve it for them

dnil:
So the thing with the proposed solution to 2) above was that it rather forced everything shown than allowed things shown. A possible solution would be to let None/all unchecked mean just that, and remove the exception for prioritized. We would then have to default all institutes to have "prioritized" selected in the settings, eg with a small db script, to keep the current behavior for our local institutes. It would have been a better design perhaps, but the change is a tad cumbersome.

dnil:
Ok, well, let's do this then. I thought they were just not happy with the graphical difference between search results and cases included for being always-shown, and would rather be able to completely disable the always-shown feature.

scout/server/blueprints/institutes/templates/overview/cases.html
```diff
@@ -227,7 +227,7 @@ <h5 class="modal-title" id="sangerModalLabel">Sanger validations to evaluate:</h
         {% set ordered_statuses = ['prioritized', 'inactive', 'active', 'archived', 'solved', 'ignored'] -%}
         {% for status in ordered_statuses %}
           {% for group_name, case_group in cases %}
-            {% if status == group_name %}
+            {% if status == group_name and case_group%}
```
dnil:
```suggestion
            {% if status == group_name and case_group %}
```

dnil (approved):
On the bottom line, I think this will be appreciated over all. Less confident it also solves the original issue, but perhaps!

PR #6300, opened 2026-05-18
Fix adding one managed variant in build 38

### LINKED ISSUE #6299: Impossible to add a single managed variant in build 38 at the moment
In the add managed variant form there is no option to select the build, which becomes '37' by default.

northwestwitch:
- Fix #6299 

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/managed_variants/controllers.py
+++ b/scout/server/blueprints/managed_variants/controllers.py
@@ -78,7 +78,7 @@ def managed_variants(request: LocalProxy) -> dict:
     managed_variants_query = store.managed_variants(
         category=categories,
         query_options=query_options,
-        build=request.form.get("build"),
+        build=filters_form.build.data,
         skip_count=skip_count,
         vars_per_page=VARS_PER_PAGE,
     )
@@ -119,6 +119,7 @@ def add_managed_variant(request: LocalProxy):
 
     managed_variant_obj = build_managed_variant(
         dict(
+            build=add_form["build"].data,
             chromosome=add_form["chromosome"].data,
             position=add_form["position"].data,
             end=add_form["end"].data,
--- a/scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
+++ b/scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
@@ -221,6 +221,7 @@
   <table class="table table-bordered table-hover">
     <thead>
       <tr>
+        <th>{{ add_form.build.label }}</th>
         <th>Chr</th>
         <th>{{ add_form.position.label }}</th>
         <th>{{ add_form.end.label }}</th>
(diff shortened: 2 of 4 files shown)


scout/server/blueprints/managed_variants/controllers.py
```diff
@@ -119,6 +119,7 @@ def add_managed_variant(request: LocalProxy):
 
     managed_variant_obj = build_managed_variant(
         dict(
+            build=add_form["build"].data,
```
dnil, line 122: 💯

scout/server/blueprints/managed_variants/controllers.py
```diff
@@ -78,7 +78,7 @@ def managed_variants(request: LocalProxy) -> dict:
     managed_variants_query = store.managed_variants(
         category=categories,
         query_options=query_options,
-        build=request.form.get("build"),
+        build=filters_form.build.data,
```
dnil, line 81: 🤔 Ok, if you say so - both should be the same in case of a regular filters form request, and I can't quite see a difference for the other two. I suppose you tested. In general, the crosstalk between the different forms is a little weird on this page, but I suspect it ever was for a shortcut to returning the just edited results after edit or add. There are plenty of uses of "direct" request.form["var"] in this function.

scout/server/blueprints/managed_variants/views.py
```diff
@@ -103,9 +103,6 @@ def add_managed_variant():
     controllers.add_managed_variant(request)
 
     return redirect(
-        url_for(
-            ".managed_variants",
-            **request.form,
-        ),
+        url_for(".managed_variants"),
```
dnil, line 106: Same here, not entirely sure this does much for the post requests, but should work just fine!

scout/server/blueprints/managed_variants/templates/managed_variants/managed_variants.html
```diff
@@ -221,6 +221,7 @@ <h5>
   <table class="table table-bordered table-hover">
     <thead>
       <tr>
+        <th>{{ add_form.build.label }}</th>
```
dnil, line 226: 💯

dnil (approved):
Yes, absolutely!

There are still a few slightly odd transitions on the page, from the shared form data, especially when working for a little while with both adding and editing variants, but they are understandable, and likely all from the preexisting design.

scout/server/blueprints/managed_variants/views.py
northwestwitch: It works fine, tested.

scout/server/blueprints/managed_variants/controllers.py
northwestwitch: I don't think it makes a difference in either case. After you insert a single new managed variant, the page returns just that - I think it's good to confirm that the process worked - then you can either reset filters or refresh the browser page to see them all.

Issue #6341, opened 2026-05-26
Loading regions (and possibly variants) fetches case panels for every variant category which is loaded


northwestwitch:
Getting panels every time is a waste of time - Let's collect them once, at the beginning of the loading. Also, why do we grab panels when loading the research category?

<img width="1281" height="754" alt="Image" src="https://github.com/user-attachments/assets/46953713-4d9c-4857-9944-363da86dde59" />

dnil:
It might also be a risk, especially with long research loading times, if versioned case panels are not locked in on the case obj, since panels could update between variant type fetches.

northwestwitch:
> It might also be a risk, especially with long research loading times, if versioned case panels are not locked in on the case obj, since panels could update between variant type fetches.

I think it still fetches the panel version associated to the case. I did a test before by adding a gene to a panel hoping that it would be picked by the load region, but the script was still fetching the previous panel version

dnil:
Yes, concur - the case obj with versions is set first, then the panels-to-genes gets built.

northwestwitch:
Fixed by #6350

PR #6350, opened 2026-05-27
Cache per-case genomic info when loading variants

northwestwitch:
- A non-invasive fix to #6341

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR


northwestwitch:
### Main branch

image



### This branch

image

scout/adapter/mongo/variant_loader.py
```diff
@@ -660,6 +660,34 @@ def _has_variants_in_file(self, variant_file: str) -> bool:
             LOG.warning("Variant file %s does not include any variants", variant_file)
             return False
 
+    def _build_cached_genomic_info(self, case_obj: dict, build: str) -> dict:
+        """Cache per-case gene_to_panels, hgncid_to_gene and hgncid_to_gene info."""
```
dnil: 💯

scout/adapter/mongo/variant_loader.py
```diff
@@ -660,6 +660,34 @@ def _has_variants_in_file(self, variant_file: str) -> bool:
...
+        if not hasattr(self, "_case_cache"):
+            self._case_cache = {}
+
+        case_id = case_obj["_id"]
+        key = (case_id, build)
+
+        if key not in self._case_cache:
+            self._case_cache[key] = self._build_cached_genomic_info(case_obj, build)
+
+        return self._case_cache[key]
```
dnil: Consider using setdefault instead: should cover most of this, no?

scout/adapter/mongo/variant_loader.py
```diff
@@ -697,11 +725,12 @@ def load_variants(
...
 
-        gene_to_panels = self.gene_to_panels(case_obj)
-        build = build if build else get_case_genome_build(case_obj)
-        genes = list(self.all_genes(build=build))
-        hgncid_to_gene = self.hgncid_to_gene(genes=genes, build=build)
-        genomic_intervals = self.get_coding_intervals(genes=genes, build=build)
+        build = build or get_case_genome_build(case_obj)
+        cache = self._get_cached_genomic_info(case_obj=case_obj, build=build)
+
+        gene_to_panels = cache["gene_to_panels"]
```
dnil, line 772: 💯

dnil (approved):
Nice with the cache on the adapter! Maybe we should still clear the cache after each case is done, like on the load research loop - it could potentially be many cases, consuming a bit of memory.

There was one helper function that looks like what python `setdefault` does, but maybe I'm missing something? It should work well at any rate!

scout/adapter/mongo/variant_loader.py
northwestwitch: You are suggesting something like this I guess

```
return self._case_cache.setdefault(
        key,
        self._build_cached_genomic_info(case_obj, build),
    )
```

For how Python works, I am afraid this won't do. setdefault is a bit problematic and will make `self._build_cached_genomic_info `run every time, even if the key is already set. See this post: 

https://stackoverflow.com/questions/22797251/why-does-setdefault-evaluate-default-when-key-is-set

northwestwitch:
> Nice with the cache on the adapter! Maybe we should still clear the cache after each case is done, like on the load research loop - it could potentially be many cases, consuming a bit of memory.

But we load one case at the time via CLI no? So the cache is clear after each run. And of course the cache should persist when you call load_variants several times for each case. Or perhaps I am missing something? 🤔

scout/adapter/mongo/variant_loader.py
dnil: Ah, right, I see. It is a tiny bit weird given the many lazy late instantiations in python, but it does make sense. I guess you could still use what they suggested on that post, like a chained or checking for existence of the cache and the key. But that will be quite a mouthful to read. The present code is fine!

dnil:
> > Nice with the cache on the adapter! Maybe we should still clear the cache after each case is done, like on the load research loop - it could potentially be many cases, consuming a bit of memory.
> 
> But we load one case at the time via CLI no? So the cache is clear after each run. And of course the cache should persist when you call load_variants several times for each case. Or perhaps I am missing something? 🤔

Scout load research should load all cases with pending research in one swoop. Let me check..

Yes, like here, loop, calling load_variants repeatedly (for each case)
https://github.com/Clinical-Genomics/scout/blob/d8ff563272f6f91ec161a514d0cfa677ddb2eae3/scout/commands/load/research.py#L91

northwestwitch:
> Yes, like here, loop, calling load_variants repeatedly (for each case)

Ah yes, quite right! I will fix. But perhaps clear case-related panel cache, but persist the other info, given that the build is still the same. I'll think of something! Good suggestion!

scout/adapter/mongo/variant_loader.py
```diff
@@ -660,6 +660,75 @@ def _has_variants_in_file(self, variant_file: str) -> bool:
...
+        """
+        Retrieve or compute case-specific gene-to-panel mapping.
+
+        This cache is scoped per case_id and is automatically cleared
+        when a new case is processed. Only the most recent case data
+        is kept in memory.
+        """
+        case_id = case_obj["_id"]
+
+        if not hasattr(self, "_last_case_id") or self._last_case_id != case_id:
```
northwestwitch, line 703: This should clean the cache of gene_to_panels every time that a case changes

PR #6351, opened 2026-05-27
Litvar links

### LINKED ISSUE #6268: Add LitVar2 search-button for SNV/Indels?
Can we add a search-button for easy use of LitVar2 (https://www.ncbi.nlm.nih.gov/research/litvar2) in Scout, to search for a particular variant (Clinical SNVs and INDELs) from its variant page in Scout? I think the tool is very useful to search the biomedical litterture for mentions of a particular variant (including in Tables and Supplements). It appears to be free to use and they provide API-connectivity (https://www.ncbi.nlm.nih.gov/research/litvar2/api). I am not sure which API-option is the best to match what you get if you do a manual search (so please give feed-back). I have used manual searching using the formats suggested on their manual search-page (https://www.ncbi.nlm.nih.gov/research/litvar2) - this however takes some time of copy & paste and I also think that if we have a button on the variant-page it would allow all users an easy acces to LitVar2 (the button could be placed as suggested in the image below, a blue button to the right side of the dbSNP-button, or where you think is a good placement on the variant page).

Image

dnil:

- Fix #6268 - add links to LitVar2 for all dbSNP ids, and gene symbol search

API calls to search for all known variants in a gene, or to check for the availability of the variant in question, e.g. treating the link as a placeholder and removing the link if unavailable. 

If there is no dbSNP id, gene-symbol autocomplete to get the first available rsID for linking (as intended in the API - the NCBI page does something very similar): 
Screenshot 2026-05-27 at 18 44 31
Screenshot 2026-05-27 at 18 44 27

If there is a dbSNP id, cheat a little by creating the expected deep link, but also check it afterwards with the Litvar sensor API, and remove the link if it doesnt exist - or add some nice info if it does.
Screenshot 2026-05-27 at 18 44 47

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN


scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -433,46 +434,9 @@ <h6 class="font-weight-bold">{{ image.title }}</h6>
     set_scrolly_table('proteins_panel_table')
     set_scrolly_table('transcripts_panel_table')
 
-    function update_gene_has_full_coverage() {
```
dnil, line 436: moved this to components as well

scout/server/blueprints/variant/templates/variant/buttons.html
```diff
@@ -11,6 +11,16 @@
   {% if variant.dbsnp_id %}
     {% for snp in variant.dbsnp_id.split(';') %}
       <a class="btn btn-sm btn-info text-white me-1 mb-1"  target="_blank" rel="noreferrer noopener" href="{{variant.snp_links[snp]}}">dbSNP ({{ snp }})</a>
+      {% if variant.litvar_snp_links and snp in variant.litvar_snp_links %}
```
dnil, line 14: Actual link will always be as rendered from links.py. The question becomes is it available for this rsID?

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+      if (existingTooltip) {
+        existingTooltip.dispose();
+      }
+      element.setAttribute('title', tooltipText);
+      element.setAttribute('data-bs-title', tooltipText);
+      element.setAttribute('data-bs-toggle', 'tooltip');
+      new bootstrap.Tooltip(element, {container: 'body'});
+    }
+
+    function verify_litvar_snp_ids() {
```
dnil: New script, moved to components, for checking if there is a link available. Runs after load completes, and removes link if it wasn't ok, or updates it with a bit of tooltip info if it was.

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
       }
     }
 
+    function set_tooltip_content(element, tooltipText) {
```
dnil: Tooltips are added to Bootstrap before load completion. Since this is run after, we have to do a little dance to get them on the list.

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+      element.setAttribute('data-bs-title', tooltipText);
+      element.setAttribute('data-bs-toggle', 'tooltip');
+      new bootstrap.Tooltip(element, {container: 'body'});
+    }
+
+    function verify_litvar_snp_ids() {
+      // Make CORS-enabled request to LitVar API by using a Scout api function for it
+      // Verify LitVar links rendered with an explicit rsID marker.
+      const snpLinks = document.querySelectorAll('.litvar-snp-link[data-litvar-rsid]');
+      const litvarSensorUrlTemplate = '{{ url_for("variant.litvar_sensor", rsid="__RSID__") }}';
```
dnil: This is the url for the new Scout endpoint. Calls made directly in js here can run into CORS issues, since they would originate with the users browser.

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+            if (result.ok && result.data.available === false) {
+              const existingTooltip = bootstrap.Tooltip.getInstance(link);
+              if (existingTooltip) {
+                existingTooltip.dispose();
+              }
+              link.remove();
+              return;
+            }
+
+            console.warn('Could not verify LitVar ID ' + rsid + ', keeping fallback link.');
```
dnil: It should not happen too often I hope - and it is potentially an issue if we can't reach them.

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+          .then(function(response) {
+            return response.json().then(function(data) {
+              return { ok: response.ok, data: data };
+            });
+          })
+          .then(function(result) {
+            if (result.ok && result.data.available && result.data.link) {
+              const tooltipText = 'Click to view ' + rsid + ' on LitVar2 (' + result.data.pmids_count + ' articles)';
+              link.href = result.data.link;
+              set_tooltip_content(link, tooltipText);
```
dnil: if its ok, update tooltip

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+              return;
+            }
+
+            if (result.ok && result.data.available === false) {
+              const existingTooltip = bootstrap.Tooltip.getInstance(link);
+              if (existingTooltip) {
+                existingTooltip.dispose();
+              }
+              link.remove();
```
dnil: If call was ok, and Litvar does not have the variant, remove the link (and tooltip).

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -628,9 +628,111 @@
...
+          })
+          .catch(function(error) {
+            console.warn('Could not verify LitVar ID ' + rsid + ', keeping fallback link.', error);
+          });
+      });
+    }
+  </script>
+{% endmacro %}
+
+{% macro variant_coverage_scripts(institute, case, variant) %}
```
dnil, line 553: I just moved it here to keep the main html uncluttered.

dnil (commented):
Thank you co-pilot - I would have stopped earlier without you.

northwestwitch (commented):
Please don't get this the wrong way but the code of this PR is seems hard to review and thinking about how the JS interacts also with the other components of the page, to maintain.

I see the difficulty you are facing here to try to build the links to litvar2: even if the dbSNP id is available the variant might be missing in the portal, if the dbSNP is missing then is a mess. But can we try to have this code handled by a Python module instead, which creates these links (after API calls and cheks) when called by the variant controller. Then the link might be visualized only if exists without complicating the rendering of the page.

The idea has always been that the heavy computation is done before the template is rendered..

dnil:
> Please don't get this the wrong way but the code of this PR is seems hard to review and thinking about how the JS interacts also with the other components of the page, to maintain.
> 
> I see the difficulty you are facing here to try to build the links to litvar2: even if the dbSNP id is available the variant might be missing in the portal, if the dbSNP is missing then is a mess. But can we try to have this code handled by a Python module instead, which creates these links (after API calls and cheks) when called by the variant controller. Then the link might be visualized only if exists without complicating the rendering of the page.
> 
> The idea has always been that the heavy computation is done before the template is rendered..

Nono, no worries - it is as foreseen from that API together with CORS limitations. 

It is not my battle really - I always found the API clunky. But I decided it wasn't that bad once I found the pattern in the deep link if there is an rsID. I now slightly regret doing this in one step - fairly confident three PRs would have been a lot easier to review.

I will try to convince you that given the API and the limitations, this is actually a somewhat sound solution.

First, the actual initial variant link based on any available rsID is generated in python. It is removed if the js module confidently interacts with the API. No link will be generated if there is no rsID, and no query fuss.

Not having an id is apparently a small limitation: all LitVar variants appear to have one.

Second, the API queries are also made in python. Notice how the URLs called from js are actually for our own Scout API endpoint. It is all modelled after how we fetch complete coverage status from chanjo2. CORS is much easier then, as the data is seen as internal to the app page.

Third, and the crucial point, adding js to call it after load is reasonable, since it saves waiting time on the page, in particular if the NCBI API would be slow on some days. I quite liked having an initial placeholder link that may work, being removed later if it clearly will not. 

The gene link is very similar, but does add complexity to the PR. Queries are actually made in python, and the js is there mainly  to postpone the query to after load, but here also to make the actual link and tooltip. It might be possible to have a better initial link guess here, but the whole idea is that we then do not know any rsID, so it is unclear where to point?

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -510,6 +510,16 @@
         {% if gene.pubmed_link %}
             <a href="{{ gene.pubmed_link }}" class="btn btn-secondary text-white" rel="noopener" referrerpolicy="no-referrer" target="_blank">PubMed</a>
         {% endif %}
+        {% set litvar_gene_query = gene.common.hgnc_symbol if gene.common else gene.hgnc_symbol or gene.hgnc_id %}
```
northwestwitch: I'm not really sure about HGNC ID (at the end). I think only gene symbol is used when it comes to significant results

scout/server/blueprints/variant/views.py
```diff
@@ -525,6 +531,106 @@ def ccv():
     return jsonify({"classification": classification, "conflicts": ccv_conflicts, **ccv_bayesian})
 
 
+@variant_bp.route("/api/v1/litvar/sensor/<rsid>")
+def litvar_sensor(rsid):
```
northwestwitch, line 536: Why don't we use the JSON format instead of the API format?

API format: https://www.ncbi.nlm.nih.gov/research/litvar2-api/sensor/rs116916706?format=api
JSON format: https://www.ncbi.nlm.nih.gov/research/litvar2-api/sensor/rs116916706?format=json

That way you don't have CORS problems and you get immediately what you are looking for instead of jsonify the response

scout/server/blueprints/variant/views.py
```diff
@@ -525,6 +531,106 @@ def ccv():
...
+            "rsid": rsid,
+            "link": link,
+            "pmids_count": litvar_payload.get("pmids_count", 0),
+        }
+    )
+
+
+@variant_bp.route("/api/v1/litvar/autocomplete")
+def litvar_autocomplete():
+    """Resolve a LitVar deep link from an autocomplete query using the first rsID match."""
```
northwestwitch, line 579: LitVar2 is apparently is supposed to be very good at understanding which variant it is given its HGVS as well (given that it is a variant in build 38.)

You don't have to do this way, but consider a logic of this type:

- Variant has rsID: fine, use it in autocomplete
- Variant has no rsID, use HGVS + gene symbol (if build is 38)
- Boh!

**Same as before, I'd use the json format in this query, so you avoid all the status code checks and CORS problems**

scout/server/blueprints/variant/templates/variant/components.html
northwestwitch: I'd avoid the issue by querying for json data (see my other comment). Also, I wonder if it's a bit of an overkill do check every single result returned by the sensor..

northwestwitch (commented):
Ok I see your point.

> First, the actual initial variant link based on any available rsID is generated in python. It is removed if the js module confidently interacts with the API. No link will be generated if there is no rsID, and no query fuss.

> I will try to convince you that given the API and the limitations, this is actually a somewhat sound solution.

I don't doubt that it's a sound solution, but looks a bit like an overkill to me?

> First, the actual initial variant link based on any available rsID is generated in python. It is removed if the js module confidently interacts with the API. No link will be generated if there is no rsID, and no query fuss.

Nice. I've seen that you send a request to the sensor (I think?) to make sure that the rsID is there. I agree that it is quicker if you send this request after the variant page is loaded, BUT: the code that does this is quite complicated. I wonder if:

1. We could just create a link for the rsID, and if it doesn't return a result it's bad bad frankly it wouldn't be the first time a link from the variant page results to a page with empty info.
2. One could send a request to the sensor while creating the link - slower, I agree, but it would simplify the code A LOT - also considering that you could run a query to return the json data directly, instead of having CORS issues.

Also LitVar is supposed to be very good at disambiguating variants given that you have the HGVS. This could be used for all 38 variants which don't have a rsID. One could search gene symbol and HGVS instead.

To recap: I don't know, this solution still looks a bit of an overkill to me. I like the fact the check is done when the page is loaded, but I'm not convinced by the queries happening in the new endpoints (do we really these new endpoints at all?). I would simplify and build rudimental queries (as we have for other cases) containing rsIDs if available or Gene symbol+HGVS (if variant os 38) and then eventually check via sensor, but in Python. how much time would a simple request would delay the page loading? 

I guess it's a tradeoff: is the performance benefit big enough to justify the new architectural complexity?

scout/server/blueprints/variant/templates/variant/components.html
dnil: No, you are right - that does not seem to bring much!

scout/server/blueprints/variant/views.py
dnil: 🤔 I don't follow you here? Why would we be retrieving their web page rather than json? Did I manage to do that when I tested the direct from js retireve?

scout/server/blueprints/variant/views.py
dnil: Yes, I did want to get the hgvs in, but did not quite manage to find a use for it. I figured it would be a relatively rare occasion that the variant did not have a known rsID, but there would be one for it in Litvar. I guess missense that lead to the same AA change would be the exceptions, and the user would have to be careful to interpret that. But lets try!

And surely we are using the json? Otw nothing should work if there was a bunch of html coming our way as the response?

dnil (commented):
Thank you for reviewing! I'll take another look and see what I can simplify without loosing  much. I did not get the part about json responses? Surely we are not retrieving their  Swagger - that should bug all over the place?

northwestwitch:
> I don't follow you here? Why would we be retrieving their web page rather than json? Did I manage to do that when I tested the direct from js retireve?

No right, it is json indeed. I was thinking about how to simplify. How about just calling sensor for variants with rsIDs, and if payload has a link use that as a link for the button? 

It should cover most of the cases no?

scout/server/blueprints/variant/views.py
dnil: Ok, good. I tried it out, and it becomes very picky if one adds both gene symbol and an hgvsp string, returning nothing unless it is an exact match. Symboll plus hgvsc replies get really weird when they are more complex. We could do it multistage, like keep firing different progressively less specific queries until we get a reply with an rsID, but I feel this is not super useful compared to just getting either the exact rsID or any with the gene symbol for further browsing.

northwestwitch: Ok, forget about it. How about just using sensor with the rsid and if the link exists use that

dnil: I removed most error handling redundancy now; still sticking to differentiating between if there was a real hit (show link), a confirmed miss (remove link), or just a general error/timeout/whatnot (keep placeholder).  We could skip the error handling completely and never remove. I dunno, I think it would be a useful service to the user, and I would love to be able to do it for more resources. Having to click to confirm existence also takes user time, especially for a resource that if available could be important to check. Like here: are there papers describing it functionally already? We'd have to have a look! But for the most part, that would  not be the case, and after a hundred non-existing replies I would probably just not click it that much anyway.

scout/server/blueprints/variant/templates/variant/components.html
```diff
@@ -510,6 +510,15 @@
...
         {% endif %}
+        <a
+          href="#"
+          class="btn btn-secondary text-white disabled litvar-gene-link"
+          rel="noopener"
+          referrerpolicy="no-referrer"
+          target="_blank"
+          aria-disabled="true"
+          data-litvar-gene-query="{{ gene.common.hgnc_symbol if gene.common else gene.hgnc_symbol }}"
+        >LitVar2</a>
```
northwestwitch, line 504: 👍🏻

scout/server/blueprints/variant/templates/variant/components.html
northwestwitch: Javascript code is becoming more and more inside this HTML page. Consider moving everything that is movable (stuff that doesn't contain direct Jinja2 code for instance, or where the jinja variable can be passed from the template instead) to a new static/script.js file, same as we have in other modules - see for instance cases or variants blueprints

scout/server/blueprints/variant/templates/variant/components.html
northwestwitch: Perhaps also this script can be moved to a scripts.js file if url_for("variant.litvar_sensor" can somehow be passed or made available to it?

scout/server/blueprints/variant/templates/variant/variant.html
```diff
@@ -433,46 +434,10 @@ <h6 class="font-weight-bold">{{ image.title }}</h6>
...
-      }).fail(function() {
-          const complete_coverage_indicator_elements = document.getElementsByClassName("complete-coverage-indicator");
-          for(let complete_coverage_indicator of complete_coverage_indicator_elements) {
-            complete_coverage_indicator.style.display = "none"
-          }
-      });
-    };
-
-    $( window ).on( "load", function() { update_gene_has_full_coverage(); })
+    globalThis.addEventListener('load', function() {
```
northwestwitch: 👍🏻

scout/server/links.py
```diff
@@ -729,6 +730,22 @@ def beacon_link(variant_obj: dict, build: int):
...
+    """LitVar SNP links from dbSNP rsIDs"""
+
+    if variant_obj.get("dbsnp_id") is None:
+        return
+    litvar_snp_links = {}
+    snp_ids = variant_obj["dbsnp_id"].split(";")
+    for snp in snp_ids:
+        if "rs" in snp:
+            litvar_snp_links[snp] = (
+                f"https://www.ncbi.nlm.nih.gov/research/litvar2/docsum?variant=litvar%40{snp}%23%23&query={snp}"  # litvar direct link, if present
```
northwestwitch:
```suggestion
                f"https://www.ncbi.nlm.nih.gov/research/litvar2/docsum?variant=litvar%40{snp}%23%23&query={snp}"  # litvar direct link, will be verified during variant page loading
```

northwestwitch: Something like that

northwestwitch (approved):
This works and it's fine to integrate it. I still feel the feature is very useful but I don't like the added code complexity - I still think it's an overkill with the postponing of the tooltips loading, all the checks if the link is valid (I think it will be valid most of the times, at least from the tests I've runned with this PR). It is fine if you want to add the code as it is, I guess, but perhaps try to move some js code to a dedicated js file if it's doable, see also my comment in the code.

scout/server/blueprints/variant/templates/variant/components.html
dnil: Mm, they basically all have some jinja, except for the pure import links, but I guess we could make the URLs etc in the macro and pass as arguments to the js. It would make the editors happier I guess! 😁 I'll give it a try.

PR #6353, opened 2026-05-29
Fix #6295 - filter dark regions/sma page for gene panels

### LINKED ISSUE #6295: Panel filter for Dark regions
Paraphase regions are currently shown regardless if the associated genes are on the case panel or not. It would be nice to be able to start viewing only clinical loci, then expand as needed.

dnil:

- Fix #6295. Case gene panels, with clinical default, dynamic HPO, and institute always-available panels.
Make the match on hgnc_id, though the Paraphrase regions are still only with gene symbols, mostly for future safety. 
Hides or shows the SMA table as well, depending on if SMN1/SMN2 are on the panel.
Add a shortcut to the HPO/dynamic list from the case page as well.

Screenshot 2026-06-01 at 21 21 58

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by DN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
+
+### Added
+- Gene panel filters to the Dark regions page
 ### Changed
 - Replaced Ensembl rest liftover service with liftover API from the Broad Institute (#6293)
 - Avoid fetching genes and panels multiple times when loading variants (#6350)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
-
 ### Added
 - LitVar2 links: variant link for variants with dbSNP rsID and a gene symbol search link (#6326)
 - Filter variants by OMIM inheritance pattern (#6355)
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -365,16 +365,6 @@ class PanelHandler:
         gene_list = [gene_obj.get(gene_format, "") for gene_obj in panel_obj.get("genes", [])]
         return gene_list
 
-    def case_panels_to_genes(
-        self, panel_names: List[str], gene_format: str = "hgnc_id"
-    ) -> Set[Union[int, str]]:
-        """Return unique genes from selected panel names.
-        For panels present on the case, and other selectable panels (e.g. institute panels),
-        fall back to latest version.
-        """
-
-        return set(self.panels_to_genes(panel_names=panel_names, gene_format=gene_format))
-
     def update_panel(self, panel_obj, version=None, date_obj=None, maintainer=None):
         """Replace a existing gene panel with a new one
 
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -366,19 +366,14 @@ class PanelHandler:
         return gene_list
 
     def case_panels_to_genes(
-        self, case_obj: dict, panel_names: List[str], gene_format: str = "hgnc_id"
+        self, panel_names: List[str], gene_format: str = "hgnc_id"
     ) -> Set[Union[int, str]]:
-        """Return unique genes from selected panel names, preferring case panel versions.
-
-        For panels present on the case, use the version recorded for that case.
-        For other selectable panels (e.g. institute panels), fall back to latest.
+        """Return unique genes from selected panel names.
+        For panels present on the case, and other selectable panels (e.g. institute panels),
+        fall back to latest version.
         """
 
-        case_panel_names = {panel["panel_name"]: panel for panel in case_obj.get("panels", [])}
-
-        panel_genes = self.panels_to_genes(panel_names=case_panel_names, gene_format=gene_format)
-
-        return set(panel_genes)
+        return set(self.panels_to_genes(panel_names=panel_names, gene_format=gene_format))
 
     def update_panel(self, panel_obj, version=None, date_obj=None, maintainer=None):
         """Replace a existing gene panel with a new one
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -4,7 +4,7 @@ import datetime as dt
 import logging
 import math
 from copy import deepcopy
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, List, Optional, Union
 
 import pymongo
 from bson import ObjectId
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -365,6 +365,21 @@ class PanelHandler:
         gene_list = [gene_obj.get(gene_format, "") for gene_obj in panel_obj.get("genes", [])]
         return gene_list
 
+    def case_panels_to_genes(
+        self, case_obj: dict, panel_names: List[str], gene_format: str = "hgnc_id"
+    ) -> Set[Union[int, str]]:
+        """Return unique genes from selected panel names, preferring case panel versions.
+
+        For panels present on the case, use the version recorded for that case.
+        For other selectable panels (e.g. institute panels), fall back to latest.
+        """
+
+        case_panel_names = {panel["panel_name"]: panel for panel in case_obj.get("panels", [])}
+
+        panel_genes = self.panels_to_genes(panel_names=case_panel_names, gene_format=gene_format)
+
+        return set(panel_genes)
+
     def update_panel(self, panel_obj, version=None, date_obj=None, maintainer=None):
         """Replace a existing gene panel with a new one
 
@@ -545,37 +560,6 @@ class PanelHandler:
             ]
         )
         return set(item["_id"] for item in query_result)
-
-    def case_panels_to_genes(
-        self, case_obj: dict, panel_names: List[str], gene_format: str = "hgnc_id"
-    ) -> Set[Union[int, str]]:
-        """Return unique genes from selected panel names, preferring case panel versions.
-
-        For panels present on the case, use the version recorded for that case.
-        For other selectable panels (e.g. institute panels), fall back to latest.
-        """
-
-        genes: Set[Union[int, str]] = set()
-        case_panels = {panel["panel_name"]: panel for panel in case_obj.get("panels", [])}
-
-        for panel_name in panel_names:
-            panel_obj = None
-            case_panel = case_panels.get(panel_name)
-            if case_panel is not None:
-                panel_obj = self.gene_panel(panel_name, version=case_panel.get("version"))
-            if panel_obj is None:
-                panel_obj = self.gene_panel(panel_name)
-            if panel_obj is None:
-                continue
-
-            panel_genes = {
-                gene.get(gene_format)
-                for gene in panel_obj.get("genes", [])
-                if gene.get(gene_format) is not None
-            }
-            genes.update(panel_genes)
-
-        return genes
 
     def search_panels_hgnc_id(self, hgnc_id: int) -> list[dict]:
         """Return all panels and versions that contain given gene,list is sorted"""
--- a/scout/adapter/mongo/panel.py
+++ b/scout/adapter/mongo/panel.py
@@ -4,7 +4,7 @@ import datetime as dt
 import logging
 import math
 from copy import deepcopy
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Set, Union
 
 import pymongo
 from bson import ObjectId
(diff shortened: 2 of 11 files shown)


scout/adapter/mongo/panel.py
```diff
@@ -546,6 +546,37 @@ def clinical_hgnc_ids(self, case_obj):
         )
         return set(item["_id"] for item in query_result)
 
+    def case_panels_to_genes(
```
northwestwitch: This function is very similar to [this one](https://github.com/Clinical-Genomics/scout/blob/fd371648ea7ee9b58ea9e06d2c5176bf9d7a9daf/scout/adapter/mongo/panel.py#L291). Would it be possible to reuse the existing code from that function (perhaps with small fixes?) to avoid repetitions? The output of the existing function is a dictionary, but we could just use the keys, and it would be the same as the output of this one

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -78,12 +86,51 @@
...
+  {% if sma_filter_form.gene_panels.choices %}
+    <div class="container_spaced col-12">
+      <div class="card">
+        <div class="card-title">Filter SMN - Dark regions</div>
+        <div class="card-body">
+          <form class="row g-2" method="get" action="{{ url_for('cases.sma', institute_id=institute._id, case_name=case.display_name) }}">
+            {{ sma_filter_form.panel_filter_applied(value="1") }}
+            <div class="col-md-8">
+              <label class="form-label mb-1" for="gene-panel-filter">Gene panels</label>
+              {{ sma_filter_form.gene_panels(
```
northwestwitch, line 118: This select could be half this long and still show quite a bunch of selected panels I think:

Image

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -78,12 +86,51 @@
...
+                id="gene-panel-filter",
+                class="selectpicker",
+                data_style="btn-secondary",
+                data_width="100%",
+                data_live_search="true",
+                title="Select gene panel(s)"
+              ) }}
```
northwestwitch, line 125: Also, the select has case panels pre-selected - I'm not sure it's good or not. What if the user wants to see all regions first and then in case apply the filter?

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -78,12 +86,51 @@
 </div>
 {% endblock %}
 
+{% macro panel_filter_card() %}
+  {% if sma_filter_form.gene_panels.choices %}
+    <div class="container_spaced col-12">
+      <div class="card">
+        <div class="card-title">Filter SMN - Dark regions</div>
```
northwestwitch: Instead of this new format of the panel, for consistency with the rest of the portal, I'd just copy the existing expandable one of the other variants. For now it's just the panels filter, but there will be also other filters to add to this panel in the future, probably

Image

scout/server/blueprints/cases/controllers.py
```diff
@@ -390,11 +420,15 @@ def _update_case_region(case_region: dict, region: dict, individual_is_affected:
...
 
-def _get_paraphrase_regions(case_obj: dict) -> dict:
+def _get_paraphrase_regions(case_obj: dict, panel_hgnc_ids: Set[int] | None = None) -> dict:
     """Check all case individuals for paraphrase region information.
     Move fixed global attributes (genes_in_region, phase_region) for display up from the individual level to case region
     level, which is returned.
+
     Set the status of the region to the highest status on any affected individual.
+
+    If given a set of panel hgnc ids, match genes in regions with this gene selection, and return a dict with regions left
```
northwestwitch, line 431: Pity that we don't turn these values in a list of HGNC IDs:

Image

Because we could have used the key/values in a direct query instead .. 🤔

northwestwitch (commented):
Nice new feature!

The overall feeling I have when I look at the new filters panel is that it's a bit disconnected from the list of dark regions shown below:

Image

This is because the other panel SMN copy number is in the middle. Have you tried moving the filters panels inside the dark region one?

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
dnil: Yes, there was space. I imagine it will shrink as we add more filter options.

scout/server/blueprints/cases/controllers.py
dnil: I know, I wanted to have the hgnc_ids from the pipeline instead, but as it is, it would probably have been at least as safe to translate them on load instead.

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
dnil: Sure, I considered it, but skipped it since it would have rather little content. But why not!

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
dnil: This is a core feature for the clinicians. We never want to show things outside the ROI unless the user explicitly asks for it. As it is the Paraphase regions are foldable, but any Paraphrase calls on them as e.g. pathogenic would be visible unless the they are filtered out.

scout/adapter/mongo/panel.py
dnil: I didn't quite find one I liked, but I will give it another read!

dnil:
> Nice new feature!
> 
> The overall feeling I have when I look at the new filters panel is that it's a bit disconnected from the list of dark regions shown below:
> 
> Image
> This is because the other panel SMN copy number is in the middle. Have you tried moving the filters panels inside the dark region one?

Mm, yes, it started off there. Since it now affects also the SMN display, it feels a little weird there. I'll see if I can't think of something, but e.g. just shifting the SMN table to after the dark regions could work.

northwestwitch:
> > Nice new feature!
> > The overall feeling I have when I look at the new filters panel is that it's a bit disconnected from the list of dark regions shown below:
> > Image
> > This is because the other panel SMN copy number is in the middle. Have you tried moving the filters panels inside the dark region one?
> 
> Mm, yes, it started off there. Since it now affects also the SMN display, it feels a little weird there. I'll see if I can't think of something, but e.g. just shifting the SMN table to after the dark regions could work.

Just assign me when you are ready for another review

scout/adapter/mongo/panel.py
northwestwitch: Regarding this one, just hear me out.
If you look at the existing functions, there is also [this other one](https://github.com/Clinical-Genomics/scout/blob/fd371648ea7ee9b58ea9e06d2c5176bf9d7a9daf/scout/adapter/mongo/panel.py#L326)

Image

For panels present in the case, you can directly pass it the panel ID (it's in the case document). For the other panels you can pass it the panel name. If you pass it gene_format="hgnc_id" you should extract exactly the info that you need. I also like the fact that the function that is called from this one (panel_to_genes) uses projection and returns only the field that is needed.

I have't tried this approach, but we could test it perhaps?
I'd really try to avoid to have an extra function in that module if you can reuse the code we already have

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
northwestwitch: Looks better a bit shorter now!

scout/server/blueprints/cases/views.py
```diff
@@ -122,11 +127,31 @@ def case(
 def sma(institute_id, case_name):
     """Visualize case SMA and other dark region data - SMN CN calls, Paraphrase/Paraphase calls"""
     institute_obj, case_obj = institute_and_case(store, institute_id, case_name)
-    data = controllers.sma_case(store, institute_obj, case_obj)
+    sma_filter_form = SmaPanelFilterForm(request.args)
+    sma_filter_form.gene_panels.choices = gene_panel_choices(store, institute_obj, case_obj)
+
+    selected_gene_panels = sma_filter_form.gene_panels.data or []
+    panel_filter_applied = request.args.get("panel_filter_applied") == "1"
```
northwestwitch, line 134: What's 1??

northwestwitch (commented):
Works as it should but I am really not fond of the new mongo panel function. Could we please try to replace it with existing functions - see my other suggestion. This way we also avoid having to introduce new tests for the added code.

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -78,12 +94,66 @@
...
+                data_live_search="true",
+                title="Select gene panel(s)"
+              ) }}
+            </div>
+            </div>
+            <div class="row mt-2">
+              <div class="col-4 d-flex  align-items-end">
+                <button type="submit" class="btn btn-primary btn-sm me-2">Apply filter</button>
+                <a class="btn btn-secondary btn-sm" href="{{ url_for('cases.sma', institute_id=institute._id, case_name=case.display_name) }}">Reset</a>
+              </div>
```
northwestwitch, line 132: Careful with the style of this button: 

Image

dnil: Are there still light mode users you think? 😄

Good catch!

scout/server/blueprints/cases/views.py
dnil: True, but a little easier to pass through view/form mangling. 🤷

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
northwestwitch: > Are there still light mode users you think? 😄

For sure! 🤣

scout/server/blueprints/cases/views.py
northwestwitch: Ok..!

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
dnil: Fixed - thank you!
Screenshot 2026-06-04 at 11 20 25
Screenshot 2026-06-04 at 11 21 39

scout/adapter/mongo/panel.py
dnil: Sure, we can skip it altogether if we let it use the latest version of the panel. And that is even better, since it would be the same behaviour as the variantS pages. I'll add the comments to the calling function docstring instead. The adapter function turned out to be untested, so I added a simple test for it instead.

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -78,12 +94,66 @@
...
+            <div class="row">
+            {{ sma_filter_form.panel_filter_applied(value="1") }}
+            <div class="col-6">
+              <label class="form-label mb-1" for="gene-panel-filter">Gene panels</label>
+              {{ sma_filter_form.gene_panels(
+                id="gene-panel-filter",
+                class="selectpicker",
+                data_style="btn-secondary",
+                data_width="100%",
+                data_live_search="true",
```
northwestwitch, line 123: What does this do?

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
```diff
@@ -230,7 +305,9 @@
         <span class="me-3">
           <form target="_blank" rel="noopener" action="{{url_for('variants.variants', institute_id=institute._id, case_name=case.display_name, variant_type='clinical') }}">
             <input type="hidden" id="hgnc_symbols" name="hgnc_symbols" value="SMN1, SMN2"></input>
-            <input type="hidden" id="gene_panels" name="gene_panels" value=""></input>
+            {% for panel_name in selected_gene_panels %}
```
northwestwitch, line 308: 👍🏻

scout/server/blueprints/cases/controllers.py
```diff
@@ -302,21 +303,51 @@ def bionano_case(store, institute_obj, case_obj) -> Dict:
...
+    """
 
     _populate_case_individuals(case_obj)
 
     case_has_alignments(case_obj)
 
+    selected_gene_panels = selected_gene_panels or []
+    panel_hgnc_ids = {
+        hgnc_id
+        for hgnc_id in store.panels_to_genes(
```
northwestwitch, line 327: Thank you! 🙏🏻

scout/server/blueprints/cases/controllers.py
```diff
@@ -390,11 +421,15 @@ def _update_case_region(case_region: dict, region: dict, individual_is_affected:
                 )
 
 
-def _get_paraphrase_regions(case_obj: dict) -> dict:
+def _get_paraphrase_regions(case_obj: dict, panel_hgnc_ids: Set[int] | None = None) -> dict:
```
northwestwitch, line 424: 💯

CHANGELOG.md
```diff
@@ -5,9 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 About changelog [here](https://keepachangelog.com/en/1.0.0/)
 
 ## [unreleased]
+
```
northwestwitch:
```suggestion
```

northwestwitch (approved):
Thank you! It looks good to me. I've tested once again - not with HPO panel - and it seems to be working nicely! 💯

scout/server/blueprints/cases/templates/cases/case_sma_dark.html
dnil: It's a really cute little bootstrap search-box-in-the-drop: actually think we should add it to most selectpickers where the dropdowns are larger than a handful of options. Try it!

northwestwitch: Ah cool!

PR #6355, opened 2026-06-01
Filter variants by gene OMIM disease inheritance

### LINKED ISSUE #6304: Gene panel/OMIM disease inheritance pattern filter for variantS
To help the analyst work quickly and focused through the variants lists, it would be helpful to filter for gene inheritance patterns among case variants. The analyst can e.g. filter for recessive disorders and dismiss all without compounds.

northwestwitch:
- Closes #6304

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -682,6 +682,10 @@ class QueryHandler(object):
                     criterion_values = query[criterion]
                     mongo_secondary_query.append({criterion: {"$in": criterion_values}})
 
+                case "omim_genetic_models":
+                    criterion_values = query[criterion]
+                    mongo_secondary_query.append({"genes.inheritance": {"$in": criterion_values}})
+
                 case "functional_annotations" | "region_annotations":
                     # filter key will be genes.[criterion (minus final char)]
                     criterion_values = query[criterion]
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -320,7 +320,7 @@
     <span style="float:right;">{{ form.gene_panels_exclude.label(class="control-label", data_bs_toggle="tooltip", data_bs_placement="left", title="Check this box to exclude from the search any gene listed in selected 'Gene Panels', 'Symbol file' or provided in the 'HGNC Symbols/Ids' field.") }} {{form.gene_panels_exclude}}</span>
     {{ form.gene_panels(class="selectpicker", data_style="btn-secondary") }}
   </div>
-  <div class="col-2">
+  <div class="col-1-5">
   {% if type == "snv" %}
     {{ form.symbol_file.label(class="control-label", data_bs_toggle="tooltip", data_bs_placement="left", title="Load an HGNC gene symbol list file; a text file with one gene symbol starting each row. Extra columns separated with tab are ignored. Comment rows starting with # are ignored.") }}
     {{ form.symbol_file(class="form-control-file") }}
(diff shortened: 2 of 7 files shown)


tests/adapter/mongo/test_query.py
```diff
@@ -165,6 +165,28 @@ def test_soft_filters_query(adapter, case_obj):
     assert mongo_query["filters"] == {"$nin": institute_soft_filters}
 
 
+def test_genetic_models_query(adapter, case_obj):
```
northwestwitch, line 168: This one is unrelated, but was missing

scout/server/static/bs_styles.css
```diff
@@ -421,3 +421,8 @@ tr[data-href] {
     opacity: 0;
   }
 }
+
+.col-1-5 {
```
northwestwitch, line 425: You can't have col-1.5 in bootstrap, so it's a custom style. Could be useful also for other pages

scout/adapter/mongo/query.py
```diff
@@ -682,6 +682,10 @@ def secondary_query(self, query: dict) -> list:
                     criterion_values = query[criterion]
                     mongo_secondary_query.append({criterion: {"$in": criterion_values}})
 
+                case "omim_genetic_models":
+                    criterion_values = query[criterion]
+                    mongo_secondary_query.append({"genes.inheritance": {"$in": criterion_values}})
```
dnil, line 687: Unindexed - fast enough?

scout/server/static/bs_styles.css
dnil: Cool! Is that all there is to the col-Ns? `flex 0 0 1/N`. I had imagined a set of intricate, Donald Knuth style layout scripts! 😄

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -320,7 +320,7 @@
     <span style="float:right;">{{ form.gene_panels_exclude.label(class="control-label", data_bs_toggle="tooltip", data_bs_placement="left", title="Check this box to exclude from the search any gene listed in selected 'Gene Panels', 'Symbol file' or provided in the 'HGNC Symbols/Ids' field.") }} {{form.gene_panels_exclude}}</span>
     {{ form.gene_panels(class="selectpicker", data_style="btn-secondary") }}
   </div>
-  <div class="col-2">
+  <div class="col-1-5">
```
dnil: We are changing 10 col space for 9 (5*2 vs 6*1.5). How about something like a single col-10 div for the final 6, divided into six col-2 - effectively like col-1.67? Still needs to sync reasonably with the rows below, so feel free to ignore if it seems impossible to get it both well aligned and screen area efficient.

dnil (approved):
I am a little worried about the unindexed query. Thoughts? It seems fast enough on stage, so I'm open to give it a try!

Let's see if we can't fix the column alignment - it looks a little seasick rn, but honestly also on main, so I will not fault it on that.

scout/adapter/mongo/query.py
northwestwitch: Possibly, variants from that case and that category are indexed, so the selection is only among those

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: I kind of like the 1.5 thing because it would be easier to reuse in other pages, instead of a custom-made col style just to fit that line in this page. I will turn back to col-2 also the second column (genes symbols file), so the space at the end of that line will be less, like this:

Image

scout/server/static/bs_styles.css
northwestwitch: Right??? 😆

northwestwitch:
> I am a little worried about the unindexed query. Thoughts? It seems fast enough on stage, so I'm open to give it a try!

Let's see how it goes

PR #6375, opened 2026-06-05
Add 'Has methylation' filter on cases page

northwestwitch:
OR

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR, automatic tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - LitVar2 links: variant link for variants with dbSNP rsID and a gene symbol search link (#6326)
 - Filter variants by OMIM inheritance pattern (#6355)
 - Gene panel filters to the Dark regions page (#6353)
+- `Has Methylation data` filter on cases page (#6375)
 ### Changed
 - Replaced Ensembl rest liftover service with liftover API from the Broad Institute (#6293)
 - Avoid fetching genes and panels multiple times when loading variants (#6350)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Enable link to previous ACMG classifications for SVs and cancer SVs (#6365)
 - Fixes bug where CCV classification ignores modifiers when submitting to db (#6369)
+- `has_methylation` field not persisting when a case document is updated (#6375)
 
 ## [4.112]
 ### Added
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -565,6 +565,7 @@ class CaseHandler(object):
             # Return the result in order of descending phenotype similarity (the same order or the _ids provided in the query)
             return sorted(list(results), key=lambda res: result_order.index(res["_id"]))
 
+        LOG.warning(query)
         return self.case_collection.find(query, projection).sort("updated_at", -1)
 
     def is_pheno_similarity_query(self, name_query: ImmutableMultiDict) -> bool:
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -374,6 +374,7 @@ class CaseHandler(object):
         finished: bool = False,
         research_requested: bool = False,
         is_research: bool = False,
+        has_methylation: bool = False,
         has_rna_data: bool = False,
         status: Any = None,
         phenotype_terms: bool = False,
@@ -454,6 +455,10 @@ class CaseHandler(object):
             condition=has_causatives,
             set_key="causatives",
             set_value={EXISTS: True, "$ne": []},
+        )
+
+        _conditional_set_query_value(
+            query=query, condition="has_methylation", set_key="has_methylation", set_value=True
         )
 
         _conditional_set_query_value(
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -458,7 +458,7 @@ class CaseHandler(object):
         )
 
         _conditional_set_query_value(
-            query=query, condition="has_methylation", set_key="has_methylation", set_value=True
+            query=query, condition=has_methylation, set_key="has_methylation", set_value=True
         )
 
         _conditional_set_query_value(
--- a/scout/adapter/mongo/case.py
+++ b/scout/adapter/mongo/case.py
@@ -565,7 +565,6 @@ class CaseHandler(object):
             # Return the result in order of descending phenotype similarity (the same order or the _ids provided in the query)
             return sorted(list(results), key=lambda res: result_order.index(res["_id"]))
 
-        LOG.warning(query)
         return self.case_collection.find(query, projection).sort("updated_at", -1)
 
     def is_pheno_similarity_query(self, name_query: ImmutableMultiDict) -> bool:
@@ -1091,6 +1090,7 @@ class CaseHandler(object):
             - gene_fusion_report_research: path to the research gene fusions report
             - genome_build: If there is a new genome build
             - has_meivariants: If there are new mei variants
+            - has_methylation: if methylation data is present
             - has_outliers: If there are new outlier variants
             - has_strvariants: If there are new strvariants
             - has_svvariants: If there are new svvariants
@@ -1164,6 +1164,7 @@ class CaseHandler(object):
                 "gene_fusion_report": case_obj.get("gene_fusion_report"),
                 "gene_fusion_report_research": case_obj.get("gene_fusion_report_research"),
                 "genome_build": case_obj.get("genome_build", "37"),
+                "has_methylation": case_obj.get("has_methylation", False),
                 "has_meivariants": case_obj.get("has_meivariants", False),
                 "has_outliers": case_obj.get("has_outliers", False),
                 "has_strvariants": case_obj.get("has_strvariants", False),
--- a/scout/server/blueprints/institutes/templates/overview/utils.html
+++ b/scout/server/blueprints/institutes/templates/overview/utils.html
@@ -40,11 +40,11 @@
     </div>
     {% if institute %} <!-- The following checkboxes will be displayed when the form is used on institute caseS page only, not on the dashboard -->
     <div class="row">
-        <div class="col-1-5 mb-3 form-check ms-3">
+        <div class="col-2 mb-3 form-check">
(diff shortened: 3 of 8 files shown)


scout/adapter/mongo/case.py
```diff
@@ -1158,6 +1164,7 @@ def update_case(self, case_obj: dict, keep_date: bool = False) -> dict:
                 "gene_fusion_report": case_obj.get("gene_fusion_report"),
                 "gene_fusion_report_research": case_obj.get("gene_fusion_report_research"),
                 "genome_build": case_obj.get("genome_build", "37"),
+                "has_methylation": case_obj.get("has_methylation", False),
```
northwestwitch, line 1167: This was actualy a bug resulting in the field not persisting when case was updated

scout/adapter/mongo/case.py
```diff
@@ -456,6 +457,10 @@ def _conditional_set_query_value(query, condition, set_key, set_value):
             set_value={EXISTS: True, "$ne": []},
         )
 
+        _conditional_set_query_value(
```
dnil, line 460: 💯

scout/server/blueprints/institutes/templates/overview/utils.html
```diff
@@ -40,23 +40,27 @@
     </div>
     {% if institute %} <!-- The following checkboxes will be displayed when the form is used on institute caseS page only, not on the dashboard -->
     <div class="row">
-        <div class="col-md-2 mb-3 form-check ms-3">
+        <div class="col-1-5 mb-3 form-check ms-3">
```
dnil: Please consider sticking with standard classes, e.g. using "col" alone for the fields that can be the same, remaining space filler size, or grouping several fields in a `<col-M><row><col-N><col-N></row></col>` block for a M/(12/N) size.

tests/adapter/mongo/test_case_handling.py
```diff
@@ -181,6 +181,14 @@ def test_case_queries(
     cases = list(adapter.cases(owner=case_obj["owner"], is_research=True))
     assert len(cases) == 1
 
+    # -----------------------------
+    # THEN search by methylation data should work
```
dnil, line 185: 🥇

dnil (approved):
Good catch, and new feature. I would prefer if we avoid non-standard layout elements like col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉) when there are reasonable other options. If it looks way too hard here... 🤷 Just give it a good second look please.

scout/server/blueprints/institutes/templates/overview/utils.html
northwestwitch: The thing is I tried having 6 elements with col-md-2 (12 cols in total should fit right?) and nope, I got one field that was moved to the next line. I can check better what's up with the formatting, could be the spacing on the sides of the row

northwestwitch:
> col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉)

Why 1-6? 1.5 columns out of 12 = [12.5%](https://github.com/Clinical-Genomics/scout/blob/aee705675cd425efea3309c9954cb3d5edbe092e/scout/server/static/bs_styles.css#L425)

🤔

dnil:
> > col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉)
> 
> Why 1-6? 1.5 columns out of 12 = [12.5%](https://github.com/Clinical-Genomics/scout/blob/aee705675cd425efea3309c9954cb3d5edbe092e/scout/server/static/bs_styles.css#L425)
> 
> 🤔

Because the first digit is in base 12, so the second should also be? 6 is half of 12, so 1.5 is 1-6. 

Like if you want half of the row you would say col-6. So half of the col-1 would be col-1-6. 🧠 💨

scout/server/blueprints/institutes/templates/overview/utils.html
dnil: Surely there was just one col hiding somewhere, like in a macro or sth? Ping if you want a second pair of eyes looking for it!

northwestwitch:
> > > col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉)
> > 
> > 
> > Why 1-6? 1.5 columns out of 12 = [12.5%](https://github.com/Clinical-Genomics/scout/blob/aee705675cd425efea3309c9954cb3d5edbe092e/scout/server/static/bs_styles.css#L425)
> > 🤔
> 
> Because the first digit is in base 12, so the second should also be? 6 is half of 12, so 1.5 is 1-6.
> 
> Like if you want half of the row you would say col-6. So half of the col-1 would be col-1-6.

Bootstrap column numbers are not digits in a base-12 numeral system, they're count of columns. Why do you have to make counter-intuitive?? 😕

dnil:
> > > > col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉)
> > > 
> > > 
> > > Why 1-6? 1.5 columns out of 12 = [12.5%](https://github.com/Clinical-Genomics/scout/blob/aee705675cd425efea3309c9954cb3d5edbe092e/scout/server/static/bs_styles.css#L425)
> > > 🤔
> > 
> > 
> > Because the first digit is in base 12, so the second should also be? 6 is half of 12, so 1.5 is 1-6.
> > Like if you want half of the row you would say col-6. So half of the col-1 would be col-1-6.
> 
> Bootstrap column numbers are not digits in a base-12 numeral system, they're count of columns. Why do you have to make counter-intuitive?? 😕

Huh? col-1-5 is totally unintuitive. If you don't like the base 12, it works just the same if you see it as counts. That is what number systems mean. Think about it! 
The reason we write 1.5 is we mean one and then half of the max count to 10 in the next less significant decimal position. We don't write 1.2 or 1.4 - or 1.6 in decimal. 
One half row is a count of six out of 12 columns. The same should be true for the less significant position, same as if you had done what we have in so many places with other count combinations: `<col-1><row><col-6>this</col></row></col>` :== `<col-1-6>this</col>`.

northwestwitch:
> The reason we write 1.5 is we mean one and then half of the max count to 10 in the next less significant decimal position. We don't write 1.2 or 1.4 - or 1.6 in decimal.

1.5 is one column and a half (referring to the unit, not a number base), where ".5" just means half of a column, not a base-12 fractional digit. 
It's not just me naming the style like this, see [here](https://stackoverflow.com/questions/30026855/can-i-give-the-col-md-1-5-in-bootstrap) or [here](https://stackoverflow.com/questions/35303782/how-can-i-use-a-half-of-a-column-on-bootstrap). Let's rename it to something like `w-12-5-percent` and close the question ok?

dnil:
> > col-1-5 (btw the name should be col-1-6 really, shouldn't it? 😉)
> 
> Why 1-6? 1.5 columns out of 12 = [12.5%](https://github.com/Clinical-Genomics/scout/blob/aee705675cd425efea3309c9954cb3d5edbe092e/scout/server/static/bs_styles.css#L425)
> 
> 🤔

If you like decimal fractions better, note that 0.8333 + 0.8333/2 is 0.125. So _1_/12 + 1/12 * _6_/12`.  "col-1-6". Same as 0.15 is _1_/10 + 1/10 * _5_/10.

northwestwitch:
> If you like decimal fractions better, note that 0.8333 + 0.8333/2 is 0.125. So _1_/12 + 1/12 * _6_/12`. "col-1-6". Same as 0.15 is _1_/10 + 1/10 * _5_/10.

No, I don't like them better

dnil:
> > The reason we write 1.5 is we mean one and then half of the max count to 10 in the next less significant decimal position. We don't write 1.2 or 1.4 - or 1.6 in decimal.
> 
> 1.5 is one column and a half (referring to the unit, not a number base), where ".5" just means half of a column, not a base-12 fractional digit. It's not just me naming the style like this, see [here](https://stackoverflow.com/questions/30026855/can-i-give-the-col-md-1-5-in-bootstrap) or [here](https://stackoverflow.com/questions/35303782/how-can-i-use-a-half-of-a-column-on-bootstrap). Let's rename it to something like `w-12-5-percent` and close the question ok?

Sure, that works for me. But I would also prefer we fix the underlying reason for the overflow; it should be fine with 12 cols total, otw something is off, like you say either with margins, padding, or perhaps with some div tag open-close mismatching somewhere?

dnil:
> > If you like decimal fractions better, note that 0.8333 + 0.8333/2 is 0.125. So _1_/12 + 1/12 * _6_/12`. "col-1-6". Same as 0.15 is _1_/10 + 1/10 * _5_/10.
> 
> No, I don't like them better

😂 How about percent then?

northwestwitch:
> Sure, that works for me. But I would also prefer we fix the underlying reason for the overflow; it should be fine with 12 cols total, otw something is off,

Fixed, there was some spacing at the beginning, which now I've removed:


image

northwestwitch:
> 😂 How about percent then?

Why not in binary?

dnil:
> > The reason we write 1.5 is we mean one and then half of the max count to 10 in the next less significant decimal position. We don't write 1.2 or 1.4 - or 1.6 in decimal.
> 
> 1.5 is one column and a half (referring to the unit, not a number base), where ".5" just means half of a column, not a base-12 fractional digit. It's not just me naming the style like this, see [here](https://stackoverflow.com/questions/30026855/can-i-give-the-col-md-1-5-in-bootstrap) or [here](https://stackoverflow.com/questions/35303782/how-can-i-use-a-half-of-a-column-on-bootstrap).

No offence, but to my limited mind that is cave bob coding style? They are changing the base mid number. It's like referring to 15:30 as 15.5. Its just 15h plus a half in decimal!? Convenient perhaps, but not correct once you think about it. Note that none of those are official packages, just random user comments. Sometimes we just express ourselves without thinking too much. Its ok. 🤷

dnil:
> > 😂 How about percent then?
> 
> Why not in binary?

Also ok, just be consistent!

CHANGELOG.md
```diff
@@ -10,13 +10,15 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - Filter variants by OMIM inheritance pattern (#6355)
 - Gene panel filters to the Dark regions page (#6353)
 - A `Me` badge to designate cases with methylation data available, on cases page (#6373)
+- `Has Methylation data` filter on cases page (#6375)
```
dnil, line 13: 💯

PR #6379, opened 2026-06-08
Fix #6376 - add an activate case button

### LINKED ISSUE #6376: Inactivate/activate status buttons
With some careful combinations of status changes, it is already possible to arrive at inactive or active as needed, so not a dramatic change. Also the events will still be logged, so there is an audit track of what happened. But in all a friendly change, that would make life easier on new users.

dnil:

- Fix #6376 - activate case button

All the infrastructure for the status change is already there, including sanity checks. It is already possible to affect inactivation as a regular user by toggling `ignored` status, and activation by e.g. toggling `prioritized` status. This change makes the user intention clear in the audit trail.

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by CR
- [x] tests executed by DN

### DIFF
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -108,7 +108,7 @@
 
                   <!-- active/inactive -->
                   <button name="status" value="{{ 'active' if case.status == 'inactive' else 'inactive' }}" type="submit" class="btn btn-dark btn-sm">
-                    {{ 'Activate' if case.status == 'inactive' else 'Inactivate' }}
+                    {{ 'Active' if case.status == 'inactive' else 'Inactive' }}
                   </button>
 
                     <!-- unused/inactive -->
--- a/scout/server/blueprints/cases/templates/cases/case.html
+++ b/scout/server/blueprints/cases/templates/cases/case.html
@@ -106,10 +106,16 @@
                     {{ 'De-prioritized' if case.status == 'prioritized' else 'Prioritized' }}
                   </button>
 
+                  <!-- active/inactive -->
+                  <button name="status" value="{{ 'active' if case.status == 'inactive' else 'inactive' }}" type="submit" class="btn btn-dark btn-sm">
+                    {{ 'Activate' if case.status == 'inactive' else 'Inactivate' }}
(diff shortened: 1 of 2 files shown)


scout/server/blueprints/cases/templates/cases/case.html
```diff
@@ -106,10 +106,16 @@ <h4 class="h4 d-inline-block">
                     {{ 'De-prioritized' if case.status == 'prioritized' else 'Prioritized' }}
                   </button>
 
+                  <!-- active/inactive -->
+                  <button name="status" value="{{ 'active' if case.status == 'inactive' else 'inactive' }}" type="submit" class="btn btn-dark btn-sm">
+                    {{ 'Activate' if case.status == 'inactive' else 'Inactivate' }}
+                  </button>
```
northwestwitch, line 112: I see the point, but then don't we want to change also the other button, for consistency? Instead of this:

Image

We could have:

- Archive
- Solve
- Prioritize

and so on?

northwestwitch (commented):
What about changing also the other buttons?

We could have instead of `Set:` `Action:` as well.. Gives more the idea of the action, obviously..

scout/server/blueprints/cases/templates/cases/case.html
dnil: Fair. I would absolutely consider a set dropdown instead. The slight complication is that two of them have modals. Also keeping it is colorful. 😁 

How about this, "active" and "inactive" on the buttons instead? That conforms to the states printed on the other buttons.

northwestwitch: > Fair. I would absolutely consider a set dropdown instead. The slight complication is that two of them have modals. Also keeping it is colorful. 😁

It would be nice, yes!

> How about this, "active" and "inactive" on the buttons instead? That conforms to the states printed on the other buttons.

Sure!

northwestwitch (approved):
👍🏻 Looks good!

PR #6385, opened 2026-06-08
Search specific var

### LINKED ISSUE #6263: Search variants, searching for specific variants and larger exports?
Hi there!

This is a hopeful request from our friends at the screening lab :)

In the preparation for the MLD screening which should begin this fall, they have sequenced a lot of individuals and observed heterozygous variants in individuals with lower enzymatic activity. To understand if such variants are actual pathogenic variants or pseudodeficiency variants, they want to look at the allele frequency in the general population. They have a list of variants that don't exist or are very rare in gnomAD, but may have been observed in a higher frequency in cases in Scout.

a) in Search variants, would it be possible to search for specific variants within genes?

b) in Search variants, would it be possible to export more than 500 variants? Some of these variants may have been observed but have a low score. The 500 export cutoff limits the possibility to look for them. 

Also, they appreciate that you've added genotype when searching for variants! <3 

Thanks,
Ingegerd

dnil:

- Fix #6263 - specifically allow searching for variant ids

We already have an index for `variant_id` (with an optional `institute`). That is a complete one, not like the sparse index. It should allow pretty quick access. The other variant search terms become irrelevant given this, though additional filtering at the case level is still possible.

See what you think. It is pretty convenient to have this on the same page, and hopefully it is intuitive enough not to cause confusion.

Screenshot 2026-06-08 at 16 25 31

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -84,7 +84,11 @@ class QueryHandler(object):
     ) -> dict:
         """Build a query to find a variant by its simple id."""
 
-        chrom, pos, ref, alt = simple_id.split("-")
+        delimiter = "-"
+        if "_" in simple_id:
+            delimiter = "_"
+
+        chrom, pos, ref, alt = simple_id.split(delimiter)
 
         variant_id_query = [
             parse_variant_id(chrom, pos, ref, alt, variant_type_element)
--- a/scout/adapter/mongo/query.py
+++ b/scout/adapter/mongo/query.py
@@ -12,6 +12,7 @@ from scout.constants import (
     TRUSTED_REVSTAT_LEVEL,
 )
 from scout.constants.filters import METHBAT_IMPRINT_LABEL, METHBAT_PROMOTER_LABEL
+from scout.parse.variant.ids import parse_variant_id
 
 CLNSIG_NOT_EXISTS = {"clnsig": {"$exists": False}}
 CLNSIG_ONC_NOT_EXISTS = {"clnsig_onc": {"$exists": False}}
@@ -78,12 +79,30 @@ class QueryHandler(object):
 
         return query
 
+    def build_simple_id_query(
+        self, simple_id: str, variant_type: list, institute_ids: Optional[list]
+    ) -> dict:
+        """Build a query to find a variant by its simple id."""
+
+        chrom, pos, ref, alt = simple_id.split("-")
+
+        variant_id_query = [
+            parse_variant_id(chrom, pos, ref, alt, variant_type_element)
+            for variant_type_element in variant_type
+        ]
+
+        mongo_variant_query = {"variant_id": {"$in": variant_id_query}}
+        if institute_ids:
+            mongo_variant_query["institute"] = {"$in": institute_ids}
+
+        return mongo_variant_query
+
     def build_variant_query(
         self,
         query: Optional[dict] = None,
         institute_ids: Optional[list] = [],
         category: Optional[Union[str, list]] = "snv",
-        variant_type: Optional[list] = ["clinical"],
+        variant_type: list = ["clinical"],
     ):
         """Build a mongo query across multiple cases.
         Translate query options from a form into a complete mongo query dictionary.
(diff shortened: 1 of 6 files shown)


scout/server/blueprints/institutes/forms.py
```diff
@@ -209,6 +218,19 @@ class GeneVariantFiltersForm(FlaskForm):
...
+    simple_id = StringField(
+        "Simple ID",
+        [
+            validators.Optional(),
+            hgnc_symbol_or_simple_id,
+            validators.Length(min=8),
+            validators.Regexp(
+                r"^(?:[1-9]|1[0-9]|2[0-2]|X|Y|MT)-[0-9]+-[ATGCN]+-[ATGCN]+$",
+                IGNORECASE,
+                message="Use format CHR-POS-REF-ALT, e.g. 1-2345-A-C",
```
northwestwitch: Nice! I would also accept underscores instead of only dash, because display_name in the database and on the portal contain underscores. It would simplify if you copy and paste from the variant's page.

scout/server/blueprints/institutes/forms.py
```diff
@@ -156,6 +158,13 @@ def process_formdata(self, valuelist):
             self.data = []
 
 
+def hgnc_symbol_or_simple_id(form, field):
+    if not form.hgnc_symbols.data and not form.simple_id.data:
+        raise validators.ValidationError(
+            "Either HGNC gene symbols or a simple variant ID is required"
+        )
```
northwestwitch, line 165: Perhaps check also that both gene or simple id are not provided together?

northwestwitch (changes requested):
Nice new feature! At the moment both gene and simple ID can be provided but only variants with simple ID are returned, which I think is confusing (even if explained in the tooltip).
See also my other comment about accepting underscores.

scout/server/blueprints/institutes/forms.py
dnil: Good call - I started with underscore only, realised we had dashes for e.g. gnomAD and switched.

scout/server/blueprints/institutes/forms.py
dnil: Yes, I simply prioritised the variant id, but maybe we could add a warning for unused filter options when that is in play.

northwestwitch: That, or just add that condition to the validation function, since you added it (the validation) - nice btw!

scout/server/blueprints/institutes/views.py
```diff
@@ -201,9 +207,37 @@ def gene_variants(institute_id):
...
+
+        if form.simple_id.data and any(
+            [
+                getattr(form, criterium).data
+                for criterium in set(["hgnc_symbols"] + SECONDARY_CRITERIA) - set(["rank_score"])
+                if hasattr(form, criterium)
+            ]
+        ):
+            flash(
+                "A Variant Simple ID was given. It cannot be used together with other variant filter options, such as gene symbol.",
```
northwestwitch:
```suggestion
                "Search by variant Simple ID cannot be combined with any other variant filter except institute.",
```

scout/adapter/mongo/query.py
```diff
@@ -78,12 +79,34 @@ def delete_variants_query(
...
         return query
 
+    def build_simple_id_query(
+        self, simple_id: str, variant_type: list, institute_ids: Optional[list]
+    ) -> dict:
+        """Build a query to find a variant by its simple id."""
+
+        delimiter = "-"
+        if "_" in simple_id:
+            delimiter = "_"
```
northwestwitch:
```suggestion
        delimiter = "_"
        simple_id = simple_id.replace("-", "_")
```

northwestwitch (approved):
Nice! 👍🏻

scout/adapter/mongo/query.py
dnil:
```suggestion
        delimiter = "-"
        simple_id = simple_id.replace("_", delimiter)
```
Or perhaps this. 🤷 I tend to want to choose the least invasive solution, with a conditional, but it doesn't matter - all should give the same end result.

northwestwitch: Sure!

PR #6399, opened 2026-06-12
Fix exporting causatives from 38 cases without --build param

### LINKED ISSUE #6396: Missing build fallback for export variants case_id
When attempting to export variants for a specific case with build 38 variants without giving any build options, the output file is empty, but it works if given a build. Did we perhaps miss using the case build as a fallback for the variants - eg by adding a common build option handling?

See 'charmingcrawfish'.

northwestwitch:
- Fix #6396 

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [x] code approved by DN
- [x] tests executed by CR

### DIFF
--- a/scout/commands/export/variant.py
+++ b/scout/commands/export/variant.py
@@ -162,7 +162,7 @@ def resolve_case(
     """
     if not case_id:
         LOG.info("Use collaborator %s", collaborator)
-        return collaborator, None
+        return collaborator, None, None
 
     case_obj = adapter.case(case_id)
     if not case_obj:
--- a/scout/commands/export/variant.py
+++ b/scout/commands/export/variant.py
@@ -154,7 +154,8 @@ def resolve_case(
     Resolve the effective collaborator and optional case object.
 
     If case_id is provided, the case is fetched and its owner is used
-    as the collaborator. If the case does not exist, the command aborts.
+    as the collaborator. Case genome build is also returned.
+    If the case does not exist, the command aborts.
 
     Returns:
         Tuple of (collaborator, case object or None)
@@ -168,7 +169,7 @@ def resolve_case(
         LOG.info("Case %s does not exist", case_id)
         raise click.Abort
 
-    return case_obj["owner"], case_obj
+    return (case_obj["owner"], case_obj, case_obj["genome_build"])
 
 
 @click.command("causatives", short_help="Export causative variants")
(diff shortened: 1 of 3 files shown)


scout/commands/export/variant.py
```diff
@@ -217,9 +218,9 @@ def causatives(
     LOG.info("Running scout export variants")
     adapter = store
 
-    collaborator, case_obj = resolve_case(adapter, case_id, collaborator)
+    collaborator, case_obj, case_build = resolve_case(adapter, case_id, collaborator)
 
-    build = "38" if build == "GRCh38" else build
```
northwestwitch, line 222: build is 37 [by default](https://github.com/Clinical-Genomics/scout/blob/436007c98c0938376b3b8fdd6fdcb6ef5d825c00/scout/commands/export/utils.py#L26) otherwise

dnil: Right. Hm, when looking at the help page, it is actually pretty clear defaulting to 37 is the intended behaviour. Let's add this to the list of v5 API changes. The sensible default is really the case_obj, it makes a lot of sense to implement it like that.

I can imagine it being super confusing for a user trying to give an explicit `--build` for some arcane reason, so we should change to make build optional and respect an explicit build given at a later point.

dnil (approved):
Yes, thank you, this seems like the reasonable way to fix it for now. But as per comment, let's add this to the list of funky CLI option combinations to fix. Doubtlessly the build option was not intended to be used together with the case_id option.

northwestwitch:
> Yes, thank you, this seems like the reasonable way to fix it for now. But as per comment, let's add this to the list of funky CLI option combinations to fix. Doubtlessly the build option was not intended to be used together with the case_id option.

I didn't get how you would like to fix the CLI on scout 5: you want to remove the default to build 37?

dnil:
> > Yes, thank you, this seems like the reasonable way to fix it for now. But as per comment, let's add this to the list of funky CLI option combinations to fix. Doubtlessly the build option was not intended to be used together with the case_id option.
> 
> I didn't get how you would like to fix the CLI on scout 5: you want to remove the default to build 37?

I would like the build option to be optional, defaulting instead to use case_obj.build primarily.

But then if given, to be respected. Actually now that I think about it, maybe a warning for the combined use would be sufficient. Like, always use case build, if available. And if the case build and build option disagree, just warn but return the variants.

PR #6491, opened 2026-08-12
Add pagination to ClinVar germline submissions page

### LINKED ISSUE #6468: Suggested ClinVar form/handling improvements
We use the ClinVar module a lot in Lund, so perhaps we notice some issues that are not obvious to everyone. I would like to report two issues, one minor and one that is more important:

1. Minor: Could we make a change so the form defaults to Condition ID type: MedGen instead of HPO as the first option? We almost always use MedGen and that is also what is officially displayed on the ClinVar variant pages.

2. More important: The page ".../institute/clinvar_germline_submissions" loads very slowly. My uneducated guess is because this page holds a very long list of past submissions. Perhaps an archive function would solve this, so only the n most recent submissions are automatically loaded?

northwestwitch:
- Fix #6468 - Add a pagination to ClinVar germline variants page

### Main branch - has a long scrolling page

image

### This branch

image

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:
1. Locally, create a germline submission, then in the database just duplicate the document many times.
2. Go to germline submissions page and compare main branch with this branch

**Expected outcome**:

**Review:**
- [ ] code approved by
- [ ] tests executed by


scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
```diff
@@ -41,28 +42,60 @@ <h3 class="mt-3">ClinVar submissions page</h3><a data-bs-target="#howto" href="#
     </div>
 
   {% if submissions %}
-
-    <div style="overflow-y: auto; height: 1000px;">
```
northwestwitch, line 45: Remove scrolling

scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
```diff
@@ -41,28 +42,60 @@ <h3 class="mt-3">ClinVar submissions page</h3><a data-bs-target="#howto" href="#
...
+</form>
 {% endmacro %}
 
 {% macro submission_panel(subm_obj) %}
 <div class="accordion accordion-flush" id="accordionFlushSubmissions">
   <div class="accordion-item">
     <h2 class="accordion-header" id="header-{{subm_obj._id}}">
-      <a class="accordion-button collapsed {% if subm_obj.status=='open' %} link-primary {% elif subm_obj.status=='submitted' %} link-success {% else %} link-secondary {% endif%}" type="button" data-bs-toggle="collapse" data-bs-target="#flush-{{subm_obj._id}}" aria-expanded="false" aria-controls="flush-{{subm_obj._id}}" style="text-decoration: none;">
-        Submission&nbsp;<strong>{{subm_obj.clinvar_subm_id or subm_obj._id}} - ({{subm_obj.status|upper}})</strong>&nbsp;/&nbsp;Created {% if subm_obj.created_by %}&nbsp;by&nbsp;<strong>{{subm_obj.created_by}}</strong>{% endif %}&nbsp;on&nbsp;<strong>{{subm_obj.created_at.strftime('%Y-%m-%d')}}</strong>&nbsp;/&nbsp;Last updated:&nbsp;<strong>{{subm_obj.updated_at.strftime('%d-%m-%Y, %T')}}</strong>
+      <a
```
northwestwitch, line 67: This is just changing the style to make it look a bit nicer

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -1240,8 +1240,8 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
   {{ pin_indicator(variant, case) }}
 {% endmacro %}
 
-{% macro pagination_hidden_div(page, result_size) %}
-  {% set per_page = 50 %}
+{% macro pagination_hidden_div(page, result_size, per_page=50) %}
```
northwestwitch, line 1243: This way is reusable from other pages

scout/server/blueprints/clinvar/views.py
```diff
@@ -103,17 +103,24 @@ def clinvar_germline_submissions(institute_id):
         if request.args.get("clinvar_id_filter")
         else None
     )
-
+    per_page = 15
+    page = request.args.get("page", 1, type=int)
+    start = (page - 1) * per_page
+    end = start + per_page
+    all_submissions = store.get_clinvar_germline_submissions(
```
northwestwitch, line 110: I think it would be even better to paginate the query results directly rather than fetching everything and slicing afterwards. But this change alone might already fix the page slowness. I'm thinking that loading a large number of submission documents, each with its own accordion, could be taking a toll on the page.

scout/server/blueprints/variants/templates/variants/utils.html
```diff
@@ -1258,13 +1258,11 @@ <h4 class="modal-title" id="filterNameModalLabel">Filter display name</h4>
     {% for p in range(1, total_pages + 1) %}
       <input type="submit" name="page" id="paginate-page-{{p}}" value="{{p}}" hidden>
     {% endfor %}
-
-    <input type="hidden" id="page" name="page" value="{{ page }}">
```
northwestwitch, line 1262: I don't think this is used?

dnil: Hm, I will not claim to understand it full - that was some chatGPT solution, was it not? We used to rely heavily on it, but for all pagination we should now be using the paginate inputs. 

Let's see, bot code or not, it should be readable. All the page inputs are named the same, but we fetch them all from the form, right, like https://github.com/Clinical-Genomics/scout/blob/aa755930e5292c0df7f153319cb350743e0a5588/scout/server/blueprints/variants/controllers.py#L92. We seemingly randomly take the first in the list, so never this original one at least. I suppose also for pagination requests maybe the one for the submitted tag gets to be first in the list, maybe even gets submitted alone with the name "page" in form. And for other reload/other action use some other, like the first page, would be first in the list. That out to make it bugged for like dismissing stuff on the third page, but I think it probably is already. 
 It should be safe to remove, and it should be easy enough to test through a few kinds of paginations to determine it, but my best suggestion is to instead move it up to the top of this function, and the dismiss-from-the-third page probably starts working again.

dnil: Ah, no, maybe the dismiss-from-a-later-page actually works on main already? I will not claim to understand the order then, but will predict it to break if this line is removed. Let's try and see! 😆

scout/server/blueprints/clinvar/views.py
```diff
@@ -103,17 +103,24 @@ def clinvar_germline_submissions(institute_id):
...
-        ),
+        "submissions": all_submissions[start:end],
         "institute": institute_obj,
         "variant_header_fields": CLINVAR_HEADER,
         "casedata_header_fields": CASEDATA_HEADER,
         "show_submit": current_user.email in institute_clinvar_submitters
         or not institute_clinvar_submitters,
         "clinvar_id_filter": clinvar_id_filter,
+        "page": page,
+        "result_size": len(all_submissions),
```
dnil, line 122: 💯

dnil (approved):
Yes, me like! Very nice with the generalisation of pagination counts - maybe we will want to have that as a user setting at some point. I'm not sure it fully solves the original issue if it's not on the mongo query as well, but I much support having it as a start at least.

Regarding the question about removing the current page hidden input, I believe it may break some use cases, but I do not fully understand the current function - it relies on flask internal treatment of which of several inputs with the same name gets to be first in an array. Try e.g. dismissing variants from later pages. Seems to work better on main than on this PR.

scout/server/blueprints/clinvar/views.py
dnil: Yes, but any secondary calls for each of the submissions, like fetching variants, genes, case info etc could also potentially be slow. Hard to tell without testing. This is a good start at least, and would be required for adding a page to that query anyway.

scout/server/blueprints/variants/templates/variants/utils.html
northwestwitch: I did some tests and it seems to work fine... 😄

northwestwitch: > that was some chatGPT solution, was it not?

Of course it was!

dnil: Ahem, ping @northwestwitch. 😜

PR #6495, opened 2026-08-12
Revert line on pagination utils and fix accordingly on CLinVar page

northwestwitch:
- Revert a line on the variant pagination hidden field, and a necessary fix on ClinVar template

<details>

**Prepare for testing**
1. `sudo -iu hiseq.clinical`
1. `ssh localhost`
</details>

<details>

**Prepare for testing**
</details>

<details>
</details>

**How to test**:

**Expected outcome**:

**Review:**
- [ ] code approved by
- [x] tests executed by CR

### DIFF
--- a/scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
+++ b/scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
@@ -368,6 +368,10 @@
   {{ super() }}
   <script type="text/javascript">
 
+  document.getElementById("pagination-form").addEventListener("submit", function () {
+      document.getElementById("page").disabled = true;
+  });
+
   $(function () {
       $('.casedata  div').hide();
       $('.vardata  div').hide();
--- a/scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
+++ b/scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
@@ -54,10 +54,22 @@
   <p>No ClinVar submissions found. You can create one by choosing one or more pinned variants on a case page and clicking the button "Submit to ClinVar".</p>
   {% endif %}
 </div>
-  <form id="pagination-form" method="get" action="{{ url_for('clinvar.clinvar_germline_submissions', institute_id=institute._id) }}">
-  {{ pagination_hidden_div(page, result_size, per_page) }}
-  {{ pagination_footer(page, result_size, per_page) }}
-</form>
+  <form
+    id="pagination-form"
+    method="get"
+    action="{{ url_for('clinvar.clinvar_germline_submissions', institute_id=institute._id) }}"
+    onclick="
+        if (event.target.matches('label.page-link[for^=paginate-]')) {
+            const input = document.getElementById(event.target.htmlFor);
+            document.getElementById('page').value = input.value;
+            event.preventDefault();
+            this.submit();
+        }
+    "
+  >
+      {{ pagination_hidden_div(page, result_size, per_page) }}
+      {{ pagination_footer(page, result_size, per_page) }}
+  </form>
 {% endmacro %}
 
 {% macro submission_panel(subm_obj) %}
@@ -367,10 +379,6 @@
 {% block scripts %}
   {{ super() }}
   <script type="text/javascript">
-
-  document.getElementById("pagination-form").addEventListener("submit", function () {
-      document.getElementById("page").disabled = true;
-  });
 
   $(function () {
       $('.casedata  div').hide();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ About changelog [here](https://keepachangelog.com/en/1.0.0/)
 - On outliers page, show inheritance patterns badges on genes cell (#6446)
 - Allow custom config human reference tracks, e.g. for local emergency copies when UCSC is unreachable (#6479)
 - Option to hide case details on the case specific gene panel extent report (#6485)
-- Pagination on ClinVar germline submissions page (#6491)
+- Pagination on ClinVar germline submissions page (#6491, #6495)
 ### Changed
 - Changelog enforcement specific to the `unreleased` changelog section (#6216)
 - Decoy GRCh37 build no longer available from igv. Add JP mirror to avoid UCSC fallback and its recent downtime issues (#6479)
--- a/scout/server/blueprints/variants/templates/variants/utils.html
+++ b/scout/server/blueprints/variants/templates/variants/utils.html
@@ -1258,6 +1258,7 @@
     {% for p in range(1, total_pages + 1) %}
       <input type="submit" name="page" id="paginate-page-{{p}}" value="{{p}}" hidden>
     {% endfor %}
+    <input type="hidden" id="page" name="page" value="{{ page }}">
   </div>
 {% endmacro %}
 


scout/server/blueprints/clinvar/templates/clinvar/clinvar_submissions.html
```diff
@@ -368,6 +368,10 @@ <h5 class="modal-title">ClinVar API submission</h5>
   {{ super() }}
   <script type="text/javascript">
 
+  document.getElementById("pagination-form").addEventListener("submit", function () {
+      document.getElementById("page").disabled = true;
+  });
+
```
northwestwitch: This is unfortunately needed now ->

### Without this code:

Image

### With this code:

Image

dnil:
Couldn't we fix the keys and getter instead? Like not depend on the proper "page" key to end up in element 0 of that getList?

northwestwitch:
> Couldn't we fix the keys and getter instead? Like not depend on the proper "page" key to end up in element 0 of that getList?

The issue is not in the getter, as the pages are updating fine when you click on the pagination. It's only annoying that you get the double parameter on the page URL. See if this other solution is better

dnil:
> > Couldn't we fix the keys and getter instead? Like not depend on the proper "page" key to end up in element 0 of that getList?
> 
> The issue is not in the getter, as the pages are updating fine when you click on the pagination. It's only annoying that you get the double parameter on the page URL. See if this other solution is better

Not super fond of this either: blocking with preventDefault can be necessary, but in this case I still feel the issue is just that there are multiple elements named the same way, and the getter doesn't care, only takes a first element 0, and the ordering is internal to Flask? I don't follow how the issue is not with the getter because of one operation working, but not quite the other?

northwestwitch:
> Not super fond of this either: blocking with preventDefault can be necessary, but in this case I still feel the issue is just that there are multiple elements named the same way, and the getter doesn't care, only takes a first element 0, and the ordering is internal to Flask? I don't follow how the issue is not with the getter because of one operation working, but not quite the other?

Mmm I don't think I understand what you mean by fixing the getter.. If I reitroduce that line the links work as expected, so no need to fixing the getter for the sake of navigating the pages on the ClinVar submissions page. The only annoying thing is the double page args in the url, originating from the fact that the page input is repeated twice in pagination_hidden_div. (This was already like this before I did any changes to the ClinVar submissions page). Did you mean another getter than the one on the ClinVar page? Could you make an example of how you would fix it otherwise?

northwestwitch:
> > Not super fond of this either: blocking with preventDefault can be necessary, but in this case I still feel the issue is just that there are multiple elements named the same way, and the getter doesn't care, only takes a first element 0, and the ordering is internal to Flask? I don't follow how the issue is not with the getter because of one operation working, but not quite the other?
> 
> Mmm I don't think I understand what you mean by fixing the getter.. If I reitroduce that line the links work as expected, so no need to fixing the getter for the sake of navigating the pages on the ClinVar submissions page. The only annoying thing is the double page args in the url, originating from the fact that the page input is repeated twice in pagination_hidden_div. (This was already like this before I did any changes to the ClinVar submissions page). Did you mean another getter than the one on the ClinVar page? Could you make an example of how you would fix it otherwise?

Like, you would rename the key to something else?

dnil:
> > > Not super fond of this either: blocking with preventDefault can be necessary, but in this case I still feel the issue is just that there are multiple elements named the same way, and the getter doesn't care, only takes a first element 0, and the ordering is internal to Flask? I don't follow how the issue is not with the getter because of one operation working, but not quite the other?
> > 
> > 
> > Mmm I don't think I understand what you mean by fixing the getter.. If I reitroduce that line the links work as expected, so no need to fixing the getter for the sake of navigating the pages on the ClinVar submissions page. The only annoying thing is the double page args in the url, originating from the fact that the page input is repeated twice in pagination_hidden_div. (This was already like this before I did any changes to the ClinVar submissions page). Did you mean another getter than the one on the ClinVar page? Could you make an example of how you would fix it otherwise?
> 
> Like, you would rename the key to something else?

Ok, three things here. 

### Naming and the getter

Renaming would be one option. Using a different name for some of it would solve it, but is going to be a little bit clunky on the python side. I don't think that is really necessary either. 

This was working fine for previous versions before removing the line, without any js extras, so reintroducing it to the same place should not on its own break anything. It's just broken on current main.

### The current pagination solution, for all pages. 
I have gripes with the original solution from chatGPT. I do trust it to find a working solution in a given case, but not necessarily make things easy to maintain after. We don't have to solve that now.

I know it is in general perfectly allowed to have multiple elements with the same name, that's the implicit way of making arrays. Like for a multiple choice checkbox. I am already getting uneasy with multiple submits with the same name, but that is kind of fine too since the value is all we care about while paginating, and only one will be the submitted one. Then things get complicated when both are used. I'm thinking it is probably possible to figure out the situation based on the submit status in the getter (we should know if a pagination was picked) and chose the correct page value from that. As it is the array order is assumed to be enough, and is apparently for the most part.

### The clinvar submissions page

If you are currently not worried about the getter taking the first value, like on the other pages, and just the URL, why not make it a POST request page? No-one has to see the array in the URL. But it should be working of course, with correct order so that the correct requested page is in element 0 all the time.

northwestwitch:
Making it a POST request was my first and obvious solution, and it was a POST request from the beginning. I switched to a GET one because it was not working - I've been spending a lot of time trying to make it a post while leaving the pagination macro as it is, since I don't want to change something that works - I kept getting an error (wrong form submitted, probably given the fact that the ClinVar submissions page  has some quirks in the structure of the table/forms.) Anyway I couldn't fix it and I didn't want to spend too much time on this either, given that that page is going to change anyway because of the refactoring. 
Any solution that fixes the bug, and especially one that is not invasive and doesn't change the original macro, is good for me. I’d be very happy to close this PR if you’d like to give it a try yourself?

> ### Naming and the getter
> Renaming would be one option. Using a different name for some of it would solve it, but is going to be a little bit clunky on the python side. I don't think that is really necessary either.
> 
> This was working fine for previous versions before removing the line, without any js extras, so reintroducing it to the same place should not on its own break anything. It's just broken on current main.
> 
> ### The current pagination solution, for all pages.
> I have gripes with the original solution from chatGPT. I do trust it to find a working solution in a given case, but not necessarily make things easy to maintain after. We don't have to solve that now.
> 
> I know it is in general perfectly allowed to have multiple elements with the same name, that's the implicit way of making arrays. Like for a multiple choice checkbox. I am already getting uneasy with multiple submits with the same name, but that is kind of fine too since the value is all we care about while paginating, and only one will be the submitted one. Then things get complicated when both are used. I'm thinking it is probably possible to figure out the situation based on the submit status in the getter (we should know if a pagination was picked) and chose the correct page value from that. As it is the array order is assumed to be enough, and is apparently for the most part.
> 
> ### The clinvar submissions page
> If you are currently not worried about the getter taking the first value, like on the other pages, and just the URL, why not make it a POST request page? No-one has to see the array in the URL. But it should be working of course, with correct order so that the correct requested page is in element 0 all the time.

dnil:
> Making it a POST request was my first and obvious solution, and it was a POST request from the beginning. I switched to a GET one because it was not working - I've been spending a lot of time trying to make it a post while leaving the pagination macro as it is, since I don't want to change something that works - I kept getting an error (wrong form submitted, probably given the fact that the ClinVar submissions page has some quirks in the structure of the table/forms.) Anyway I couldn't fix it and I didn't want to spend too much time on this either, given that that page is going to change anyway because of the refactoring. Any solution that fixes the bug, and especially one that is not invasive and doesn't change the original macro, is good for me. I’d be very happy to close this PR if you’d like to give it a try yourself?

🙄 Sure. I would not enjoy having it that brittle: if that breaks it, maybe just switching browsers would too. But please keep this open in case it turns out to be super subtle and I give up. 😆

northwestwitch:
Fixed by #6496. Closing

